PR #19, opened 2019-09-10
Don't recommend setting default rust toolchain

prestwich:
Changes the README to recommend setting a local override, rather than setting the default toolchain. Also removes some leftover whitespace


fckt:
Isn't the best way to just use a `rust-toolchain` file?

prestwich:
Yes, but I'd rather not make tooling decisions for your repo :)

PR #23, opened 2019-09-17
Fix tests race condition

fckt:
Tests are concurrent for BLOCKCHAIN_INTERFACE. Moved BLOCKCHAIN_INTERFACE into thread_local variable to not share it b/w threads.

### DIFF
--- a/near-bindgen/src/environment/env.rs
+++ b/near-bindgen/src/environment/env.rs
@@ -41,8 +41,8 @@ const STATE_KEY: &[u8] = b"STATE";
 /// A simple macro helper to read blob value coming from host's method.
 macro_rules! try_method_into_register {
     ( $method:ident ) => {{
-        BLOCKCHAIN_INTERFACE.with(|b_i| unsafe {
-            b_i.borrow()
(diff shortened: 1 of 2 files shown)


near-bindgen/src/environment/env.rs
```diff
@@ -36,10 +41,12 @@ const STATE_KEY: &[u8] = b"STATE";
 /// A simple macro helper to read blob value coming from host's method.
 macro_rules! try_method_into_register {
     ( $method:ident ) => {{
-        unsafe { BLOCKCHAIN_INTERFACE
-            .as_ref()
-            .expect(BLOCKCHAIN_INTERFACE_NOT_SET_ERR)
-            .$method(ATOMIC_OP_REGISTER) };
+        BLOCKCHAIN_INTERFACE.with(|b_i| unsafe {
```
nearmax: Here it is `b_i` but in some other places it is `b`. Some inconsistent naming

fckt:
> Could you rebuild the example contract and also try them with near-vm-runner-standalone to make sure it actually works?

checked, it works!

PR #28, opened 2019-10-02
Travis ci

ailisp:
Add a travis config which build four cases:
- rust nightly ~~2019-05-22~~ 2019-09-03
- latest nightly

each combine with

- git borsh/near-vm-logic
- release borsh/near-vm-logic

Add a helper script actually run the test. This script also works locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ rust:
   - nightly
 env:
   - NEAR_RELEASE=true
-  - NEAR_RELAESE=false
+  - NEAR_RELEASE=false
 
 matrix:
   allow_failures:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: rust
 rust:
   - nightly-2019-05-22
+  - nightly-2019-09-03
   - nightly
 env:
   - NEAR_RELEASE=true
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,19 @@
+language: rust
+rust:
+  - nightly-2019-05-22
+  - nightly
+env:
+  - NEAR_RELEASE=true
+  - NEAR_RELAESE=false
+
+matrix:
+  allow_failures:
+    - rust: nightly
+  fast_finish: true
+
+branches:
+  only:
+  - master
+
+script:
+  - ./ci-test.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: rust
 rust:
-  - nightly-2019-05-22
   - nightly-2019-09-03
   - nightly
 env:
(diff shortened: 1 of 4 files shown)


ailisp:
@nearmax our way to compare compiler error message is very depending on specific version of rust compiler. I the following successful/failed build: nightly 2019-09-03 success, 2019-05-22 (nearcore) and latest nightly failed. 
Unfortunately travis doesn't allow more than two nightly rust so I put 09-03 and latest nightly to make it pass.

https://travis-ci.com/nearprotocol/near-bindgen/builds/130112329
https://travis-ci.com/nearprotocol/near-bindgen/builds/130113538

ailisp:
Also latest nightly build fails for a different reason, those smart contract supposed to pass also failed.

ci-test.sh
```diff
@@ -0,0 +1,28 @@
...
+    echo "Test with release version of borsh and near-vm-logic"
+    sed -n '/^borsh/p' near-bindgen/Cargo.toml 
+    sed -n '/^near-vm-logic/p' near-bindgen/Cargo.toml
+    cargo test --all
+else
+    echo "Test with git version of borsh and near-vm-logic"
+
+    cp near-bindgen/Cargo.toml{,.bak}
+    cp Cargo.lock{,.bak}
+    borsh_git='borsh = { git = "https://github.com/nearprotocol/borsh", branch = "master" }'
```
nearmax: I think using commented out `[patch]` is more maintainable because it requires only one sed and people can also test it manually on their machine if they see CI failing.

.travis.yml
```diff
@@ -0,0 +1,19 @@
+language: rust
+rust:
+  - nightly-2019-09-03
+  - nightly
+env:
+  - NEAR_RELEASE=true
+  - NEAR_RELAESE=false
```
nearmax: That's a typo

ci-test.sh
ailisp: Make sense

.travis.yml
ailisp: :+1:

ailisp:
> > Also latest nightly build fails for a different reason, those smart contract supposed to pass also failed.
> 
> I don't see from this link (2019-05-22 rust: https://travis-ci.com/nearprotocol/near-bindgen/builds/130112329 ) failing contracts. Also, the tests for each contract should be run with:
> 
> ```
> cargo test --features env_test
> ```
> 
> This is until we have
> #29 fixed

rust 20190522 is still problematic (because of different compilation message. Try this locally:
```
echo "nightly-2019-05-22" > rust-toolchain
cargo test --features env_test
```

ailisp:
@nearmax fixed

PR #32, opened 2019-10-23
Bump to 0.3

evgenykuzyakov:
* Bump version to 0.3 (new logic, new context)
* Add `panic_utf8` and `block_timestamp`
* Update borsh dep version
* Adding a new example that works with collections
* Update collections API to use references instead of consuming K,V
* Adding test all, build all for examples


near-bindgen/src/environment/blockchain_interface.rs
```diff
@@ -14,6 +14,7 @@ pub trait BlockchainInterface {
     unsafe fn predecessor_account_id(&self, register_id: u64);
     unsafe fn input(&self, register_id: u64);
     unsafe fn block_index(&self) -> u64;
+    unsafe fn block_timestamp(&self) -> u64;
```
nearmax: Please expose new methods in `env.rs`. Do not wrap timestamp into `Instant` or `DateTime`, just keep it as `u64`.

examples/test_all.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
nearmax: I am not sure CI is actually running tests from examples. We should add it to CI config if it does not.

near-bindgen/src/collections/set.rs
```diff
@@ -39,7 +39,7 @@ where
     T: BorshSerialize + BorshDeserialize,
 {
     /// Serializes element into an array of bytes.
-    fn serialize_element(&self, element: T) -> Vec<u8> {
+    fn serialize_element(&self, element: &T) -> Vec<u8> {
```
nearmax: CC @mikhailOK since, he is using near-bindgen collections for EVM.

evgenykuzyakov: yep, I'll update evm contract later. Currently it depends on the 0.2.5 version, so shouldn't be an issue

examples/test_all.sh
```diff
@@ -0,0 +1,15 @@
...
+set -e
+
+pushd $(dirname $0)
+
+for d in */ ; do
+    pushd $d
+    echo "Checking $d compiles"
+    RUSTFLAGS='-C link-arg=-s' cargo +nightly check --target wasm32-unknown-unknown --release
```
ailisp: I think try remove `+nightly`. So in travis there's two build, nightly-2019-09-03 and nightly (latest), both contain nightly rust only. but doing explicit "+nightly" will override in nightly-2019-09-03 build.
Doing two build is to tell a specific rust version that works, and let you know if it still works on latest nightly

.travis.yml
```diff
@@ -1,14 +1,12 @@
 language: rust
 rust:
-  - nightly-2019-09-03
```
ailisp: so i don't recommend remove this, but update it to a newer working version is fine

examples/test_all.sh
evgenykuzyakov: I don't think we need to test against `nightly-2019-09-03` at all. Going to switch it to latest nightly from nearcore

.travis.yml
```diff
@@ -1,14 +1,12 @@
...
   - nightly
```
mikhailOK: should actually switch to stable, near-bindgen itself doesn't require nightly

evgenykuzyakov: good point

examples/test_all.sh
ailisp: There was a time near-bindgen doesn't pass with latest nightly and this was needed to make it pass. In future still likely new nightly breaks test so it's better to test with both a pinned version (2019-10-21 for nearcore now) and latest nightly.

near-bindgen/src/environment/env.rs
```diff
@@ -119,24 +116,24 @@ pub fn input() -> Option<Vec<u8>> {
 }
 /// Current block index.
 pub fn block_index() -> BlockIndex {
+    unsafe {
+        BLOCKCHAIN_INTERFACE
+            .with(|b| b.borrow().as_ref().expect(BLOCKCHAIN_INTERFACE_NOT_SET_ERR).block_index())
+    }
+}
+/// Current block timestamp.
+pub fn block_timestamp() -> BlockIndex {
```
nearmax:
```suggestion
pub fn block_timestamp() -> u64 {
```

examples/test_all.sh
ailisp: just saw you replace with stable. if stable works too that's better

near-bindgen/src/environment/env.rs
```diff
@@ -79,10 +79,7 @@ pub fn read_register(register_id: u64) -> Option<Vec<u8>> {
 /// Returns the size of the register. If register is not used returns `None`.
```
nearmax: That's a lot of reformating. We should figure out why our rustfmt format things differently. Do you have https://github.com/nearprotocol/nearcore/blob/master/rustfmt.toml in ~/.config/rustfmt/ ?

evgenykuzyakov: Yes. I think it's because the original file was not reformatted at all. Rustfmt is enabled per project in clion

nearmax: My rustfmt was uninstalled :(

PR #38, opened 2019-10-25
Fun token example

evgenykuzyakov:
I need a test that has different context within a single test, so I can test locks and allowance properly. @nearmax  

Also looks like I need to update comments.


examples/fun-token/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+        if account.balance < lock_amount {
+            env::panic(b"Not enough unlocked balance");
+        }
+        account.balance -= lock_amount;
+
+        // If locking by escrow, need to check and update the allowance.
+        if escrow_account_id != owner_id {
+            let allowance = account.get_allowance(&escrow_account_id);
+            if allowance < lock_amount {
+                env::panic(b"Not enough allowance");
```
ailisp: i have a question here, if paniced here the before `account.balance -= lock_amount` won't commit?

examples/fun-token/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+    }
+    fn bob() -> AccountId {
+        "bob.near".to_string()
+    }
+    fn carol() -> AccountId {
+        "carol.near".to_string()
+    }
+
+    fn get_context(predecessor_account_id: AccountId) -> VMContext {
+        VMContext {
```
ailisp: these fixtures seems appear a lot of place, would be helpful to provide as a helper?

examples/fun-token/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+        testing_env!(get_context(bob()), Config::default());
+        contract.lock(carol(), lock_amount);
+        assert_eq!(contract.get_allowance(carol(), bob()), allowance - lock_amount);
+        assert_eq!(contract.get_unlocked_balance(carol()), total_supply - lock_amount);
+        assert_eq!(contract.get_total_balance(carol()), total_supply);
+        // Acting as alice now
+        testing_env!(get_context(alice()), Config::default());
+        contract.lock(carol(), lock_amount);
+    }
+}
```
ailisp: :+1 a lot of interesting cases

examples/fun-token/src/lib.rs
evgenykuzyakov: Yes

examples/fun-token/src/lib.rs
evgenykuzyakov: We can provide some default implementation and just modify details.

PR #39, opened 2019-10-26
Cleanup and usability improvements

nearmax:
* Script for publishing crates;
* Removed unnecessary lazy_static;
* Remove env_test altogether, now that we have a workaround for cargo bug;
* Removing check_all.sh which fails now for no apparent reason, even though build succeeds;
* Bumping version.


ci-test.sh
```diff
@@ -25,8 +25,6 @@ else
     fi
 
     # Only testing it for one configuation to avoid running the same tests twice
-    echo "Checking compilation of wasm32 for all examples"
-    ./examples/check_all.sh
```
evgenykuzyakov: It was actually helpful cause it was building wasm32 target. We should instead investigate why it fails to build now?

nearmax: ./build_all.sh is building it anyway.

evgenykuzyakov: But `build_all.sh` is not part of CI

nearmax: Fixed.

PR #41, opened 2019-11-18
Allow having conditional compilation and other method level attributes

nearmax:


near-bindgen-core/src/initializer_attribute.rs
```diff
@@ -54,6 +54,13 @@ pub fn process_init_method(
     impl_type: &Type,
     is_trait_impl: bool,
 ) -> syn::Result<TokenStream2> {
+    let attrs = method.attrs.iter().fold(TokenStream2::new(), |mut acc, attr| {
+        let attr_str = attr.path.to_token_stream().to_string();
+        if attr_str != "callback_args_vec".to_string() && attr_str != "callback_args".to_string() {
```
frol: You don't need `&str.to_string()` to compare. `if attr_str != "callback_args_vec" && attr_str != "callback_args" {` should be just fine.

nearmax: Fixed.

near-bindgen-core/src/initializer_attribute.rs
```diff
@@ -54,6 +54,13 @@ pub fn process_init_method(
     impl_type: &Type,
     is_trait_impl: bool,
 ) -> syn::Result<TokenStream2> {
+    let attrs = method.attrs.iter().fold(TokenStream2::new(), |mut acc, attr| {
+        let attr_str = attr.path.to_token_stream().to_string();
+        if &attr_str != "callback_args_vec" && &attr_str != "callback_args" {
```
frol: Hmm, why `&attr_str`?

nearmax: Fixed. That's weird it works, but CLion complains: https://www.dropbox.com/s/m1yo93wbtrnhf3o/Screenshot%202019-11-18%2016.15.19.png?dl=0

PR #45, opened 2019-11-22
Some rustdocs

nearmax:
This code also introduces some changes and bumps the version. I have marked some rustdoc test as ignored to be able to submit this PR.


examples/status-message-collections/src/lib.rs
```diff
@@ -37,14 +37,14 @@ impl StatusMessage {
 mod tests {
     use super::*;
     use near_bindgen::MockedBlockchain;
-    use near_bindgen::{testing_env, Config, VMContext};
+    use near_bindgen::{testing_env, VMContext};
 
     fn get_context(input: Vec<u8>, is_view: bool) -> VMContext {
         VMContext {
-            current_account_id: "alice.near".to_string(),
```
fckt: motivation? (just want to be sure I'm on track with)

nearmax: We prohibited `.` in account names long time ago. I simple have not updated this code yet.

PR #49, opened 2019-12-30
Emulate iterator in near_bindgen. Add Vector collection. Remove iterator API.

nearmax:


nearmax:
Pushed because I need to unblock Flux. Will address the comments later.

near-bindgen/src/collections/vector.rs
```diff
@@ -0,0 +1,413 @@
...
+            self.len -= 1;
+
+            // Swap the last element with the current one.
+            let raw_last_value = if env::storage_remove(&last_lookup_key) {
+                match env::storage_get_evicted() {
+                    Some(x) => x,
+                    None => env::panic(ERR_INCONSISTENT_STATE),
+                }
+            } else {
+                env::panic(ERR_INCONSISTENT_STATE)
```
fckt: curious, how it compiles if the return type of `env::panic` is `()`?..

bowenwang1996: The return type of `env::panic` is `!`

PR #53, opened 2020-01-09
Log str instead of bytes

### LINKED ISSUE #52: panic and log should take string rather than [u8]
`env::panic` and `env::log` should take &str rather than &[u8] for better developer experience.

bowenwang1996:
Fixes #52


nearmax:
I am going to undo this change. It is not explained in https://github.com/nearprotocol/near-bindgen/issues/52 why it improves development experience. Most of the time panic uses constants and the following two lines:
```rust
const ERR_KEY_SERIALIZATION: &[u8] = b"Cannot serialize key with Borsh";
const ERR_KEY_SERIALIZATION: &str = "Cannot serialize key with Borsh";
```
have exactly the same development experience.

Also, we intentionally avoided strings because we wanted near-bindgen support tiny optimized contracts and pulling dependency on strings increases the code size when compiled to Wasm, so we discourage users from using them in the first place. https://rustwasm.github.io/book/reference/code-size.html#avoid-string-formatting

PR #58, opened 2020-01-23
Add a simple log to examples for better devx

janedegtiareva:


janedegtiareva:
It looks like error messages are usually done by e.g.  env::panic(b"Not enough allowance");
Do we want to switch that to log?

nearmax:
> It looks like error messages are usually done by e.g. env::panic(b"Not enough allowance");
> Do we want to switch that to log?

I made a type. What I meant is:

> Let's have some meaningful log message, that reflects what is happening at the given moment of time.

janedegtiareva:
> > It looks like error messages are usually done by e.g. env::panic(b"Not enough allowance");
> > Do we want to switch that to log?
> 
> I made a type. What I meant is:
> 
> > Let's have some meaningful log message, that reflects what is happening at the given moment of time.
Are you thinking something generic like "so and so fn got called" (see latest commit for example), or more specific to the individual examples?

evgenykuzyakov:
@janedegtiareva Is it still needed?

evgenykuzyakov:
Closing for now

PR #60, opened 2020-01-23
Update build command in build.sh to ensure optimized build

janedegtiareva:
(otherwise the .wasm file might be too large to deploy)


janedegtiareva:
done (removed .cargo dirs)

PR #63, opened 2020-01-24
Replace serde_json

nearmax:


janedegtiareva:
Should we wait until this works on testnet? (current version on testnet does not work)

PR #66, opened 2020-01-29
[WIP] A contract that implements user-centric open-web contract

evgenykuzyakov:


examples/open-web/src/lib.rs
```diff
@@ -0,0 +1,367 @@
...
+            b'-' | b'_' | b'.' => (),
+            _ => env::panic(
+                b"Unsupported character in the app ID. Only allowed to use `-.|` and 0-9 a-z",
+            ),
+        }
+    }
+}
+
+fn app_key(app_id: &AppId, key: &Key) -> Vec<u8> {
+    let mut res = Vec::with_capacity(app_id.len() + key.len() + 1);
```
willemneal: Why is this mutable?

examples/open-web/src/lib.rs
```diff
@@ -0,0 +1,367 @@
...
+    res
+}
+
+// One can provide a name, e.g. `ext` to use for generated methods.
+#[ext_contract(remote)]
+pub trait RemoteSelf {
+    fn post_message(&mut self, app_id: AppId, message: Message);
+}
+
+#[near_bindgen(init => new)]
```
willemneal: What does this mean?

examples/open-web/src/lib.rs
evgenykuzyakov: Because I modify it below. In rust you can't modify content of immutable collections.

examples/open-web/src/lib.rs
evgenykuzyakov: It's a macro with param. Indicates that we use `new` as a constructor for the object. `near-bindgen` behind the scene always deserializes the state of the object (from persistent storage) before calling a public function and serializes the state back at the end of the function execution.

examples/open-web/src/lib.rs
```diff
@@ -0,0 +1,373 @@
+use borsh::{BorshDeserialize, BorshSerialize};
```
nearmax: This contract also needs a thorough documentation explaining the usage.

examples/open-web/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+type AccountId = String;
+
+const GAS: u64 = 1_000_000_000_000_000;
+
+const APP_METHODS: &[u8] = b"set,remove,pull_message,send_message";
+
+#[near_bindgen]
+#[derive(BorshDeserialize, BorshSerialize)]
+pub struct UserData {
+    /// App specific PublicKey -> App ID prefix.
```
nearmax: Why App ID prefix and not just App ID?

examples/open-web/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+        self.messages.insert(&app_id, &q);
+    }
+
+    pub fn send_message(
+        &mut self,
+        receiver_id: AccountId,
+        app_id: AppId,
+        message: Message,
+    ) -> PromiseOrValue<()> {
+        let _my_app_id = self.auth_app_id();
```
nearmax:
```rust
self.auth_app_id();
```

evgenykuzyakov:
Moved to https://github.com/metanear/

PR #73, opened 2020-02-03
Add panic hook

### LINKED ISSUE #72: Expose errors
If there is a deserialization error the `near-bindgen` contract fails with `unknown trap`. Instead we can use `panic_utf8` to throw a meaningful error.

Alternative is to wrap panic into a hook. Use something like: https://github.com/rustwasm/console_error_panic_hook

evgenykuzyakov:
Added STD panic hook. Now we expose Rust errors from Wasm through `panic_utf8`:

Examples:

unreachable macro:
```
Scheduling a call: panic-test.test_panic({})
Error:  TypedError: Transaction Byy8Hg5te6i9LtgCMcg8uVHzHnJmfTKBZJT5Aj7wPnEL failed. Smart contract panicked: panicked at 'internal error: entered unreachable code: AHAHA', src/lib.rs:29:9
```

invalid arguments:
```
Scheduling a call: panic-test.set_allowance({})
Error:  TypedError: Transaction 3GE7tCd2bg1GYhgXXUE7Y5wArLXKYujCQ8qvn6mVwjnC failed. Smart contract panicked: panicked at 'called `Result::unwrap()` on an `Err` value: Error("invalid type: null, expected a string", line: 0, column: 0)', src/libcore/result.rs:1084:5
```

Fixes #72


evgenykuzyakov:
I also need to add a hook for `initializer`, 1 sec

PR #74, opened 2020-02-03
0.4.0 Allow specifying serialization per input, output, callbacks

nearmax:
Refactors the code to make it more readable and maintainable by splitting code generation and information extraction. This allows more granular testing (will add more tests after this PR) and ability to extend the functionality. In the future it will also allow strict typing with `Promises` such that it won't be possible to construct chain of callbacks with incompatible types.

### Init method

Previously, when someone wants to have a method that accepts arguments and initializes the entire contract that would need to decorate the `impl` section like this:
```rust
#[near_bindgen(init => my_impl_method)]
impl MyContract {
  pub fn my_impl_method(arg0: SomeType) -> Self {
  }
  pub fn regular_method(&self, arg0: SomeOtherType) {
  }
}
```
when this feature was used every other non-init method would assume that the contract is initialized and would panic if it wasn't. If the feature is not used every method would assume that contracts implements `Default` trait and will call if it finds out the contract was not initialized before.

This approach had the following disadvantages:
* We did not allow multiple initialization methods;
* It was not possible to allow both default initialization and initialization using init methods;
* An addition of a single init method will modify the behavior (generated code) of non-init methods. Image Alice implementing a smart contract with methods M1, M2, ... . Imagine, later Bob decides to add an initializer method I1, when he does it the wrapping code generated for M1,M2, ... will change to not use `Default` trait overturning the expectations that Alice had when she was implementing her methods;
* Nit. Duplication of the method name, `my_impl_method` had to be written twice.

The new syntax is the following:
```rust
#[near_bindgen]
impl MyContract {
  #[init]
  pub fn my_impl_method1(arg0: SomeType) -> Self {
  }
  #[init]
  pub fn my_impl_method2(arg0: SomeType1, arg1: SomeType2) -> Self {
  }
  pub fn regular_method(&self, arg0: SomeOtherType) {
  }
}
```
It allows multiple init methods, but it still requires the contract to implement the `Default` trait. If contract developer wants to prohibit default initialization and have the old behavior when the contract panics if it was not initialized they can implement `Default` trait like this:
```rust
impl Default for MyContract {
  fn default() -> Self {
    panic!("This contract prohibits default initialization")
  }
}
```
Note, that for most Rust contracts it is safe to auto-derive this trait: `#[derive(Default)]`.

### Callbacks

Previously when a method expects an information passed as a callback it needs to decorate itself like this specifying which arguments should be read from the callback rather than `env::input()`:
```rust
#[near_bindgen]
impl MyContract {
   #[callback_args(arg0, arg2)]
   pub fn subscribed_to_callback(&self, arg0: A, arg1: B, arg2: C) {
   }
}
```
If it unknown what is the number of the callback arguments then one can use the following syntax to get them all if they are of the same type:
```rust
#[near_bindgen]
impl MyContract {
   #[callback_args_vec(arg1)]
   pub fn subscribed_to_callback(&self, arg0: A, arg1: Vec<B>, arg2: C) {
   }
}
```

This had the following disadvantages:
* We cannot have multiple different ways to access the callback arguments (for instance have both `callback_args` and `callback_args_vec`);
* Nit. Duplication of argument names.

The new format is the following:
```rust
#[near_bindgen]
impl MyContract {
   pub fn subscribed_to_callback(&self, arg0: A, #[callback_vec] arg1: Vec<B>, #[callback]  arg2: C, arg3: D, #[callback] arg4: E) {
   }
}
```

Also, in addition one can use the same signature for the traits that describe the external contracts:
```rust
trait SomeExternalContract {
  fn method_we_call_with_callback(&self, &self, arg0: A, #[callback_vec] arg1: Vec<B>, #[callback]  arg2: C, arg3: D, #[callback] arg4: E);
}
```

### Flexible serialization

Previously all contract API was serialized with JSON which had the following issues:
* JSON consumed too much gas. Some smart contracts would have 80% of their gas spent only on parsing JSON, even making some of the contract unfeasible because they would exceed the maximum gas limit, e.g. Eth bridge;
* JSON does not allow any Rust type;
* Binary blobs are difficult with JSON. E.g. if someone want to pass a binary blob they would need to base64 it;
* JSON is a lot of code that needs to be compiled into a contract.

The alternative to JSON that does not have the above disadvantages is binary serialization, e.g. [Borsh](http://borsh.io). Unfortunately binary serialization (not just Borsh) has the following disadvantages:
* It is not self-descriptive;
* Unlike JSON it is not a well-known format natively supported by many frameworks and languages.

Therefore we want to allow specifying various serializations in a single contract.

Now each contract method can has a different serialization/deserialization format for the following component independenty:
* All input args can have a different format (note, input args should all have the same format within a given contract method);
* Callback args can have different formats;
* Result of the method can have a different format.
E.g.:
```rust
#[near_bindgen]
impl MyContract {
  #[init]
  pub fn f1(#[serializer(borsh)] arg0: SomeType) -> Self {
  }

  #[result_serializer(borsh)]
  pub fn f2(#[serializer(json)] arg0: SomeType1, #[serializer(json)] arg1: SomeType2) -> Something{
  }

  #[result_serializer(borsh)]
  pub fn f3(&self, #[callback] #[serializer(borsh)] callback0: A, #[callback] #[serializer(json)] 
 callback1: B, #[serializer(borsh)] arg0: C, #[serializer(borsh)] arg1: D, callback2: E) ->  F {
  }
}
```
If serializer is not specified it is assumed to be JSON.

## TODO:
- [x] Update examples to use the new syntax.
- [x] Update documentation in this repo

## After new version is published:
- [ ] Update https://github.com/nearprotocol/workshop

## Next PR:
- [ ] Add metadata function that returns JS schema for nearlib.


evgenykuzyakov:
Super cool.

Is it possible to specify `#[serializer(borsh)]` for entire method or for entire impl?

nearmax:
> Super cool.
> 
> Is it possible to specify `#[serializer(borsh)]` for entire method or for entire impl?

This is not the case yet. Currently one needs to specify this attribute on each individual argument, and where JSON is chosen by default when nothing is specified. We can add this functionality later, currently it is a nice to have but not a requirement for EthDenver.

nearmax:
@evgenykuzyakov Please review. I have fixed examples, docs and tested cross contract calls with Borsh.

examples/fun-token/src/lib.rs
```diff
@@ -59,8 +59,15 @@ pub struct FunToken {
...
+impl Default for FunToken {
+    fn default() -> Self {
+        panic!("Fun token should be initialized before usage")
+    }
+}
+
+#[near_bindgen]
 impl FunToken {
+    #[init]
     pub fn new(owner_id: AccountId, total_supply: Balance) -> Self {
```
evgenykuzyakov: This seems like a duplicated code that most contract will have to implement if there is a constructor.
There are 2 issues:
1) Manually disabling `Default` using panic. I think before this, the `Default` was not required, so if you wanted to avoid default, you didn't have to implement it.
2) `init` doesn't prevent reinitializing again. You have to manually implement state checking. As part of the `ethbridge` I used the following line in the `init` wasm function
```
assert!(near_bindgen::env::state_read::<EthBridge>().is_none(), "Already initialized");
```

nearmax: We can later add feature `#[init_once]`.

nearmax:
> Actually. Going to approve to unblock the current version. But ideally we should have a flag that makes `Default` disabled.
> If you don't want to address issues mentioned above, just create TODOs

@evgenykuzyakov  Thank you. I have tried to rationalize the change in the description (I have updated it since your last message).

PR #103, opened 2020-03-04
Fix cross contract high level, add epoch_height and fix broken build

### LINKED ISSUE #102: cross-contract high level example doesn't work
When following the cross contract high level example readme, at this step
```
near call cross_contract simple_call "{\"account_id\": \"status_message\", \"message\":\"bonjour\"}" --accountId=test_near --gas 1000000000000000
```
the call failed because it exceeds the prepaid gas, but I already give the maximum prepaid gas here (10^16).

bowenwang1996:
Fixes #102. Also fixes the same bug in cross contract low level example.
Fixes nearprotocol/nearcore#2062
Fixes https://github.com/nearprotocol/nearcore/pull/2278#issuecomment-604727305


nearmax:
It is still failing with `Error:  InvalidTxError: {}`

bowenwang1996:
@nearmax I just tried again and it works for me. I am using master of nearcore and near-shell version 0.20.1

PR #104, opened 2020-03-05
add nightly to prerequisites

icerove:
when I run build command without nightly, it has errors like this:
can't find crate for `core`


nearmax:
As discussed before, Rust contracts can be built without nightly, and it should not be a pre-requisite.

PR #105, opened 2020-03-10
clean up readme for high level cross contract example

mikedotexe:
Added few changes that were necessary in my mind:

- Info on running the `stop.py` script afterwards so Docker doesn't keep running
- Some differences in how `create-near-app`'s `config.js` looks
- instead of `…"{\"key\":…` use single quotes without escaping `…'{"key":…`

Also added more help text generally.


nearmax:
Have you tried following the instructions and made sure they work?

mikedotexe:
> Have you tried following the instructions and made sure they work?

absolutely. I wouldn't dream of pushing up instructions I haven't tried. unless it was a nightmare.
ran it all again starting with `./build.sh` and noticed that the two lock files and wasms needed to be updated, so pushed those up as well.

nearmax:
@bowenwang1996 says this example does not run anymore : https://github.com/nearprotocol/near-sdk-rs/issues/102 Are you saying it actually runs correctly?

mikedotexe:
I was aware of the issue Bowen made and did not run into it.
If I may, I'll be assigning it back to Bowen.
I tested this example over 10 times in the process of breaking it out into a standalone example here:
https://github.com/near-examples/rust-high-level-cross-contract
so I'm confident I didn't run into that issue during all those trials.

bowenwang1996:
@mikedotexe which version of nearcore are you using? Can you check the field `max_total_prepaid_gas` in your genesis?

mikedotexe:
> @mikedotexe which version of nearcore are you using? Can you check the field `max_total_prepaid_gas` in your genesis?

I've been on `master` for `nearcore`. Just pulled again, got some changes, and ran the instructions again. Still works.

```
mike@near-mp ~/near/nearcore/near/res -- master -> grep max_total_prepaid_gas *
testnet_genesis_config.json:        "max_total_prepaid_gas": 10000000000000000,
```

bowenwang1996:
@mikedotexe I just tried again and it still fails. Also if I am not crazy, 10^16 (max prepaid gas) is smaller than 10^18 (which is what used in the contract for cross contract gas) and it makes sense that you cannot spend more than what you have. But maybe I am hallucinating since no one else has reproduced the error I encountered yet....

mikedotexe:
I think we can close this, actually, as we have this fleshed out in detail for end users on the example:
https://github.com/near-examples/rust-high-level-cross-contract

PR #108, opened 2020-03-13
standardize license to be dual license MIT and Apache

mikedotexe:
Discussions have landed that while `nearcore` will have the GNU license, many other repositories will have a dual license of MIT and Apache. This is one of those repositories.


mikedotexe:
Closing in favor of https://github.com/near/near-sdk-rs/pull/136

PR #109, opened 2020-03-13
Update to 0.6.0

evgenykuzyakov:
- Bump version to 0.6.0
- Replace *_with_schema with regular calls for Borsh (disable Schema requirements for input/results same as 0.4.0 version)
- Use new vm-* version at 0.6.0


evgenykuzyakov:
Oh. Wait. It needs near-vm-logic with new borsh

PR #110, opened 2020-03-17
Update fungible token contract

evgenykuzyakov:
Simplifying fungible token contract by removing locks.
Introducing  `U128` wrapper and adding comments


examples/fungible-token/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+
+pub struct U128(u128);
+
+impl From<u128> for U128 {
+    fn from(v: u128) -> Self {
+        Self(v)
+    }
+}
+
+impl Into<u128> for U128 {
```
nearmax: `Into` is usually not implemented directly. Instead we need to implement `From<U128>`

examples/fungible-token/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    /// Helper method to set the account details for `owner_id` to the state.
+    fn set_account(&mut self, owner_id: &AccountId, account: &Account) {
+        let account_hash = env::sha256(owner_id.as_bytes());
+        self.accounts.insert(&account_hash, &account);
+    }
+}
+
+metadata! {}
```
nearmax: This will not produce metadata. `metadata!` should be wrapping all impl section with contract methods. We should either remove it or wrap everything.

evgenykuzyakov: Oh. Might be a leftover from `borsh`

PR #111, opened 2020-03-19
#[payable] attribute added

fckt:
To prevent contract clients to accidentally attach deposit to methods which doesn't consume value (in case of Functional with deposit it means loss of tokens).

By default we don't allow contract methods to accept deposits. Method must be opted-it for it explicitly.

### Discussion
https://commonwealth.im/near/proposal/discussion/329-payable-functions


fckt:
btw, link to travis is broken
https://travis-ci.com/github/nearprotocol/near-sdk-rs/jobs/299961767

near-bindgen-core/src/code_generator/impl_item_method_info.rs
```diff
@@ -52,8 +52,20 @@ impl ImplItemMethodInfo {
...
             ..
         } = attr_signature_info;
+        let deposit_check = if *is_payable {
+            // No check if method is payable
+            quote! {}
+        } else {
+            // If method is not payable, do a check to make sure that it doesn't consume deposit
+            quote! {
+                if near_bindgen::env::attached_deposit() != 0 {
+                    panic!("Method doesn't accept deposit");
```
evgenykuzyakov: I think we should panic with env instead in case we switch to no_std in the future, so we don't need to capture inner panic thought the hook.
```suggestion
                    near_bindgen::env::panic(b"Method doesn't accept deposit");
```

fckt: there is a https://doc.rust-lang.org/nomicon/panic-handler.html

Actually it's just generated some user-land code. Should we recommend users to not use `panic!`? 

Btw, there is a #[no_std] alternative https://doc.rust-lang.org/core/panic/struct.PanicInfo.html 

But I actually ok with both versions.

fckt:
@nearmax 
something broke:

➜  near-bindgen git:(fckt/payable) git push
ERROR: Permission to nearprotocol/near-bindgen.git denied to fckt.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights

near-bindgen-core/src/code_generator/impl_item_method_info.rs
evgenykuzyakov: It's probably faster to panic with `env` than with macro.

evgenykuzyakov:
> ➜ near-bindgen git:(fckt/payable) git push
ERROR: Permission to nearprotocol/near-bindgen.git denied to fckt.
fatal: Could not read from remote repository.

It was renamed to `near-sdk-rs`

nearmax:
> @nearmax
> something broke:
> 
> ➜ near-bindgen git:(fckt/payable) git push
> ERROR: Permission to nearprotocol/near-bindgen.git denied to fckt.
> fatal: Could not read from remote repository.
> 
> Please make sure you have the correct access rights

We moved it to a different organization and renamed it: https://github.com/near/near-sdk-rs

near-bindgen-core/src/code_generator/impl_item_method_info.rs
fckt: third-party will never use the `near_bindgen::env::panic`, they are using a standard functions. I'm fine with `near_bindgen::env::panic` here, but slightly into to use a standard interfaces and adopt them as a default strategy.

PR #113, opened 2020-03-20
Fix ci

ailisp:
The problem was travis not installed for near organization, installed now


ailisp:
@nearmax Seems near-sdk-rs refer to near/borsh, but borsh is still under nearprotocol now.

ailisp:
Should I move borsh to near now? Based on moves i did recently original
nearprotocol one would be redirect by github, so it's probably not breaking
any code using nearprotocol/borsh

evgenykuzyakov:
> Should I move borsh to near now? Based on moves i did recently original
nearprotocol one would be redirect by github, so it's probably not breaking
any code using nearprotocol/borsh

Yes

PR #114, opened 2020-03-21
Rename #[near_sdk] to #[near_bindgen]

nearmax:


nearmax:
> Do we need to include the .wasm files? They should not have changed as far as I understand.

Wasm compiler works in mysterious ways.

PR #123, opened 2020-04-02
[WIP] Add cross contract tests

evgenykuzyakov:
Adding a cross-contract testing framework based on standalone runner.
It can only do function call actions right now. Need to add other actions a more testing.


evgenykuzyakov:
> Why are tests written in python 🤔

Standalone runner right now can only do function calls using VM Logic. There are efforts to publish `runtime` crate so we can run entire chunk input. Before this is done, I've made a simple python runtime that uses standalone and routes receipts and cross-contract dependencies.

fckt:
Still actual?

evgenykuzyakov:
Outdated

PR #124, opened 2020-04-03
Add non-mutable reference to mocked blockchain

willemneal:


willemneal:
@nearmax I'm not sure why there is no travis check on this.  Also Pika was meant to be added when I finished the mock-vm.

PR #125, opened 2020-04-03
Add support for near-mock-vm

willemneal:
- Return outcome, receipts
- Add standalone binary for running wasm generated by as.


willemneal:
@nearmax It's failing the example test since the abort feature isn't on for non-relase test, but I'm not sure what changed where that would matter?  Could it be the removal of the "not wasm target"s?

willemneal:
@evgenykuzyakov Yes, I'm finally moving this to the AS sdk as it's used in unit testing. However, there were still a couple of changes to this sdk that were required.

frol (approved):
I cannot reason enough about the implications, but the code looks good to me

near-sdk-macros/res/near_blockchain.rs
```diff
@@ -105,6 +107,8 @@ pub mod near_blockchain {
             sys::log_utf16(len, ptr)
         }
 
+        unsafe fn abort(&self, msg_ptr: u32, filename_ptr: u32, line: u32, col: u32) {}
```
evgenykuzyakov, line 110: Is there `sys::abort` ?

evgenykuzyakov (approved):

near-sdk-macros/res/near_blockchain.rs
willemneal: Not sure what you mean.  This function is present in `near-vm-logic` so it should be mocked here as well and is used by errors in AS.

Issue #126, opened 2020-04-03
Update `insert` logic in `Set`


evgenykuzyakov:
Comment saying:
```
 /// Adds a value to the set. If the set did not have this value present, `true` is returned. If
    /// the set did have this value present, `false` is returned. Note, the elements that have the same hash value are undistinguished by
    /// the implementation.
```

But the current logic is inverse. It returns `true` if the value is already present.

PR #129, opened 2020-04-06
0.8.0: Fix set insert returned value

evgenykuzyakov:
Bumping to 0.8.0, because `Set` API has changed to reflect the comment and be similar to `HashSet` API.
Small refactoring, to remove unnecessary storage writes.

Fixes:
#126 

## Test plan
- Updated tests to match the new logic


near-sdk/src/collections/set.rs
```diff
@@ -63,27 +63,21 @@ impl<T> Set<T> {
         env::storage_has_key(&index_lookup)
     }
 
-    /// Adds a value to the set. If the set did not have this value present, `true` is returned. If
-    /// the set did have this value present, `false` is returned. Note, the elements that have the same hash value are undistinguished by
-    /// the implementation.
+    /// Adds a value to the set.
+    /// If the set did not have this value present, `true` is returned.
+    /// If the set did have this value present, `false` is returned.
+    /// Note, the elements that have the same hash value are undistinguished by the implementation.
```
nearmax, line 69: Since we do not hash state keys this comment is incorrect.

near-sdk/src/collections/vector.rs
```diff
@@ -73,18 +73,7 @@ impl<T> Vector<T> {
             }
         } else {
             let lookup_key = self.index_to_lookup_key(index);
-            let last_lookup_key = self.index_to_lookup_key(self.len - 1);
-            self.len -= 1;
```
nearmax, line 77: This breaks the `Vector::len()` because we need to store the real decreased length in `self.len`

evgenykuzyakov: `pop_raw` decreases length

near-sdk/src/collections/set.rs
evgenykuzyakov: removed

nearmax:
Please publish the libs.

Issue #128, opened 2020-04-06
Include common serialization classes


evgenykuzyakov:
The common classes are:
- Primitive types: U64, I64, U128, I128 - base-10 string encoding 
- StrPublicKey - to parse public key from string base58 format.
- ValidAccount - AccountID that verifies Account ID validity.
- Base64Vec - `Vec<u8>` as Base64

PR #131, opened 2020-04-06
Adding helper types

evgenykuzyakov:
- Added `U128`, `U64`, `I128`, `I64`.
- Added `Base58PublicKey`
- Added `Base64VecU8`

Addresses some classes from #128

## Test plan
- Added unit tests for ser/de


near-sdk/src/types/public_key.rs
```diff
@@ -0,0 +1,162 @@
...
+        let s = <String as serde::Deserialize>::deserialize(deserializer)?;
+        s.try_into()
+            .map_err(|err: Box<dyn std::error::Error>| serde::de::Error::custom(err.to_string()))
+    }
+}
+
+impl From<&StrPublicKey> for String {
+    fn from(str_public_key: &StrPublicKey) -> Self {
+        match str_public_key.0[0] {
+            0 => "ed25519:".to_string() + &bs58::encode(&str_public_key.0[1..]).into_string(),
```
nearmax, line 84: We are considering switching from b58 encoding in the future https://github.com/nearprotocol/nearcore/issues/2100
Let's not add it here since we won't be able to remove it from the compiled smart contracts in the future. Also, `bs58` is not a very popular library and I would suggests us to be conservative when deciding to add dependencies to `near-sdk`. I don't see a good reason why we won't `PublicKey` as a regular enum with `Vec<u8>` stringified as hex or base64.

near-sdk/src/types/integers.rs
```diff
@@ -0,0 +1,115 @@
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+/// Implements a wrapped integer type that serializes into and deserializes from base10 string.
+macro_rules! impl_str_type {
```
nearmax, line 4: Let's name the module to reflect the purpose of these types, e.g. `stringifiable_types` or `str_types`.

near-sdk/src/types/public_key.rs
evgenykuzyakov: I don't think we'll ever switch from bs58 for public keys, but it would make sense to rename `StrPublicKey` to `PublicKeyBase58` to make it clear and forward compatible.
The format right now is used by our tools and RPCs, so it's a standard right now.

near-sdk/src/types/integers.rs
evgenykuzyakov: `serde_types`?

evgenykuzyakov: Going to use `json_types`

near-sdk/src/json_types/vector.rs
```diff
@@ -0,0 +1,83 @@
+/// Helper class to serialize/deserialize `Vec<u8>` to base64 string.
```
nearmax, line 1: Module level docstring should use `//!`

near-sdk/src/json_types/vector.rs
```diff
@@ -0,0 +1,83 @@
...
+use serde::de::Error;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Base64VecU8(pub Vec<u8>);
```
nearmax, line 6: Please add docstring here. If there is a repetition between the module-level docstring and the struct-level docstring, use the latter one.

evgenykuzyakov:
Bumping version to `0.9.1`. Will wait for https://github.com/near/near-sdk-rs/pull/135 and publish. There are no new API changes. Only fixes

PR #133, opened 2020-04-07
Adding 2 helper methods towards env

evgenykuzyakov:
- Adding account ID validity check.
- Adding `state_exists` to use in `#[init]`.


near-sdk/src/environment/env.rs
```diff
@@ -708,3 +712,131 @@ pub fn state_write<T: borsh::BorshSerialize>(state: &T) {
...
+pub fn state_exists() -> bool {
+    storage_has_key(STATE_KEY)
+}
+
+// ##################
+// # Helper methods #
+// ##################
+
+/// Returns `true` if the given account ID is valid and `false` otherwise.
+pub fn is_valid_account_id(account_id: &[u8]) -> bool {
```
nearmax, line 726: Ideally, this logic should live in `near-vm-logic` or `primitives` and we should be importing it here.

evgenykuzyakov: Yep, good point. I also wanted to validate account ID in the `near-vm-logic`, so we can throw errors earlier.

examples/fungible-token/src/lib.rs
```diff
@@ -111,8 +111,9 @@ impl FungibleToken {
     /// Initializes the contract with the given total supply owned by the given `owner_id`.
     #[init]
     pub fn new(owner_id: AccountId, total_supply: U128) -> Self {
```
fckt, line 113: Btw, why not use the deployer (env::predecessor_account_id()) as an owner?

evgenykuzyakov: It to be able to deploy from the factory for example. The token owner doesn't have to be the caller.

near-sdk/src/environment/env.rs
```diff
@@ -698,6 +703,18 @@ pub fn storage_has_key(key: &[u8]) -> bool {
         _ => panic!(RETURN_CODE_ERR),
     }
 }
+/// Accessing receipts created by the contract. Only available in unit tests.
+pub fn created_receipts() -> Vec<Receipt> {
```
fckt, line 707:
```suggestion
#[cfg(test)]
pub fn created_receipts() -> Vec<Receipt> {
```

evgenykuzyakov: This isn't going to work, because `#[cfg(test)]` only works for local module. So if you are using a library and want to use test-only method, it's not going to be available within your tests.

PR #135, opened 2020-04-13
Do not check for attached_deposit if method is not mutable

### LINKED ISSUE #134: `#[payable]` change might break all view calls
Seems like function that are not marked as payable during the view call will try to call `attached_deposit()` and this is prohibited method in view. So every view call is going to fail.

We should add `#[view]` or detect methods as `view` if for example they doesn't use `&mut self`

fckt:
fixes #134


near-sdk-core/src/info_extractor/attr_sig_info.rs
```diff
@@ -97,13 +99,27 @@ impl AttrSigInfo {
...
+            let is_view = receiver.mutability.is_none();
+            if is_view && is_payable {
+                return Err(Error::new(
+                    receiver.span(),
+                    "Payable method must be mutable (not view)",
+                ));
+            }
+            is_view
+        } else {
+            false
```
evgenykuzyakov, line 112: Should this be a `view` as well. @nearmax ?

fckt: I think you're right

fckt: oh, actually, #init methods is also an exception

near-sdk-core/src/info_extractor/attr_sig_info.rs
```diff
@@ -20,13 +20,15 @@ pub struct AttrSigInfo {
     pub is_payable: bool,
     /// The serializer that we use for `env::input()`.
     pub input_serializer: SerializerType,
+    /// Whether the method is not mutate state
```
nearmax, line 23: Please fix grammar.

near-sdk-core/src/info_extractor/attr_sig_info.rs
```diff
@@ -97,13 +96,36 @@ impl AttrSigInfo {
             }
         }
 
+        let is_view = if let Some(ref receiver) = receiver {
```
nearmax, line 99: According to this code static method (the one that does not use `self`) that is not decorated with `#[init]` is not a view method, which is wrong.

near-sdk-core/src/code_generator/impl_item_method_info.rs
```diff
@@ -84,7 +90,7 @@ impl ImplItemMethodInfo {
                 method_invocation = quote! {
                     contract.#ident(#arg_list)
                 };
-                if mutability.is_some() && reference.is_some() {
+                if !is_view && reference.is_some() {
```
nearmax, line 93: Why do we need this change?

near-sdk-core/src/info_extractor/attr_sig_info.rs
fckt: How so? Else branch executes if receiver is None and if `is_init` is false `is_view` sets to true.

near-sdk-core/src/code_generator/impl_item_method_info.rs
fckt: just small refactoring to make the code more self-descriptive

near-sdk-core/src/info_extractor/attr_sig_info.rs
nearmax: You are right.

PR #137, opened 2020-04-17
0.9.2: Fix payable to be non native macro

evgenykuzyakov:
Fixing the following issue:
```
error: cannot find attribute `payable` in this scope
  --> src/lib.rs:17:7
   |
17 |     #[payable]
   |       ^^^^^^^

error: aborting due to previous error
```

#Test plan:

Added to status_message


examples/status-message/src/lib.rs
```diff
@@ -14,6 +14,7 @@ pub struct StatusMessage {
 
 #[near_bindgen]
 impl StatusMessage {
+    #[payable]
```
fckt, line 17: but why?

PR #143, opened 2020-05-01
Fix typo in payable code example

khorolets:


README.md
```diff
@@ -112,7 +112,7 @@ impl Default for StatusMessage {
 
 To declare a payable method simply use `#[payable]` decorator:
 ```rust
-user near_sdk::payable;
+use near_sdk::payable;
```
evgenykuzyakov, line 115: I don't think we need this line at all. `#[payable]` is inner proc-macro, so it's not required to be imported. Similar to `#[init]`

khorolets:
@nearmax I don't think small change to README is the cause of broken tests. FYI tried to run `./ci-test.sh` locally on master branch and it also fails.

PR #147, opened 2020-05-08
0.9.3: Update near-sdk. Remove wasm dependencies for tests

evgenykuzyakov:
- Bump to version 0.9.3
- Removing nearcore dependencies on types and fees from `wasm32` target. Makes wasm compilation faster.
- Replacing dependencies in `example` with `*` to reuse versions.
- Recompile everything


evgenykuzyakov:
That's too much out of date and we shouldn't make some usage non-wasm-32 because it affects wasm-vm

PR #149, opened 2020-05-12
Add RunnerStandalone and quickcheck tests

willemneal:


examples/fungible-token/Cargo.toml
```diff
@@ -24,3 +24,25 @@ panic = "abort"
...
+
+[dev-dependencies.near-primitives]
+git = "https://github.com/nearprotocol/nearcore.git"
+version = "0.1.0"
+branch = "fckt/standalone-runitme-patch"
+
+[dev-dependencies.near-crypto]
+git = "https://github.com/nearprotocol/nearcore.git"
+version = "0.1.0"
+branch = "fckt/standalone-runitme-patch"
```
evgenykuzyakov, line 48: This can now be replaced with nearcore:
```suggestion
near-crypto = { git = "https://github.com/nearprotocol/nearcore.git" }
near-primitives = { git = "https://github.com/nearprotocol/nearcore.git" }
near-runtime-standalone = { git = "https://github.com/nearprotocol/nearcore.git" }
```

examples/fungible-token/rust-toolchain
```diff
@@ -0,0 +1 @@
+nightly
```
evgenykuzyakov, line 1: I think you need to specify `nearcore` specific chain: `nightly-2020-03-19`

But instead we should use it only for the test folder. E.g. using `cargo +nightly-2020-03-19 ...`

examples/fungible-token/tests/utils.rs
```diff
@@ -0,0 +1,195 @@
+
```
evgenykuzyakov, line 1: Can common utils be put in `near-sdk` to be reusuable? What's the plan?
cc @fckt

examples/status-message/src/lib.rs
```diff
@@ -29,6 +29,10 @@ impl StatusMessage {
         env::log(b"A");
         self.records.get(&account_id).cloned()
     }
+
```
evgenykuzyakov, line 32: Please revert status-message changes

willemneal: Oops!

examples/fungible-token/rust-toolchain
willemneal: So `cargo +nightly-2020-03-19 test` ? or add `nightly-2020-03-19` to a file in the test folder?

willemneal: I added `+nightly-2020-03-19` to the `tests_all.sh` file and they worked.  Should I push that?

examples/fungible-token/tests/utils.rs
fckt: That's a good question. As you can see these utils here are not reusable.

At some point, we should have a general tool which generate an external client for contracts which will create transactions (or just actions) based on the contract metadata.

examples/fungible-token/rust-toolchain
fckt: There is already `rust_toolchain` in the root.

examples/fungible-token/tests/general.rs
```diff
@@ -0,0 +1,49 @@
...
+    assert_eq!(
+        root.get_balance(&mut runtime, root.account_id().into()),
+        deposit_amount
+    );
+    return true;
+}
+
+#[quickcheck]
+fn qc_test_transfer(initial_supply: Balance) -> bool {
+    let deposit_amount = ntoy(initial_supply);
```
fckt, line 33: How quickcheck generates values? You're doing `let deposit_amount = ntoy(initial_supply);` for some reason that's why I'm asking..

willemneal: It seems to start low. You can add `info!("{}",..)` and the `RUST_LOG=info cargo test`

fckt: But why it can't overflow? I mean `ntoy(initial_supply)`

fckt: Btw, you can do `cargo test -- --nocapture` to see output of `println!` or `dbg!()` for debug. No need for RUST_LOG..

examples/fungible-token/rust-toolchain
willemneal: Yes the issue was that the test won't build without nightly, so if it's ok for all the tests to be tested with nightly I can push or I just need to have a special case for this directory in `test_all.sh`.

examples/fungible-token/tests/general.rs
willemneal: Ah good to know.  The other reason for using info was because that's what quickcheck uses and you can get more info from it.

examples/fungible-token/rust-toolchain
fckt: Oh, actually I was thinking that rust_toolchain in the root is `nightly-2020-03-19` (like in https://github.com/near/initial-contracts/blob/master/rust-toolchain). Looks like it's not a case. 

Having a local rust_toolchain with `nightly-2020-03-19` in the examples/fungible-token directory should be ok.

Issue #146, opened 2020-05-07
Implement ordered map for near sdk


nearmax:
We need to implement an ordered map for `near_sdk::collections::*`. Currently we only have an unordered map, see `near_sdk::collections::Map` for the reference https://github.com/near/near-sdk-rs/blob/master/near-sdk/src/collections/map.rs. Unlike `std::collections::*` this collections should be mapping elements onto our key-value storage, using `env::storage_*` operations. This is because this collection will be used to store large number of elements that we cannot entirely load into the smart contract memory.

You can choose which type of data-structure to implement, as long as it satisfies the following properties that we prioritize:
* Simplicity and cleanliness of the implementation -- the code needs to be as simple and self-descriptive as possible;
* Minimum number of `env::storage_*` operations -- each atomic operation should perform as few `env::storage_*` operations as possible, because they are expensive;
* Assuming there is `N` elements in the map, there each storage record should consume `O(1)` space.
* Tests should cover all corner cases;
* Average case complexity should be the same as the worst-case complexity, so skip-list or BST will not do;
* Search/insert/delete operations should have O(log n) complexity. It should also occupy O(N) records in the key-value storage.

Hint, consider the following data-structures: B-tree, red black tree.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 500.0 DAI (500.0 USD @ $1.0/DAI)  attached to it as part of the http://github.com/near fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/near/near-sdk-rs/146/4350).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/near/near-sdk-rs/146/4350).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $136,979.89 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 1 week, 5 days from now.
Please review their action plans below:


[...]

PR #154, opened 2020-05-16
TreeMap impl based on AVL-tree

sergey-melnychuk:
TreeMap based on AVL-tree

| Method | Complexity (worst case) | 
|---------|-----------------------|
| `get`/`contains_key` | O(1) // UnorderedMap lookup |
| `insert`/`remove` | O(log(N)) |
| `min`/`max` | O(log(N)) |
| `above`/`below` | O(log(N)) |
| `range` of K elements | O(Klog(N)) |

Closes https://github.com/near/near-sdk-rs/issues/146


sergey-melnychuk:
@nearmax , please check out the draft (please note it is a draft: I'll catch up with all TODOs) - does it look like what's expected? 

`Heap` incapsulates indices, thus `HeapMap` only maps keys to values, letting `Heap` do all housekeeping of min-heap. In final version `insert` & `remove` will be O(log(N)), and `iterator` will be O(Nlog(N)), as promised.

Half of CI checks fail (for `env: NEAR_RELEASE=false`) with error `HostError(GasLimitExceeded)`, am I doing something wrong? :)

nearmax:
Hey @sergey-melnychuk ! Thank you for working on it. Sorry for not reviewing it yesterday, I will review it today.

sergey-melnychuk:
@nearmax , no problem, PR is ready for review now.

near-sdk/src/collections/heap_map.rs
```diff
@@ -0,0 +1,183 @@
...
+    keys: Heap<K>,
+    _values: PhantomData<V>,
+}
+
+impl<K, V> HeapMap<K, V>
+    where
+        K: Ord + BorshSerialize + BorshDeserialize,
+        V: BorshSerialize + BorshDeserialize,
+{
+    pub fn new(id: Vec<u8>) -> Self {
```
nearmax, line 28: Please add method that exposes the largest key (the top element from the heap).

near-sdk/src/collections/heap_map.rs
```diff
@@ -0,0 +1,183 @@
...
+        opt
+    }
+
+    // Note: iterator mutates underlying indexes when sorts the keys.
+    pub fn keys<'a>(&'a mut self) -> impl Iterator<Item = K> + 'a {
+        self.keys.iter()
+    }
+
+    // Note: mutable borrow is required for iterator.
+    pub fn entries<'a>(&'a mut self) -> impl Iterator<Item = (K, V)> + 'a {
```
nearmax, line 74: Please rename to `sort_iter`

near-sdk/src/collections/heap_map.rs
```diff
@@ -0,0 +1,183 @@
...
+        let opt = self.get(key);
+        self.keys.insert(key);
+        env::storage_write(
+            &append_slice(&self.key_prefix, &serialize(key)),
+            &serialize(value));
+        opt
+    }
+
+    // Note: iterator mutates underlying indexes when sorts the keys.
+    pub fn keys<'a>(&'a mut self) -> impl Iterator<Item = K> + 'a {
```
nearmax, line 69: Please rename to `sort_keys`

near-sdk/src/collections/heap.rs
```diff
@@ -0,0 +1,392 @@
...
+        let idx = idx_opt.unwrap();
+        let last = self.elements.len();
+        swap(&mut self.elements, idx, last, &mut self.indices);
+        self.elements.pop_raw();
+        let n = self.len();
+        sink(&mut self.elements, idx, n, &mut self.indices);
+        self.indices.remove(&value);
+    }
+
+    pub fn iter<'a>(&'a mut self) -> impl Iterator<Item = T> + 'a {
```
nearmax, line 69: Please rename to `sort_iter`.

near-sdk/src/collections/heap.rs
```diff
@@ -0,0 +1,392 @@
...
+        sort(&mut self.elements, &mut self.indices);
+        self.elements.iter()
+    }
+}
+
```
nearmax, line 74: Please add implementation of `Default` trait like in `UnorderedMap`.

near-sdk/src/collections/heap.rs
```diff
@@ -0,0 +1,392 @@
...
+    pub fn len(&self) -> u64 {
+        self.elements.len()
+    }
+
+    pub fn clear(&mut self) {
+        self.elements.clear();
+        self.indices.clear();
+    }
+
+    pub fn at(&self, idx: u64) -> Option<T> {
```
nearmax, line 40: Please add docstrings explaining that index is 1-based. Same for `lookup`.

near-sdk/src/collections/heap_map.rs
```diff
@@ -0,0 +1,183 @@
...
+    pub fn insert(&mut self, key: &K, value: &V) -> Option<V> {
+        let opt = self.get(key);
+        self.keys.insert(key);
+        env::storage_write(
+            &append_slice(&self.key_prefix, &serialize(key)),
+            &serialize(value));
+        opt
+    }
+
+    // Note: iterator mutates underlying indexes when sorts the keys.
```
nearmax, line 68: Please use `///`

nearmax:
My follow up request was to implement BTree on the key-value storage. This issue does not close entirely #146 but for the sake of gitcoin management we can mark it as such. HeapMap does not solve the problem, since users want to iterate over the collection with O(log n) or amortized O(1) per step. However, I see value in Heap and HeapMap, because someone might find it useful to have access in the top-most element (e.g. for the cases when there are order books, but not when entire order book needs to be retrieved).

nearmax:
@sergey-melnychuk 
> Half of CI checks fail (for env: NEAR_RELEASE=false) with error HostError(GasLimitExceeded), am I doing something wrong? :)

Your code is fine. We broke our CI recently: https://github.com/near/near-sdk-rs/issues/156

SkidanovAlex:
`Heap::insert` doesn't check if the key is already present, and `HeapMap::insert` doesn't do this check either. Thus, if the same key is inserted twice, it will be twice in the `elements` and once in the `indices`. If it is then removed, it will be once in `elements` and zero times in `indices`. At this point looking up by key will report the key missing, but iterating will report it present, i.e. the map will be inconsistent.

nearmax:
@sergey-melnychuk We will be happy to pay out the bounty, given your code matches your proposal that we accepted. Unfortunately, we haven't realized at the time of acceptance that the heap is not a suitable data structure for the ordered map, and will not be merging the implementation in, seeking further to have a sorted map implementation.
Alternatively, we will also be happy to pay extra $250 ($750 total) if you want to implement the proper sorted map, based on AVL, RedBlack, or some other sort of self-balancing trees.
Also note that your Heap implementation was not covering a rather trivial corner case. Should you decide to implement the balanced tree, we will expect higher attention to details

sergey-melnychuk:
@nearmax , @SkidanovAlex , thank you for the review, I will look into it today in the evening.

This is the initial approach, prototype, so work only starts here, not ends :) I was trying to prototype quickly - to get feedback and start for next iteration(s).

To summarize:
- double-insertion bug - my bad, thanks for pointing out, I will fix it today (fix/test must be simple).
- requirements: now desired features are more or less defined, so I can address them:
  - O(Klog(N)) for first K keys: I believe it is possible, heap seems like a good fit.
  - `floor`/`ceil` queries - unexpected in current impl, thus O(Nlog(N)).
- O(Nlog(N)) for sorted view without the heap is totally valid point, any array/vector would do.

Next impl must fit these:
- `min`/`max` queries in O(log(N)) worst case
- `floor`/`ceil` (find closes key above/below) queries in O(log(N)) worst case
- `insert`/`remove`/`lookup` in O(log(N)) worst case
- iterate keys in sorted order in O(Nlog(N)) worst case

IMHO AVL is the best fit in performance/complexity tradeoff. Let me come up with a prototype of AVL based on Maps.

nearmax:
@sergey-melnychuk 

> IMHO AVL is the best fit in performance/complexity tradeoff. Let me come up with a prototype of AVL based on Maps.

Thank you, AVL would be the best. And sorry for the mess up.

near-sdk/src/collections/heap_map.rs
sergey-melnychuk: done

sergey-melnychuk:
@nearmax , nothing to be sorry about - it is much easier to clarify requirements in front of a prototype, than out of thin air :)

Added `TreeMap` skeleton, I believe in a few days I will make it ready for review. It looks promising, most operations seem to fit O(log(N)).

near-sdk/src/collections/heap.rs
```diff
@@ -0,0 +1,464 @@
...
+            heap.insert(&x);
+        }
+
+        let n = vec.len();
+        for (i, _) in vec.iter().enumerate() {
+            assert_eq!(heap.remove_max(),  Some((n - i) as u8));
+            assert_eq!(heap.len() as usize, n - 1 - i);
+        }
+
+        assert_eq!(heap.len(), 0);
```
SkidanovAlex, line 459: Add an extra remove_maxp to test removing max from an empty heap.

sergey-melnychuk: Nice catch, thanks.

sergey-melnychuk:
@nearmax , @SkidanovAlex PR is ready for review. Access/mutation takes O(log(N)) in worst case and iteration over K sorted elements takes O(Klog(N)) in worst case, as balance is restored in O(log(N)) after insertions and deletions. Now I see I should have started with AVL-tree from the beginning :) 

Though I was trying to keep code as clean as readable as possible, still added some comments to tricky parts: mostly related to removing a node, balancing and rotations.

Also HeapMap is removed, it was totally useless.

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1300 @@
...
+        loop {
+            match self.rgt.get(&at) {
+                Some(rgt) => at = rgt,
+                None => break
+            }
+        }
+        self.key.get(&at)
+    }
+
+    fn floor_at(&self, mut at: u64, key: &K) -> Option<K> {
```
SkidanovAlex, line 188: From the code it appears that `floor_at` returns the key that is strictly smaller or `None` if there's no key smaller, and `ceil_at` returns the key which is strictly larger than `K` and `None` if there's no key larger.

It is more common to have `lower_bound` and `upper_bound`, where `lower_bound` returns the smallest key that is greater or equal, and `upper_bound` returns the smallest key that is strictly greater.

The current `ceil_at` already behaves as `upper_bound`. Can we add `upper_bound` as an alias for `ceil_at` and `lower_bound` that matches the description above?

And then expose corresponding iterators (iterate from lower_bound, iterate from upper_bound)

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1300 @@
...
+            self.ht.remove(&r_id);
+
+            // removing node might have caused a imbalance - balance the tree up to the root,
+            // starting from lowest affected key - the parent of a leaf node in this case
+            self.check_balance(root, &p_key)
+
+        } else {
+            // non-leaf node, select subtree to proceed with
+            let b = self.get_balance(r_id);
+            if (b >= 0 && lft_opt.is_some()) || rgt_opt.is_none() {
```
SkidanovAlex, line 471: It should be enough to check `b >= 0` right? If we know that at least one child is set, and the left tree is no smaller than the right tree, than the left tree must exist.
Similarly, if `b < 0` rgt_opt will never be `None`

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1300 @@
...
+                // proceed with left subtree
+                let lft = lft_opt.unwrap();
+
+                // k - max key from left subtree
+                let k = self.max_at(lft).unwrap();
+
+                // n - id of a node that holds key k, p - id of immediate parent of n
+                let (n, p) = self.lookup_at(r_id, &k).unwrap();
```
SkidanovAlex, line 479: We do two lookups here, that will traverse the exact same path, could it be merged into one lookup?

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1300 @@
...
+                        Some(l) => {
+                            self.rgt.insert(&p, &l);
+                            self.lft.remove(&n);
+                        },
+                        None => {
+                            self.rgt.remove(&p);
+                        }
+                    };
+                } else {
+                    // n is on left link of p
```
SkidanovAlex, line 497: `n` corresponds to the largest element in the subtree of `lft`. For it to be on the left link of its parent, it's parent can't be in the subtree of `lft` (otherwise it's parent is larger than `n` but `n` is the largest). Thus, the parent of `n` is `r_id` itself. Thus, below it must be that `p == r_id`. As such, the `self.lft.insert(p, l)` below is a no-op, and should be removed.

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1300 @@
...
+                        }
+                    }
+                }
+
+                // removing node might have caused a imbalance - balance the tree up to the root,
+                // starting from the lowest affected key (min key from right subtree in this case)
+                self.check_balance(root, &k)
+
+            } else {
+                root
```
SkidanovAlex, line 556: This must be unreachable, right?
And then the previous `else if` can become `else`

SkidanovAlex:
I didn't check all the tests, but we definitely need a test that runs many random operations (inserts, deletes) against the AVL tree, and the vanilla Rust BTree, and then checks that the contents of the two trees match, and so does the order of them.

Also, if I understand the code correctly, when you insert you assume that all the existing node ordinals are contiguous (you use `len` as the next ID, presuming that it's available). However, in the code of `do_remove` it doesn't appear that you actually make any effort not to create a hole in the ids. You replace the key of `r_id` with the key of `n`, but `n` is not the largest ID, it's the ID of one of the two adjacent elements in the tree. Probably need to make `do_remove` return the ID of the node removed, and then swap it with the `len - 1`.

near-sdk/src/collections/tree_map.rs
sergey-melnychuk: done

near-sdk/src/collections/tree_map.rs
sergey-melnychuk: done

near-sdk/src/collections/tree_map.rs
sergey-melnychuk: done

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1300 @@
...
+        V: Copy + BorshSerialize + BorshDeserialize,
+{
+    pub fn new(id: Vec<u8>) -> Self {
+        let h_prefix = append(&id, b'h');
+        let l_prefix = append(&id, b'l');
+        let r_prefix = append(&id, b'r');
+        let k_prefix = append(&id, b'k');
+        let v_prefix = append(&id, b'v');
+
+        let root: u64 = env::storage_read(&id)
```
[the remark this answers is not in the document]
sergey-melnychuk: done

near-sdk/src/collections/mod.rs
```diff
@@ -58,3 +68,27 @@ pub(crate) fn next_trie_id() -> Vec<u8> {
...
+
+pub(crate) fn append(id: &[u8], chr: u8) -> Vec<u8> {
+    let mut result = Vec::with_capacity(id.len() + 1);
+    result.extend(id);
+    result.push(chr);
+    result
+}
+
+pub(crate) fn append_slice(id: &[u8], extra: &[u8]) -> Vec<u8> {
+    let mut result = Vec::with_capacity(id.len() + extra.len());
```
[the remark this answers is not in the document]
sergey-melnychuk: done

near-sdk/src/collections/tree_map.rs
sergey-melnychuk: ack

sergey-melnychuk:
@SkidanovAlex , thank you for review.

> (you use len as the next ID, presuming that it's available)

This is interesting catch, I'll try to come up with minimum test to reproduce, and make some property-based tests against `RBTree` or `Map` + sorting.

near-sdk/src/collections/tree_map.rs
sergey-melnychuk: done

sergey-melnychuk:
Approach to address "use len as the next ID":

1. Increasing only `ops: u64` counter as a source of new ID for inserted node. Is only incremented when node is inserted. IMHO, not the most elegant, but with least housekeeping overhead. Problem is possible overflow when `ops` is getting close to `std::u64::MAX`. 

I doubt that such limit can ever be reached in practice (`std::u64::MAX` ~= 10^18): quick math shows that inserting an item *every nanosecond* starting now avoids overflow for next 500+ years. Yet this approach still brings a limit to keep in mind.

2. After node `n` is removed, swap node `len-1` (containing highest id) with `n`, as suggested by @SkidanovAlex . This would require updating at most 4 links: parents left or right, left, right and key. Yet to efficiently perform such update, a reverse cache `rev: Map<u64, K>` is required to avoid `O(N)` lookup of a key for node `len-1`.

This would still fit in required O(N) space and done properly allows not to keep in mind any limits.

I'll go with option 2.

sergey-melnychuk:
@SkidanovAlex , PTAL, all issues are resolved.

nearmax:
Thank you @sergey-melnychuk ! We will review it soon.

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1637 @@
...
+
+        assert!(map.height() <= max_tree_height(n));
+        map.clear();
+    }
+
+    #[test]
+    fn insert_n_random() {
+        test_env::setup();
+
+        for k in 1..5 {
```
SkidanovAlex, line 952: 5 => 500? 5 is just not serious :)

SkidanovAlex:
- Do we have a fuzzy test that executes many random operations and compares TreeMap against BTreeMap (including order)?

- Also, given BTreeMap has `Range`, would be great to test some random ranges in such a test as well.

near-sdk/src/collections/tree_map.rs
sergey-melnychuk: I can't agree more! Unfortunately, even with `1..6` results in `HostError(GasLimitExceeded)`. Any suggestions how to address it?

sergey-melnychuk:
> * Do we have a fuzzy test that executes many random operations and compares TreeMap against BTreeMap (including order)?
> * Also, given BTreeMap has `Range`, would be great to test some random ranges in such a test as well.

Property-based test against `std::collection::BTreeMap`:
https://github.com/near/near-sdk-rs/pull/154/files#diff-5cc570a79aec0b36a7062792ee6b3069R1605

~~Sky~~ `HostError(GasLimitExceeded)` is the limit :)

near-sdk/src/collections/tree_map.rs
nearmax: When we create `MockedBlockchain` in `test_env::setup()` we pass `Default::default()` as a `VMConfig`. In this specific test instead of calling `test_env::setup();` we can unwrap that code and initialize `VMConfig` with `limit_config::max_gas_burnt` equal to `u64::MAX`.

Also, @SkidanovAlex `5` is not the number of items we insert in the tree here, `let n = 1 << k;` is.

sergey-melnychuk:
Update:
- added method to setup test env with unlimited gas
- increased upper bound of tree size in `test_remove_n_random` to 1024
- added property-based test for range queries against BTreeMap
  - covered all 4 combination on inclusive/exclusive range bounds

Tests pass, tree seems to work properly :) Please let me know where I can improve it more!

nearmax:
@mikhailOK is currently reviewing it.

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1681 @@
...
+        self.iter().collect()
+    }
+
+    //
+    // Internal utilities
+    //
+
+    fn set_root(&mut self, root: u64) {
+        env::storage_write(&self.tree_prefix, &serialize(&root));
+        self.root = root;
```
mikhailOK, line 184: we shouldn't have to `storage_write`: `Vector` doesn't do it with `len` field. CC @nearmax

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1681 @@
...
+        let remove = self.len - 1;
+        if target == remove {
+            return self.root;
+        }
+
+        let k_opt = self.key.get(&remove);
+        if k_opt.is_none() {
+            return self.root;
+        }
+        let key = k_opt.unwrap();
```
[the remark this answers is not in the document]
sergey-melnychuk: Suggested code won't compile. If there is no key to remove, no call to `do_remove` (and thus to `swap_with_last`) is made - see `remove` impl.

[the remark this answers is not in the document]
sergey-melnychuk: Oh, I see now - applying the suggestion.

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1681 @@
...
+        }
+
+        // move references from `ht`, `lft`, `rgt` from `r` to `t`.
+        for map in [&mut self.ht, &mut self.lft, &mut self.rgt].iter_mut() {
+            let val = map.remove(&remove);
+            match val {
+                Some(x) => {
+                    map.insert(&target, &x);
+                },
+                None => ()
```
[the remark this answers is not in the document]
sergey-melnychuk: In fact we do, thanks.

near-sdk/src/collections/tree_map.rs
sergey-melnychuk: `root` is rather important property of a tree, not saving it would require special kind of lookup (most likely in O(N)) to restore it for an existing tree.

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1681 @@
...
+use crate::collections::UnorderedMap;
+use crate::env;
+
+/// AVL tree implementation
+///
+/// Runtime complexity (N = number of entries):
+/// - `lookup`/`insert`/`remove`: O(log(N)) worst case
+/// - `min`/`max`: O(log(N)) worst case
+/// - `floor`/`ceil` (find closes key above/below): O(log(N)) worst case
+/// - iterate K elements in sorted order: O(Klog(N)) worst case
```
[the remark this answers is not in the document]
sergey-melnychuk: Please elaborate, I assume O(log(N)) is the lookup of the `min`. How exactly AVL-tree can be traversed in sorted order in O(K)?

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1681 @@
...
+
+    #[test]
+    fn prop_avl_vs_rb() {
+        fn prop(insert: Vec<(u32, u32)>, remove: Vec<u32>) -> bool {
+            let a = avl(&insert, &remove);
+            let b = rb(&insert, &remove);
+            let v1: Vec<(u32, u32)> = a.iter().collect();
+            let v2: Vec<(u32, u32)> = b.into_iter().collect();
+            v1 == v2
+        }
```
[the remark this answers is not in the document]
sergey-melnychuk: ack

near-sdk/src/collections/heap.rs
```diff
@@ -0,0 +1,477 @@
...
+        let n = self.len();
+        swap(&mut self.elements, 1, n, &mut self.indices);
+        sink(&mut self.elements, 1, n - 1, &mut self.indices);
+        self.elements.pop();
+        max
+    }
+
+    pub fn lookup(&self, value: &T) -> Option<u64> {
+        self.indices.get(value)
+    }
```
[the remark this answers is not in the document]
sergey-melnychuk: removed `lookup`, it was not used really

near-sdk/src/collections/heap.rs
```diff
@@ -0,0 +1,477 @@
...
+{
+    // Sort vector in-place in O(Nlog(N))
+    let n = vec.len();
+    let mut k = n;
+    while k > 1 {
+        swap(vec, 1, k, indices);
+        k -= 1;
+        sink(vec, 1, k, indices);
+    }
+}
```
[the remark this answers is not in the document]
sergey-melnychuk: replaced `sort` (which IMHO must not be included in public API of max heap) with `into_iter`

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1681 @@
...
+#[derive(BorshSerialize, BorshDeserialize)]
+pub struct TreeMap<K, V> {
+    tree_prefix: Vec<u8>,
+
+    len: u64,
+    root: u64,                      // ID of a root node of the tree
+    ht: UnorderedMap<u64, u64>,     // height of a subtree at a node
+    lft: UnorderedMap<u64, u64>,    // left link of a node
+    rgt: UnorderedMap<u64, u64>,    // right link of a node
+    key: UnorderedMap<u64, K>,      // key stored in a node
```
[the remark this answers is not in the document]
sergey-melnychuk: in such case, `nodes: Vector<Node<K>>` instead of 4 maps looks more suitable - potentially, number of lookups per node can be reduced by 4.

near-sdk/src/collections/tree_map.rs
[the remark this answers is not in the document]
sergey-melnychuk: Answering my own question: with `nodes: Vector<Node<K>>` (as suggested in comment above), such iteration can be done in exactly O(log(N) + K), as `nodes` will always be sorted.

Yet insertions will degrade to O(N) worst case.

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1681 @@
...
+        self.min_at(self.root, self.root).map(|(k, _, _)| k)
+    }
+
+    pub fn max(&self) -> Option<K> {
+        self.max_at(self.root, self.root).map(|(k, _, _)| k)
+    }
+
+    pub fn floor(&self, key: &K) -> Option<K> {
+        self.floor_at(self.root, key)
+    }
```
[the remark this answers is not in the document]
sergey-melnychuk: ack

near-sdk/src/collections/tree_map.rs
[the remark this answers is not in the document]
mikhailOK: near_sdk Vector is pretty much an UnorderedMap with indices as keys, there is no resizing

near-sdk/src/collections/tree_map.rs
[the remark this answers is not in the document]
mikhailOK: lookup the `min` and simply traverse the edges. Since we don't have parent pointer, we can store the stack of parents.
O(K + log(N)) nodes visited

near-sdk/src/collections/tree_map.rs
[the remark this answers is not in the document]
mikhailOK: How come the tests didn't catch that?
Does `prop_avl_vs_rb` catch it if we do deletions between insertions?

near-sdk/src/collections/tree_map.rs
mikhailOK: It's a bit confusing:
- simple classes don't have to handle their own storage they are stored by the owner (some key determined by the owner -> serialized object)
- near_sdk collections manage their storage and their serialization doesn't include elements
- they still have fields like `len` in Vector, and rely on those fields to be stored/read by the owner serializing the collection: `Vector::new(id)` sets `len=0`, so it can only be used for creating a new vector as opposed to providing access to an existing collection prefixed by `id`
- `TreeMap::new` calls `UnorderedMap::new` which in turn calls `Vector::new`. So we can't use `TreeMap::new` to access an existing map: even with reading `root` the `len` field is broken.

sergey-melnychuk: makes sense, removing store/load of root id

near-sdk/src/collections/tree_map.rs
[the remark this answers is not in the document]
sergey-melnychuk: I don't know why I got confused - this would be plain in-order traversal, O(log(N) + K) is doable

near-sdk/src/collections/tree_map.rs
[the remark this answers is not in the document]
sergey-melnychuk: Good question, seems like this is rather rare case that requires very specific input to be caught. I will try to come up with such test that must fail without `map.remove(&target)`.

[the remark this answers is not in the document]
mikhailOK: We should improve the random test until it can catch this type of bugs:
- how does quickcheck decide the range of values? if it uses whole u32 then all insertions are distinct and deletions are no-op, need meaningful values
- how does it decide the ratio of insertions of deletions? with too many deletions the map is always empty, with almost all insertions some other case may be missed
- makes sense to test lookups between those operations

near-sdk/src/collections/tree_map.rs
sergey-melnychuk: done

near-sdk/src/collections/tree_map.rs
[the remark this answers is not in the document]
sergey-melnychuk: done

near-sdk/src/collections/tree_map.rs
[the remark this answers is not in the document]
sergey-melnychuk: done

near-sdk/src/collections/tree_map.rs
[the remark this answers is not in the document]
sergey-melnychuk: done

near-sdk/src/collections/tree_map.rs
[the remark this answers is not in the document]
sergey-melnychuk: Now I see that that test do fine, there is just no bug to catch: `remove` is unnecessary - as `target`'s height and links (max 1 link) are reset in `do_remove` before `swap_with_last` is called.

Adding comment. Sorry for confusion.

near-sdk/src/collections/tree_map.rs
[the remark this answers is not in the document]
sergey-melnychuk: done

nearmax:
@mikhailOK Is this collection good to go, can we merge it?

near-sdk/src/collections/tree_map.rs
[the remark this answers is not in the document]
sergey-melnychuk: While it is possible to iterate over full tree in O(N) with in-order traversal, I don't think it holds for any range of K elements. For a range, finding successor node remains O(log(N)) worst case, example: range of [max(root.left).value, root.value, min(root.right).value].

[wiki](https://en.wikipedia.org/wiki/AVL_tree#Traversal)

Leaving iterator as it is for now with O(Klog(N)) worst case. In separate PR I can introduce full tree asc/desc iterators that have O(N) complexity in worst case.

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1772 @@
...
+            self.tree.replace(node.id, node);
+        } else {
+            self.tree.push(node);
+        }
+    }
+
+    /// Return height of the tree - number of nodes on the longest path starting from the root node.
+    pub fn height(&self) -> u64 {
+        self.node(self.root).map(|n| n.ht).unwrap_or_default()
+    }
```
mikhailOK, line 96: This should be private I think in case we want to switch from AVL

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1772 @@
...
+        self.above_at(self.root, key)
+    }
+
+    /// Returns the largest key that is strictly less than key given as the parameter
+    pub fn below(&self, key: &K) -> Option<K> {
+        self.below_at(self.root, key)
+    }
+
+    /// Returns the smallest key that is greater or equal to key given as the parameter
+    pub fn floor(&self, key: &K) -> Option<K> {
```
mikhailOK, line 144: this is still backwards compared to java TreeMap `floorKey`, `ceilingKey`.
Also one of the doc comments should be `less or equal`

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1772 @@
...
+        self.min_at(self.root, self.root).map(|(n, _)| n.key)
+    }
+
+    /// Returns the largest stored key from the tree
+    pub fn max(&self) -> Option<K> {
+        self.max_at(self.root, self.root).map(|(n, _)| n.key)
+    }
+
+    /// Returns the smallest key that is strictly greater than key given as the parameter
+    pub fn above(&self, key: &K) -> Option<K> {
```
mikhailOK, line 134: Maybe `above,below` -> `higher,lower`?

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1772 @@
...
+    }
+
+    #[test]
+    fn prop_avl_height() {
+        test_env::setup_free();
+
+        fn prop(insert: Vec<(u32, u32)>, remove: Vec<u32>) -> bool {
+            let map = avl(&insert, &remove);
+            map.height() <= max_tree_height(map.len())
+        }
```
mikhailOK, line 1703: I add this check to this test:
```    
    fn check_all_balances<K, V>(map: &TreeMap<K, V>, root: u64)
    where
        K: Ord + Copy + BorshSerialize + BorshDeserialize,
        V: Copy + BorshSerialize + BorshDeserialize,
    {
        if let Some(node) = map.node(root) {
            let balance = map.get_balance(&node);
            assert!(-1 <= balance && balance <= 1, "BAD BALANCE: balance {}, node id={}, ht={}, l={:?}, r={:?}", balance, node.id, node.ht, node.lft, node.rgt);
            if let Some(id) = node.lft {
                map.check_all_balances(id);
            }
            if let Some(id) = node.rgt {
                map.check_all_balances(id);
            }
        }
    }

    #[test]
    fn prop_avl_height() {
        test_env::setup_free();

        fn prop(insert: Vec<(u32, u32)>, remove: Vec<u32>) -> bool {
            let map = avl(&insert, &remove);
            check_all_balances(&map, map.root);
            map.height() <= max_tree_height(map.len())
        }

        QuickCheck::new()
            .tests(300)
            .quickcheck(prop as fn(std::vec::Vec<(u32, u32)>, std::vec::Vec<u32>) -> bool);
    }
```
and it fails, is my assert wrong?

near-sdk/src/collections/tree_map.rs
sergey-melnychuk: done

near-sdk/src/collections/tree_map.rs
sergey-melnychuk: [Java](https://docs.oracle.com/javase/8/docs/api/java/util/TreeMap.html)

```
ceilingKey(K key)
Returns the least key greater than or equal to the given key, or null if there is no such key.

floorKey(K key)
Returns the greatest key less than or equal to the given key, or null if there is no such key.
```

Updating existing to:

```
/// Returns the smallest key that is greater or equal to key given as the parameter
    pub fn ceil_key(&self, key: &K) -> Option<K>

/// Returns the largest key that is less or equal to key given as the parameter
    pub fn floor_key(&self, key: &K) -> Option<K>
```

near-sdk/src/collections/tree_map.rs
sergey-melnychuk: done

near-sdk/src/collections/tree_map.rs
```diff
@@ -0,0 +1,1776 @@
...
+    fn default() -> Self {
+        Self::new(next_trie_id())
+    }
+}
+
+
+impl<K, V> TreeMap<K, V>
+    where
+        K: Ord + Copy + BorshSerialize + BorshDeserialize,
+        V: Copy + BorshSerialize + BorshDeserialize,
```
mikhailOK, line 61: I think `V` doesn't need `Copy` at all
And for `K` it makes sense to remove `Copy` and require `Clone` or change some interfaces to `&K`

near-sdk/src/collections/tree_map.rs
sergey-melnychuk: looking into it

sergey-melnychuk: Assertion `-1 <= balance <= 1` describes perfectly balanced tree (at each level), height of such tree would always be `log(N+1)`. So assertion is not wrong, just too strict for AVL tree.

AVL gives no guarantees of being perfectly balanced at each level after each insert/remove, thus `log(N+1)` is a lower bound of tree height, and upper bound is: `C * log2(n + D) + B` according to [wiki](https://en.wikipedia.org/wiki/AVL_tree).

mikhailOK: Quote from wiki:
```
of its two child sub-trees. A binary tree is defined to be an AVL tree if the invariant

    BalanceFactor ( n o d e ) ∈ { − 1 , 0 , 1 } {\displaystyle {\text{BalanceFactor}}(node)\in {\{-1,0,1\}}} {\displaystyle {\text{BalanceFactor}}(node)\in {\{-1,0,1\}}}[7]

holds for every n o d e {\displaystyle node} {\displaystyle node} in the tree. 
```

sergey-melnychuk: still looking into it then

sergey-melnychuk: `remove` indeed does not restore AVL-tree invariants, investigating further

near-sdk/src/collections/tree_map.rs
sergey-melnychuk: ack

near-sdk/src/collections/tree_map.rs
sergey-melnychuk: Thank you, @mikhailOK ! I was wrong relying only on tree height to validate balance, as well as about definition of a perfectly balanced tree (it is only perfectly balanced when balance == 0 in all nodes), and also about AVL-tree fundamental property - balance among all subtrees.

Fixed now, `prop_avl_balance` passes.

near-sdk/src/collections/tree_map.rs
sergey-melnychuk: PTAL, now `K: Clone + Ord + <ser/de>` and `V: <ser/de>` only.
I was trying to keep `tree` as `Vector<Node<&K>>`, but lost this fight to borrow checker, thus put a few `.clone()`s when moving keys around.

nearmax:
@mikhailOK said he found a test that surfaces some bug in the implementation. @mikhailOK could you please provide it here?

nearmax:
@SkidanovAlex @mikhailOK LGTMed this PR. Is there anything from your side you want to comment on?

SkidanovAlex:
The AVL tree looks good to me with all the new tests.

Can we remove Heap and the HeapMap? AVL tree is a strict superset of functionality.

Let's also get LGTM from Mikhail.

nearmax:
> The AVL tree looks good to me with all the new tests.
> 
> Can we remove Heap and the HeapMap? AVL tree is a strict superset of functionality.
> 
> Let's also get LGTM from Mikhail.

AFAIU heap has smaller constant in O(log N) than AVL, so it is beneficial to keep around. @mikhailOK already LGTMed, see above.

mikhailOK:
Heap is kind of weird right now, it's a priority queue but that also doesn't allow duplicate values. Make sense to remove it or make private until we decide what we want from it.

nearmax:
@sergey-melnychuk Could you remove the heap? After that we will merge and I will release the bounty.

sergey-melnychuk:
@nearmax , heap is removed - I'll submit separate PR with priority queue, then you can decide if it is useful for you.

nearmax:
@SkidanovAlex could you change your review to "approved" so that I don't force merge?

nearmax:
@sergey-melnychuk Thank you very much for spending your time on implementing this collection together with us! I have paid out the bounty, please let us know if there any issues with payments. We are very satisfied with the result and hoping for the future collaboration together with you.

sergey-melnychuk:
@nearmax , Thank you, nice to know! Please feel free to ping me directly on GitHub if you need my help. 

As promised, submitting a PR with Priority Queue: https://github.com/near/near-sdk-rs/pull/170

PR #161, opened 2020-05-22
Update near-sdk-rs to 0.9.3

bowenwang1996:


evgenykuzyakov:
It has to be `0.10` because it has methods that might not be available on every chain.

bowenwang1996:
@evgenykuzyakov fixed

evgenykuzyakov:
@bowenwang1996 Can you also run `./build_all.sh && ./check_all.sh && ./test_all.sh` in the examples, please before committing. It rebuilds and tests all examples.

Issue #165, opened 2020-05-28
Add Debug for Base58PublicKey


ilblackdragon:
Otherwise can't use it or structs that contain it in assets.

PR #166, opened 2020-05-28
Add Debug to Base58PublicKey and add ValidAccountId

evgenykuzyakov:
Fixes: #165


near-sdk/src/json_types/account.rs
```diff
@@ -0,0 +1,84 @@
...
+use crate::AccountId;
+
+/// Helper class to validate account ID during serialization and deserializiation
+#[derive(
+    Debug, Clone, PartialEq, PartialOrd, Ord, Eq, BorshDeserialize, BorshSerialize, Serialize,
+)]
+pub struct ValidAccountId(pub AccountId);
+
+impl ValidAccountId {
+    pub fn is_valid(&self) -> bool {
```
bowenwang1996, line 15: this doesn't make sense. So something of type `ValidAccountId` can still be invalid?

evgenykuzyakov: I guess I can make it private. In theory I can make inner field private, which will hide the constructor, so you can only construct it from the `TryFrom`.

bowenwang1996: >  In theory I can make inner field private, which will hide the constructor, so you can only construct it from the TryFrom.

Sounds good.

nearmax:
FYI In the past @frol raised an important suggestion that we should make our structures self-validating. E.g. we should not be able to construct an invalid instance of `AccountId` in nearcore.

near-sdk/src/json_types/account.rs
```diff
@@ -0,0 +1,92 @@
...
+impl ValidAccountId {
+    fn is_valid(&self) -> bool {
+        is_valid_account_id(&self.0.as_bytes())
+    }
+
+    pub fn account_id(&self) -> &AccountId {
+        &self.0
+    }
+
+    pub fn into_account_id(self) -> AccountId {
```
nearmax, line 23: Seems like `impl From<ValidAccountId> for AccountId` already implements this functionality.

evgenykuzyakov: Ok. Should we rename `account_id()` into `.as_ref()`?

nearmax:
Is this PR ready to submit @evgenykuzyakov

evgenykuzyakov:
Yes, will submit after #200  Want to make sure

Issue #122, opened 2020-03-31
Support "#[self_call]" for calls 


evgenykuzyakov:
It should check whether the current account equals to the predecessor account.
```
    assert_eq!(env::current_account_id(), env::predecessor_account_id());
``` 

Most of the time when you implement a callback function, you have to lock it for calls from the contract itself. This method assert check is very common and probably should be supported by the SDK

evgenykuzyakov:
Alternative name can be `#[private]`

fckt:
Makes a lot of sense! I'm for `#[private]`

PR #168, opened 2020-05-29
feat: #[private] attribute

chain-police:
fixes https://github.com/near/near-sdk-rs/issues/122


olonho:
@evgenykuzyakov shall we abandon that one?

evgenykuzyakov:
Yes

PR #170, opened 2020-06-02
[Archive] PriorityQueue based on Max-Heap

sergey-melnychuk:
Priority Queue based on Max-Heap.

Iterator consumes the queue and returns all elements in descending order.

Runtime complexity (worst case):
- `offer`:  O(log(N))
- `poll`:   O(log(N))
- `peek`:   O(1)

("leftover" from https://github.com/near/near-sdk-rs/pull/154)


evgenykuzyakov:
archived

PR #171, opened 2020-06-03
[Archive] Proposed implementation of the Red-Black tree for an ordered map

nearmax:
One of our external contributors proposed an alternative to our AVL-based ordered map implementation using red-black trees. Quote from them:
> The implementation is based on this pseudocode for a left leaning rbtree http://opendatastructures.org/ods-python/9_2_RedBlackTree_Simulated_.html it's apparently the same reference that was used for Rust std's BTree implementation http://cglab.ca/~abeinges/blah/rust-btree-case/

We need to do the following work items in the following order:
- [ ] Identify if there is any value in the proposed implementation, if there is some then give some reward. We don't need to perform the detailed code review for it;
- [ ] Whether this implementation is better than AVL based implementation;
- [ ] If it is strictly better than replace AVL based implementation with it. Otherwise keep this code as an archive.


evgenykuzyakov:
archived

PR #177, opened 2020-06-04
Miscellaneous comments

evgenykuzyakov:
Addresses fungible-token comments from https://github.com/near/core-contracts/issues/58


examples/fungible-token/src/lib.rs
```diff
@@ -5,11 +5,15 @@
 *  - JSON calls should pass U128 as a base-10 string. E.g. "100".
 *  - The contract optimizes the inner trie structure by hashing account IDs. It will prevent some
 *    abuse of deep tries. Shouldn't be an issue, once NEAR clients implement full hashing of keys.
-*  - This contract doesn't optimize the amount of storage, since any account can create unlimited
-*    amount of allowances to other accounts. It's unclear how to address this issue unless, this
-*    contract limits the total number of different allowances possible at the same time.
-*    And even if it limits the total number, it's still possible to transfer small amounts to
-*    multiple accounts.
+*  - The contract tracks storage difference before and after the call. If the storage increases,
```
willemneal, line 8:
```suggestion
*  - The contract tracks the change in storage before and after the call. If the storage increases,
```

examples/fungible-token/src/lib.rs
```diff
@@ -5,11 +5,15 @@
...
+*    the contract requires the caller of the contract to attach enough deposit to the function call
+*    to cover the storage stake difference.
```
willemneal, line 10: "the storage stake difference"?

examples/fungible-token/src/lib.rs
```diff
@@ -5,11 +5,15 @@
...
+*    It's done to prevent denial of service attack on the contract by taking all available storage.
```
willemneal, line 11:
```suggestion
*    This is done to prevent a denial of service attack on the contract by taking all available storage.
```

examples/fungible-token/src/lib.rs
```diff
@@ -5,11 +5,15 @@
...
+*    If the storage decreases, the contract will issue a refund for the storage stake difference.
+*    The unused tokens from the attached deposit are also going to be refunded, so it's safe to
```
willemneal, line 13:
```suggestion
*    The unused tokens from the attached deposit are also refunded, so it's safe to
```

examples/fungible-token/src/lib.rs
```diff
@@ -5,11 +5,15 @@
...
+*    attach more deposit than required.
+*  - The deployed contract has to be locked. It means it should not have any access keys on the
```
willemneal, line 15: Since we have a `lockup` contract, `locked` might confuse people.  Perhaps rewrite the whole two lines like.
- To prevent the deployed contract from being modified or deleted, it should not have any access keys on its account.

examples/fungible-token/src/lib.rs
evgenykuzyakov: replaced to "the storage fee difference"

examples/fungible-token/src/lib.rs
```diff
@@ -5,15 +5,15 @@
 *  - JSON calls should pass U128 as a base-10 string. E.g. "100".
 *  - The contract optimizes the inner trie structure by hashing account IDs. It will prevent some
 *    abuse of deep tries. Shouldn't be an issue, once NEAR clients implement full hashing of keys.
-*  - The contract tracks storage difference before and after the call. If the storage increases,
+*  - The contract tracks the change in storage before and after the call. If the storage increases,
 *    the contract requires the caller of the contract to attach enough deposit to the function call
 *    to cover the storage stake difference.
```
willemneal, line 10:
```suggestion
*    to cover the storage cost.
```

evgenykuzyakov: If the storage decreases, the contract will issue a refund for the cost of the released storage.

Issue #95, opened 2020-02-22
Current `#[init]` doesn't allow to check whether the contract is already initialized.


evgenykuzyakov:
We should support `#[init_once]` that would check that the current contract state is empty, otherwise you have to rely on the manual check from the state. Something like this:
```rust
assert!(near_bindgen::env::state_read::<EthBridge>().is_none(), "Already initialized");
```

PR #178, opened 2020-06-04
Add best practices

evgenykuzyakov:


mikedotexe:
Would love to see guidance in terms of `panic!` vs `env::panic`

HELP.md
```diff
@@ -0,0 +1,109 @@
+# Contract best practices
+
+## Init once
+
+Make sure you check that state doesn't exists in `init` using
+
+```rust
+assert!(!env::state_exists(), "The contract is already initialized");
```
willemneal, line 8: Should this become an argument on the `init` marco? e.g. `init(once)`?

evgenykuzyakov: Or `#[init_once]` https://github.com/near/near-sdk-rs/issues/95

HELP.md
```diff
@@ -0,0 +1,109 @@
...
+NEAR currently expects contracts to support JSON serialization. JSON can't handle large integers (above 2**53 bits).
+That's why you should use helper classes from the `json_types` in `near_sdk` for `u64` and `u128`.
+We provide types `U64` and `U128`, that wraps the integer into a struct and implements JSON serialization and
+deserialization as a base-10 strings.
+
+E.g.
+
+```rust
+pub fn add(&self, a: U128, b: U128) -> U128 {
+    (a.0 + b.0).into()
```
willemneal, line 55: I found this to be annoying.  Is there really not a way to handle this with the bindgen?

evgenykuzyakov: Not easily. We don't want to make it a requirement to serialize long numbers as strings. Hopefully JSON standard can improve in the future and we can use regular long numbers with string quotes.

HELP.md
```diff
@@ -0,0 +1,109 @@
...
+}
+```
+
+
+## View vs Change method
+
+By default, `near_sdk` assumes that the method is a `view` if it has `&self` and method is `change` if it has `&mut self`.
+View methods usually don't have ability to change state and have limited context.
+Change methods will save the modified state back to the contract.
```
willemneal, line 64: I would also mention that you can modify storage the contract's state.

HELP.md
evgenykuzyakov: We can implement math on `U128`, so you'll be able to do simply `a + b`

willemneal: Yeah that's a good idea.

HELP.md
```diff
@@ -0,0 +1,109 @@
...
+Try to validate the input, context, state and access first before making any actions. This will safe gas for the caller to fail earlier.
+
+
+## Use `env::log`
+
+Use logging for debugging and notifying user.
+When you need a formatted message, you can use the following:
+
+```rust
+env::log(format!("Transferred {} tokens from {} to {}", amount, sender_id, receiver_id).as_bytes());
```
willemneal, line 94: I've always found this API strange.  Why not just make log a macro so that you don't have to `.as_bytes` every time?

HELP.md
```diff
@@ -0,0 +1,109 @@
...
+
+If your method makes a cross-contract call, you may want to return the `Promise` that the contract creates.
+It allows the caller to wait for the result of the promise instead of the returning the result immediately.
+Because if the promise fails for some reason, the caller will not know about this or the caller may display the result earlier.
+
+E.g.
+
+```rust
+pub fn withdraw(&mut self, amount: U128) -> Promise {
+    Promise::new(self.owner_id.clone()).transfer(amount.0)
```
willemneal, line 107: Perhaps we should also change the promise API so that it uses `U128`

HELP.md
evgenykuzyakov: Good idea, please add an issue for this.

HELP.md
evgenykuzyakov: No, it's good for now. I still have a hope that we'll use borsh instead of JSON. Inner numbers don't have to be wrapped in `U128`

willemneal: Do you mean use borsh for input serialization too?

HELP.md
```diff
@@ -0,0 +1,716 @@
...
+The prefix is used to generate internal keys to store data in the persistent storage.
+These internal keys has to be unique to avoid collisions (even with key `STATE`).
+
+## Generating unique prefixes for persistent collections
+
+When a contract gets complicated, there may be multiple different
+collections that may not be all part of the main structure, but instead be part of sub-structure or nested collections.
+They all need to have unique prefixes.
+
+#### The traditional way
```
mattlockyer, line 42: should we make this a small stub and link to appendix of doc just to keep the flow and emphasis on "the new way"?

HELP.md
```diff
@@ -0,0 +1,716 @@
...
+            // This is needed to differentiate across accounts.
+            prefix.extend(env::sha256(account_id.as_bytes()));
+            UnorderedSet::new(prefix)
+        });
+        tokens
+    }
+}
+```
+
+#### The new way
```
mattlockyer, line 75: like

HELP.md
```diff
@@ -0,0 +1,717 @@
...
+When a contract gets complicated, there may be multiple different
+collections that may not be all part of the main structure, but instead be part of sub-structure or nested collections.
+They all need to have unique prefixes.
+
+We can introduce an `enum` for tracking storage prefixes and keys.
+And then use borsh serialization to construct a unique prefix for every collection.
+It's as efficient as manually constructing them, because Borsh enum only takes one byte.
+
+```rust
+#[derive(BorshSerialize)]
```
chadoh, line 47: Why no need for `BorshDeserialize` here?

HELP.md
```diff
@@ -0,0 +1,717 @@
...
+            metadata: LazyOption::new(StorageKeys::Metadata.try_to_vec().unwrap()),
+        }
+    }
+    
+    fn get_tokens(&self, account_id: &AccountId) -> UnorderedSet<String> {
+        let tokens = self.accounts.get(account_id).unwrap_or_else(|| {
+            UnorderedSet::new(
+                StorageKeys::SubAccount { account_hash: env::sha256(account_id.as_bytes()) }
+                    .try_to_vec()
+                    .unwrap(),
```
chadoh, line 71: is there a way to avoid all the `.try_to_vec().unwrap()`? It really clutters up a great interface.

HELP.md
```diff
@@ -0,0 +1,717 @@
...
+        // Create the new contract using the data from the old contract.
+        Self { owner_id: old_contract.owner_id, data: old_contract.data, new_data }
+    }
+}
+```
+
+## Use `PanicOnDefault`
+
+By default `near_sdk` allows a contract to be initialized with the default state.
+Usually, if you have an initializer, you may want to prevent it.
```
chadoh, line 114: We should say _why_ here. Or give an example of something bad that could happen if you forget this.

HELP.md
evgenykuzyakov: we can add `to_key` impl that does this:
```rust
impl StorageKeys {
  pub fn to_vec(&self) -> Vec<u8> {
    self.try_to_vec().unwrap()
  }
}
```

evgenykuzyakov: We can also change `UnorderedSet::new` to take `IntoVec` and then it may simplify the code.

HELP.md
evgenykuzyakov: We're only serializing keys, but not deserializing them.

HELP.md
```diff
@@ -0,0 +1,762 @@
...
+
+```rust
+#[near_bindgen]
+impl Contract {
+    pub fn take_my_money(&mut self) {
+        env::log(b"Thanks!");
+    }
+
+    pub fn do_not_take_my_money(&mut self) {
+        let error = format!("", ident.to_string());
```
evgenykuzyakov, line 322:
```suggestion
```

HELP.md
chadoh: Would we have to do that with every kind of collection? Is it worth it? What would this look like, then? This?

```rust
StorageKeys::SubAccount { account_hash: env::sha256(account_id.as_bytes()) }
    into(),
```

HELP.md
```diff
@@ -0,0 +1,761 @@
...
+}
+```
+
+## `Base64VecU8` JSON type
+
+Often the contract needs to receive or return a binary data.
+Encoding a `Vec<u8>` with JSON will lead to an integer array, e.g. `[110, 101, 97, 114]`
+This is inefficient in both compute and space.
+
+`Base64VecU8` is a wrapper on top of `Vec<u8>` that allows to pass it as arguments or result.
```
chadoh, line 236: Why is this more efficient in either compute or space? 

We store it internally as `Vec<u8>` anyhow, so the space consideration seems irrelevant.

And for compute on in/out, base64 is [about 33% larger than binary](https://stackoverflow.com/questions/33811673/why-the-size-of-base64-encoded-string-is-larger-than-the-original-file). How then is serializing & deserializing base64 more compute-efficient than using binary?

If neither of these are true, is there some other situation that should make people reach for `Base64VecU8`?

HELP.md
```diff
@@ -0,0 +1,761 @@
...
+
+```rust
+#[near_bindgen]
+impl Contract {
+    /// View method. Requires cloning the account id.
+    pub fn get_owner_id(&self) -> AccountId {
+        self.owner_id.clone()
+    }
+
+    /// View method. More efficient, but can't be reused internally, because it consumes self.
```
chadoh, line 279: Is there a reason to prefer Rust's doc comments with three slashes to non-doc comments with only two?

HELP.md
```diff
@@ -0,0 +1,761 @@
...
+/// Main contract structure serialized with Borsh
+#[near_bindgen]
+#[derive(BorshDeserialize, BorshSerialize, PanicOnDefault)]
+pub struct Contract {
+    pub pair: Pair,
+}
+
+/// Implements both `serde` and `borsh` serialization.
+#[derive(Serialize, Deserialize, BorshDeserialize, BorshSerialize)]
+#[serde(crate = "near_sdk::serde")]
```
chadoh, line 472: Is it common to implement both? If this is not common, can we change the example to something that is? If it is common, can we explain why it's useful?

HELP.md
```diff
@@ -0,0 +1,761 @@
...
+
+    pub fn get_pair(self) -> Pair {
+        self.pair
+    }
+}
+```
+
+## Use `setup_alloc!`
+
+The SDK provides a helper macro to setup a global allocator from `wee_alloc` crate:
```
chadoh, line 495: Can we add a sentence or a link about what a global allocator is and why this is needed? 

Also, is there ever a situation when this is not needed, or when a contract author may prefer a different allocator? If not, why even make people include this boilerplate?

HELP.md
```diff
@@ -0,0 +1,761 @@
...
+`LookupMap` has a better performance and stores less data comparing to the `UnorderedMap`.
+
+- `UnorderedMap` requires `2` storage reads to get the value and `3` storage writes to insert a new entry.
+- `LookupMap` requires only one storage read to get the value and only one storage write to store it.
+
+### Storage space
+
+`UnorderedMap` requires more storage for an entry comparing to a `LookupMap`.
+
+- `UnorderedMap` stores the key twice (once in the first map and once in the vector of keys) and value once. It also has higher constant for storing the length of vectors and prefixes.
```
chadoh, line 655: "higher constant"? Seems like maybe there's a missing word?

HELP.md
```diff
@@ -0,0 +1,761 @@
...
+
+It's a type of persistent collection that only stores a single value.
+The goal is prevent contract from deserializing the given value until it's needed.
+An example can be a large blob of metadata that is only needed when it's requested in a view call,
+but not needed for the majority of contract operations.
+
+It acts like an `Option` that can either hold a value or not and also requires a unique prefix (a key in this case)
+like other persistent collections.
+
+Comparing to other collections, the `LazyOption` allows to initialize the value during the constructor.
```
chadoh, line 668: Mixing grammar and substantive changes with this suggestion; not sure if I understand the substantive parts enough to convey them accurately

```suggestion
Compared to other collections, `LazyOption` only allows you to initialize the value during initialization.
```

chadoh:
@evgenykuzyakov I mostly fixed typos here, a4d4899, but also added some extra detail and I'm not sure I got the code in my examples 100% right. Can you double-check it for me?

mikedotexe (approved):
awesome work

HELP.md
mikedotexe: There's a good example here, and I think this is typically helpful as it is here right above a function. A non-doc comment might be in the implementation if something seems unintuitive and it's helpful to add a note.
https://doc.rust-lang.org/reference/comments.html#doc-comments

HELP.md
mikedotexe: pushed link to global allocators

HELP.md
mikedotexe: added a comment about serde being useful

HELP.md
mikedotexe: Looking at the `fn insert_raw` from `LookupMap` and `UnorderedMap` seems like the latter has to incorportate `u64` storage. I'll add a word or two here for readability.

HELP.md
```diff
@@ -0,0 +1,859 @@
...
+
+/// Main contract structure serialized with Borsh
+#[near_bindgen]
+#[derive(BorshDeserialize, BorshSerialize, PanicOnDefault)]
+pub struct Contract {
+    pub pair: Pair,
+}
+
+/// Implements both `serde` and `borsh` serialization.
+/// `serde` is typically useful when returning a struct in JSON form for a frontend.
```
willemneal, line 566:
```suggestion
/// `serde` is typically useful when returning a struct in JSON format for a frontend.
```

mikedotexe:
I'm going to merge this. I removed a section that might be confusing how it was, and I think it'd be great to flesh out cross-contract calls (along with this addition) separately later. See https://github.com/near/near-sdk-rs/issues/356

HELP.md
chadoh: Great info about serde, thanks.

I'm also curious about using both "Serialize/Deserialize" and "BorshSerialize/BorshDeserialize."

HELP.md
chadoh: Ok, but the docs are not really _used_ yet, in the NEAR ecosystem. 

I'm guessing that someday the Metadata API might use doc comments.

mikedotexe: I don't think that's accurate, we just haven't fully fleshed it out yet, which I hope we'll do this quarter.
Screen Shot 2021-04-09 at 2 02 01 PM

HELP.md
mikedotexe: Sure, feel free to make an issue. Here are some helpful links:
- https://borsh.io/
- https://docs.near.org/docs/concepts/storage-staking#use-a-binary-serialization-format-rather-than-json

HELP.md
```diff
@@ -0,0 +1,814 @@
...
+Usually, if you have an initializer, you will want to prevent this.
+There is a helper derive macro `PanicOnDefault` to do this, e.g.
+
+```rust
+#[near_bindgen]
+#[derive(BorshDeserialize, BorshSerialize, PanicOnDefault)]
+pub struct Contract {
+    pub data: String,
+}
+```
```
matklad, line 126: Why do we need a separate proc macro for this? seems like `near_bindgen` can generate `Default` impl  which panics if ther's a method annotated with `#[init]`? Or maybe, better still, the code-gen can be changed to not rely on `Default` at all if there's init? 

It feels a bit that this is a work-around rather a best practice :)

PR #184, opened 2020-06-08
feat(near-sdk/collections): Implemented Debug trait for Vector collection

frol:
This PR is inspired by [a StackOverflow question](https://stackoverflow.com/questions/62261302/how-to-debug-print-near-protocol-collections).

Other collections will also benefit from having Debug trait implementations, in my opinion. I would like to leave this exercise to the question author.

NOTE: This PR looks bigger than it should be due to the fact that we put more requirements than needed on some of the methods and I had to split the `impl`s into separate ones to put different sets of requirements.

# Test Plan

* Added a new test for Debug trait implementation ensuring that it works fine even for the collections that implement only BorshDeserialize (i.e. you don't need BorshSerialize to have Debug trait)


near-sdk/src/collections/vector.rs
```diff
@@ -210,41 +244,31 @@ where
...
-    /// Extends vector from the given collection.
-    pub fn extend<IT: IntoIterator<Item = T>>(&mut self, iter: IT) {
-        for el in iter {
-            self.push(&el)
-        }
+impl<T: fmt::Debug + BorshDeserialize> fmt::Debug for Vector<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.to_vec().fmt(f)
     }
+}
```
frol, line 253: This is the core part of this PR. Just 4 new lines of code.

near-sdk/src/collections/vector.rs
```diff
@@ -210,41 +244,31 @@ where
...
-    }
-
-    /// Extends vector from the given collection.
-    pub fn extend<IT: IntoIterator<Item = T>>(&mut self, iter: IT) {
-        for el in iter {
-            self.push(&el)
-        }
+impl<T: fmt::Debug + BorshDeserialize> fmt::Debug for Vector<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.to_vec().fmt(f)
```
bowenwang1996, line 251: this could potentially be expensive. If a user log this causally in the code, it might consume a lot of gas.

frol: I know. What do you suggest? I don’t think it should be used in production code, so it should be fine

bowenwang1996: Not sure. @evgenykuzyakov what do you think?

frol:
As per the discussion during the last Contract Runtime meeting, we decided to go with a feature-flag for the expensive Debug implementation while still implementing minimum useful Debug trait.

frol:
@evgenykuzyakov Do you know how to resolve the CI failure?

```
error: failed to select a version for `borsh`.
    ... required by package `near-vm-errors v0.9.1 (https://github.com/nearprotocol/nearcore#bbe0137b)`
    ... which is depended on by `near-vm-logic v0.9.1 (https://github.com/nearprotocol/nearcore#bbe0137b)`
    ... which is depended on by `near-sdk v0.11.0 (/home/travis/build/near/near-sdk-rs/near-sdk)`

versions that meet the requirements `^0.6.2` are: 0.6.2
all possible versions conflict with previously selected packages.
  previously selected package `borsh v0.6.1`
    ... which is depended on by `near-sdk v0.11.0 (/home/travis/build/near/near-sdk-rs/near-sdk)`
failed to select a version for `borsh` which could resolve this conflict

The command "./ci-test.sh" exited with 101.
```

frol:
Tests got recovered. Please, review and merge.

near-sdk/src/collections/vector.rs
```diff
@@ -13,6 +15,7 @@ const ERR_INDEX_OUT_OF_BOUNDS: &[u8] = b"Index out of bounds";
 /// An iterable implementation of vector that stores its content on the trie.
 /// Uses the following map: index -> element.
 #[derive(BorshSerialize, BorshDeserialize)]
+#[cfg_attr(not(feature = "expensive-debug"), derive(Debug))]
```
frol, line 18: This non-`expensive-debug` derive and the explicit `expensive-debug` implementation below are the meat of this PR (all the rest is just moving things around to relax the requirements enabling usage of certain methods when only BorshSerialize or BorshDeserialize implemented and not both.

PR #193, opened 2020-06-21
fix cross contract examples

bowenwang1996:
Cross contract examples were broken because we didn't transfer enough tokens to deploy the status message contract.


mikedotexe:
Related to https://github.com/near-examples/rust-high-level-cross-contract/issues/31
I will re-add the tag so it is supposed to show up on the examples page after this is complete

PR #200, opened 2020-07-10
fix CI tests

### LINKED ISSUE #196: Master branch CI failure
CI jobs with `NEAR_RELEASE=false` fail: https://travis-ci.com/github/near/near-sdk-rs/builds/174852125

@evgenykuzyakov Do you have any idea on how this can be resolved?

bowenwang1996:
fixes #196

### DIFF
--- a/near-sdk/src/collections/set.rs
+++ b/near-sdk/src/collections/set.rs
@@ -187,7 +187,7 @@ mod tests {
         test_env::setup();
         let mut set = UnorderedSet::default();
         let mut rng = rand_xorshift::XorShiftRng::seed_from_u64(0);
-        for _ in 0..1000 {
+        for _ in 0..500 {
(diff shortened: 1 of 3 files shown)


bowenwang1996:
@chefsale @ailisp I am getting compilation errors on CI due to ` the `wasm32-unknown-unknown` target may not be installed`. Is CI environment set up properly?

ailisp:
It might not installed for stable rust. I just add it and rebuild

nearmax:
Thank you!

frol:
Does this mean that we might have broken the existing contracts since they got more expensive?

bowenwang1996:
@frol yes it's possible, although I don't think it's breaking in most cases unless you wrote a function that no matter what the input is, it always consumes a lot of gas.

PR #206, opened 2020-07-17
Allow minification of contracts.

olonho:


minifier/minify_examples.sh
```diff
@@ -0,0 +1,25 @@
...
+# done
+
+cd out
+for p in ./base/*.wasm ; do
+  w=$(basename -- $p)
+  wasm-snip $p --snip-rust-fmt-code --snip-rust-panicking-code -p core::num::flt2dec::.* -p core::fmt::float::.*  \
+     --output snipped-$w
+  wasm-gc snipped-$w
+  wasm-strip snipped-$w
+  wasm-opt -Oz snipped-$w --output opt-snipped-$w
```
willemneal, line 21: Where is this installed for CI?  I had trouble trying to find the best way.

olonho: Not sure if it's installed on CI.

minifier/README.md
```diff
@@ -0,0 +1,15 @@
...
+     * cargo install wasm-snip wasm-gc
+     * apt install binaryen wabt
+
+*WARNING*: minification could be rather aggressive, so you *must* test the contract after minificaion.
+Standalone NEAR runtime (https://github.com/nearprotocol/nearcore/tree/master/runtime/near-vm-runner) could be helpful
+here.
+
+Current approach to minification is the following:
+    * snip (i.e. just replace with `unreachable` instruction) few known fat functions from the standard library
+     (such as float formatting and panic related)
```
evgenykuzyakov, line 12: float formatting can actually happen, because we don't prohibit floats. Not that it's needed to be included most of the time.

willemneal: So do we just prohibit the use of float operations?  If so, I'll need to update sdk-as to allow for them.

olonho: Think float computations can happen in contract, we can just forbid float formatting.

PR #212, opened 2020-07-25
docs: recommend a decimal-handling library

chadoh:
[On StackOverflow], @robert-zaremba asked if the NEAR team recommends a specific library for handling decimals. This question was closed because it does not meet StackOverflow's guidelines, given the imprecise nature of recommendations. I thought this might make a good addition to `near_sdk` docs. Robert prefers `decimate`, so that's what I've recommended here.

* Is there a better place for this information to live in the `near_sdk` docs?
* Do more accomplished NEAR Rustaceans recommend a different way to handle large numbers?

  [On StackOverflow]: https://stackoverflow.com/questions/62990352/whats-the-recommended-library-for-handling-decimals-in-near-smart-contracts

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,9 @@ The general workflow is the following:
     }
     ```
 
+## Tips
+When working with NEAR smart contracts you will need to handle many large numbers (the minimum allowed amount of NEAR is 1/10<super>24</super> of one NEAR token – put another way, one NEAR token is 1,000,000,000,000,000,000,000,000 yoctoNEAR). To simplify handling such large numbers, we recommend [decimate](https://docs.rs/decimate/).
+
 ## Building Rust Contract
 We can build the contract using rustc:
 ```bash
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ The general workflow is the following:
     ```
 
 ## Tips
-When working with NEAR smart contracts you will need to handle many large numbers (the minimum allowed amount of NEAR is 1/10<super>24</super> of one NEAR token – put another way, one NEAR token is 1,000,000,000,000,000,000,000,000 yoctoNEAR). To simplify handling such large numbers, we recommend [decimate](https://docs.rs/decimate/).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -174,6 +174,9 @@ The general workflow is the following:
     }
     ```
 
+## Tips
+When working with NEAR smart contracts you will need to handle many large numbers (the minimum allowed amount of NEAR is 1/10<super>24</super> of one NEAR token – put another way, one NEAR token is 1,000,000,000,000,000,000,000,000 yoctoNEAR). To simplify handling such large numbers, we recommend [decimate](https://docs.rs/decimate/).
+
```
robert-zaremba, line 179: It's not only for large numbers but also for decimal numbers (eg: `1.25`) which are not represented as `float`.

Let's specify this as well.

chadoh: Good call; let me know if you like the updated version

evgenykuzyakov:
What's the security and support status of `decimate`? Is it actively supported and whether there are some security reviews on it. By recommending a crate we might be risking funds of projects that build contract on it based on our recommendations. 

Within `nearcore` we use `num_rational::Rational` and `num_bigint::BigUint` for handling large numbers.

evgenykuzyakov:
Also for more performant math but fixed length there are `uint` crate. It allows define structs like `U256` or `U512`

robert-zaremba:
Good point about `U256` and `U512` - we should mention it as well.

`num_rational::Rational` is more general version of `decimate`:
+ `decimate` is much simpler, less code, and supports only power of 10 denominators (1, 10, 100).... - which makes the computation simpler and faster, and code smaller
+ `decimate` is robust enough - as the nominator  (mantissa in decimate notaion) is generic, so it can be `u128` as well as `BigUint`. 
+ `decimate` doesn't support precise division (eg `1/3`), 
+ `Rational` is more robust, you can represent any rational number and use different types for 


I still think it's good to give some overview to the user about handling numeric numbers. Some users may not know that they will have exception if using big numbers with `u64` type... If not in this page, we can have a community page somewhere where we can add this tips (can we have community wiki in official docs?)  (Can anyone link Peter  "Potato" here?)

We can also add a disclaimer, that we don't take any responsibility for this recommendation and this is only for information purpose.

evgenykuzyakov:
@potatodepaulo

evgenykuzyakov:
outdated. Probably should go to docs

PR #217, opened 2020-08-20
[2.0.0] feat: LookupMap and LookupSet

evgenykuzyakov:
### Contract changes

* Updated `status-message-collections` to use `LookupMap`
* **BREAKING** Updated `fungible-token` implementation to use `LookupMap`. It changes storage layout.

### API changes

* Introduce `LookupMap` and `LookupSet` that are faster implementations of `UnorderedMap` and `UnorderedSet`, but without support for iterators.
Most read/lookup/write are done in 1 storage access instead of 2 or 3 for `Unordered*` implementations.
* **BREAKING** `Default` is removed from `near_sdk::collections` to avoid implicit state conflicts.
Collections should be initialized by explicitly specifying prefix using `new` method.
* **BREAKING** `TreeMap` implementation was updated to use `LookupMap`.
Previous `TreeMap` implementation was renamed to `LegacyTreeMap` and was deprecated.
It should only be used if the contract was already deployed and state has to be compatible with the previous implementation.


near-sdk/src/collections/legacy_tree_map.rs
```diff
@@ -0,0 +1,1769 @@
...
+
+                // removing node might have caused a imbalance - balance the tree up to the root,
+                // starting from the lowest affected key (min key from right subtree in this case)
+                self.check_balance(self.root, &p.key)
+            }
+        }
+    }
+
+    // Move content of node with id = `len - 1` (parent left or right link, left, right, key, height)
+    // to node with give `id`, and remove node `len - 1` (pop the vector of nodes).
```
olonho, line 557: given?

near-sdk/src/collections/legacy_tree_map.rs
```diff
@@ -0,0 +1,1769 @@
...
+    // This ensures that among n nodes in the tree, max id is n-1, so when new node is inserted,
+    // it gets and id as it's position in the vector.
```
olonho, line 559: Not sure if understand the last part of sentence.

near-sdk/src/collections/legacy_tree_map.rs
```diff
@@ -0,0 +1,1769 @@
+//! Legacy `TreeMap` implementation that is using `UnorderedMap`.
+//! DEPRECATED. This implementation is deprecated and may be removed in the future.
```
olonho, line 2: Hmm, but what it the point of having such implementation then?

evgenykuzyakov: Because some contracts might still be using it.

evgenykuzyakov:
Going to change version to `2.0.0`

near-sdk/src/collections/legacy_tree_map.rs
evgenykuzyakov: updated

near-sdk/src/collections/tree_map.rs
```diff
@@ -16,54 +16,37 @@ use crate::collections::UnorderedMap;
...
-}
-
-impl<K, V> Default for TreeMap<K, V>
-    where
-        K: Ord + Clone + BorshSerialize + BorshDeserialize,
-        V: BorshSerialize + BorshDeserialize,
-{
-    fn default() -> Self {
-        Self::new(next_trie_id())
+        Self { id, key, lft: None, rgt: None, ht: 1 }
```
olonho, line 37: Maybe left, right and height would be more readable?

olonho (approved):

near-sdk/src/collections/tree_map.rs
evgenykuzyakov: Yep. Bad naming.

PR #227, opened 2020-09-15
Add init_once

evgenykuzyakov:
Introduced `init_once` that is an initializer that requires the contract was not initialized yet.

Fixes: #95


bowenwang1996 (commented):
what is an example where we would prefer `#[init]` over `#[init_once]`?

evgenykuzyakov:
> `#[init]` can be use for migration.

willemneal (approved):
LGTM

near-sdk-core/src/info_extractor/attr_sig_info.rs
```diff
@@ -14,8 +14,12 @@ pub struct AttrSigInfo {
     pub non_bindgen_attrs: Vec<Attribute>,
     /// All arguments of the method.
     pub args: Vec<ArgInfo>,
-    /// Whether method can be used as initializer.
+    /// Whether the method is an initializer.
+    /// NOTE: It's non-inclusive to `is_init_once`.
     pub is_init: bool,
+    /// Whether the method is an initializer and the contract state should't exist.
+    /// NOTE: If it's true, it means `is_init` is true as well.
```
nearmax, line 21: It sounds cleaner to make it non-overlapping, e.g. `is_init` would correspond to `[init]` attribute only. Then we can have a method on this struct like `is_some_init_method`

ailisp (approved):
Looks a useful API!

nearmax (changes requested):
Please add compilation integration test.

README.md
```diff
@@ -83,17 +83,17 @@ impl StatusMessage {
...
 
 * **Initialization methods.** We can define an initialization method that can be used to initialize the state of the
-contract.
+contract. NOTE: `#[init_once]` verifies that the contract has not been initialized (the contract state doesn't exist).
 
     ```rust
     #[near_bindgen]
     impl StatusMessage {
-      #[init]
+      #[init_once]
```
frol, line 96: Should the documentation and an example remain for `#[init]`?

frol (approved):

near-sdk-core/src/info_extractor/attr_sig_info.rs
evgenykuzyakov: It's easier to manage internally because there are bunch of places that depend on `is_init`, while `is_init_once` is optional for `is_init`.

README.md
evgenykuzyakov: The default option for people to use should be `init_once`. We'll keep `init` as is for backward compatibility reasons, but most `new` functions expect `init_once` behavior

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## `Pending`
+
+* Introduced `init_once` that is an initializer that requires the contract was not initialized yet.
```
olonho, line 5: We may want to document the exact state machine of contract lifetime, i.e. smth along lines of 

```
 Deploy -> First run -> Init once -> Init -> method call
              -> Subsequent runs -> Init -> method call
```

README.md
olonho: Can we mark init as obsolete, so that people could migrate?

olonho (approved):

README.md
frol: We need to communicate that. If I read an old example and it has `#[init]` and only can read about `#[init_once]` I would get confused. Let's improve the docs and the changelog.

ilblackdragon:
What's the status on this?

evgenykuzyakov:
Will update and request the review again

ilblackdragon:
What's the status on this?

mikedotexe:
> What's the status on this?

We're discussing this right now and would like to put it in the next release, likely this week.

PR #228, opened 2020-09-15
Add #[private] method decorator

### LINKED ISSUE #122: Support "#[self_call]" for calls

evgenykuzyakov:
Based on #168 by @chain-police 

Introduced `#[private]` method decorator, that verifies `predecessor_account_id() == current_account_id()`.
NOTE: Usually, when a contract has to have a callback for a remote cross-contract call, this callback method should
only be called by the contract itself. It's to avoid someone else calling it and messing the state. Pretty common pattern
is to have an assert that validates that the direct caller (predecessor account ID) matches to the contract's account (current account ID).

Fixes #122


ailisp (approved):
looks useful! but
`#[private] pub fn ...` does looks a bit strange. Maybe `#[owner]` from eth world is a better name?

nearmax (approved):
Please add a compilation test.

frol (approved):
Beautiful!

olonho (approved):
Cool!

README.md
```diff
@@ -122,6 +122,21 @@ pub fn my_method(&mut self) {
...
+is to have an assert that validates that the direct caller (predecessor account ID) matches to the contract's account (current account ID).
+Macro `#[private]` simplifies it, by making it a single line macro instead and improves readability.
+
+To declare a private method use `#[private]` decorator:
+```rust
+
+#[private]
+pub fn my_method(&mut self) {
+...
+}
```
willemneal, line 136:
```suggestion
}
/// Which is equivalent to

pub fn my_method(&mut self ) {
    if env::current_account_id() != env::predecessor_account_id() {
        near_sdk::env::panic("Method method is private".as_bytes());
    }
...
}
```

willemneal (approved):
LGTM

willemneal:
> looks useful! but
> `#[private] pub fn ...` does looks a bit strange. Maybe `#[owner]` from eth world is a better name?

Technically the macro could add the `pub` to make it feel more private.

frol:
> Technically the macro could add the pub to make it feel more private.

Great point! How about even dropping the need to have the #[private] marker at all?

evgenykuzyakov:
> How about even dropping the need to have the #[private] marker at all?

This will make all methods to be exported under `near_bindgen` which will probably generate JSON schema when it's needed. Unless we can explain that people need to have a separate `impl` section for non exported methods, then it might be confusing.

PR #230, opened 2020-09-16
Add test contract

evgenykuzyakov:


olonho (approved):

examples/test-contract/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+
+#[near_bindgen]
+impl TestContract {
+    #[init]
+    pub fn new() -> Self {
+        assert!(!env::state_exists(), "Already initialized");
+        Self {}
+    }
+
+    #[init]
```
willemneal, line 25: Why the second init?

evgenykuzyakov: The idea for `migrate` with `init` is to be able to upgrade contract and migrate it from the old data scheme to the new data scheme. Assuming we can't do this natively with borsh, we can use `migrate` call to manually deserialize the state using the old struct/schema and then save the state serialized with the current struct/schema.

While `new` is for original initialization, when the state is not available.

PR #231, opened 2020-09-16
Add `#[derive(NoDefault)]` macro

evgenykuzyakov:
* Added `#[derive(PanicOnDefault)]` that automatically implements `Default` trait that panics when called.
This is helpful to prevent contracts from being initialized using `Default` by removing boilerplate code.


near-sdk-macros/src/lib.rs
```diff
@@ -137,3 +137,25 @@ pub fn metadata(item: TokenStream) -> TokenStream {
         )
     }
 }
+
+#[proc_macro_derive(NoDefault)]
```
nearmax, line 141: Please add docstring to explain its purpose in depth, so that docs.rs have nice documentation.

near-sdk/compilation_tests/no_default.rs
```diff
@@ -0,0 +1,23 @@
+//! Smart contract with initialization function.
+
+use borsh::{BorshDeserialize, BorshSerialize};
+use near_sdk::{near_bindgen, NoDefault};
+
+#[near_bindgen]
+#[derive(NoDefault, BorshDeserialize, BorshSerialize)]
```
nearmax, line 7: I am not sure whether `NotDefault` is more grammar-friendly.

nearmax (approved):

ailisp (commented):
Why this is needed, isn't when user didn't `#[derive(default)]` rust would give compilation error?
image

frol (approved):
I like the idea, though it is a bit confusing, and the downside is that this generates more code to the final contract. Also, how useful is this versus just not deriving Default?

evgenykuzyakov:
> Why this is needed, isn't when user didn't #[derive(default)] rust would give compilation error?

It's because `Default` is required by `near_bindgen` for non-init methods.

olonho (approved):
Would be great to know size implications here.

frol:
> It's because Default is required by near_bindgen for non-init methods.

I wonder if we want to have NoDefault used by default by `#[near_bindgen]` unless we have a special flag (e.g. `#[near_bindgen(allow_custom_default)]`)

> Would be great to know size implications here.

Would be great to add some tests with reports how every PR changes the size of contracts.

near-sdk/compilation_tests/no_default.rs
willemneal: Yeah the phrasing makes it seems that by not adding `derive(Default)` means there is no default.  How about `EnsureNoDefault` or `FailingDefault`?

willemneal (approved):

near-sdk/compilation_tests/no_default.rs
evgenykuzyakov: Switched to `PanicOnDefault`

PR #232, opened 2020-09-16
Update Cargo.toml to use overflow and opt-s

evgenykuzyakov:
Adding `overflow-checks = true` and changing to `opt-level = "s"`

From my experience, `opt-level = "s"` produces smaller binary than `opt-level = "z"`. But `overflow-checks` are usually good for the contracts, especially financial contracts, even though it increases the binary size likely due to standard library.


olonho:
Could you please provide the information of how contract size changes? IIRC, @ailisp enabled contract size measurements on CI some time ago, so it's pretty easy to size size changes and section breakdown.

evgenykuzyakov:
I guess I need to rebuild them with `z` to compare costs of simple `overflow-checks = true`

ailisp (approved):
Some contracts like lockable_fungible_token.wasm has increased size, is this because overflow-checks=true?

evgenykuzyakov:
> Some contracts like lockable_fungible_token.wasm has increased size, is this because overflow-checks=true?

I think so.

evgenykuzyakov:
Here is `opt-z` to compare. Some of them are larger, but some of them are smaller. Still most of them are slightly larger.

mikedotexe:
Are we still wanting to make this change?

austinabell:
Although I don't really see the motivation for doing this for examples:
- `opt-level = "s"` shouldn't be applied by default I don't think unless a contract is stable, guaranteed not to change, and in the edge case where it's smaller code size. Seems weird to do this across the board when it depends on the contract itself if contract size or potential speed is preferred.
- `overflow-checks` again, this seems like it probably should only be applied on some contracts, but might be fine if added, I don't have an opinion.

I updated the contracts with the merge, so the difference can be seen

agostbiro:
Thanks for all the input in this PR. I'm closing it as the last comment was almost two years ago. Please open a new issue to discuss if the change is still desired.

PR #234, opened 2020-09-20
added an assertion to ensure tokens aren't transfer to the token contract itself.

jasperdg:
Sending tokens to the contract would essentially burn said tokens. On Ethereum millions upon millions have been lost by not having a similar check in the ERC20 implementation.


evgenykuzyakov (changes requested):
Good idea. Can you recompile with `./build.sh` and then add a unit test to verify the error.
You can check the assert using `#[should_panic(expected = "Can't transfer tokens to token contract itself")]`

jasperdg:
@evgenykuzyakov Done

evgenykuzyakov:
@jasperdg So we're trying to use reproducible builds with Docker. I can take over the PR, rebuild the binary with it and later merge it. Or if you want, you can do this. I'll add instructions today

jasperdg:
*@evgenykuzyakov Happy to do that! That way if I want to contribute in
the future I’m familiar* w/ the process.

PR #235, opened 2020-10-01
Feat: Add better simulation API and macros

### LINKED ISSUE #233: Improve testing / introduce Rust API
So currently we are wriitng method names in string and passing arguments in JSON. In integration tests this leads to a weird json parsing errors when things don't match.

We are also introducing more borsh arguments for method names. Which means soon we will end up with ABIs pretty much.

Suggestion to add a set of marcos and tools to infer the API for using contracts.
This also will serve not just for testing but also as rust-api-rs, which we currently don't have.

```
#[derive_json_abi]
struct FungibleToken {
    pub fn new();
    pub fn transfer(&mut self);
    pub fn transferFrom(&mut self, …);
    pub fn balanceOf(&self, owner_id: AccountId) -> Balance;
}
```
OR
```#[derive_borsh_abi]```
OR
``` 
load_abi!(FungibleToken, 'path/to/abi');
```
and then
```
let mut contract = FungibleToken::???(runtime, contract_account_id); // this would just create a contract object for given 
```
OR
```
// This would deploy contract, call new and create a object.
let mut contract = FungibleToken::deploy(runtime, signer_id, code, /* args for new call */);
```

Can use then:
```
contract.transfer(signer_id, to_id, amount);
contract.balanceOf(owner_id);
etc
```

We can hide signer into contract when creating one, but it’s a bit confusing in JS IMO.

Given runtime has very similar API to our RPC, this APIs should take a trait in which StandaloneRuntime and RustRPCProvider can implement.

### PR DESCRIPTION
**Pull request by willemneal**
This is an improvement to the now defunct #226 and fixes #233.

### CODE REVIEW

**Comment by willemneal** [comment]
Now you can import the following:

```rust
use fungible_token::FungibleTokenContract;
```

Which has the same methods as the contract, but looks like:

```rust
struct FungibleTokenContract {
    account_id: AccountId,
}

impl FungibleTokenContract {
    pub fn new(&self, owner_id: AccountId, total_supply: Balance) -> PendingContractTx {
        let balance: U128 = total_supply.into();
        PendingContractTx::new(
            &self.account_id,
            "new",
            json!({
              "owner_id": owner_id,
              "total_supply": balance
            }),
            false,
        )
    }
   ...
}

```

However, to run the test in `examples/fungible-token`, you need to use nightly, e.g. `cargo +nightly test`.

Currently there is an issue with building the fungible contract.  It fails on a dependency of `near-sdk-sim`, so placing it in `[dev-dependencies]` fixes it, but then when running the tests fails to find the dependency.  So I'm not sure what the issue is or how best to make the dependency dependent on the test environment.

near-sdk-macros/src/lib.rs
```diff
@@ -27,7 +27,9 @@ pub fn near_bindgen(_attr: TokenStream, item: TokenStream) -> TokenStream {
             }
         };
         let generated_code = item_impl_info.wrapper_code();
+        let marshalled_code = item_impl_info.marshall_code();
```
evgenykuzyakov, line 30: This may increase filesize unless it's used only for testing. Can we remove it from the wasm32 target?

willemneal: Yes it's only testing. It injects calls to ignore in wasm32 target. But it _should_ be here!  Thanks!

examples/fungible-token/src/lib.rs
```diff
@@ -269,8 +269,7 @@ impl FungibleToken {
     }
 }
 
-#[cfg(not(target_arch = "wasm32"))]
-#[cfg(test)]
+#[cfg(all(test, not(target_arch = "wasm32")))]
```
evgenykuzyakov, line 272: Did this change decreased size by 7Kb or was it something else?

examples/fungible-token/tests/general.rs
```diff
@@ -0,0 +1,111 @@
...
+        }),
+        false,
+    );
+    let contract = root.deploy_and_init(&TOKEN_WASM_BYTES, initial_tx);
+    let value = contract.view(PendingContractTx::new(
+        &contract.account_id,
+        "get_total_supply",
+        json!({}),
+        true,
+    ));
```
evgenykuzyakov, line 69: I thought this should come from `FungibleTokenContract ` API.

near-sdk-core/src/code_generator/impl_item_method_info.rs
```diff
@@ -158,4 +158,64 @@ impl ImplItemMethodInfo {
             }
         }
     }
+
+    pub fn marshal_method(&self) -> TokenStream2 {
```
evgenykuzyakov, line 162: Please add a comment to mention that it doesn't output any extra code for `wasm32` target.
Also why we can't have 2 implementations? One for `wasm32` and one for not. Does target test have to be post macro expansion?

near-sdk-sim/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "near-sdk-sim"
+version = "2.0.0"
+authors = ["Illia Polosukhin <illia.polosukhin@gmail.com>"]
```
evgenykuzyakov, line 4: Replace it with `hello@near.org`

near-sdk-sim/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+edition = "2018"
+publish = false
+
+
+[dependencies]
+serde_json = "*"
+serde = { version = "*", features = ["derive"] }
```
evgenykuzyakov, line 11: `serde` should be reused from `near_sdk`

evgenykuzyakov (commented):
Would be great to add docs about how to use it to near-sdk-sim README

examples/fungible-token/tests/general.rs
willemneal: It does, this was just an example of how to do it without it.

evgenykuzyakov: Got it. Please add a comment on how to do this same with generated API. Or add the same test with generated API.

rust-toolchain
```diff
@@ -1 +1 @@
-stable
+stable-2020-10-08
```
evgenykuzyakov, line 1: Why do we need this? Is this from nearcore?

examples/fungible-token/tests/general.rs
```diff
@@ -0,0 +1,83 @@
...
+    let root = runtime.get_root();
+    // let balance: U128 = initial_balance.into();
+    let contract = FungibleTokenContract::_new("contract".to_string());
+    let contract_user = root.deploy_and_init(
+        &TOKEN_WASM_BYTES,
+        contract.new(root.account_id.clone(), initial_balance.into()),
+    );
+    let alice = runtime.create_user("alice".to_string(), to_yocto("100"));
+    (runtime, contract, alice)
+}
```
chadoh, line 26: Things I like about the stuff inside `init`:

* No unnecessary ceremony with Rust unwrapping, options, or error returns
* `deploy_and_init` matches the concept from near-cli
* Preserving function calls as-written, with `contract.fn_name(...args)`, vs something like `contract.call("fn_name", ...args)`

Some mostly-syntactic things I find confusing:

* needing to pass `None` into `init_test_runtime` even when you don't want to change any settings. If we make this a macro could we avoid that?
* "root" instead of matching our currently-more-common name, `master_account`
* the need to initialize `contract`
* having `WhateverContract::_new` be the way to initialize a contract
* having both `contract` and `contract_user`
* the word "user" being present at all, since NEAR vernacular usually only refers to "accounts"
* what is `&TOKEN_WASM_BYTES`?
* why do we need to "deploy" the contract?
* the difference in how an "account with a contract" is initialized vs how an account without one is initialized

So let me play with some possible alternative syntax.

```rust
fn init(initial_balance: u128) -> (TestRuntime, FungibleTokenContract, User) {
    let runtime = init_test_runtime();
    let master_account = runtime.get_master_account();
    let contract = create_account!({
        id: "contract".to_string(),
        balance: initial_balance.into(),
        contract: FungibleTokenContract,
    });
    // if we use a macro, can we return different types for "AccountWithContract" and "AccountWithoutContract"?
    let alice = create_account!({
        id: "alice".to_string(),
        balance: to_yocto("100"),
    });
    master_account.call(contract.new()); // this loses deploy_and_init, but maybe that's ok?
    let contract = FungibleTokenContract::_new("contract".to_string());
    (runtime, contract, alice)
}
```

chadoh: Or maybe:

```rust
fn init(initial_balance: u128) -> (TestRuntime, FungibleTokenContract, User) {
    let runtime = init_test_runtime!();
    let master_account = runtime.get_master_account();
    let alice = master_account.create_account("alice".to_string(), to_yocto("100"));
    let contract = master_account.create_account("contract".to_string(), initial_balance.into());
    let contract = master_account.deploy(contract, FungibleTokenContract);
    master_account.call(contract.new(master_account.id.clone(), initial_balance.into()));
    (runtime, contract, alice)
}

File: rust-toolchain (hunk 7 above)
Comment (willemneal) [reply rust-toolchain:1]: The reason is that the fix that stabilizes wasmer requires that stable be above a certain version, I think 1.44, so this is just to make sure it uses the most recent version.

**Comment by willemneal** [comment]
@chadoh @amgando I incorporated your feedback as best I could.  I added a `ExecutionResult` struct that holds a reference to the runtime so that it can handle all of the methods around intermediate promises.  I also added two new macros: `deploy!` and `deploy_default!`.  They both either just deploy a contract or if the method and args are provided at the end, it initializes it too.
 The later uses default values for initial transfer and gas.

```rust
let master_account = init_simulator(None);
    // default values for deposit and gas
    let contract_user = deploy_default!(
        // Contract Proxy
        FungibleTokenContract,
        // Contract account id
        "contract",
        // Bytes of contract
        &TOKEN_WASM_BYTES,
        // User deploying the contract,
        master_account,
        // init method
        new,
        // Args to initialize contract
        master_account.account_id(),
        initial_balance.into()
    );

```
In the high level example:
```rust
    let contract = contract_user.contract;
    let res = master_account.call(
        contract.deploy_status_message(status_id.clone(), status_amt.into()),
        STORAGE_AMOUNT,
        DEFAULT_GAS,
    );

    let promise_outcomes = res.get_receipt_outcomes();
    println!("{:#?}\n{:#?}", promise_outcomes, &res);
...
    let v1: Vec<u8> = vec![42];
    let res = master_account.call(contract.merge_sort(v1), 0, DEFAULT_GAS * 3);
    let value = res.get_borsh_value::<Vec<u8>>().unwrap();
```

willemneal:
@chadoh 

Now it for high level example it is:

```rust

fn init(
    initial_balance: u128,
) -> (UserAccount, ContractAccount<CrossContractContract>, UserAccount) {
    let master_account = init_simulator(None);
    let contract_account =
        deploy!(CrossContractContract, "contract", &TOKEN_WASM_BYTES, master_account);
    let alice = master_account.create_user("alice".to_string(), initial_balance);
    (master_account, contract_account, alice)
}

#[test]
fn test_sim_transfer() {
    // let transfer_amount = to_yocto("100");
    let initial_balance = to_yocto("100000");
    let (master_account, contract, _alice) = init(initial_balance);
    let status_id = "status".to_string();
    let status_amt = to_yocto("35");
    let res = call!(
        master_account,
        contract.deploy_status_message(status_id.clone(), status_amt.into()),
        STORAGE_AMOUNT,
        DEFAULT_GAS
    );
...
```

And examples from fungible token:

```rust
    let contract_user = deploy!(
        // Contract Proxy
        contract: FungibleTokenContract,
        // Contract account id
        contract_id: "contract",
        // Bytes of contract
        bytes: &TOKEN_WASM_BYTES,
        // User deploying the contract,
        signer_account: master_account,
        // init method
        init_method: new(master_account.account_id(), initial_balance.into())
    );
```
and to use call with default gas and calling a view method:

```rust 
    let (master_account, contract, alice) = init(initial_balance);
    let res = call!(
        master_account,
        contract.transfer(alice.account_id.clone(), transfer_amount.into()),
        deposit = STORAGE_AMOUNT
    );
    println!("{:#?}", res.status());
    assert!(res.is_ok());

    let value = view!(contract.get_balance(master_account.account_id()));
    let value: String = value.from_json_value().unwrap();
```

willemneal:
Currently getting the following errors:
```
building cross-contract-high-level
    Updating git repository `https://github.com/near/nearcore.git`
error: failed to get `near-crypto` as a dependency of package `near-sdk-sim v2.0.0 (/home/travis/build/near/near-sdk-rs/near-sdk-sim)`
    ... which is depended on by `cross-contract-high-level v0.1.0 (/home/travis/build/near/near-sdk-rs/examples/cross-contract-high-level)`
Caused by:
  failed to load source for dependency `near-crypto`
Caused by:
  Unable to update https://github.com/near/nearcore.git?branch=stable_wasmer#86cd2847
Caused by:
  object not found - no match for id (86cd2847359f930e5a873314b9dddd121d56306f); class=Odb (9); code=NotFound (-3)

```

examples/fungible-token/tests/general.rs
```diff
@@ -0,0 +1,75 @@
...
+        // Contract account id
+        contract_id: "contract",
+        // Bytes of contract
+        bytes: &TOKEN_WASM_BYTES,
+        // User deploying the contract,
+        signer_account: master_account,
+        // init method
+        init_method: new(master_account.account_id(), initial_balance.into())
+    );
+    let alice = master_account.create_user("alice".to_string(), to_yocto("100"));
```
chadoh, line 30: I really want accounts-with-a-contract and accounts-without-a-contract to be created with the same interface. They are the same thing in NEAR (even if represented by different structs in the machinery of the simulation tester), but this current interface makes them seem very different.

Maybe this:

```rust
let alice = create_account!(id: "alice".to_string(), initial_balance: to_yocto("100"), master_account: master_account);
let contract = create_account!(
    id: "contract",
    contract: &TOKEN_WASM_BYTES, // or TokenFactoryContract
    master_account: master_account,
    init_method: new(master_account.account_id(), initial_balance.into())
);
```

Or this, since the value of `deploy_and_init` in the context of simulation tests still seems pretty minimal to me, and I weigh helping people think clearly more heavily:

```rust
let alice = master_account.create_user("alice".to_string(), to_yocto("100"));
let contract = master_account.create_user("contract".to_string(), to_yocto("100"));
let contract = master_account.deploy(&TOKEN_WASM_BYTES);
call!(master_account, contract.new(master_account.account_id(), initial_balance.into());
```

(Holy smokes, this version finally made me realize that the `initial_balance` argument is being used to set the `total_supply`! Should we rename that argument?)

Or a hybrid approach:

```rust
let alice = master_account.create_user("alice".to_string(), to_yocto("100"));
let contract = master_account.create_user("contract".to_string(), to_yocto("100"));
let contract = deploy!(
    master_account: master_account,
    contract_account: contract,
    wasm_file: "../res/fungible_token.wasm", // could the bytes be loaded within the `deploy!` macro?
    init_function: new(master_account.account_id(), initial_balance.into())
);
```

(Note that this version uses `wasm_file` and `init_function` to match the `near-cli` arguments to `near deploy`. I think it makes sense to use `init_function` instead of `init_method`. If it's possible to pass just a string pointing to a file as above, then I also think it makes sense to call the argument `wasm_file` instead of `bytes` or `contract`.)

I think I lean toward the `create_account!` macro, out of all of these, with the argument name improvements from the third version. What do you think, @willemneal?

examples/cross-contract-high-level/tests/general.rs
```diff
@@ -0,0 +1,64 @@
...
+near_sdk_sim::lazy_static::lazy_static! {
+    static ref TOKEN_WASM_BYTES: &'static [u8] = include_bytes!("../res/cross_contract_high_level.wasm").as_ref();
+}
+
+fn init(
+    initial_balance: u128,
+) -> (UserAccount, ContractAccount<CrossContractContract>, UserAccount) {
+    let master_account = init_simulator(None);
+    let contract_account =
+        deploy!(CrossContractContract, "contract", &TOKEN_WASM_BYTES, master_account);
```
chadoh, line 17: Why does this version use positional arguments instead of named? I don't find it useful to offer these alternatives. It will be easier to maintain & explain if we only have one way to pass these arguments, and I think we should go with the named arguments as used in the FungibleToken tests.

examples/cross-contract-low-level/tests/general.rs
```diff
@@ -0,0 +1,74 @@
...
+fn init(
+    initial_balance: u128,
+) -> (UserAccount, ContractAccount<CrossContractContract>, UserAccount) {
+    let master_account = init_simulator(None);
+    let contract_account = deploy! {
+    contract: CrossContractContract,
+    contract_id: "contract",
+    bytes: &TOKEN_WASM_BYTES,
+    signer_account: master_account
+    };
```
chadoh, line 21: This is my least favorite version of how to pass arguments to `deploy!` shown here, given the improper indentation on lines 17-20.

chadoh (changes requested):
I'm leaving a few comments here that I would love to discuss right in-line where I've left them. Let's please have these discussions before you do any work to address my comments, as my requests are not yet specific and require discussion to be formalized.

After we discuss and you address them either by refusing or updating the code, we can mark them as "resolved", and it will be easier to follow the discussion. It's getting a bit unwieldy; this is a large PR.

examples/fungible-token/tests/general.rs
```diff
@@ -0,0 +1,75 @@
...
+    call, deploy, init_simulator, to_yocto, view, ContractAccount, UserAccount, STORAGE_AMOUNT,
+};
+use std::str::FromStr;
+
+extern crate fungible_token;
+use fungible_token::FungibleTokenContract;
+
+near_sdk_sim::lazy_static::lazy_static! {
+    static ref TOKEN_WASM_BYTES: &'static [u8] = include_bytes!("../res/fungible_token.wasm").as_ref();
+}
```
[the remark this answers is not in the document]
willemneal: The tricky part here is that I don't know where the artifact of the contract will end up.  Rust puts things in `target/wasm32-unknown-unknown/..`, but the `target` directory could be in the project root or a workspace root.  Perhaps this can be determined when running the procedural macro, but as you point out this could be a future improvement/simplification.  

Another solution is that we can include the type information in the wasm binary so that reading it will generate the proxy. This has the added benefit of allowing for JS objects in NAJ to be generated as well.

examples/cross-contract-high-level/tests/general.rs
willemneal: The reason was that all I had at first was positional and then I added the named arguments and this didn't get updated.  Underneath the main case for the a macro is all positional arguments so this is one of them that doesn't take gas/deposit. I agree that we should change this.

examples/cross-contract-low-level/tests/general.rs
willemneal: I'm confused, do you mean that you just don't like the indentation?

examples/fungible-token/tests/general.rs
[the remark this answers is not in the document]
willemneal: So my conclusion with this is we should aim to try to create a solution where the type info comes from the wasm file.  This way we get AS support for free because Rust would be generated from the type info.

I think this feature is a big win too for creating the JS wrappers automatically.

examples/fungible-token/tests/general.rs
willemneal: My reservation about making account creation and deploying the same is that it differs from the CLI semantics.  Though I agree it is important to convey that contracts are just account + contract, but I think it's okay to treat them differently in this context.

chadoh: That's fair

examples/cross-contract-low-level/tests/general.rs
chadoh: It seems/seemed like you're demonstrating three ways to call the `deploy!` macro:

1. with positional arguments, `deploy!(master_account, "contract", ...)`
2. with named arguments, using parentheses: `deploy!(signer_account: master_account, contract_id: "contract", ...)`
3. with named arguments, using curly brackets and no parentheses: `deploy! { signer_account: master_account, contract_id: "contract", ... }`

I'm saying that I don't think there's value in supporting more than one, and I think the parentheses seem like a fine way to go. Judging by the improper indentation here, it seems like rustfmt doesn't automatically set indentation using this curly-bracket version.

examples/fungible-token/tests/general.rs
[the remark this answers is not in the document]
chadoh: That sounds great – we can improve this once we have type info stored in the wasm

chadoh (approved):
I am satisfied with the interface for how to use these tests, but I did not look deeper into how the testing infrastructure was written

near-sdk/src/utils/mod.rs
```diff
@@ -8,6 +10,25 @@ macro_rules! log {
...
+    pub args: Vec<u8>,
+    pub is_view: bool,
+}
+
+impl PendingContractTx {
+    pub fn new(receiver_id: &str, method: &str, args: serde_json::Value, is_view: bool) -> Self {
+        Self {
+            receiver_id: receiver_id.to_string(),
+            method: method.to_string(),
+            args: args.to_string().as_bytes().to_vec(),
```
evgenykuzyakov, line 26:
```suggestion
            args: args.to_string().into_bytes(),
```

evgenykuzyakov (commented):
Looks good. Mostly comments

near-sdk-sim/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+
+[dependencies]
+near-sdk = { path = "../near-sdk", version = "2.0.0"}
+near-crypto = { git = "https://github.com/near/nearcore.git", tag = "stable_wasmer_0.17.2" }
+near-primitives = { git = "https://github.com/near/nearcore.git", tag = "stable_wasmer_0.17.2" }
+lazy_static = "1.4.0"
+near-pool = { git = "https://github.com/near/nearcore.git", tag = "stable_wasmer_0.17.2" }
+near-runtime-configs = { git = "https://github.com/near/nearcore.git", tag = "stable_wasmer_0.17.2" }
+near-store = { git = "https://github.com/near/nearcore.git", tag = "stable_wasmer_0.17.2" }
+node-runtime = { git = "https://github.com/near/nearcore.git", tag = "stable_wasmer_0.17.2" }
```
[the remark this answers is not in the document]
willemneal: Yeah I think we should make a tag for sure.  I also realized that having a github dependency isn't so bad since it's a dev dependency, not that a lot of projects will depend on other contract crates.

near-sdk-sim/README.md
```diff
@@ -0,0 +1,95 @@
...
+          // init method
+          init_method: new(master_account.account_id(), initial_balance.into())
+      );
+    let alice = master_account.create_user("alice".to_string(), to_yocto("100"));
+
+    let res = call!(
+        master_account,
+        contract.transfer(alice.account_id.clone(), transfer_amount.into()),
+        deposit = STORAGE_AMOUNT
+    );
```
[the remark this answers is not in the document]
willemneal: Yes if you only provide positional arguments than you need `..., deposit_amount, gas_amount)`.  If you don't provide either then defaults are used.  Otherwise, you can just specify one or the other this way.

near-sdk-sim/src/outcome.rs
```diff
@@ -0,0 +1,253 @@
...
+}
+
+impl ExecutionResult {
+    #[doc(hidden)]
+    pub fn new(outcome: ExecutionOutcome, runtime: &Rc<RefCell<RuntimeStandalone>>) -> Self {
+        Self { runtime: Rc::clone(runtime), outcome }
+    }
+
+    /// Interpret the SuccessValue as a JSON value
+    pub fn get_json_value(&self) -> serde_json::Result<Value> {
```
willemneal, line 51: @evgenykuzyakov Should these methods also unwrap the value instead of returning a result?

evgenykuzyakov: yep

near-sdk-sim/src/outcome.rs
```diff
@@ -250,6 +245,6 @@ mod tests {
         let mut outcome = ExecutionOutcome::default();
         outcome.status = status;
         let result = outcome_into_result(outcome, &Rc::new(RefCell::new(init_runtime(None).0)));
-        assert_eq!(value, result.get_json_value().unwrap());
+        assert_eq!(value, result.unwrap_json_value().unwrap());
```
evgenykuzyakov, line 248: It shouldn't need `unwrap()`

willemneal: Ah darn!

evgenykuzyakov (approved):

PR #249, opened 2020-11-18
add with_nonce to promise key actions

mattlockyer:
Not sure if this is the right way to implement? Passes CI. Doesn't break anything.

Added:

- add_full_access_key_with_nonce
- add_access_key_with_nonce


evgenykuzyakov (commented):
Looks good. Please add comments to the new methods

mattlockyer:
@evgenykuzyakov comments added.

Are there any additional steps to expose these methods in the contract runtime, or is this sufficient for the following call in a smart contract?
```
 promise.add_access_key_with_nonce(...)
```

frol (approved):
Should we mark old methods as deprecated? Rust has a built-in attribute for that: https://doc.rust-lang.org/edition-guide/rust-2018/the-compiler/an-attribute-for-deprecation.html

mattlockyer:
@frol I'm not sure we want to deprecate them. In fact, if nearcore implements a nonce fix we'll likely deprecate the methods that I just added because contracts/users shouldn't supply nonce for keys. Chain should enforce nonce for repeat use of key or alter receipts, etc... but not sure what that fix will be right now.
CC @bowenwang1996  

@evgenykuzyakov 
How can we push a release of this so I can use it to compile a new contract? Timeline?

evgenykuzyakov:
> How can we push a release of this so I can use it to compile a new contract? Timeline?

For now you can link this commit and repo for testing purposes and compile the contract.

evgenykuzyakov:
Now you just need to rebuild contracts, so they don't take 5 Mb each.

You can check https://github.com/near/near-sdk-rs/pull/236/files#diff-1e93dc714a056b6bc276c323012066287064b7e422d34c6675d0c6da92e2a870 for contract-builder docker instructions

evgenykuzyakov:
Rebuilt examples with the contract helper.

willemneal (approved):

PR #256, opened 2020-12-06
DEMONSTRATION: trying to get another example w/ sim tests

mikedotexe:
This is for demonstration purposes as we try to get simulation tests working for another "example" project here.


mikedotexe:
Hey @willemneal I have updated this repo with some changes that I think would be relevant in terms of using Simulation Testing for gas estimation. In the example, run `./build.sh` (just to be sure) and then `./test.sh` and it'll slowly go through and add 2048 words into an AVL Tree, recording the output in two files of the amount of gas burned during the **add** and **view**.
What's curious is this gets stuck. Usually around ~800 entries:

Screen Shot 2020-12-07 at 9 38 54 PM

Any idea what would make this hang?

PR #258, opened 2020-12-11
feat: Add ProfileData to ExecutionResult

willemneal:
This uses the new updates to nearcore to pass a reference to ProfileData to ApplyState when executing a transaction.


evgenykuzyakov:
@willemneal We can close this one right?

Issue #265, opened 2021-01-14
bug: near-sdk 2.0.0


evgenykuzyakov:
seems the issue with `syn` crate that was somehow triggered by the `borsh` version bump.

For now let's issue `near-sdk 2.0.1` without any changes, because it created a bunch of side effects from nearcore dependency. Let's create a new branch with only `syn` version to be fixed to `"=1.0.57"`

artob:
Previously in nearcore: https://github.com/near/nearcore/pull/3785

evgenykuzyakov:
Published `2.0.1` crate and released a tag

PR #266, opened 2021-01-14
[2.0.1]: Pin syn version to <=1.0.57

evgenykuzyakov:
Fixes: https://github.com/near/near-sdk-rs/issues/265

We'll not merge this, but instead release it as a `2.0.1` patch


willemneal (approved):

evgenykuzyakov:
Published `2.0.1` and released a tag

frol:
> We'll not merge this, but instead release it as a 2.0.1 patch

This is not intuitive. What is the deal-breaker preventing us from merging it?

BTW, it would be more permissive to use `<=1.0.57` instead of `=1.0.57` (e.g. in case if some other dependency requires `<=1.0.50` with the current approach the upstream project will end up with two versions of `syn` crate)

evgenykuzyakov:
> > We'll not merge this, but instead release it as a 2.0.1 patch
> This is not intuitive. What is the deal-breaker preventing us from merging it?

It's more like we can merge the change, but to do this we need to sync up to the master. Better to do this in a separate branch to avoid losing the pointer for the latest code pushed to crates.io as we did with 2.0.0. But since the tag and release is published on github, it should be okay to update it and sync to master and then merge

> BTW, it would be more permissive to use <=1.0.57 instead of =1.0.57 (e.g. in case if some other dependency requires <=1.0.50 with the current approach the upstream project will end up with two versions of syn crate)

Sounds good.

evgenykuzyakov:
I get the following error during tests or just check:
```
rror[E0061]: this function takes 7 arguments but 12 arguments were supplied
   --> near-sdk-sim/src/runtime.rs:312:29
    |
312 |         let result = viewer.call_function(
    |                             ^^^^^^^^^^^^^ expected 7 arguments
313 |             trie_update,
    |             -----------
314 |             self.cur_block.block_height,
    |             ---------------------------
315 |             self.cur_block.block_timestamp,
    |             ------------------------------
316 |             &CryptoHash::default(),
    |             ----------------------
317 |             self.cur_block.epoch_height,
    |             ---------------------------
318 |             &EpochId::default(),
    |             -------------------
319 |             &account_id.to_string(),
    |             -----------------------
320 |             method_name,
    |             -----------
321 |             args,
    |             ----
322 |             &mut logs,
    |             ---------
323 |             self.epoch_info_provider.as_ref(),
    |             ---------------------------------
324 |             PROTOCOL_VERSION,
    |             ---------------- supplied 12 arguments
```

olonho:
Think arguments were packed into the structure.

willemneal (approved):

PR #269, opened 2021-01-16
clarifying note, where does the xxxxxxxContract struct comes from in sim testing

luciotato:


willemneal (approved):
Thanks!

examples/cross-contract-high-level/tests/general.rs
```diff
@@ -3,6 +3,7 @@ use near_sdk_sim::{
     STORAGE_AMOUNT,
 };
 extern crate cross_contract_high_level;
+//Note: the struct xxxxxxContract is created by #[near_bindgen] (near_skd_rs~2.0.4)
```
chadoh, line 6:
```suggestion
//Note: the struct xxxxxxContract is created by #[near_bindgen] (near_sdk_rs~2.0.4)
```

chadoh: Does this need to mention `2.0.4`? Seems possibly overly specific & quickly out-of-date.

chadoh: Actually, how about this:

```suggestion
// Note: the struct xxxxxxContract is created by #[near_bindgen] from near-sdk in combination with near-sdk-sim
```

@willemneal does this seem more accurate? I think it's important to understand that `#[near_bindgen]` only creates this `xxxxxContract` struct when the file is compiled with `near-sdk-sim`.

luciotato: Yes that's a better comment. The key thing is that for new developers trying to understand what the code is doing, the fact that `struct CrossContractContract` is nowhere to be found in the code is baffling. It's important to let the reader know that this is struct generated on the fly by a macro, so they don't get confused when they can't find the struct declaration in the code.

willemneal: Yeah it's the trickiest part for sure.  Currently the new type is created only if the target is not wasm.  So technically you could use it in unit tests too.

mikedotexe (approved):

PR #275, opened 2021-01-24
feat(contract-standards): Adding near-contract-standards library and more tools

### LINKED ISSUE #254: Standards library
We are keep adding more and more standards.
Many of which are surfaced in the examples in the direct way, instead of using a library.

I've worked on a prototype of such library here - https://github.com/ilblackdragon/balancer-near/tree/master/near-lib-rs/src

For all smart contract standrads (like upgradability pattern, factory pattern, fungible token and non-fungible token) it should provide an easy to include functionality (that is already well tested) & testing harnesses.

Additional tools like test tokens, any special testing tools should be also included there.

ilblackdragon:
`near-contract-standards` is a library with implementations of contract standards.

Additional tools include testing and implementation tools that reduce contract code and tests size by 2x.
Also added `env::storage_byte_code()` which right now returns a constant but in the future will be switched to return current storage cost from environment.

Fixes #254


near-std/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "near-std"
```
frol, line 2: I would like to discuss:

Do you believe it makes sense to have a separate crate for this?

Do you think it makes sense to have all the standards in a single crate?

FYI, my initial thought was that this is NEAR implementation of Rust std, so I would prefer to have another name, which will also reflect the fact that it is a crate for contracts (currently I would consider it to be a part of nearcore   rather than near-sdk)

examples/fungible-token/src/lib.rs
```diff
@@ -16,263 +16,65 @@
...
-    /// (`predecessor_id`).
-    /// Requirements:
-    /// * Caller of the method has to attach deposit enough to cover storage difference at the
-    ///   fixed storage price defined in the contract.
-    #[payable]
-    pub fn transfer(&mut self, new_owner_id: AccountId, amount: U128) {
-        // NOTE: New owner's Account ID checked in transfer_from.
-        // Storage fees are also refunded in transfer_from.
-        self.transfer_from(env::predecessor_account_id(), new_owner_id, amount);
+// #[near_bindgen]
```
frol, line 42: How does it work without this attribute?

ilblackdragon: It doesn't work with or without actually. Filed an issue in #276

near-std/Cargo.toml
ilblackdragon: Yes, I think separate crate make sense, containing all official compositional standards - standards that will be extended by other logic and composed (e.g. tokens, upgradability, etc). Some utils can go into sdk crate proper, but it then will end with overloading SDK with tons of utils. I think release schedule can be pretty different between these two crates.

This is a crate I've been using between all my contracts and actually other folks are starting to use them as well: https://github.com/search?l=TOML&q=near-lib&type=Code

My initial name was near-lib. We can call it near-sdk-lib?

frol (commented):
@ilblackdragon near-lib is even more confusing (we had already had `nearlib` which was the JS API client library, and we finally renamed it to `near-api-js` to avoid ambiguity).

> Some utils can go into sdk crate proper, but it then will end with overloading SDK with tons of utils.

SDK should be a one-stop-shop of all the compatible utils by design:

> A software development kit (SDK) is a collection of software development tools in one installable package.

[Wikipedia](https://en.wikipedia.org/wiki/Software_development_kit)

I believe that SDK can have a different release schedule, but given the nature of SDK it should provide or re-export (potentially behind a feature flag) the NEP standards and various helpers.

I propose `near-contract-standards` or it would be even better to split them apart into dedicated crates (e.g. `near-contract-nft-standard`), so it is easy to track the evolution of certain standards (if that is necessary).

frol (approved):

Cargo.toml
```diff
@@ -16,6 +17,6 @@ exclude = [
 
 # Special triple # comment for ci.
 [patch.crates-io]
-### borsh = { git = "https://github.com/near/borsh-rs", branch = "master" }
-### near-vm-logic = { git = "https://github.com/near/nearcore", branch = "master" }
-### near-primitives-core = { git = "https://github.com/near/nearcore", branch = "master" }
+### borsh = { git = "https://github.com/near/borsh", branch = "master" }
+### near-vm-logic = { git = "https://github.com/nearprotocol/nearcore", branch = "master" }
+### near-runtime-fees = { git = "https://github.com/nearprotocol/nearcore", branch = "master" }
```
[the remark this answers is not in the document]
ilblackdragon: merge conflict resolved badly. reverted.

near-sdk/src/test_utils/context.rs
```diff
@@ -56,13 +65,13 @@ impl VMContextBuilder {
         self
     }
 
-    pub fn epoch_height(&mut self, epoch_height: EpochHeight) -> &mut Self {
-        self.context.epoch_height = epoch_height;
+    pub fn block_timestamp(mut self, block_timestamp: u64) -> Self {
```
frol, line 68: Is it intentional that this method does not follow the rest of the methods API? (`self -> Self` instead of `&mut self -> &mut Self`)

ilblackdragon: I changed all of them to Self in the other PR #293 
I don't really understand why `&mut Self` is used, as that really makes that object not very usable.

frol: `&mut Self` eliminates unnecessary copies and simplifies the conditional parameters:

```rust
let mut builder = Builder::new();
if something {
  builder.block_height(10);
}
if something_else {
  builder.epoch_height(10);
}
```

with `Self`:

```rust
let mut builder = Builder::new();
if something {
  builder = builder.block_height(10);
}
if something_else {
  builder = builder.epoch_height(10);
}
```

You can read more about cons and pros: http://web.archive.org/web/20161030171809/https://aturon.github.io/ownership/builders.html (`&mut Self` was recommended)

evgenykuzyakov:
It looks like we should lock our inter-connected crate versions by using `=3.0.0-pre.1`, otherwise the higher version overwrites the dependency. This usually doesn't happen, but may break some stuff if someone relies on near-sdk = `=3.0.0` and we public near-sdk-core with `3.1.0`, then their `near-sdk` create will depend on `near-sdk-core = 3.1.0`, which it shouldn't

But because the previous version was published under `3.0.0-pre-release`, it overwrites `3.0.0-pre.1` for some alphabetical reason.

Just published `near-contract-standards v0.1.0-pre.1` and `near-sdk v3.0.0-pre.2`

mikedotexe (approved):
I really like this approach and organization. One thing I don't know how to do with the way the macros work: pass in metadata fields. I think the Bridge will need this.
Approving, and would like to see how to use these macros in other examples

mikedotexe:
For instance, when I try to use the macro in a public function, it'll break:

```rust
#[near_bindgen]
impl Contract {
    …

    pub fn add_metadata_symbol(&mut self, symbol: String) { //} -> FungibleTokenMetadata {
        near_contract_standards::impl_fungible_token_metadata!(
            Contract,
            String::from("1.0.0"),
            String::from("Example NEAR fungible token"),
            symbol,
            String::from("https://github.com/near/near-sdk-rs/tree/master/examples/fungible-token"),
            24
        );
    }
}

near_contract_standards::impl_fungible_token_core!(
    Contract,
    token
);
```

I get:

```bash
error[E0434]: can't capture dynamic environment in a fn item
  --> /Users/mike/near/near-sdk-rs/near-contract-standards/src/fungible_token/macros.rs:61:9
   |
61 |         #[near_bindgen]
   |         ^^^^^^^^^^^^^^^
   |
   = help: use the `|| { ... }` closure form instead
```

chadoh (approved):
A few nits, but I really like the approach and look forward to people being able to use it and get us more feedback.

near-contract-standards/src/fungible_token/metadata.rs
```diff
@@ -0,0 +1,15 @@
+use near_sdk::serde::Serialize;
+
+#[derive(Serialize)]
+#[serde(crate = "near_sdk::serde")]
+pub struct FungibleTokenMetadata {
+    pub version: String,
+    pub name: String,
+    pub symbol: String,
+    pub reference: String,
```
mikedotexe, line 9: actually, I do think we want to add `reference_hash` as sha256 of `reference`. I do believe this makes sense, like:

    pub reference_hash: [u8; 32],

We have this in the Nomicon NEP here, too:
https://nomicon-pr-158.onrender.com/Standards/Tokens/FungibleTokenMetadata.html#reference-level-explanation

near-contract-standards/src/fungible_token/implementation.rs
```diff
@@ -0,0 +1,256 @@
...
+        let amount = env::attached_deposit();
+        assert_eq!(
+            amount,
+            self.storage_minimum_balance().0,
+            "Requires attached deposit of the exact storage minimum balance"
+        );
+        let account_id =
+            account_id.map(|a| a.into()).unwrap_or_else(|| env::predecessor_account_id());
+        self.internal_register_account(&account_id);
+        AccountStorageBalance { total: amount.into(), available: amount.into() }
```
mikedotexe, line 218: Also want to point out that there's a bit of a confusion on this.
Al believes this should essentially return `available: 0` and believes this method is separate from account registration. He voiced it here: https://github.com/near/NEPs/pull/158#discussion_r582090178
Looking at this method and `storage_withdraw` (where we have `…self.accounts.remove(…`) it's pretty clear these are account registration and unregistration. Just want to connect dots here, as I don't think we've brought this up with Eugene directly.

mikedotexe:
@mfornet and @djsatok for your info, this is where we're at. An example of a `mint` function looks like this after some cursory testing:

```rust
#[near_bindgen]
impl Contract {

    …

    #[payable]
    pub fn mint(&mut self, account_id: ValidAccountId, amount: U128) {
        let deposit = env::attached_deposit();
        if let Some(account_balance) = self.token.accounts.get(&account_id.as_ref()) {
            // Add balance to the account, return deposit
            Promise::new(env::predecessor_account_id()).transfer(deposit);
            let new_balance = account_balance.saturating_add(amount.0);
            self.token.accounts.insert(&account_id.as_ref(), &new_balance);
        } else {
            // Insert the account, use the deposit for storage
            self.token.storage_deposit(Option::from(account_id.clone()));
            self.token.accounts.insert(&account_id.as_ref(), &amount.0);
        }
    }
}
```

examples/fungible-token/src/lib.rs
```diff
@@ -1,553 +1,117 @@
...
-            attached_deposit
-                + Balance::from(initial_storage - current_storage) * STORAGE_PRICE_PER_BYTE
-        };
-        if refund_amount > 0 {
-            env::log(format!("Refunding {} tokens for storage", refund_amount).as_bytes());
-            Promise::new(env::predecessor_account_id()).transfer(refund_amount);
-        }
-    }
-}
+near_contract_standards::impl_fungible_token!(
```
ilblackdragon, line 44: I think this should split into 3 impl still:
for example if contract wants to programmatically set the metadata, would need to implement all these functions separately.
Also in contract that is token + something else, impl of StorageManager would probably be more sophisticated.

ilblackdragon: I'm curious if there is a way to have a generic proxy for traits with composing. I feel like I saw a crate like this.
Something like `#[compose(token)] impl StorageManager for Contract {}`

.gitignore
```diff
@@ -14,3 +14,6 @@ target
 
 # Key pairs generated by the NEAR shell
 neardev
+
+# OS
+.DS_Store
```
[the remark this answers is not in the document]
evgenykuzyakov: Good idea

examples/fungible-token/src/lib.rs
```diff
@@ -1,553 +1,117 @@
...
+    Contract,
+    token,
+    String::from("1.0.0"),
+    String::from("Example NEAR fungible token"),
+    String::from("EXAMPLE"),
+    String::from("https://github.com/near/near-sdk-rs/tree/master/examples/fungible-token"),
```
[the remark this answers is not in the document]
evgenykuzyakov: Reworked metadata

examples/fungible-token/src/lib.rs
```diff
@@ -1,553 +1,117 @@
...
-            output_data_receivers: vec![],
-            epoch_height: 0,
-        }
+    fn get_context(predecessor_account_id: ValidAccountId) -> VMContextBuilder {
+        let mut builder = VMContextBuilder::new();
+        builder
+            .current_account_id(accounts(0))
+            .signer_account_id(accounts(1))
+            .predecessor_account_id(predecessor_account_id);
+        builder
```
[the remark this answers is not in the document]
evgenykuzyakov: It looks like it returns `&mut Self` instead of `Self` or `mut Self`, so we can't do this

near-contract-standards/src/fungible_token/metadata.rs
evgenykuzyakov: Switched to base58 hash

near-contract-standards/src/fungible_token/implementation.rs
evgenykuzyakov: Let's discuss on the meeting tomorrow.

examples/fungible-token/src/lib.rs
evgenykuzyakov: I've decided to remove Metadata macro completely, since it doesn't help much, but instead hides named arguments from the user, so it's easier to mess up.

There were 3 independent macro before combined `impl_fungible_token` macro. Now only `_core` and `_storage` left. Updated the example

near-contract-standards/src/fungible_token/implementation.rs
```diff
@@ -0,0 +1,256 @@
...
+        &mut self,
+        sender_id: AccountId,
+        receiver_id: AccountId,
+        amount: U128,
+    ) -> U128;
+}
+
+#[ext_contract(ext_fungible_token_receiver)]
+pub trait FungibleTokenReceiver {
+    fn ft_on_transfer(&mut self, sender_id: AccountId, amount: U128, msg: String) -> Promise;
```
mikedotexe, line 29: I think we decided on `PromiseOrValue` yes?

evgenykuzyakov: For this Trait it doesn't matter

chadoh: Is it worth documenting it according to the spec, for people who implement their FT-friendly contracts by reading the source?

ilblackdragon: It should be PromiseOrValue. 
Otherwise how does one return that it consumed X amount of tokens out of amount sent?

And yes, please add detailed doc string explaining what is happening there. Or at least linking to the standard.

near-contract-standards/src/fungible_token/receiver.rs
```diff
@@ -0,0 +1,23 @@
...
+    /// - The transferred amount was `500`, but the action in `msg` field has expired and the transfer must be
+    ///   cancelled. The method must return `500` or panic.
+    ///
+    /// Arguments:
+    /// - `sender_id` - the account ID that initiated the transfer.
+    /// - `amount` - the amount of tokens that were transferred to this account in a decimal string representation.
+    /// - `msg` - a string message that was passed with this transfer call.
+    ///
+    /// Returns the amount of unused tokens that should be returned to sender, in a decimal string representation.
+    fn ft_on_transfer(&mut self, sender_id: ValidAccountId, amount: U128, msg: String) -> U128;
```
ilblackdragon, line 22: Hm, sorry I mean should here be PromiseOrValue? How does one chain more stuff after on_trasnfer receival?

near-contract-standards/src/fungible_token/implementation.rs
```diff
@@ -0,0 +1,256 @@
...
+}
+
+impl FungibleToken {
+    pub fn new(prefix: &[u8]) -> Self {
+        let mut this = Self {
+            accounts: LookupMap::new(prefix.to_vec()),
+            total_supply: 0,
+            account_storage_usage: 0,
+        };
+        let initial_storage_usage = env::storage_usage();
```
willemneal, line 60: Does the line above add anything to storage?  If not, perhaps the lines should be switched to be explicit that it's the storage used before running the constructor.

near-contract-standards/src/fungible_token/implementation.rs
```diff
@@ -0,0 +1,256 @@
...
+        let tmp_account_id = unsafe { String::from_utf8_unchecked(vec![b'a'; 64]) };
+        this.accounts.insert(&tmp_account_id, &0u128);
+        this.account_storage_usage = env::storage_usage() - initial_storage_usage;
+        this.accounts.remove(&tmp_account_id);
```
willemneal, line 64: Why is this tmp account added and then deleted?

near-contract-standards/src/fungible_token/implementation.rs
```diff
@@ -0,0 +1,256 @@
...
+        this
+    }
+
+    pub fn internal_deposit(&mut self, account_id: &AccountId, amount: Balance) {
```
willemneal, line 68: Should these internal methods be public?  I know they won't become exported on the Wasm module, just a thought.

near-contract-standards/src/fungible_token/implementation.rs
mikedotexe: to see how much storage it took and keep track of how much each future user will take up

near-contract-standards/src/fungible_token/implementation.rs
```diff
@@ -0,0 +1,256 @@
...
+    fn ft_transfer_call(
+        &mut self,
+        receiver_id: ValidAccountId,
+        amount: U128,
+        memo: Option<String>,
+        msg: String,
+    ) -> Promise {
+        assert_one_yocto();
+        let sender_id = env::predecessor_account_id();
+        let amount = amount.into();
```
willemneal, line 131: A non-rust dev would find this line very strange (or even rust devs).  Though I love that the type inference from how amount used later works, it's still a little confusing.
```suggestion
```

near-contract-standards/src/fungible_token/implementation.rs
```diff
@@ -0,0 +1,256 @@
...
+        self.internal_transfer(&sender_id, receiver_id.as_ref(), amount, memo);
```
willemneal, line 132:
```suggestion
        self.internal_transfer(&sender_id, receiver_id.as_ref(), amount.into(), memo);
```

near-contract-standards/src/fungible_token/implementation.rs
willemneal: Perhaps it makes more sense to remove this line, then `amount.into()` in the function call.  This also removes two `into`s later.

near-contract-standards/src/fungible_token/implementation.rs
```diff
@@ -0,0 +1,256 @@
...
+        receiver_id: ValidAccountId,
+        amount: U128,
+    ) -> U128 {
+        let amount: Balance = amount.into();
+        let sender_id: AccountId = sender_id.into();
+        let receiver_id: AccountId = receiver_id.into();
+
+        // Get the unused amount from the `ft_on_transfer` call result.
+        let unused_amount = match env::promise_result(0) {
+            PromiseResult::NotReady => unreachable!(),
```
willemneal, line 174: This should panic with an error.

near-contract-standards/src/fungible_token/implementation.rs
```diff
@@ -0,0 +1,256 @@
...
+                if let Ok(unused_amount) = near_sdk::serde_json::from_slice::<U128>(&value) {
+                    std::cmp::min(amount, unused_amount.0)
+                } else {
+                    amount
+                }
+            }
+            PromiseResult::Failed => amount,
+        };
+
+        if unused_amount > 0 {
```
willemneal, line 185:
```suggestion
        if 0 < unused_amount {
```

near-contract-standards/src/fungible_token/implementation.rs
```diff
@@ -0,0 +1,256 @@
...
+            let receiver_balance = self.accounts.get(&receiver_id).unwrap_or(0);
+            if receiver_balance > 0 {
```
willemneal, line 187:
```suggestion
            if 0 < receiver_balance {
```

near-contract-standards/src/fungible_token/implementation.rs
```diff
@@ -0,0 +1,256 @@
...
+
+    fn storage_minimum_balance(&self) -> U128 {
+        (Balance::from(self.account_storage_usage) * env::storage_byte_cost()).into()
+    }
+
+    fn storage_balance_of(&self, account_id: ValidAccountId) -> AccountStorageBalance {
+        if let Some(balance) = self.accounts.get(account_id.as_ref()) {
+            AccountStorageBalance {
+                total: self.storage_minimum_balance(),
+                available: if balance > 0 { 0.into() } else { self.storage_minimum_balance() },
```
willemneal, line 250:
```suggestion
                available: if 0 < balance { 0.into() } else { self.storage_minimum_balance() },
```

near-contract-standards/src/fungible_token/implementation.rs
```diff
@@ -0,0 +1,256 @@
...
+        assert_one_yocto();
+        let amount: Balance = amount.unwrap_or_else(|| self.storage_minimum_balance()).into();
+        assert_eq!(
+            amount,
+            self.storage_minimum_balance().0,
+            "The withdrawal amount should be the exact storage minimum balance"
+        );
+        let account_id = env::predecessor_account_id();
+        if let Some(balance) = self.accounts.remove(&account_id) {
+            if balance > 0 {
```
willemneal, line 231:
```suggestion
            if 0 < balance {
```

examples/fungible-token/src/lib.rs
```diff
@@ -1,553 +1,155 @@
...
-    }
-    fn bob() -> AccountId {
-        "bob.near".to_string()
-    }
-    fn carol() -> AccountId {
-        "carol.near".to_string()
-    }
+    const REFERENCE: &str =
+        "https://github.com/near/near-sdk-rs/tree/master/examples/fungible-token";
+    const REFERENCE_HASH: &str = "Aa4hsn9vdMetr2WDvYWduLCFpi6VZqJ3AzDm16VmSibG";
```
willemneal, line 85: How is this computed?  Is this the commit hash?

near-contract-standards/src/fungible_token/implementation.rs
mikedotexe: is there some reason for these suggestions?

willemneal: Yes it was linked above.  It's a code style outlined by [here](https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/style.md#comparisons), which argues that it's better to match the real number line, e.i.  `x < y` is preferred over `y > x`.

frol: I find it hard to read "if zero is less than amount" (0 < amount) while "if amount is greater than zero" sounds organic to me.

@matklad I wonder to hear your input on the rust-analyzer rule.

willemneal: @frol good point.  I'm fine to drop all of the suggestions. Just wanted consistency.

near-contract-standards/src/fungible_token/implementation.rs
```diff
@@ -0,0 +1,256 @@
...
+
+impl FungibleToken {
+    pub fn new(prefix: &[u8]) -> Self {
+        let mut this = Self {
+            accounts: LookupMap::new(prefix.to_vec()),
+            total_supply: 0,
+            account_storage_usage: 0,
+        };
+        let initial_storage_usage = env::storage_usage();
+        let tmp_account_id = unsafe { String::from_utf8_unchecked(vec![b'a'; 64]) };
```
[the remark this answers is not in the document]
evgenykuzyakov: Maybe `repeat` will optimize at the run time. Need to try

near-contract-standards/src/fungible_token/implementation.rs
evgenykuzyakov: `env::storage_usage()` returns the value in bytes that is immediately updates when we modify internal value for inserting the temp account into the structure. We should probably refactor it a little to create a method that explains the logic below.

near-contract-standards/src/fungible_token/implementation.rs
evgenykuzyakov: Yes, because the contract that include `FungibleToken` struct needs to call them (without exporting). And since they are in a separate crate, we can't use `pub(crate)` visibility here.

near-contract-standards/src/fungible_token/implementation.rs
evgenykuzyakov: But it's unreachable, which panics with the error if it would ever happen. This is proper Rust

near-contract-standards/src/fungible_token/implementation.rs
evgenykuzyakov: I agree with @frol here on the readability when comparing `x` to `0`. I think when you use both variables e.g. `x` and `y` this is when `x < y` is better than `y > x`.

examples/fungible-token/src/lib.rs
evgenykuzyakov: It's just a random hash. Was replaced with obscure `vec[1; 32].into()`

ilblackdragon:
Hm.... How does one put the hash of the contract (code) into the contract itself?

near-contract-standards/src/fungible_token/implementation.rs
matklad: `x > 0` is definitely better that `0 < x` most of the time :)

The background behind the rule is that I've found myself doing a lot of bugs in the code involving *multiple* comparisons, like `x < y && x >= z`, or when checking if two intervals overlap. With "always use `<` / `<=`" rule, such comparisons, for me, became less bug prone.
`5 < x && x <= 10` is more intuitive than `x > 5 && x <= 10`. 
`10 < x && x <= 5` is obviously visibly buggy. 

I've clarified the rule to mention *multiple* comparisons in general. 

As usual, common sense should prevail over any written maxims :) Sometimes small wins in readability are outweighed by their discussion.

willemneal:
@ilblackdragon I think you need it to be part of the initialization.

near-contract-standards/src/fungible_token/implementation.rs
willemneal: Ah I see.

mikedotexe:
> Hm.... How does one put the hash of the contract (code) into the contract itself?

The `reference_hash` is the hash of the external JSON file that contains additional, free-form details.
https://nomicon-pr-158.onrender.com/Standards/Tokens/FungibleTokenMetadata.html#reference-level-explanation

near-contract-standards/src/fungible_token/receiver.rs
mikedotexe: done

near-contract-standards/src/fungible_token/implementation.rs
evgenykuzyakov: I've updated it to be `let amount: Balance = amount.into()` to give it a type.

near-contract-standards/src/fungible_token/receiver.rs
evgenykuzyakov: Updated

evgenykuzyakov:
Let's merge it and iterate

PR #283, opened 2021-02-02
[WIP] Contract only

evgenykuzyakov:


evgenykuzyakov:
Worth doing it once we have strict types

PR #288, opened 2021-02-11
[3.0.0] Release

evgenykuzyakov:
Updating to `near-vm-logic` version `3.0.0`

Bumping dependencies to match `near-vm-logic`. Updating `near-sdk-sim` to point to the commit of `near-vm-logic` release.

Rebuilt all examples using docker contract builder.

## Test plan:
- CI
- Verified by releasing `3.0.0-pre-release` and testing a few contracts with both building and unit tests.


near-sdk-sim/Cargo.toml
```diff
@@ -1,22 +1,21 @@
 [package]
 name = "near-sdk-sim"
-version = "2.0.0"
+version = "3.0.0-pre-release"
```
frol, line 3: I am not sure how "-pre-release" version work; I know for sure that "-alpha.N" and "-beta.N" (e.g. "-alpha.1", "-beta.7") work fine

frol (approved):

willemneal (approved):

near-sdk-sim/Cargo.toml
mikedotexe: @matklad do you have a suggestion or thought on this one?

willemneal: This one is still waiting on a release from the node-runtime crate so we can fix this when actually trying to do a prerelease.

matklad: the convention is I think `-pre.1` or `-beta.1`. It has affordance for `-pre.2` if you do need to make changes. 

Generally, it is `-anysuffix`, sever doesn't assign any particular meaning there.

PR #291, opened 2021-02-11
feat: add setup_alloc macro to remove boilerplate and add cfg flag to allocator

willemneal:
`setup_alloc` macro generates the same boilerplate as before, but also adds a `#[cfg(target_arch = "wasm32")]`, which prevents the allocator from being used when the contract's main file is used in simulation testing.


evgenykuzyakov:
Please add it to CHANGELOG.md into PENDING

evgenykuzyakov (approved):

frol (approved):

PR #293, opened 2021-02-21
feat(contract-standards) Switch contract standards to use NEP-141 for FungibleToken

ilblackdragon:
Based on the work https://github.com/near/core-contracts/pull/128


ilblackdragon:
Going to merge it into add-near-std to reduce number of PRs.

PR #296, opened 2021-02-25
fix: bitvec dep issue

willemneal:
A work around for this issue: https://github.com/bitvecto-rs/bitvec/issues/105


frol (approved):
Yeah, I bumped into it on nearcore yesterday as well and fixed exactly as you did.

evgenykuzyakov:
Fixed in another PR

PR #297, opened 2021-02-26
refactor: add more simulation tests to ft example

chadoh:
Add cross-contract simulation tests for fungible token example to demonstrate `ft_transfer_call` behavior


chadoh:
@evgenykuzyakov tests currently failing with this message:

```
running 6 tests
test simulate_transfer_call_when_called_contract_not_registered_with_ft ... ok
test simulate_total_supply ... ok
test simulate_simple_transfer ... ok
test simulate_transfer_call_with_immediate_return_and_no_refund ... ok
thread 'simulate_transfer_call_with_promise_and_refund' panicked at 'assertion failed: `(left == right)`
  left: `950000000000000000000000000`,
 right: `1000000000000000000000000000`', tests/general.rs:195:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test simulate_transfer_call_with_promise_and_refund ... FAILED
test simulate_transfer_call_promise_panics_for_a_full_refund ... ok

failures:

failures:
    simulate_transfer_call_with_promise_and_refund

test result: FAILED. 5 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

Not sure if the error is in my test code or in the contract. See simulation test [here](https://github.com/near/near-sdk-rs/pull/297/files#diff-8e0c6aa876ec9748fe024aeb196d18b71ab8212cf813f2fb1b13b242a5688ecbR162) and `defi` contract code at https://github.com/near/near-sdk-rs/pull/297/files#diff-bfaab05b377a2624e84970f7afe31204a9ef6dd272be06ee08759e20d217a1cc.

examples/fungible-token/tests/contracts/defi/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+        amount: U128,
+        msg: String,
+    ) -> PromiseOrValue<U128> {
+        log!("in ft_on_transfer, msg = {:?}", msg);
+        match msg.as_str() {
+            "take-my-money" => PromiseOrValue::Value(U128::from(0)),
+            _ => {
+                let prepaid_gas = env::prepaid_gas();
+                let account_id = env::current_account_id();
+                me::value_please(msg, &account_id, 0, prepaid_gas / 4).into()
```
chadoh, line 43: Should this use `prepaid_gas / 4`? It seems a bit arbitrary. I stole it from the High Level Cross Contracts example. I'm not sure what else to set it to, and this worked, so I left it.

Zooming out, in the general case, how can a contract developer determine a reasonable `gas` amount to attach to a cross contract call?

willemneal: @amgando and I discussed this the other day.  It would be great to have a `gas_left` from `env`, which you can do currently by subtracting prepaid and from used, but you need to make sure to leave a buffer for the current contract.

evgenykuzyakov: The issue is that promises are not resolved until the end of the execution. And you also need to keep some for writing the serialized contract state back.

PR #304, opened 2021-03-02
Mike/141 movement with sim

mikedotexe:
mike is hurrying


near-contract-standards/src/fungible_token/implementation.rs
```diff
@@ -201,57 +198,52 @@ impl FungibleTokenResolver for FungibleToken {
                 }
             }
         }
-        amount.into()
+        PromiseOrValue::Value(U128::from(amount)).into()
```
chadoh, line 201: Does this need `PromiseOrValue::Value` _and_ `into()`?

PR #305, opened 2021-03-03
ft metadata: make reference & reference_hash optional

chadoh:
@mikedotexe and I believe that these two fields should be optional. If a FT contract includes all other metadata on-chain, there may be no point in including an external reference file.


evgenykuzyakov (commented):
Yo need to update fungible_token example as well

PR #306, opened 2021-03-04
Fix ext_contract with serialize borsh

ilblackdragon:
Doesn't compile when ext_contract on trait with `serialize(borsh)` is used (see repro in PR):
```
error[E0658]: `let` expressions in this position are experimental
   --> src/lib.rs:138:1
    |
138 | #[ext_contract(ext_test)]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #53667 <https://github.com/rust-lang/rust/issues/53667> for more information
    = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: expected expression, found statement (`let`)
   --> src/lib.rs:138:1
    |
138 | #[ext_contract(ext_test)]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: variable declaration using `let` is a statement
    = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0308]: mismatched types
   --> src/lib.rs:138:1
    |
138 | #[ext_contract(ext_test)]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `std::vec::Vec`, found `bool`
    |
    = note: expected struct `std::vec::Vec<u8>`
                 found type `bool`
    = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to 3 previous errors
```


willemneal:
I'm adding a test for the macro using your example.  I get the following generated code:

```rust

pub mod test_ext {
    use super::*;
    use near_sdk::{AccountId, Balance, Gas, Promise};
    use std::string::ToString;
    pub fn test<T: ToString>(
        v: Vec<String>,
        __account_id: &T,
        __balance: near_sdk::Balance,
        __gas: near_sdk::Gas,
    ) -> near_sdk::Promise {
        #[derive(near_sdk :: borsh :: BorshSerialize)]
        struct Input {
            v: Vec<String>,
        }
        let args = Input { v };
        let args = let result = near_sdk :: borsh :: BorshSerialize :: try_to_vec (& args) . expect ("Failed to serialize the cross contract args using Borsh.") ;
        near_sdk::Promise::new(__account_id.to_string()).function_call(
            b"test".to_vec(),
            args,
            __balance,
            __gas,
        )
    }
}

```
The error message highlights `let args = let result ...`. So it seems this is a case where the `let args` is generated twice.  Will have a fix soon.

.idea/$CACHE_FILE$
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
frol, line 1: BTW, you can add a global gitignore config on your laptop, this way you won't commit commonly created files by various tools and IDEs

willemneal:
Pushed the fix!

willemneal (approved):
LGTM.  Some might dislike the `.idea` part, but I think it's okay to have for future users who don't have it is their global `.gitignore`.

.gitignore
```diff
@@ -8,3 +8,5 @@ target
 
 # Key pairs generated by the NEAR shell
 neardev
+
+.idea
```
evgenykuzyakov, line 12: Yeah, @chadoh suggested to use global `.gitignore` since it's not related to the project.

ilblackdragon: I find this suboptimal, as pretty much you are saying that every single developer in the world should have this configured?... Or be very careful about their adding code?

evgenykuzyakov: Yes. It's correct. You add this once for your machine and then don't need to pollute gitignore of every project: https://docs.github.com/en/github/using-git/ignoring-files

The reason is if I don't use `clion` or other `.idea` IDE, then it's not needed for me in .gitignore

ilblackdragon: this seems like a really weird discussion... having extra 5 bytes in the gitignore doesn't cost much and saves globally hours of work seems like a worthy exchange.

chadoh: I've gotten yelled at when I've added `.DS_Store` to the `.gitignore` of other open source projects, which made me think it's a bit of a standard practice to avoid adding every env- and editor-specific file to every project's `.gitignore`

PR #308, opened 2021-03-04
Separate ft-basic/ft-with-metadata examples

mikedotexe:
I know we're currently in the process of implementing this standard:
https://gist.github.com/evgenykuzyakov/09ca17cec161e5b0878454b1afef17f8

but I want to unblock the EVM, Bridge, and DevRel. The first two need a fungible token with no metadata, and DevRel needs a fungible token with metadata set on initialization.

Previously we were only taking `reference` and `reference_hash` which are optional. I've made it so the init function can take all the metadata fields and the optionality matches our latest spec, located on a PR here: https://nomicon-pr-158.onrender.com/Standards/Tokens/FungibleTokenMetadata.html#reference-level-explanation

Updated the simulation tests and unit tests on both basic and metadata implementations. At first I thought I'd make it so there was one FT example, but it's actually less efficient to try to have one example that does both.


examples/fungible-token/ft-with-metadata/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+impl FungibleTokenMetadataProvider for Contract {
+    fn ft_metadata(&self) -> FungibleTokenMetadata {
+        FungibleTokenMetadata {
+            spec: FT_METADATA_SPEC.to_string(),
+            name: self.name.clone().unwrap(),
+            symbol: self.symbol.clone().unwrap(),
+            icon: self.icon.clone(),
+            reference: self.reference.clone(),
+            reference_hash: self.reference_hash.clone(),
+            decimals: self.decimals.clone().unwrap(),
```
evgenykuzyakov, line 97: seems wrong. You need to update `FungibleTokenMetadata` to optional too

mikedotexe:
Closing as I believe we no longer have a need for an example where Metadata isn't used, as per discussions in Discord.

PR #309, opened 2021-03-05
Merge AR and SM standards as discussed today

evgenykuzyakov:
The merge described in the following NEP: https://github.com/near/NEPs/pull/166


near-contract-standards/src/fungible_token/storage_impl.rs
```diff
@@ -0,0 +1,102 @@
...
+            self.internal_register_account(&account_id);
+            let refund = amount - min_balance;
+            if refund > 0 {
+                Promise::new(env::predecessor_account_id()).transfer(refund);
+            }
+        }
+        self.internal_storage_balance_of(&account_id).unwrap()
+    }
+
+    fn storage_withdraw(&mut self, amount: Option<U128>) -> StorageBalance {
```
chadoh, line 71: Should `assert_one_yocto`. @oysterpack agreed with your original leaning that, since this function moves user funds, we should ensure the user confirms. See https://github.com/near/NEPs/pull/166/commits/beb30876f6027ada4db0cf5a63de9aa4e3e086b9

This should probably be done in some `internal_storage_withdraw`, which I don't see here. Or at least, in the core `StorageManagement` implementation, not the specific FT version of StorageManagement.

near-contract-standards/src/fungible_token/storage_impl.rs
```diff
@@ -0,0 +1,102 @@
...
+            self.internal_storage_balance_of(&env::predecessor_account_id())
+        {
+            match amount {
+                Some(amount) if amount.0 > 0 => {
+                    env::panic(b"The amount is greater than the available storage balance");
+                }
+                _ => storage_balance,
+            }
+        } else {
+            env::panic(b"The account does not exist");
```
chadoh, line 82: "account does not exist" sounds to me like "invalid account id". Can we make this "The account is not registered"?

near-contract-standards/src/fungible_token/storage_impl.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    fn storage_withdraw(&mut self, amount: Option<U128>) -> StorageBalance {
+        if let Some(storage_balance) =
+            self.internal_storage_balance_of(&env::predecessor_account_id())
+        {
+            match amount {
+                Some(amount) if amount.0 > 0 => {
+                    env::panic(b"The amount is greater than the available storage balance");
+                }
+                _ => storage_balance,
```
chadoh, line 79: Am I reading this correctly? "If called with `storage_withdraw(amount: [some positive number])`, panic. Otherwise, return `StorageBalance` object, but still don't send predecessor the amount they asked for."

near-contract-standards/src/storage_management/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    /// Withdraw specified amount of available Ⓝ for predecessor account.
+    ///
+    /// This method is safe to call. It MUST NOT remove data.
+    ///
+    /// `amount` is sent as a string representing an unsigned 128-bit integer. If
+    /// omitted, contract MUST refund full `available` balance. If `amount` exceeds
+    /// predecessor account's available balance, contract MUST panic.
+    ///
+    /// If predecessor account not registered, contract MUST panic.
```
chadoh, line 35:
```suggestion
    /// If predecessor account not registered, contract MUST panic.
    ///
    /// MUST require exactly 1 yoctoNEAR attached balance to prevent restricted
    /// function-call access-key call (UX wallet security)
```

As added in https://github.com/near/NEPs/pull/166/commits/beb30876f6027ada4db0cf5a63de9aa4e3e086b9

chadoh (changes requested):

near-contract-standards/src/fungible_token/storage_impl.rs
ilblackdragon: Can we make errors more parsable by software in general?
E.g. format!("ERR_NO_USER:{}", user)

Or have enum of these errors or something?

We have this thing everywhere, which is really hard later to show any meaningful errors in frontends.

near-contract-standards/src/fungible_token/storage_impl.rs
oysterpack: you are reading correctly ... `storage_withdraw` logic is not even close to correct

examples/fungible-token/ft/src/lib.rs
```diff
@@ -62,7 +62,7 @@ impl Contract {
 
 near_contract_standards::impl_fungible_token_core!(Contract, token, on_tokens_burned);
 // near_contract_standards::impl_fungible_token_ar!(Contract, token);
```
willemneal, line 64:
```suggestion
```

near-contract-standards/src/fungible_token/storage_impl.rs
evgenykuzyakov: It's correct for the fugible token, since `max == min` and if the account exists, then we have `available == 0`, so if amount is positive, we panic.

near-contract-standards/src/fungible_token/storage_impl.rs
evgenykuzyakov: I think this should come through the event NEP: https://github.com/near/NEPs/issues/130
It looks like the goal is to use Borsh for events, should we try something new unparsable here?

chadoh (approved):

mikedotexe (approved):
Strangely, even without the `Deserialize` I can see the JSON result using NEAR CLI and hitting the RPC directly.

near-contract-standards/src/storage_management/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+use near_sdk::serde::Serialize;
+
+#[derive(Serialize)]
+#[serde(crate = "near_sdk::serde")]
+pub struct StorageBalance {
+    pub total: U128,
+    pub available: U128,
+}
+
+#[derive(Serialize)]
```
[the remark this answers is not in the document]
mikedotexe: Resolving this. I just added `Deserialize` and used NEAR CLI as well as a direct RPC call.

```bash
http post https://rpc.testnet.near.org jsonrpc=2.0 id=dontcare method=query \
  params:='{
    "request_type": "call_function",
    "finality": "final",
    "account_id": "141.mike.testnet",
    "method_name": "storage_balance_of",
    "args_base64": "eyJhY2NvdW50X2lkIjogIm1pa2UudGVzdG5ldCJ9"
  }'
  ```
  
  And confirm that with and without that, they return the same thing. Weird. Oh well, moving on.

near-contract-standards/src/fungible_token/storage_impl.rs
mikedotexe: I'd like to continue this discussion about error message later and affirm that this isn't blocking for the PR. 🙏🏼

near-contract-standards/src/storage_management/mod.rs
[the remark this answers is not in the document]
evgenykuzyakov: `Deserialize` is needed to deserialize an object, e.g. when `json` is given in method args. For returns we only need `Serialize` to serialize towards json

willemneal (approved):
LGTM.  Was a little confused with `on_account_closed` but realized that the macro is adding `self.`.

PR #315, opened 2021-03-05
Update metadata to match the standard and example to support default metadata

evgenykuzyakov:
Making `reference` and `reference_hash` optional. Updating example to match new metadata.
Allow example to be initialized with a given metadata.


near-contract-standards/src/fungible_token/metadata.rs
```diff
@@ -1,20 +1,29 @@
...
+#[derive(BorshDeserialize, BorshSerialize, Clone, Deserialize, Serialize)]
 #[serde(crate = "near_sdk::serde")]
 pub struct FungibleTokenMetadata {
     pub spec: String,
     pub name: String,
     pub symbol: String,
     pub icon: Option<String>,
-    pub reference: String,
-    pub reference_hash: Base64VecU8,
+    pub reference: Option<String>,
```
mikedotexe, line 14: 👍🏼

examples/fungible-token/ft/src/lib.rs
```diff
@@ -29,23 +29,44 @@ near_sdk::setup_alloc!();
...
+    metadata: FungibleTokenMetadata,
 }
 
 #[near_bindgen]
 impl Contract {
-    /// Initializes the contract with the given total supply owned by the given `owner_id`.
+    /// Initializes the contract with the given total supply owned by the given `owner_id` with
+    /// default metadata (for example purposes only).
+    #[init]
+    pub fn new_default_meta(owner_id: ValidAccountId, total_supply: U128) -> Self {
```
willemneal, line 40: What's the effect of multiple `#[init]` methods?  Can you only initialize with one of them?

near-sdk-sim/src/user.rs
```diff
@@ -319,7 +319,7 @@ pub fn init_simulator(genesis_config: Option<GenesisConfig>) -> UserAccount {
 ///   contract_id: "contract",
 ///   bytes: &TOKEN_WASM_BYTES,
 ///   signer_account: master_account,
-///   init_method: new(master_account_id, initial_balance.into(), "".to_string(), vec![1;32].into())
+///   init_method: new_default_meta(master_account_id, initial_balance.into())
```
willemneal, line 322: Although I learned that doc tests compile individually, it's still great to see them here ensuring that the examples stay up to date.

willemneal (approved):
LGTM, just one question above.

near-contract-standards/src/fungible_token/metadata.rs
```diff
@@ -1,20 +1,29 @@
...
+    pub reference_hash: Option<Base64VecU8>,
     pub decimals: u8,
 }
 
 pub trait FungibleTokenMetadataProvider {
     fn ft_metadata(&self) -> FungibleTokenMetadata;
 }
+
+impl FungibleTokenMetadata {
+    pub fn assert_valid(&self) {
```
mikedotexe, line 24: I think we also need to make sure that someone doesn't just throw in `reference` and expect it to be there. I was under the impression that if you want to use the Metadata standard, you have to include the required fields (`spec`, `name`, `symbol`, `decimals`) and we wouldn't allow them to disregard those.

mikedotexe:
For those wanting to test with NEAR CLI, the command will look something like this:
`near deploy 141.mike.testnet --wasmFile ./res/fungible_token.wasm --initFunction new --initArgs '{"owner_id": "141.mike.testnet", "total_supply": "1000000", "metadata": {"spec": "ft-1.0.0", "name": "Mochi Rewards", "symbol": "MOCHI", "decimals": 19, "reference": "https://example.com/mochi.svg", "reference_hash": "MDIwNmE5Nzg0M2IxYmE0ZmJiMTQ3ZDQ3MjU1MGVjM2I="}}'`

or

`near deploy 141.mike.testnet --wasmFile ./res/fungible_token.wasm --initFunction new_default_meta --initArgs '{"owner_id": "141.mike.testnet", "total_supply": "1000000"}'`

followed by:
    
    near view 141.mike.testnet ft_metadata

examples/fungible-token/ft/src/lib.rs
mikedotexe: You can add multiple macros like this. I believe under the hood the main thing that's happening is the contract doesn't try to retrieve storage and fit it into the struct. Eugene has used this to do fancy contract upgrades by using this tactic.

near-contract-standards/src/fungible_token/metadata.rs
evgenykuzyakov: These are required fields on the structure. What we can enforce is if `reference` is given, then we require `reference_hash`

evgenykuzyakov: I also forgot to enforce SPEC

mikedotexe: Ah I see. Yes, it'll panic and tell you the next field that you forgot to add.

`message: 'Smart contract panicked: panicked at &#39;Failed to deserialize input from JSON.: Error(&quot;missing field &#x60;symbol&#x60;&quot;, line: 1, column: 117)&#39;, ft&#x2F;src&#x2F;lib.rs:35:1',`
even though I also don't have decimals.

Doesn't tell you "you forgot these fields" which isn't as great, but I'm not going to be a stick in the mud.

examples/fungible-token/ft/src/lib.rs
evgenykuzyakov: `init` is a marker for a constructor

mikedotexe (approved):

PR #318, opened 2021-03-06
security bug: check at pub fn callback entry point

luciotato:
We need to check that `pub fn ft_resolve_transfer` is being called from this same contract.


near-contract-standards/src/fungible_token/implementation.rs
```diff
@@ -224,6 +224,8 @@ impl FungibleTokenResolver for FungibleToken {
         receiver_id: ValidAccountId,
         amount: U128,
     ) -> U128 {
+        //check: MUST only be called as .then() callback from this contract
+        assert_eq!(env::predecessor_account_id(), env::current_account_id());
```
ilblackdragon, line 228: there is actually `assert_self();` shortcut in near_sdk.

ilblackdragon (approved):

near-contract-standards/src/fungible_token/implementation.rs
willemneal: You can also use the `#[private]` macro above the method declaration, which adds this check.  Good catch @luciotato!

willemneal (approved):
LGTM!

near-contract-standards/src/fungible_token/implementation.rs
evgenykuzyakov: Good catch. Let's switch to `#[private]` as @willemneal suggested

luciotato:
tried #[private] but I get a compilation error. Maybe you can check that @willemneal

luciotato:
Ok looks like #[private] is being added by the fungible_token macro implementation

The missing #[private] is in a piece of code that never gets used. The macro re-implements that code with #[private] (a few lines to call the internal resolver). 

Also you can't add #[private] to this superfluous code to avoid confusion, because #[private] depends on #[near_bindgen] and this internal `FungibleToken` struct MUST NOT be "bindgended". So the next best thing is to remove the code to avoid confusion

I'll drop this PR because in the current version there's no security bug.

PR #320, opened 2021-03-08
feat: add borrowed mutable and referenced runtime

joshuajbouw:
For the evm tests we require the ability to produce blocks from the account, which wraps the runtime. It would be fantastic if it is possible to get a borrowed `Ref` or `MutRef` so that we can do such.

What this does is exactly that, with the bonus of fixing a potential issue with `Runtime::current_block` which had returned a mutable `Block` and not a referenced `Block`. I'm sure it is ideal for there not to be mutable access to it as it is generated.


willemneal (approved):
LGTM and thanks for adding doc tests!

frol (approved):

evgenykuzyakov:
Looks like @mhalambek fixed the CI. So we can merge this branch now.

PR #323, opened 2021-03-09
Switching to crate version for near-sdk-sim. Fix test schedule to run faster

evgenykuzyakov:
@willemneal


willemneal (approved):

willemneal:
So I have found that breaking up the tests too much is actually slower since a lot of the jobs queue up.  My next try is to break up the tests into just two jobs, with the longer running ones in one and the shorter ones in the other.

Also some questions I have:

- Do we still need the `NEAR_RELEASE` since from now on we'll depend on released versions of nearcore crates?
- Do we need to test with nightly or beta?

willemneal:
Okay so now the max of each job is ~20 minutes.  The current issue is that the 12 Github Action jobs two were still queued so the total time is now ~32 minutes, slightly better than the 75 min we had before.

I just added caching, which might need some more info to cache the examples better.

willemneal:
With caching and only checking stable, testing is now down to ~17 min

willemneal:
So currently because we have multiple PR's being tested the testing since your merge has stalled:

Screen Shot 2021-03-12 at 1 13 27 PM


It's annoying because if there is parallelism it's great, but more jobs mean more queued.  So there is some middle ground that will guarantee a time.

PR #325, opened 2021-03-10
feat: make `user.call` not use PendingContractTx

willemneal:
BREAKING CHANGE.  
The following methods were renamed:

`view` ->  `view_method_call`
`call` -> `function_call`
`deploy_and_init` -> `deploy_and_initialize`

And in their places are new functions which don't take a `PendingContractTx` and rather the arguments directly.


near-sdk-sim/src/user.rs
```diff
@@ -154,18 +154,35 @@ impl UserAccount {
...
+        
+    }
+
+    pub fn call(
+      &self,
+      receiver_id: AccountId,
+      method: &str, 
+      args: &[u8],
+      gas: Gas,
+      deposit: Balance,
```
chadoh, line 178: Could we make these `Option`s so I can pass `None, None` to get the defaults?

frol: Well, `None` should not mean "default". If you want something default in Rust, you have it - `Default` trait (`Gas::default()`, `Balance::default()`), though I would still consider explicit useful constants instead of `default` since I cannot really tell what is the right "default balance" or "default gas", so I suggest we introduce some baseline consts in near-sdk. The default deposit to a function call is usually zero.

near-sdk-sim/src/user.rs
```diff
@@ -154,18 +154,35 @@ impl UserAccount {
...
+      self.call(pending_tx.receiver_id.clone(),
+          &pending_tx.method,
+          &pending_tx.args,
+          gas,
+          deposit,
+      )
+        
+    }
+
+    pub fn call(
```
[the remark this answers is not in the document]
frol: Heads up: This change is going to break the public API. What is the state of `near-sdk-sim` in terms of API stability? Is this fine to introduce breaking changes?

frol (approved):

near-sdk-sim/src/user.rs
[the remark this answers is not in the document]
willemneal: @frol, we are about to publish for the first time so I thought this would be the opportunity to break the API.  As it is written now, you would just need to rename the methods so it's not too bad of a breaking change.

near-sdk-sim/src/user.rs
willemneal: We already have two  `near_sdk_sim::DEFAULT_GAS` and `near_sdk_sim::STORAGE_AMOUNT`. Personally I think it's best to just use defaults in the `call/deploy` macros.

PR #328, opened 2021-03-10
FungibleToken: improve error messages for non-registered accounts

sept-en:
Currently, we are receiving just _"The account is not registered"_ error when there's no registered account and these functions of FT are called:
* `FungibleToken::internal_deposit()`
* `FungibleToken::internal_withdraw()`
* `StorageManagement::storage_deposit()`
* `StorageManagement::storage::withdraw()`

This is not very convenient and could be sometimes not that easy to find out (especially having complex contracts with cross-contracts calls) which account doesn't exist indeed. We recently ran into this problem when writing some sim tests and spent a lot amount of time trying to track the issue. Making these messages more descriptive by specifying the account id solves the issue.

### DIFF
--- a/near-contract-standards/src/fungible_token/core_impl.rs
+++ b/near-contract-standards/src/fungible_token/core_impl.rs
@@ -72,7 +72,7 @@ impl FungibleToken {
     }
 
     pub fn internal_deposit(&mut self, account_id: &AccountId, amount: Balance) {
-        let balance = self.accounts.get(&account_id).expect("The account is not registered");
+        let balance = self.accounts.get(&account_id).expect(format!("The account {} is not registered", &account_id).as_str());
         if let Some(new_balance) = balance.checked_add(amount) {
             self.accounts.insert(&account_id, &new_balance);
             self.total_supply =
@@ -83,7 +83,7 @@ impl FungibleToken {
     }
 
     pub fn internal_withdraw(&mut self, account_id: &AccountId, amount: Balance) {
-        let balance = self.accounts.get(&account_id).expect("The account is not registered");
+        let balance = self.accounts.get(&account_id).expect(format!("The account {} is not registered", &account_id).as_str());
         if let Some(new_balance) = balance.checked_sub(amount) {
             self.accounts.insert(&account_id, &new_balance);
             self.total_supply =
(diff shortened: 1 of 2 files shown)


near-contract-standards/src/fungible_token/storage_impl.rs
```diff
@@ -23,7 +23,7 @@ impl FungibleToken {
                 env::panic(b"Can't unregister the account with the positive balance without force")
             }
         } else {
-            env::log(b"The account is not registered");
+            env::log(format!("The account {} is not registered", &account_id).as_bytes());
```
willemneal, line 26:
```suggestion
            log!("The account {} is not registered", &account_id);
```

near-contract-standards/src/fungible_token/storage_impl.rs
```diff
@@ -80,7 +80,7 @@ impl StorageManagement for FungibleToken {
                 _ => storage_balance,
             }
         } else {
-            env::panic(b"The account is not registered");
+            env::panic(format!("The account {} is not registered", &env::predecessor_account_id()).as_bytes());
```
willemneal, line 83:
```suggestion
            log!("The account {} is not registered", &env::predecessor_account_id());
```

mfornet: This should not be `log!` but `env::panic`

near-contract-standards/src/fungible_token/storage_impl.rs
sept-en: Thanks, this looks better!

near-contract-standards/src/fungible_token/core_impl.rs
```diff
@@ -72,7 +72,7 @@ impl FungibleToken {
     }
 
     pub fn internal_deposit(&mut self, account_id: &AccountId, amount: Balance) {
-        let balance = self.accounts.get(&account_id).expect("The account is not registered");
+        let balance = self.accounts.get(&account_id).expect(format!("The account {} is not registered", &account_id).as_str());
```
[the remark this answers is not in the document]
evgenykuzyakov: This does run `format!` even when expect is not called. Maybe we should encapsulate this into `internal_balance_of`, so you don't have to run `format!` in the common case when the balance is present

PR #330, opened 2021-03-10
Add state existence check to #[init]

evgenykuzyakov:
Introducing a **BREAKING** change to the default behavior of `#[init]`. It will now check that the state doesn't exist.

Fixes: #95
Replacing PR: #227 

Since we haven't published `3.0.0` we can bundle this change together


frol (approved):
LGTM!

willemneal (approved):

PR #334, opened 2021-03-12
fix warnings

Kouprin:
Screen Shot 2021-03-12 at 5 18 12 PM


near-sdk/src/environment/env.rs
```diff
@@ -619,7 +619,7 @@ pub fn promise_result(result_idx: u64) -> PromiseResult {
             PromiseResult::Successful(data)
         }
         2 => PromiseResult::Failed,
-        _ => panic!(RETURN_CODE_ERR),
+        _ => panic!("{}", RETURN_CODE_ERR),
```
evgenykuzyakov, line 622: I think this should be `unreachable!()` instead of `panic!()` and same below

evgenykuzyakov (approved):

willemneal (approved):

PR #340, opened 2021-03-15
contract standards: add ext_fungible_token

robert-zaremba:
When implementing contracts which operate with fungible tokens, it's useful to use `#[ext_contract(ext_fungible_token)]` macro.

This PR adds a canonical implementation of that wrapper to near-contract-standards.

### DIFF
--- a/near-contract-standards/src/fungible_token/core_impl.rs
+++ b/near-contract-standards/src/fungible_token/core_impl.rs
@@ -33,6 +33,25 @@ pub trait FungibleTokenReceiver {
     ) -> PromiseOrValue<U128>;
 }
 
+#[ext_contract(ext_fungible_token)]
+pub trait FungibleTokenContract {
+    fn ft_transfer(&mut self, receiver_id: AccountId, amount: U128, memo: Option<String>);
+
+    fn ft_transfer_call(
+        &mut self,
+        receiver_id: ValidAccountId,
+        amount: U128,
+        memo: Option<String>,
+        msg: String,
+    ) -> PromiseOrValue<U128>;
+
+    /// Returns the total supply of the token in a decimal string representation.
+    fn ft_total_supply(&self) -> U128;
+
+    /// Returns the balance of the account. If the account doesn't exist must returns `"0"`.
+    fn ft_balance_of(&self, account_id: ValidAccountId) -> U128;
+}
+
 /// Implementation of a FungibleToken standard.
 /// Allows to include NEP-141 compatible token to any contract.
 /// There are next traits that any contract may implement:
--- a/near-contract-standards/src/fungible_token/core_impl.rs
+++ b/near-contract-standards/src/fungible_token/core_impl.rs
@@ -39,7 +39,7 @@ pub trait FungibleTokenContract {
 
     fn ft_transfer_call(
         &mut self,
-        receiver_id: ValidAccountId,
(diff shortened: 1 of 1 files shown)


evgenykuzyakov:
I wonder if it adds all methods to contract storage even if you're using only one method. But because they are static, it probably doesn't.

near-contract-standards/src/fungible_token/core_impl.rs
```diff
@@ -33,6 +33,25 @@ pub trait FungibleTokenReceiver {
     ) -> PromiseOrValue<U128>;
 }
 
+#[ext_contract(ext_fungible_token)]
+pub trait FungibleTokenContract {
+    fn ft_transfer(&mut self, receiver_id: AccountId, amount: U128, memo: Option<String>);
```
evgenykuzyakov, line 38: Should we use `ValidAccountId` for consistency? I wish we didn't need to do conversion from `AccountId` to `ValidAccountId`, so you don't have to every operate on just `AccountId` unless you do sting operations.
```suggestion
    fn ft_transfer(&mut self, receiver_id: ValidAccountId, amount: U128, memo: Option<String>);
```

robert-zaremba: yeah, in smart contract, we usually are after validation, so converting back to `ValidAccountId` is usually just an unnecessary step. If you think it's better to have it with a type system, then feel free to commit the suggestion. I don't have a strong opinion.

robert-zaremba:
> I wonder if it adds all methods to contract storage even if you're using only one method. But because they are static, it probably doesn't.

You mean to the binary? Everything what is static and not used  (structures, static functions, const functions) should be removed in the compile time.

near-contract-standards/src/fungible_token/core_impl.rs
evgenykuzyakov: Yeah, let's use `AccountId` then in 2 places below

near-contract-standards/src/fungible_token/core_impl.rs
```diff
@@ -33,6 +33,25 @@ pub trait FungibleTokenReceiver {
...
+
+    fn ft_transfer_call(
+        &mut self,
+        receiver_id: ValidAccountId,
```
evgenykuzyakov, line 42:
```suggestion
        receiver_id: AccountId,
```

near-contract-standards/src/fungible_token/core_impl.rs
```diff
@@ -33,6 +33,25 @@ pub trait FungibleTokenReceiver {
...
+        amount: U128,
+        memo: Option<String>,
+        msg: String,
+    ) -> PromiseOrValue<U128>;
+
+    /// Returns the total supply of the token in a decimal string representation.
+    fn ft_total_supply(&self) -> U128;
+
+    /// Returns the balance of the account. If the account doesn't exist must returns `"0"`.
+    fn ft_balance_of(&self, account_id: ValidAccountId) -> U128;
```
evgenykuzyakov, line 52:
```suggestion
    fn ft_balance_of(&self, account_id: AccountId) -> U128;
```

evgenykuzyakov (changes requested):

evgenykuzyakov (approved):

PR #343, opened 2021-03-16
docs(contract-standards): add clarifying comment

### LINKED ISSUE #342: Fungible token example doesn't transfer Ⓝ with storage_withdraw
It appears that we didn't finish fleshing out the `storage_withdraw` function for the fungible token contract standard.
Particularly here:
https://github.com/near/near-sdk-rs/blob/661a14ead9c98c6a892937531d84418fe2af9baf/near-contract-standards/src/fungible_token/storage_impl.rs#L79

It looks like we're just returning the storage balance instead of using a Promise to send back the requested amount.

chadoh:
Fixes #342

Multiple people, me included, have been confused on their first read of this function. To save future readers from confusion, let's add a comment explaining why this implementation is correct.

### DIFF
--- a/near-contract-standards/src/fungible_token/storage_impl.rs
+++ b/near-contract-standards/src/fungible_token/storage_impl.rs
@@ -68,15 +68,12 @@ impl StorageManagement for FungibleToken {
         self.internal_storage_balance_of(&account_id).unwrap()
     }
 
-    /**
-     * While storage_withdraw normally allows the caller to retrieve `available` balance, the basic
-     * Fungible Token implementation sets storage_balance_bounds.min == storage_balance_bounds.max,
-     * which means available balance will always be 0. So this implementation:
-     *
-     * * panics if `amount > 0`
-     * * never transfers Ⓝ to caller
-     * * returns a `storage_balance` struct iff `amount` is 0
-     */
+    /// While storage_withdraw normally allows the caller to retrieve `available` balance, the basic
+    /// Fungible Token implementation sets storage_balance_bounds.min == storage_balance_bounds.max,
+    /// which means available balance will always be 0. So this implementation:
+    /// * panics if `amount > 0`
+    /// * never transfers Ⓝ to caller
+    /// * returns a `storage_balance` struct if `amount` is 0
     fn storage_withdraw(&mut self, amount: Option<U128>) -> StorageBalance {
         assert_one_yocto();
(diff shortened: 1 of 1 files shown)


near-contract-standards/src/fungible_token/storage_impl.rs
```diff
@@ -68,6 +68,15 @@ impl StorageManagement for FungibleToken {
...
     }
 
+    /**
+     * While storage_withdraw normally allows the caller to retrieve `available` balance, the basic
+     * Fungible Token implementation sets storage_balance_bounds.min == storage_balance_bounds.max,
+     * which means available balance will always be 0. So this implementation:
+     *
+     * * panics if `amount > 0`
+     * * never transfers Ⓝ to caller
+     * * returns a `storage_balance` struct iff `amount` is 0
```
frol, line 78:
```suggestion
     * * returns a `storage_balance` struct if `amount` is 0
```

near-contract-standards/src/fungible_token/storage_impl.rs
```diff
@@ -68,6 +68,15 @@ impl StorageManagement for FungibleToken {
         self.internal_storage_balance_of(&account_id).unwrap()
     }
 
+    /**
```
frol, line 71: Is it meant to be a documentation comment? In rust `///` is used for that and `//` for regular comments. I find `* *` difficult to read and I never observe this style in Rust source code, so I would like to ask you to chose either `///` or `//` for the comment blocks.

frol (approved):

near-contract-standards/src/fungible_token/storage_impl.rs
chadoh: "iff" meaning "[if and only iff](https://en.wikipedia.org/w/index.php?title=Iff&redirect=no)" 😃

near-contract-standards/src/fungible_token/storage_impl.rs
chadoh: My JS-trained hands did this automatically and my editor formatted it as a documentation comment and I figured I'd leave it in and see if anyone hollered 😄

near-contract-standards/src/fungible_token/storage_impl.rs
chadoh: But I don't seem to be using that correctly so I'll change it 🙃

PR #345, opened 2021-03-17
fix(near-sdk-sim): allow borsh serialized arguments in sim macro

willemneal:


near-sdk-core/src/code_generator/trait_item_method_info.rs
```diff
@@ -43,4 +22,28 @@ impl TraitItemMethodInfo {
...
+      let value_ser = match serializer {
+          SerializerType::JSON => quote! {
+              let args = near_sdk::serde_json::to_vec(&args).expect("Failed to serialize the cross contract args using JSON.");
+          },
+          SerializerType::Borsh => quote! {
+              let args = near_sdk::borsh::BorshSerialize::try_to_vec(&args).expect("Failed to serialize the cross contract args using Borsh.");
+          },
+      };
+      
+      quote ! {
```
willemneal, line 43:
```suggestion
      quote! {
```

near-sdk-core/src/code_generator/impl_item_method_info.rs
```diff
@@ -202,37 +201,52 @@ impl ImplItemMethodInfo {
...
+      Some(match acc {
+          None => quote! { #ident_str: #ident },
+          Some(a) => quote! { #a, #ident_str: #ident },
+      })
+  })
+  .unwrap();
+  quote! {
+    let args = near_sdk::serde_json::json!({#args}).to_string().into_bytes();
+  }
+}
```
evgenykuzyakov, line 252: The formatting looks wrong. It uses 2 spaces for some reason instead of 4

near-sdk-core/src/code_generator/impl_item_method_info.rs
```diff
@@ -202,37 +201,52 @@ impl ImplItemMethodInfo {
...
         }
     }
 }
+
+fn json_serialize(attr_signature_info: &AttrSigInfo) -> TokenStream2 {
+  let args: TokenStream2 = attr_signature_info
+  .input_args()
+  .fold(None, |acc: Option<TokenStream2>, value| {
+      let ident = &value.ident;
+      let ident_str = format!("{}", ident.to_string());
```
evgenykuzyakov, line 242: Also why do we need format here?

near-sdk-core/src/code_generator/item_impl_info.rs
```diff
@@ -657,8 +657,36 @@ mod tests {
...
+        assert_eq!(expected.to_string(), actual.to_string());
+    }
+
+    #[test]
+    fn marshall_borsh() {
+        let impl_type: Type = syn::parse_str("Hello").unwrap();
+        let mut method: ImplItemMethod = syn::parse_str(
+r#"
+pub fn borsh_test(&mut self, #[serializer(borsh)] a: String) {}
+"#
```
evgenykuzyakov, line 675: Is it possible to indent?

near-sdk-core/src/code_generator/item_impl_info.rs
```diff
@@ -657,8 +657,36 @@ mod tests {
...
+                  .to_string()
+                  .into_bytes();
+                  near_sdk::PendingContractTx::new_from_bytes(&self.account_id, "method", args, true)
+                }
+        );
+        assert_eq!(expected.to_string(), actual.to_string());
+    }
+
+    #[test]
+    fn marshall_borsh() {
```
evgenykuzyakov, line 670: Can you also add a test for using `#result_serializer(borsh)`

near-sdk-core/src/code_generator/trait_item_method_info.rs
```diff
@@ -43,4 +22,28 @@ impl TraitItemMethodInfo {
             }
         }
     }
+
+    pub fn generate_serialier(attr_sig_info: &AttrSigInfo, serializer: &SerializerType) -> TokenStream2 {
+      let has_input_args = attr_sig_info.input_args().next().is_some();
+      if !has_input_args {
+        return quote! {let args = vec![]; }
+      }
```
evgenykuzyakov, line 30: I think for json we need `b"{}".to_vec()`. Hmm, how did it work before? Do we ever have no args?

near-sdk-core/src/code_generator/item_impl_info.rs
willemneal: This isn't needed because the macros for the simulator transform methods to return a `PendingContractTx`.  Thus the result serialization happens when the contract gets run and then the `ExecutionResult` takes care of deserializing.

near-sdk-core/src/code_generator/trait_item_method_info.rs
willemneal: This update assumes that the local `args` will be a byte vector.  So in the case where there are no arguments this is the default value.  Then the other paths for serializing each create an `args` with the same type.

near-sdk-core/src/code_generator/impl_item_method_info.rs
willemneal: This was the way I got `"ident": ident` for creating the json macro.

near-sdk-core/src/code_generator/impl_item_method_info.rs
```diff
@@ -230,9 +220,26 @@ impl ImplItemMethodInfo {
...
         }
     }
 }
+
+fn json_serialize(attr_signature_info: &AttrSigInfo) -> TokenStream2 {
+    let args: TokenStream2 = attr_signature_info
+        .input_args()
+        .fold(None, |acc: Option<TokenStream2>, value| {
+            let ident = &value.ident;
+            let ident_str = format!("{}", ident.to_string());
```
evgenykuzyakov, line 235: shouldn't `.to_string()` be enough
```suggestion
            let ident_str = ident.to_string();
```

evgenykuzyakov (approved):

near-sdk-core/src/code_generator/impl_item_method_info.rs
evgenykuzyakov: Doesn't `ident.to_string()` already provides a `String`?

PR #346, opened 2021-03-18
change license from dual apache/MIT to GPL as it uses nearcore

### LINKED ISSUE #333: Change license to GPL
near-sdk-rs depends on nearcore which has GPL license, which means it should also be GPL.

mikedotexe:
Fixes #333


willemneal (approved):

evgenykuzyakov:
Does GPL allows building contracts without revealing the source code? If not what's the crates from nearcore that we use that requires GPL?

I'd be against using GPL if we can avoid it. This is worst license than iTunes

evgenykuzyakov:
Ok, it looks like we'll be able to get out of GPL once nearcore is out of GPL which should be high-pri, because it GPL is terrible and most software companies avoid it for a reason. They can't use it internally and that why can't contribute to the open source of the given GPL software

evgenykuzyakov (approved):

PR #347, opened 2021-03-20
fix env::block_timestamp documentation

robert-zaremba:
The current `env::block_timestamp` documentation says that the return value is denominated in nanoseconds. This is wrong. The value is denominated in seconds (did empirical tests).


robert-zaremba:
sorry, my I made a mistake.

PR #348, opened 2021-03-23
[3.0.1]: Publish, fix license to GPL, rebuild

evgenykuzyakov:
- Published `3.0.1-pre.1`. Will publish `3.0.1` once this is approved. (May need to publish nearcore crates as real versions)
- Rebuilt examples using contract builder
- Updated licenses of the crates that depend on `nearcore` crates to `GPL-3.0`


README.md
```diff
@@ -33,7 +33,7 @@
 
 ## Release notes
 
-### Version `3.0.0`
```
frol, line 36: What happen to 3.0.0 version? Why the changelog now belongs to 3.0.1.pre-1? (btw, it seems that it is commonly used `3.0.1-per.1` over `3.0.1.pre-1`)

frol (approved):

willemneal (approved):

README.md
evgenykuzyakov: Oh, that's a typo

frol: Should we just remove the git tag? :thinking:

PR #351, opened 2021-03-31
fix: remove unsound `unsafe impl Sync` from sim's contract cache 

matklad:


willemneal (approved):
Thanks! I picked up this bad habit from earlier examples.

frol (approved):

PR #352, opened 2021-04-05
[3.1.0] Introduce BorshStorageKey

evgenykuzyakov:
### `3.1.0`

* Introduce trait `IntoStorageKey` and updating all persistent collections to take it instead of `Vec<u8>`.
  It's a non-breaking change.
* Introduce a macro derive `BorshStorageKey` that implements `IntoStorageKey` using borsh serialization. Example:
```rust
use near_sdk::BorshStorageKey;

#[derive(BorshSerialize, BorshStorageKey)]
enum StorageKey {
    Records,
    UniqueValues,
}

#[near_bindgen]
#[derive(BorshDeserialize, BorshSerialize, PanicOnDefault)]
pub struct StatusMessage {
    pub records: LookupMap<String, String>,
    pub unique_values: LookupSet<String>,
}

#[near_bindgen]
impl StatusMessage {
    #[init]
    pub fn new() -> Self {
        Self {
            records: LookupMap::new(StorageKey::Records),
            unique_values: LookupSet::new(StorageKey::UniqueValues),
        }
    }
}
```


chadoh:
```rust
#[derive(BorshSerialize)]
enum StorageKey {
    Records,
    UniqueValues,
}

// Implement `BorshStorageKey`, so `StorageKey` gets `IntoStorageKey`
// trait implementation based on `BorshSerialize`.
impl BorshStorageKey for StorageKey {}
```

This feels sort of like two lines of boilerplate. How much work would it require to make the macro StorageKey-specific? Something like this:

```rust
#[derive(BorshStorageKey)]
enum StorageKey {
    Records,
    UniqueValues,
}
```

frol:
@evgenykuzyakov I love it! :+1:

I agree with Chad about the boilerplate, though I am unsure if the derive macro worth it.

Another point, the example is not very compelling; I had to read it twice to understand what exactly is proposed here, and that is for a person who has a prior background with near-sdk-rs and Rust, so I think it is worth exploring the space for a better example.

evgenykuzyakov:
> Another point, the example is not very compelling; I had to read it twice to understand what exactly is proposed here, and that is for a person who has a prior background with near-sdk-rs and Rust, so I think it is worth exploring the space for a better example.

It's explained in the https://github.com/near/near-sdk-rs/pull/178 how to use it more details, but I can provide more context.

> This feels sort of like two lines of boilerplate. How much work would it require to make the macro StorageKey-specific? 

Macro derives requires to introduce a new macro `near-sdk-macro`, but looking at our zero boilerplate goal, it maybe worth it :)

evgenykuzyakov:
Actually, making it with `#[derive]` will still require `BorshSerialize`, which is not looking nice:
```rust
#[derive(BorshStorageKey, BorshSerialize)]
enum StorageKey {
    Records,
    UniqueValues,
}
```

It's also possible to make it with `proc_macro_attribute`, e.g.:
```rust
#[storage_key]
enum StorageKey {
    Records,
    UniqueValues,
}
```
But then the shortest one is probably a macro:
```rust
storage_key! {
    Records,
    UniqueValues,
}
```

But maybe the macro will not be autocompleted nicely.

frol:
```
#[derive(BorshSerialize, BorshStorageKey)]
```

It looks perfectly fine to me, and the least magical. `storage_key!` macro looks short, but it is hard to reason about it.

evgenykuzyakov:
For now, I've decided to go with `storage_key! { ... }`. It does auto-complete for me through macro-extention, so this is not a concern anymore. So long as we can explain how this works, it should be easy to use.

frol (approved):

CHANGELOG.md
```diff
@@ -1,11 +1,45 @@
 # Changelog
 
+## `3.1.0`
+
+* Updated dependencies for `near-sdk`
+* Introduce trait `IntoStorageKey` and updating all persistent collections to take it instead of `Vec<u8>`.
+  It's a non-breaking change.
+* Introduce a macro `storage_key!` that crates a `StorageKey` enum, which can be passed to persistent collections to create unique prefixes. Example:
```
chadoh, line 8:
```suggestion
* Introduce a macro `storage_key!` that creates a `StorageKey` enum, which can be passed to persistent collections to create unique prefixes. Example:
```

chadoh:
I actually prefer `#[derive(BorshSerialize, BorshStorageKey)]` or `#[storage_key]` to the macro. Feels surprising to have a `StorageKey` variable show up "out of nowhere" with the macro version.

evgenykuzyakov:
Ok, let's switch to `#[derive(BorshSerialize, BorshStorageKey)]`

chadoh (approved):
Should probably fix the typo in the CHANGELOG, and I'd prefer we use a less surprising syntax as mentioned above. But I don't feel a need to re-review after these changes are made, and I leave it up to your discretion whether to make further changes at all.

bowenwang1996 (approved):

PR #361, opened 2021-04-12
add NFT to near-contract-standards

chadoh:
fixes near/berry-pickers#2

_work in progress_


chadoh:
No need to review yet, @mikedotexe & @evgenykuzyakov; I failed to mark as Draft when I first created, which made it request review automatically

examples/non-fungible-token/README.md
```diff
@@ -0,0 +1,38 @@
...
+### `1.0.0`
+
+- Switched form using [NEP-21](https://github.com/near/NEPs/pull/21) to [NEP-141](https://github.com/near/NEPs/issues/141).
+
+### `0.3.0`
+
+#### Breaking storage change
+
+- Switching `UnorderedMap` to `LookupMap`. It makes it cheaper and faster due to decreased storage access.
+
```
austinabell, line 38: this should probably be updated?

examples/non-fungible-token/nft/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+};
+use near_contract_standards::non_fungible_token::token::{Token, TokenId};
+use near_contract_standards::non_fungible_token::NonFungibleToken;
+use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
+use near_sdk::collections::LazyOption;
+use near_sdk::json_types::ValidAccountId;
+use near_sdk::{
+    env, near_bindgen, AccountId, BorshStorageKey, PanicOnDefault, Promise, PromiseOrValue,
+};
+use std::collections::HashMap;
```
austinabell, line 29: a weird side effect of the macro that requires importing this, should it not just use the reference itself?

near-sdk/src/collections/unordered_set.rs
```diff
@@ -9,7 +9,7 @@ const ERR_INCONSISTENT_STATE: &[u8] = b"The collection is an inconsistent state.
 const ERR_ELEMENT_SERIALIZATION: &[u8] = b"Cannot serialize element with Borsh";
 
 /// An iterable implementation of a set that stores its content directly on the trie.
-#[derive(BorshSerialize, BorshDeserialize)]
+#[derive(BorshSerialize, BorshDeserialize, Debug)]
```
austinabell, line 12: This debug impl is broken because `Vector` doesn't have the same conditional implementation. The case where it is different is when `expensive-debug` feature is set and `T` is not constrained by `BorshDeserialize`

near-contract-standards/src/non_fungible_token/macros.rs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        #[near_bindgen]
+        impl NonFungibleTokenResolver for $contract {
+            #[private]
+            fn nft_resolve_transfer(
+                &mut self,
+                previous_owner_id: AccountId,
+                receiver_id: AccountId,
+                approved_account_ids: Option<HashMap<AccountId, u64>>,
```
austinabell: This is the line that uses the previously mentioned import, which is probably worthwhile to reference directly here or add a `use` above?

near-contract-standards/src/non_fungible_token/metadata.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub fn assert_valid(&self) {
+        assert_eq!(self.media.is_some(), self.media_hash.is_some());
+        if let Some(media_hash) = &self.media_hash {
+            assert_eq!(media_hash.0.len(), 32, "Media hash has to be 32 bytes");
+        }
+
+        assert_eq!(self.reference.is_some(), self.reference_hash.is_some());
+        if let Some(reference_hash) = &self.reference_hash {
+            assert_eq!(reference_hash.0.len(), 32, "Reference hash has to be 32 bytes");
+        }
```
austinabell, line 60: should there not be any validation of anything else, for example, the timestamps? Or is this only handled off-chain and it's fine if the format is incorrect on chain?

austinabell (commented):
A quick pass-through. Didn't look in-depth at the core details against a spec but will look to do that soon

examples/non-fungible-token/nft/src/lib.rs
mikedotexe: Are you highlighting the `use near_sdk::…` part? Curious what the suggestion here is. I see this is how we did it for the FT example, too.

austinabell: nah just the `HashMap` reference, since you don't directly use it in the code. Let me create a commit to show you what I mean

near-contract-standards/src/non_fungible_token/core_impl.rs
```diff
@@ -0,0 +1,442 @@
...
+
+        let mut this = Self {
+            owner_id: owner_id.into(),
+            extra_storage_in_bytes_per_token: 0,
+            owner_by_id: UnorderedMap::new(owner_by_id_prefix),
+            token_metadata_by_id: if let Some(prefix) = token_metadata_prefix {
+                Some(LookupMap::new(prefix))
+            } else {
+                None
+            },
```
austinabell, line 111:
```suggestion
            token_metadata_by_id: token_metadata_prefix.map(|prefix| Some(LookupMap::new(prefix))),
```

near-contract-standards/src/non_fungible_token/core_impl.rs
```diff
@@ -0,0 +1,442 @@
...
+            tokens_per_owner: if let Some(prefix) = enumeration_prefix {
+                Some(LookupMap::new(prefix))
+            } else {
+                None
+            },
```
austinabell, line 116:
```suggestion
            tokens_per_owner: enumeration_prefix.map(|prefix| Some(LookupMap::new(prefix))),
```

near-contract-standards/src/non_fungible_token/core_impl.rs
```diff
@@ -0,0 +1,442 @@
...
+        token_owner_id: ValidAccountId,
+        token_metadata: Option<TokenMetadata>,
+    ) -> Token {
+        // TODO: charge someone for storage
+        assert_at_least_one_yocto();
+        assert_eq!(env::predecessor_account_id(), self.owner_id, "Unauthorized");
+        if self.token_metadata_by_id.is_some() && token_metadata.is_none() {
+            env::panic(b"Must provide metadata");
+        }
+        if let Some(_) = self.owner_by_id.get(&token_id) {
```
austinabell, line 345:
```suggestion
        if self.owner_by_id.get(&token_id).is_some() {
```
Seems like checking key existence should be a function on unordered_map, this is doing an unnecessary read

near-contract-standards/src/non_fungible_token/core_impl.rs
```diff
@@ -0,0 +1,442 @@
...
+    ) -> bool {
+        // Get whether token should be returned
+        let must_revert = match env::promise_result(0) {
+            PromiseResult::NotReady => unreachable!(),
+            PromiseResult::Successful(value) => {
+                if let Ok(yes_or_no) = near_sdk::serde_json::from_slice::<bool>(&value) {
+                    yes_or_no
+                } else {
+                    true
+                }
```
austinabell, line 397:
```suggestion
               near_sdk::serde_json::from_slice::<bool>(&value).unwrap_or(true)
```

austinabell (commented):
Few more comments as I'm scanning through core impl

austinabell:
I'll jump back in tomorrow to look more in-depth, but one final comment I will note is that there is a lot exposed (basically all modules and fields of structures) and this will definitely be challenging to maintain and update without making breaking changes. Depends on how much technical debt is acceptable if this is a rush to come in.

I'll try to look into why certain things need to be exposed tomorrow.

mikedotexe:
> I'll jump back in tomorrow to look more in-depth, but one final comment I will note is that there is a lot exposed (basically all modules and fields of structures) and this will definitely be challenging to maintain and update without making breaking changes. Depends on how much technical debt is acceptable if this is a rush to come in.
> 
> I'll try to look into why certain things need to be exposed tomorrow.

Curious about this, too. I think there are some `pub` functions that aren't actually exposed methods to external users from, say, NEAR CLI, but are public simply because of code organization. Would love to sync on this.

near-contract-standards/src/non_fungible_token/core_impl.rs
mikedotexe: This was lost during the refactor to TreeMap

near-contract-standards/src/non_fungible_token/core_impl.rs
mikedotexe: This one also is now outdated

examples/non-fungible-token/tests/sim/utils.rs
```diff
@@ -0,0 +1,123 @@
...
+    NFT_WASM_BYTES => "res/non_fungible_token.wasm",
+    TOKEN_RECEIVER_WASM_BYTES => "res/token_receiver.wasm",
+    APPROVAL_RECEIVER_WASM_BYTES => "res/approval_receiver.wasm",
+}
+
+const NFT_ID: &str = "nft";
+const TOKEN_RECEIVER_ID: &str = "token-receiver";
+const APPROVAL_RECEIVER_ID: &str = "approval-receiver";
+
+// TODO: how to export String instead of &str? Way too much `into`/`to_string` with &str.
```
austinabell, line 20: This can be done with [lazy_static](https://docs.rs/lazy_static/1.4.0/lazy_static/) or some equivalent. I don't know if it fixes anything, because then it would always have to clone if being moved, but can revisit this later

examples/non-fungible-token/nft/src/lib.rs
mikedotexe: Now I totally get it, great catch. I just moved the import your highlighted and another I just added (for `U128`) into the `macros.rs` file. So much better. 🤙🏼

near-contract-standards/src/non_fungible_token/macros.rs
mikedotexe: 👍🏼 done, mahalo

near-contract-standards/src/non_fungible_token/enumeration_impl.rs
```diff
@@ -0,0 +1,99 @@
...
+use near_sdk::json_types::{U128, ValidAccountId};
+use crate::non_fungible_token::token::Token;
+use std::collections::HashMap;
+use near_sdk::AccountId;
+
+type TokenId = String;
+
+impl NonFungibleToken {
+    // Helper function used by a enumerations methods
+    // Note: this method is not exposed publicly to end users
```
austinabell, line 12:
```suggestion
    /// Helper function used by a enumerations methods
    /// Note: this method is not exposed publicly to end users
```

near-contract-standards/src/non_fungible_token/enumeration_impl.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub fn enum_get_token(&self, owner_id: AccountId, token_id: TokenId) -> Token {
```
austinabell, line 13:
```suggestion
    fn enum_get_token(&self, owner_id: AccountId, token_id: TokenId) -> Token {
```

near-contract-standards/src/non_fungible_token/enumeration_impl.rs
```diff
@@ -0,0 +1,99 @@
...
+        let metadata = self.token_metadata_by_id.as_ref().unwrap().get(&token_id);
+        let approved_account_ids = self
+            .approvals_by_id.as_ref()
+            .unwrap().get(&token_id).or_else(|| Some(HashMap::new()));
```
austinabell, line 17:
```suggestion
        let approved_account_ids =
            Some(self.approvals_by_id.as_ref().unwrap().get(&token_id).unwrap_or_default());
```

Is this intentionally defaulting the hashmap instead of leaving empty? Why is this done with this and not metadata?

near-sdk/src/collections/unordered_set.rs
mikedotexe: Another good catch. I know my IDE sometimes will suggest things and if you select an option, do modifications for you. My guess is this is what happened to Chad. https://github.com/near/near-sdk-rs/blame/439126d603dc04effd82d53606a24fac68fc95aa/near-sdk/src/collections/unordered_set.rs#L12
I've removed this and tests are passing and all is well.

near-contract-standards/src/non_fungible_token/metadata.rs
mikedotexe: I think we'll let the NFT owners decide this. I could see a case where an NFT might be migrated across the Rainbow Bridge from another L1, for instance, and owners may need the option to input time details without such restrictions.

near-contract-standards/src/non_fungible_token/core_impl.rs
mikedotexe: Get:

>expected `LookupMap<TokenId, TokenMetadata>`, found `Option<LookupMap<<unknown>, <unknown>>>`

near-contract-standards/src/non_fungible_token/enumeration_impl.rs
```diff
@@ -0,0 +1,99 @@
...
+
+        for token_id in token_set.unwrap().iter().skip(from_index_val as usize) {
+            tokens.push(self.enum_get_token(account_id.as_ref().parse().unwrap(), token_id));
+            if has_limit {
+                decrementing_limit -= 1;
+                if decrementing_limit == 0 { break; }
+            }
+        }
+
+        tokens
```
austinabell, line 97:
```suggestion
        let tokens_per_owner = self.tokens_per_owner.as_ref().expect(
            "Could not find tokens_per_owner when calling a method on the enumeration standard.",
        );
        let token_set = if let Some(token_set) = tokens_per_owner.get(account_id.as_ref()) {
            token_set
        } else {
            return vec![];
        };

        let limit = limit.map(|v| v as usize).unwrap_or(usize::MAX);
        assert_ne!(limit, 0, "Cannot provide limit of 0.");

        let start_index: u128 = from_index.map(From::from).unwrap_or_default();
        assert!(token_set.len() as u128 >= start_index);

        token_set
            .iter()
            .skip(start_index as usize)
            .take(limit)
            .map(|token_id| self.enum_get_token(account_id.as_ref().clone(), token_id))
            .collect()
```

austinabell (commented):
sorry for splitting review, GH is buggy

near-contract-standards/src/non_fungible_token/core_impl.rs
austinabell:
```suggestion
            token_metadata_by_id: token_metadata_prefix.map(|prefix| LookupMap::new(prefix)
```
oops, forgot to remove the `Some(..)`, this is it

near-contract-standards/src/non_fungible_token/enumeration_impl.rs
mikedotexe: Good question, would be curious to know more about this. Got me reading more about it here:
https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#1123-1132
This commit suggestion got outdated due to some fmt corrections, but I've pushed it up.

near-contract-standards/src/non_fungible_token/enumeration_impl.rs
```diff
@@ -0,0 +1,99 @@
...
+
+        for (token_id, owner_id) in self.owner_by_id.iter().skip(start_index as usize) {
+            tokens.push(self.enum_get_token(owner_id, token_id));
+            if has_limit {
+                decrementing_limit -= 1;
+                if decrementing_limit == 0 { break; }
+            }
+        }
+
+        tokens
```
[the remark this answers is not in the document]
mikedotexe: wow, what's the `From::from` here, v interesting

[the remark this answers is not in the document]
mikedotexe: Thanks for fleshing out that `.map` sequence. I was trying to do that, then using `fold` and it wasn't cooperating. Appreciate it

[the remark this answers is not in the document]
mikedotexe: I pushed this up, too, except had to remove the `>=` and place back with `>`

near-contract-standards/src/non_fungible_token/core_impl.rs
mikedotexe: Neat, looks like we were missing a trailing `)` but I wanted to push up a commit anyway since I did this exact pattern right below it after adding the enumeration standard. Thanks!

near-contract-standards/src/non_fungible_token/enumeration_impl.rs
mikedotexe: great thanks!

near-contract-standards/src/non_fungible_token/enumeration_impl.rs
austinabell: Well it would just change how this is handled externally, or how it's serialized. It's just the difference between `Some` empty map and `None`. I'll try to find out what the differences are in how it's handled

near-contract-standards/src/non_fungible_token/core_impl.rs
austinabell: yeah that's fair, but there is still no need for the `if let Some(_)` pattern here, so I think my suggestion is still valid

near-contract-standards/src/non_fungible_token/enumeration.rs
```diff
@@ -0,0 +1,48 @@
+use crate::non_fungible_token::token::Token;
+use near_sdk::json_types::{ValidAccountId, U128};
+
+pub trait NonFungibleTokenEnumeration {
```
austinabell, line 4: Just noting that all of these public traits (won't comment on all) have no doc comment. I don't think it matters much, but might be helpful and probably good practice since it's exported

near-contract-standards/src/non_fungible_token/metadata.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub media: Option<String>, // URL to associated media, preferably to decentralized, content-addressed storage
+    pub media_hash: Option<Base64VecU8>, // Base64-encoded sha256 hash of content referenced by the `media` field. Required if `media` is included.
+    pub copies: Option<u64>, // number of copies of this set of metadata in existence when token was minted.
+    pub issued_at: Option<String>, // ISO 8601 datetime when token was issued or minted
+    pub expires_at: Option<String>, // ISO 8601 datetime when token expires
+    pub starts_at: Option<String>, // ISO 8601 datetime when token starts being valid
+    pub updated_at: Option<String>, // ISO 8601 datetime when token was last updated
+    pub extra: Option<String>, // anything extra the NFT wants to store on-chain. Can be stringified JSON.
+    pub reference: Option<String>, // URL to an off-chain JSON file with more info.
+    pub reference_hash: Option<Base64VecU8>, // Base64-encoded sha256 hash of JSON from reference field. Required if `reference` is included.
```
austinabell, line 33: I know this might be an annoying comment, but switching those comments to doc comments and adding a quick description of the two structures would be useful imo

near-contract-standards/src/non_fungible_token/mod.rs
```diff
@@ -0,0 +1,16 @@
...
+pub mod core;
+pub mod core_impl;
+pub mod enumeration;
+pub mod enumeration_impl;
+pub mod macros;
+pub mod metadata;
+pub mod receiver;
+pub mod resolver;
+pub mod token;
+pub mod utils;
```
austinabell, line 13: While I think it makes sense to export some of these, might want to audit which are exported for making it more future proof and maintainable

near-contract-standards/src/non_fungible_token/token.rs
```diff
@@ -0,0 +1,15 @@
...
+pub type TokenId = String;
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(crate = "near_sdk::serde")]
+pub struct Token {
+    pub token_id: TokenId,
+    pub owner_id: AccountId,
+    pub metadata: Option<TokenMetadata>,
+    pub approved_account_ids: Option<HashMap<AccountId, u64>>,
+}
```
austinabell, line 15: I think docs here would be really useful, if the struct is imported, it might not even be clear to a developer what type of token this is

near-contract-standards/src/non_fungible_token/enumeration_impl.rs
mikedotexe: Cool thanks! No sweat on digging deeper unless you feel like it; I was itching the scratch of curiosity. Good thoughts

near-contract-standards/src/non_fungible_token/core_impl.rs
```diff
@@ -0,0 +1,451 @@
...
+            token_metadata_by_id: token_metadata_prefix.map(|prefix| LookupMap::new(prefix)),
+            tokens_per_owner: enumeration_prefix.map(|prefix| LookupMap::new(prefix)),
+            approvals_by_id,
+            next_approval_id_by_id,
+        };
+        this.measure_min_token_storage_cost();
+        this
+    }
+
+    // TODO: does this seem reasonable?
```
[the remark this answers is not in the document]
mikedotexe: I'm quite sure this is correct, but am double-checking with Eugene now.

near-contract-standards/src/non_fungible_token/core_impl.rs
```diff
@@ -0,0 +1,451 @@
...
+        from: &AccountId,
+        to: &AccountId,
+    ) {
+        // update owner
+        self.owner_by_id.insert(token_id, to);
+
+        // if using Enumeration standard, update old & new owner's token lists
+        if let Some(tokens_per_owner) = &mut self.tokens_per_owner {
+            // owner_tokens should always exist, so call `unwrap` without guard
+            let mut owner_tokens = tokens_per_owner.get(from).unwrap();
```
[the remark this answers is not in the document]
mikedotexe: Good idea, and yes, `expect` messages will propagate all the way back to the caller like a panic message does. 👍🏼 
Pushed up that change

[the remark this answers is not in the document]
austinabell: yeah I just mean if the assumption that this can't be hit otherwise a bug is upheld. Awesome!

near-contract-standards/src/non_fungible_token/enumeration.rs
mikedotexe: Thanks for noting this. I went through and added a couple dozen lines of doc comments. I think we'll still need to comb through this whole crate. I really appreciate you pointing this out. I would like to have docs be an integral part of the code review process.

near-contract-standards/src/non_fungible_token/mod.rs
mikedotexe: Are we cool with doing this in a separate pass? I think it's likely we rethink this "zero boilerplate" approach. I did notice when I was writing docs just now that `token`, for instance, didn't make all that much sense in how it showed up in the Rust docs. Kind of hoping we can clear this up in a sweep of all the contract standards at some point.

near-contract-standards/src/non_fungible_token/metadata.rs
mikedotexe: Totally. I got ambitious just now thinking I could Comment All The Things™ and realize this is actually quite a task to do. I'm hoping our whole team can lean into this soon.

near-contract-standards/src/non_fungible_token/core.rs
```diff
@@ -3,6 +3,13 @@ use crate::non_fungible_token::token::{Token, TokenId};
 use near_sdk::json_types::ValidAccountId;
 use near_sdk::PromiseOrValue;
 
+/// Trait used for all non-fungible tokens. The specification for the
+/// [core non-fungible token standard] lays out the reasoning for each method.
+/// It's important to check out [NonFungibleTokenReceiver](crate::non_fungible_token::receiver::NonFungibleTokenReceiver)
+/// and [`crate::non_fungible_token::resolver::NonFungibleTokenResolver`] to
```
austinabell, line 9: maybe also worth it to shorten this link as well (you can even add the alias with the link below if it clutters the comment)

near-contract-standards/src/non_fungible_token/mod.rs
```diff
@@ -1,15 +1,36 @@
...
+/// implemented on a receiving account separate from the NFT contract. This is
+/// part of the core standard.
 pub mod receiver;
+/// Trait for the resolver of an `nft_transfer_call` method call. This method is
+/// implemented on the NFT contract and used to resolve how the transfer-and-call
+/// process went. This is part of the core standard.
 pub mod resolver;
+/// The Token struct for the non-fungible token.
 pub mod token;
+/// NFT utility functions
```
austinabell: if only the trait or specific types are needed by a developer, why expose all modules themselves? would just cause a more annoying import process if it doesn't autocomplete imo.

This just means it will be harder to maintain without breaking change because you can never move anything out of those modules or delete any of them. Might make sense for things like receiver, resolver, token but it's hard for me to justify exporting things like `core`, `core_impl`, `enumeration_impl`

austinabell: also just read https://github.com/near/near-sdk-rs/pull/361#discussion_r640283800 comment. While I think it can definitely be done in another pass, we have to make sure that pass is done before a release or else there would be API breaking changes for this sort of re-export

mikedotexe: Cool thanks, can we sync on this? I believe I could use a little guidance on how it might be best laid out. Super happy to shift this around and thanks for pointing out the breaking change aspect.

near-contract-standards/src/non_fungible_token/core.rs
mikedotexe: good catch, pushed up. thanks

near-contract-standards/src/non_fungible_token/mod.rs
mikedotexe: Resolving conversation as the changes have been moved to https://github.com/near/near-sdk-rs/pull/435

near-sdk/src/utils/mod.rs
```diff
@@ -51,6 +51,11 @@ pub fn assert_one_yocto() {
     assert_eq!(env::attached_deposit(), 1, "Requires attached deposit of exactly 1 yoctoNEAR")
 }
 
+/// Assert that at least 1 yoctoNEAR was attached.
+pub fn assert_at_least_one_yocto() {
```
austinabell, line 55: So one note about this is that we would need to bump the SDK version to be compatible with the NFT standard. If we want to make this accessible ASAP to people using the current version of the SDK, maybe it's worth either redefining in the contract-standards for now or just calling the assert directly?

Also just unclear the greater usage of this, we might want to be careful before adding another public function as #380 was also confusing to me.

mikedotexe (approved):

austinabell (approved):

PR #368, opened 2021-04-21
Add helper for gas value

ilyar:


near-sdk-sim/src/units.rs
```diff
@@ -18,3 +18,30 @@ pub fn to_yocto(value: &str) -> u128 {
         part1
     }
 }
+
+pub fn to_gas(tera_gas: &str) -> u64 {
```
matklad, line 22: An alternative would be 

```rust
pub fn tera_gas(tera_gas: f64) -> u64
```

Long term (definitelly out of scope for this PR), for SDK specifically, I feel like we should have a dedicated `Gas` type and write `Gas::from_tera_f64(1.1)`, to follow std's Duration type: https://doc.rust-lang.org/stable/std/time/struct.Duration.html#method.from_secs_f32.

ilyar:
after https://github.com/near/nearcore/pull/4252
@matklad please check it out

frol:
@ilyar thank you for the contribution! I feel that float precision may cause certain unexpected issues, thus I would suggest considering `giga` helper in addition to `tera`, so one can use `giga(1_100)` in case when they want `1.1 TGas`.

austinabell:
Constants have been added for `ONE_TERA`, which can do any arithmetic and avoids a specific pattern suggested here. Hope this is sufficient, please feel free to open it back up if this is not.

PR #375, opened 2021-04-23
load dump state from testnet/mainnet to init simulator

ailisp:
Details and usage of this is in updated readme

### DIFF
--- a/near-sdk-sim/Cargo.toml
+++ b/near-sdk-sim/Cargo.toml
@@ -21,6 +21,8 @@ near-vm-logic = "=4.0.0-pre.1"
 near-pool = "=0.1.0-pre.1"
 near-store = "=0.1.0-pre.1"
 near-runtime = "=4.0.0-pre.1"
+serde = {version = "1", features=["derive"]}
+serde_json = "1"
 
 lazy-static-include = "3"
 # Temporary workaround see https://github.com/bitvecto-rs/bitvec/issues/105
--- a/near-sdk-sim/README.md
+++ b/near-sdk-sim/README.md
@@ -6,9 +6,9 @@ But the true power of blockchains & smart contracts comes from cross-contract ca
 
 As a first step, you can use this library! With it, you can:
 
-* Test cross-contract calls
-* Profile [gas](https://docs.near.org/docs/concepts/gas) & [storage](https://docs.near.org/docs/concepts/storage-staking) usage for your contract, establishing lower bounds for costs of deployed contracts and rapidly identifying problematic areas prior to deploying.
-* Inspect intermediate state of all calls in a complicated chain of transactions
+- Test cross-contract calls
+- Profile [gas](https://docs.near.org/docs/concepts/gas) & [storage](https://docs.near.org/docs/concepts/storage-staking) usage for your contract, establishing lower bounds for costs of deployed contracts and rapidly identifying problematic areas prior to deploying.
+- Inspect intermediate state of all calls in a complicated chain of transactions
 
 To view this documentation locally, clone this repo and from this folder run `cargo doc --open`.
 
@@ -36,7 +36,6 @@ near-sdk = "=3.1.0"
 
 Note that you need to add the `tag` (or `commit`) of the version.
 
-
 ## Workspace setup
 
 If you want to check gas & storage usage of one Rust contract, you can add the above dependencies to `Cargo.toml` in the root of your project. If you want to test cross-contract calls, we recommend setting up a cargo [workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html). Here's how it works:
@@ -51,7 +50,6 @@ mv contract contract-wrap
 ```
 
 Now in the root of the project (`contract-wrap`), create a new `Cargo.toml`. You'll have to add the normal `[package]` section, but unlike most projects you won't have any `dependencies`, only `dev-dependencies` and a `workspace`:
-
 
 ```toml
 [dev-dependencies]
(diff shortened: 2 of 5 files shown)


near-sdk-sim/README.md
```diff
@@ -461,3 +453,31 @@ pub fn init () {
     let root = init_simulator(Some(genesis));
 }
 ```
+
+## Loading dumped state from testnet/mainnet contract
+
+This is useful when your test depends on real contract state that exists on testnet and mainnet. For example, storage cost and some gas cost on testnet is depends on the real size of the state that your contract used on testnet and mainnet. So download and start simulator with them gives you a more accurate understanding of gas/storage cost. Another use case is you'd like to migrate your contract state format from old struct defintion to new, in this case fetch testnet states and run integration tests locally will give you confident that your migration works without breaking contract methods.
```
matklad, line 459: >in this case fetch testnet states and run integration tests locally will give you confident that your migration works without breaking contract methods

Hey, that's a cool use-case I did think of before!

near-sdk-sim/README.md
```diff
@@ -461,3 +453,31 @@ pub fn init () {
...
+
+First you'll need a tool to download state of given contract. Note, the `view_state` rpc doesn't give you the same format that you can passed to simulator as initial state. The major difference is intitial state need every state record typed and has its account id. You can use this [script](https://github.com/near/repro-near-funcall/blob/master/collect-state-records.js) to download multiple contract account state, at given height, in a format that can be used as initial state of near-sdk-sim:
```
matklad, line 461: >You can use this [script]

That's OK as a first-cut, but it's definitely not the experience we want to ship to users. I think we'll need to teach sdk-sim itself to HTTP testnet. This'll help not only this issue, but also with getting realistic costs.

near-sdk-sim/src/runtime.rs
```diff
@@ -83,6 +85,26 @@ impl GenesisConfig {
...
+        });
+        self
+    }
+
+    pub fn load_state_records(mut self, path: &Path) -> Self {
+        let data = fs::read(path).unwrap();
+        let state: Vec<_> = serde_json::from_slice(&data).unwrap();
+        self.state_records.extend(state);
+        self
+    }
```
matklad, line 107: Unlike the API's in near core, this one is a public function that would be called by the contract developers, so the rules here are different. It makes sense to spend some time polishing the API here. I suggest the following signature:

```
pub fn load_state_records(mut self, path: impl AsRef<Path>) -> io::Result<Self> {
```

* this allows the user to pass a string literal as a path
* it moves unwrap to the call-site, which is especially important for libraries.

matklad: Is `state_records` the best naming here? Is `StateRecord` a concept which is exposed to the contract developers? Or is it some near-core internal detail? If the latter, it might be better to call this just `load_state` or `with_initial_state`?

near-sdk-sim/src/runtime.rs
```diff
@@ -517,4 +539,13 @@ mod tests {
         runtime.force_account_update("root".into(), &bob_account);
         assert_eq!(runtime.view_account(&"root").unwrap().locked, 10000);
     }
+
+    #[test]
+    fn test_load_states() {
+        let mut genesis = GenesisConfig::default()
+            .add_account("simple-state.repro.testnet".to_string(), 165437999999999999999000)
+            .add_account("multiple-state.repro.testnet".to_string(), 165437999999999999999000)
+            .load_state_records(Path::new("./res/state_records.json"));
```
matklad, line 548: I am not feeling good about dumping testing .json in this repo. There's nothing specifically wrong with it, but it is a minor technical dept. For the test, I suggest the following:

* create a Vec<StateRecord> in memory
* dump it to a temporary file
* use temp file path in the test

near-sdk-sim/Cargo.toml
```diff
@@ -21,6 +21,8 @@ near-vm-logic = "=4.0.0-pre.1"
 near-pool = "=0.1.0-pre.1"
 near-store = "=0.1.0-pre.1"
 near-runtime = "=4.0.0-pre.1"
+serde = {version = "1", features=["derive"]}
```
matklad, line 24: Hm, wait, are we actually using serde here?

near-sdk-sim/src/runtime.rs
ailisp: This json isdumped from testnet, and when user use this API he would also do dump a json from testnet then load it. Sounds more like an integration test here? So i just put in the real dumped json

near-sdk-sim/Cargo.toml
ailisp: Hmm i thought i need this to use serde_json, not?

near-sdk-sim/README.md
ailisp: I plan to postpone this until we have a plan of make a near-sdk-sim also a cli, after that this script's functionality can be implemented as part of near-sdk-sim-cli

near-sdk-sim/src/runtime.rs
ailisp: GenericConfig.state_record is public api of near-sdk-sim, so user of near-sdk-sim should be aware of state_records any way? And a mental difference to me is:
- state is a key-value store of the blockchain state
- state_records is a vector of initial state to populate

And the format we take here is vector of records. Does this fit your mental model?

matklad: >GenericConfig.state_record is public api of near-sdk-sim,

Ah, right, its a public API of the simulator. Then the naming is indeed right, but the argument should be changed to be `AsRef`

near-sdk-sim/Cargo.toml
matklad: Not really, no. You only need serde if you want to `derive(Deserialize)`. For `serde_json::from_slice`, you only need `serde_json`.

ailisp: :+! ah that's precise!

near-sdk-sim/src/runtime.rs
ailisp: Cool, AsRef allows pass &str is great

willemneal (approved):
This is awesome!

austinabell:
sim is being deprecated and this functionality exists in workspaces. Feel free to re-open if you'd like to update the PR to get this in

@ChaoticTempest this could be useful to pull relevant docs into workspaces, at a glance it seems like it applies the same

PR #376, opened 2021-04-24
Upgrade rust toolchain

evgenykuzyakov:
- Delete `rust-toolchain`
- Upgrade contract builder to use Rust `1.51.0`
- Fix compilation issue with the macro of ft token. Thanks to @matklad 
- Rebuild examples with the new contract builder
- Update `near-contract-standards` to `3.1.1`

# Test plan:
- CI


rust-toolchain
```diff
@@ -1 +1 @@
-stable-2020-10-08
+stable-2021-03-25
```
matklad, line 1: 1.51.0

matklad: Alternatively, I think we might consider removing this file altogether. For stable, it's not super valuable, and it might be beneficial to use not exactly the specified toolchain. For example, I use `rustup default beta`, so that, when hacking of the projects, I can notice minor breakages.

evgenykuzyakov: Basically, remove specified toolchain from the `near-sdk-rs`, but still use rust version for the `contract-builder`?

matklad: Yup

near-contract-standards/src/fungible_token/macros.rs
```diff
@@ -1,6 +1,6 @@
 #[macro_export]
 macro_rules! impl_fungible_token_core {
-    ($contract: ident, $token: ident, $on_tokens_burned_block: block,) => {
+    ($contract: ident, $token: ident $(, $on_tokens_burned_fn:ident)?) => {
```
mikedotexe, line 3: whoa 👀

mikedotexe (approved):

PR #385, opened 2021-04-28
fix: simulator doesn't fail with long chains

### LINKED ISSUE #377: Time travel simulation
In some cases we want to test things happen only after a specific amount of time, for example:
- in staking pool contract, staking reward is distributed after one epoch
- in berry farm, reward is distributed after a minute

One way of doing this is let near-sdk-sim runtime process N empty blocks. It currently have way of process block but not exposed

### LINKED ISSUE #379: produce more than 5640 blocks in simulator cause SIGSEGV: invalid memory reference
Add in any simulation test, call 
```
    master_account
        .borrow_runtime_mut()
        .produce_blocks(86400)
        .unwrap();
```

On my test, it always fail after block of height 5640 get produced (by add a println after `cur_block.produce`) of this error:
```
thread 'test_staking_reward' panicked at 'not within a catch_unsafe_unwind scope', /home/bo/.cargo/registry/src/github.com-1ecc6299db9ec823/wasmer-runtime-core-near-0.17.1/src/fault.rs:212:10
stack backtrace:
error: test failed, to rerun pass '-p art --test general'

Caused by:
  process didn't exit successfully: `/home/bo/workspace/artificial-coin/art/target/debug/deps/general-b394ce9ac5486239 test_staking_reward --exact --nocapture` (signal: 11, SIGSEGV: invalid memory reference)
```

matklad:
The core issue is that Block is essentially a linked list, so:

* dropping it may SO
* cloning it is linear and may SO

Fix the drop manually and optimize the clone to be O(1)

closes #379


willemneal (approved):
Awesome!

ailisp:
Awesome+1!

ailisp (approved):

near-sdk-sim/src/runtime.rs
```diff
@@ -87,7 +87,7 @@ impl GenesisConfig {
 
 #[derive(Debug, Default, Clone)]
 pub struct Block {
-    prev_block: Option<Box<Block>>,
+    prev_block: Option<Arc<Block>>,
```
bowenwang1996, line 90: Do we actually need `Arc` here? Would `Rc` suffice?

matklad: Good question, I haven't actually thought about this. I'd say that:

* both would probably work
* it doesn't really matter which one we pick here (like, this worked before by just copying the whole thing)
* Arc seems to be a better default choice, as that preserves Sync-ness of the Block

evgenykuzyakov: I wonder if we use `Box` recursively somewhere in nearcore as well

evgenykuzyakov (approved):

near-sdk-sim/src/runtime.rs
matklad: @evgenykuzyakov nerdsniped! Here's the list of recursive types from nearcore: 

https://github.com/rust-analyzer/rust-analyzer/commit/ffd92831327b4cb9c19bc1fc440f2a0eb6842c89

Nothing there seems problematic. 

Though, the tool is rather simplistic, it might've misssed some

PR #386, opened 2021-04-28
docs: add documentation and examples for sdk macros

### LINKED ISSUE #360: Document Macros
View the current Rust docs here:
https://docs.rs/near-sdk/3.1.0/near_sdk/index.html#macros

- [ ] Provide description for `log`. When you click inside log, provide a description and an example. Something like, "You can easily log with a variable and have the log show up in something like NEAR Explorer to look like this…"
- [ ] Provide description for `testing_env` and show example and/or link to a simple example of this. (Perhaps in this repo for the cross contract high level exmaple)
- [ ] Extra credit for documenting other parts of this module that would benefit developers and might be frequently visited.

In the right sidebar under **Epics**, click on the link for info on running Rust docs locally and good examples of documentation.

austinabell:
- Documents `log!` macro
- Documents `testing_env!` macro
- Fixes some warnings relating to docs
- Update `MockedBlockchain` reference to remove the need to import outside of macro. (there might be a better way to handle the SDK references, but I'm trying to keep changes contained) 
- Documents other parts of the utils/test_utils modules (extra credit from #360)

Please let me know thoughts, changes to wording, or any ideas on what is in need of documenting for this issue.

I tried to keep the scope of changes within the utils/test_utils modules, aside from the changes to fix the warnings. Also, the intra doc links don't seem to work on the current version of stable. I couldn't find if docs are published through CI anywhere here, but if there is, might be best to generate on nightly or double-check the links don't whiff.

closes #360


austinabell:
Also, is there a reason for not supporting including only the VM context and config, where the match goes from 1 to 3 params that's worthwhile documenting? Or is this a change acceptable to make in this/another PR?

near-sdk/src/test_utils/test_env.rs
```diff
@@ -32,11 +33,13 @@ pub fn setup_with_config(vm_config: VMConfig) {
...
 }
 
+/// Setup the blockchain interface with a default configuration.
 pub fn setup() {
     setup_with_config(VMConfig::default());
 }
 
 // free == effectively unlimited gas
+/// Sets up the blockchain interface with a [`VMConfig`] which sets the gas costs to zero.
 pub fn setup_free() {
```
austinabell, line 43: Didn't want to add a comment on this, but this doesn't update the runtime configuration, which also contains gas costs. I still haven't resolved when each is used or why they are separate, can someone shed some light?

mikedotexe:
Thank you, @austinabell ! You are very appreciated. Today is my last day of PTO, I'll be looking into this more tomorrow, hopefully with more stable internet.

near-sdk/src/test_utils/mod.rs
```diff
@@ -7,10 +7,68 @@ mod context;
...
+/// [`RuntimeFeesConfig`]: crate::RuntimeFeesConfig
+/// [`AccountId`]: crate::AccountId
+/// [`Balance`]: crate::Balance
+/// [`PromiseResult`]: crate::PromiseResult
+/// [`HashMap`]: std::collections::HashMap
 #[macro_export]
 macro_rules! testing_env {
-    ($context:expr, $config:expr, $fee_config:expr, $validator:expr, $promise_results:expr) => {
-        near_sdk::env::set_blockchain_interface(Box::new(MockedBlockchain::new(
+    ($context:expr, $config:expr, $fee_config:expr, $validators:expr, $promise_results:expr $(,)?) => {
```
mikedotexe, line 70: Could I get some knowledge share on what's going on with `$(,)?`?

near-sdk/src/utils/mod.rs
```diff
@@ -2,6 +2,34 @@ pub(crate) mod storage_key_impl;
 
 use crate::{env, AccountId, PromiseResult};
 
+/// Helper macro to log a message through [`env::log`].
```
mikedotexe, line 5: ❤️

mikedotexe (approved):

near-sdk/src/test_utils/mod.rs
austinabell: So what that syntax does is allow an optional comma to trail the last element in the macro. This does not functionally change anything, but when the macro is spread over a few lines, having the trailing comma is inline with rust fmt default which includes the trailing comma on multiline.

Is not a breaking change, just allows a more flexible syntax for a user.

near-sdk/src/test_utils/mod.rs
```diff
@@ -7,10 +7,68 @@ mod context;
...
+/// - `promise_results`(optional): a [`Vec`] of [`PromiseResult`] which mocks the results
+/// of callback calls during the execution.
+///
+/// Any argument not included will use the default implementation of each.
+///
+/// # Example use
+///
+/// ```
+/// #[macro_use]
+/// extern crate near_sdk;
```
matklad, line 34: This ideally should be `use near_sdk::testing_env`, `extern crate` shouldn't be used in modern Rust code. Though, it also requires that the macro uses `$crate` internally and doesn't just rely on everything being in scope.

near-sdk/src/utils/mod.rs
```diff
@@ -2,6 +2,34 @@ pub(crate) mod storage_key_impl;
...
+/// log!("test");
+/// let world: &str = "world";
+/// log!(world);
+/// log!("Hello {}", world);
+/// log!("x = {}, y = {y}", 10, y = 30);
+/// # }
+/// ```
+///
+/// [`env::log`]: crate::env::log
+/// [`std::format`]: https://doc.rust-lang.org/std/macro.format.html
```
matklad, line 31:
```suggestion
```

Intra doc links work for std. You might be able to get rid of some other explicit links as well.

near-sdk/src/utils/mod.rs
```diff
@@ -63,6 +91,7 @@ impl PendingContractTx {
 }
 
 /// Boilerplate for setting up allocator used in Wasm binary.
+/// This initializes the allocator with [`WeeAlloc`](crate::wee_alloc::WeeAlloc).
```
matklad, line 94: Let's maybe rephrase this to something like `Sets up the [global allocator](link to std) to WeeAlloc`. The current phrasing lacks the explanation/like for what an allocator is.

near-sdk/src/test_utils/mod.rs
austinabell: there was a reason I couldn't do that, I'll try again to see if I was wrong or to pinpoint what that issue was

matklad: Yeah, it's true that `near_sdk` could be written in such a way that extern crate is required. I *think* it should be possible to fix that in a backwards compatible way. 

I think this'll do the trick:

```rust
diff --git a/near-sdk/src/test_utils/mod.rs b/near-sdk/src/test_utils/mod.rs
index 3506e20..a8ac175 100644
--- a/near-sdk/src/test_utils/mod.rs
+++ b/near-sdk/src/test_utils/mod.rs
@@ -68,12 +68,12 @@ use near_vm_logic::mocks::mock_external::Receipt;
 #[macro_export]
 macro_rules! testing_env {
     ($context:expr, $config:expr, $fee_config:expr, $validators:expr, $promise_results:expr $(,)?) => {
-        near_sdk::env::set_blockchain_interface(Box::new(near_sdk::MockedBlockchain::new(
+        $crate::env::set_blockchain_interface(Box::new($crate::MockedBlockchain::new(
             $context,
             $config,
             $fee_config,
             $promise_results,
-            match near_sdk::env::take_blockchain_interface() {
+            match $crate::env::take_blockchain_interface() {
                 Some(mut bi) => bi.as_mut_mocked_blockchain().unwrap().take_storage(),
                 None => Default::default(),
             },
@@ -82,14 +82,14 @@ macro_rules! testing_env {
         )));
     };
     ($context:expr, $config:expr, $fee_config:expr, $validators:expr $(,)?) => {
-        testing_env!($context, $config, $fee_config, $validators, Default::default());
+        $crate::testing_env!($context, $config, $fee_config, $validators, Default::default());
     };
 
     ($context:expr, $config:expr, $fee_config:expr $(,)?) => {
-        testing_env!($context, $config, $fee_config, Default::default());
+        $crate::testing_env!($context, $config, $fee_config, Default::default());
     };
     ($context:expr) => {
-        testing_env!($context, Default::default(), Default::default());
+        $crate::testing_env!($context, Default::default(), Default::default());
     };
 }
 
```

austinabell: hmm, yeah I don't see an issue with this. I remember I tried this, but I might have scrapped because I wasn't sure on the implications and didn't want to have any breaking changes. I'll commit this change now

matklad (commented):
LGTM!

makes sense to grep the core for the remaining `extern crate`s, if there are any

evgenykuzyakov (approved):

Issue #388, opened 2021-04-29
near-sdk-sim calculates storage incorrectly 


skyward-dev:
For example, I'm trying to deploy wrap near contract with 10 NEAR. It works on the testnet, but doesn't work in the simulator tests.
```
let w_near = near.deploy_and_init(
            &W_NEAR_WASM_BYTES,
            WRAP_NEAR_ID.to_string(),
            "new",
            b"{}",
            to_yocto("10"),
            BASE_GAS,
        );
```
It looks like the simulator uses different storage per byte and fails with the following:
```
Outcome ExecutionOutcome {
    logs: [],
    receipt_ids: [
        `7qy1rGatQE1fCVd7V3neku5Hm9dvakY2LHiC8L3jERxh`,
        `5CBACqrLyeTEeVr3wB5dnyZ2LkYy2LAhospjfSMJjW9Z`,
    ],
    burnt_gas: 4673903491021,
    tokens_burnt: 467390349102100000000,
    status: Failure(Action #0: The account wrap.near wouldn't have enough balance to cover storage, required to have 6896582900000000000000000 yoctoNEAR more),
} was a failure
```

evgenykuzyakov:
Hmm. It looks like it uses old protocol version for some reason. Let me investigate.
cc @willemneal Have you seen it before?

willemneal:
@evgenykuzyakov Yeah I think we need to republish the nearcore crates.

evgenykuzyakov:
I think the issue is the `RuntimeStandalone` doesn't upgrade the `RuntimeConfig`.
It's done in `RuntimeConfig::getConfig(config, protocol_version)` (or smth like that), which we don't call. We probably just need to call it at start and give default config.

evgenykuzyakov:
Publishing [3.2.0] now. It should have correct storage cost per byte.

PR #390, opened 2021-04-30
near-sdk-sim 3.2.0: Expose cur block and fix storage costs

evgenykuzyakov:
* Expose `cur_block` and `genesis_config` from `RuntimeStandalone`. This allows to manipulate block time, e.g. set to a given timestamp.
* Use `RuntimeConfig::from_protocol_version` that fixes storage costs issue. Fixes: #388 
* Set root account balance to one billion tokens.

## Test plan:
- Modify FT test to verify storage cost fix by limiting initial deposit to 5 NEAR. It is not enough with old storage cost (50kb) and enough with new storage cost (500kb)


near-sdk-sim/src/runtime.rs
```diff
@@ -268,7 +271,7 @@ impl RuntimeStandalone {
             random_seed: Default::default(),
             epoch_id: EpochId::default(),
             current_protocol_version: PROTOCOL_VERSION,
-            config: Arc::from(self.genesis.runtime_config.clone()),
+            config: RuntimeConfig::from_protocol_version(&self.runtime_config, PROTOCOL_VERSION),
```
matklad, line 274: The thing with lazy_static that happens inside `RuntimeConfig::from_protocol_version` feels a bit odd. Seems like tests might start with different genesiss_config, and then config of one test will override config of the other?

matklad: :thinking: I think that's exactly the reason why https://github.com/near/nearcore/pull/4263 fails

matklad: Added a fix for that to https://github.com/near/nearcore/pull/4263 (not that we can immediately use that in the SDK)

evgenykuzyakov: It was mostly for mainnet upgrade. I guess we can also override Default for GenesisConfig to use this instead of applying it in the produce block. This way the user can override the config if needed.

near-sdk-sim/src/runtime.rs
```diff
@@ -146,17 +146,18 @@ impl Block {
...
     outcomes: HashMap<CryptoHash, ExecutionOutcome>,
     profile: HashMap<CryptoHash, ProfileData>,
-    cur_block: Block,
+    pub cur_block: Block,
     runtime: Runtime,
     tries: ShardTries,
     pending_receipts: Vec<Receipt>,
     epoch_info_provider: Box<dyn EpochInfoProvider>,
     pub last_outcomes: Vec<CryptoHash>,
+    pub runtime_config: Arc<RuntimeConfig>,
```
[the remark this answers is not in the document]
evgenykuzyakov: Since this is for tests only, it worth give more control to the user. `cur_block` should be exposed as is and there is no reason to hide it from the user. The main idea is you should have granular control over block time. For example when testing lockup contracts we need to check what would be the balance after a year. For staking pool we want to have access to epoch ID. But if we rely only on produce block advancing this variables, it's hard to write assert statements because you either should replicate math in tests, or just hardcode the output without verifying math. But if you have precise time control, then you can just use round numbers.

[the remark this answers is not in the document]
austinabell: yeah I was just curious what the interactions would look like to explore if there is an alternative that's easier to maintain/use

mikedotexe (approved):
This is perfect timing for something I'm helping with on the weekends.

PR #391, opened 2021-04-30
feat(json_types): implement FromStr for ValidAccountId

austinabell:
Just saw https://stackoverflow.com/questions/67332443/how-to-convert-accountid-into-validaccountid-in-near-protocol-contracts

and this makes it easier to convert as it doesn't require importing the `TryFrom` or `TryInto` trait

Also changes the logic a bit to avoid allocating a `String` from `&str` before it's checked as valid.

### DIFF
--- a/near-sdk/src/json_types/account.rs
+++ b/near-sdk/src/json_types/account.rs
@@ -6,9 +6,6 @@ use std::fmt;
 use crate::env::is_valid_account_id;
 use crate::AccountId;
 
-// TODO: this should probably be a specific error type instead of a string.
-const INVALID_ACCOUNT_ID_MSG: &str = "The account ID is invalid";
-
 /// Helper class to validate account ID during serialization and deserializiation.
 /// This type wraps an [`AccountId`].
 ///
@@ -61,11 +58,11 @@ impl TryFrom<&str> for ValidAccountId {
     }
 }
 
-fn validate_account_id(id: &str) -> Result<(), Box<dyn std::error::Error>> {
+fn validate_account_id(id: &str) -> Result<(), ParseAccountIdError> {
     if is_valid_account_id(id.as_bytes()) {
         Ok(())
     } else {
-        Err(INVALID_ACCOUNT_ID_MSG.into())
+        Err(ParseAccountIdError::InvalidAccountId)
     }
 }
 
@@ -92,6 +89,22 @@ impl From<ValidAccountId> for AccountId {
         value.0
     }
 }
+
+#[non_exhaustive]
+#[derive(Debug)]
+pub enum ParseAccountIdError {
+    InvalidAccountId,
+}
+
+impl std::fmt::Display for ParseAccountIdError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::InvalidAccountId => write!(f, "The account ID is invalid"),
+        }
+    }
+}
+
+impl std::error::Error for ParseAccountIdError {}
 
 #[cfg(test)]
 mod tests {
--- a/near-sdk/src/json_types/account.rs
+++ b/near-sdk/src/json_types/account.rs
@@ -2,7 +2,6 @@ use borsh::{BorshDeserialize, BorshSerialize};
 use serde::{de, Deserialize, Serialize};
 use std::convert::TryFrom;
 use std::fmt;
-use std::str::FromStr;
 
 use crate::env::is_valid_account_id;
 use crate::AccountId;
@@ -58,7 +57,7 @@ impl TryFrom<&str> for ValidAccountId {
     type Error = Box<dyn std::error::Error>;
 
     fn try_from(value: &str) -> Result<Self, Self::Error> {
-        Self::from_str(value)
+        value.parse()
     }
 }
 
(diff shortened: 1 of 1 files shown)


near-sdk/src/json_types/account.rs
```diff
@@ -1,21 +1,31 @@
...
 #[derive(
     Debug, Clone, PartialEq, PartialOrd, Ord, Eq, BorshDeserialize, BorshSerialize, Serialize,
 )]
 pub struct ValidAccountId(AccountId);
 
 impl ValidAccountId {
-    fn is_valid(&self) -> bool {
-        is_valid_account_id(&self.0.as_bytes())
-    }
     pub fn to_string(&self) -> String {
```
austinabell, line 29: Also, this function does nothing, since `to_string()` comes with `ToString` which is auto impl from `fmt::Display`. I kept it in though because technically it's a breaking change although I don't know if you can actually trigger it because `ValidAccountId::to_string(&id)` still works.

frol (approved):
This makes total sense to me :+1:

near-sdk/src/json_types/account.rs
matklad: Yeah, we should remove this once we are ready to to a semver-breaking change.

near-sdk/src/json_types/account.rs
```diff
@@ -33,35 +43,47 @@ impl AsRef<AccountId> for ValidAccountId {
...
+    fn deserialize<D>(deserializer: D) -> Result<Self, <D as de::Deserializer<'de>>::Error>
     where
-        D: serde::Deserializer<'de>,
+        D: de::Deserializer<'de>,
     {
-        let s = <String as serde::Deserialize>::deserialize(deserializer)?;
-        s.try_into()
-            .map_err(|err: Box<dyn std::error::Error>| serde::de::Error::custom(err.to_string()))
+        let s: String = Deserialize::deserialize(deserializer)?;
+        Self::try_from(s).map_err(de::Error::custom)
```
matklad, line 52:
```suggestion
        s.parse().map_err(de::Error::custom)
```

I imagine we want to remove TryFrom one day

near-sdk/src/json_types/account.rs
```diff
@@ -33,35 +43,47 @@ impl AsRef<AccountId> for ValidAccountId {
...
     type Error = Box<dyn std::error::Error>;
 
     fn try_from(value: String) -> Result<Self, Self::Error> {
-        let res = Self(value);
-        if res.is_valid() {
-            Ok(res)
-        } else {
-            Err("The account ID is invalid".into())
-        }
+        validate_account_id(value.as_ref())?;
```
matklad, line 76:
```suggestion
        validate_account_id(value.as_str())?;
```

To keep this more specific. In general, `as_ref` is a comaptability hazard when used on concrete types, as new impl of AsRef might be added in the future and can regress type inference. That is, use `AsRef` when you have `T: AsRef`, otherwise prefer concrete functions. 

To clarify, this is just a rule of thumb, ther's no specific problems in this case.

near-sdk/src/json_types/account.rs
```diff
@@ -1,21 +1,31 @@
...
+/// This type wraps an [`AccountId`].
+///
+/// # Example
+/// ```
+/// use near_sdk::AccountId;
+/// use near_sdk::json_types::ValidAccountId;
+///
+/// let id: AccountId = "bob.near".to_string();
+/// let validated: ValidAccountId = id.parse().unwrap();
+/// ```
```
matklad, line 22: :heart: , examples are great, we need more of those!

near-sdk/src/json_types/account.rs
```diff
@@ -1,21 +1,31 @@
 use borsh::{BorshDeserialize, BorshSerialize};
-use serde::Serialize;
-use std::convert::{TryFrom, TryInto};
+use serde::{de, Deserialize, Serialize};
 use std::fmt;
+use std::{convert::TryFrom, str::FromStr};
```
matklad, line 4: Nit, but let's keep uses separate, to be consistent with the rest of the code. z

matklad (approved):

near-sdk/src/json_types/account.rs
austinabell: I kept it here because we have a string, and we don't want to call `FromStr` because it forces the string to be reallocated after. 

Sure, this clone might be optimized out but I was airing on the side of not having that be possible for now.

near-sdk/src/json_types/account.rs
austinabell: That's a really good point, I'll do that now!

near-sdk/src/json_types/account.rs
matklad: Ah, yeah, you are absolutely correct!

evgenykuzyakov (approved):

near-sdk/src/json_types/account.rs
```diff
@@ -87,7 +121,7 @@ mod tests {
 
     #[test]
     fn test_ser() {
-        let key: ValidAccountId = "alice.near".try_into().unwrap();
+        let key: ValidAccountId = "alice.near".parse().unwrap();
```
mikedotexe, line 124: Cool, this is new to me. 👍🏼

mikedotexe (approved):

PR #392, opened 2021-05-03
fix(collections): range panic on lo/hi excluded and flipped

austinabell:
In the case of lo > hi and both are excluded, this does not panic as the function comment suggests.

If this is intended, let me know and I can update the docs or add docs on the `Cursor` to indicate it will panic in this case (if it does).


matklad:
sgtm, do we have tests for this?

austinabell:
> sgtm, do we have tests for this?

Is now! Not sure why I didn't check before

evgenykuzyakov (approved):

mikedotexe:
I didn't catch this. Great attn to detail, and moar tests always good.

PR #393, opened 2021-05-03
refactor: cleanup legacy tree map

austinabell:
- remove unnecessary clones
- Clean up logic blocks to be more readable/concise
- Remove unnecessary explicit lifetimes

I tried to make the changes the least invasive as possible and just making a quick pass-through. There should be no logical changes here

### DIFF
--- a/near-sdk/src/collections/legacy_tree_map.rs
+++ b/near-sdk/src/collections/legacy_tree_map.rs
@@ -182,6 +182,8 @@ where
         Cursor::range(&self, lo, hi).into_iter()
     }
 
+    /// Helper function which creates a [`Vec<(K, V)>`] of all items in the [`LegacyTreeMap`].
+    /// This function collects elements from [`LegacyTreeMap::iter`].
     pub fn to_vec(&self) -> Vec<(K, V)> {
         self.iter().collect()
     }
@@ -197,7 +199,7 @@ where
         let mut parent: Option<Node<K>> = self.node(p);
         loop {
             let node = self.node(at);
-            match node.clone().and_then(|n| n.lft) {
+            match node.as_ref().and_then(|n| n.lft) {
                 Some(lft) => {
                     at = lft;
                     parent = node;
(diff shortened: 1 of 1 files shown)


matklad:
Do we actually want to cleanup it? Seems like it's better to just freeze the code for this particular type, as it is clearly designated as deprecated?

austinabell:
> Do we actually want to clean up it? Seems like it's better to just freeze the code for this particular type, as it is clearly designated as deprecated?

Yeah, it wasn't marked explicitly as deprecated, so these are just some changes I made while scanning through what it does. Can drop this PR, but I don't see a downside to bringing it in, even if it is being deprecated. I'm just going alphabetical, and there are these same patterns in the tree map.

matklad:
>I don't see a downside to bringing it in, even if it is being deprecated.

There's always a risk that an innocent behavior-preserving refactor actually changes the behavior in a subtle way, not caught by the tests. 

To be clear, I don't have enough context here to make a judgment call (we need to wait for @mikedotexe for that), just wanted to raise this consideration!

austinabell:
> > I don't see a downside to bringing it in, even if it is being deprecated.
> 
> There's always a risk that an innocent behavior-preserving refactor actually changes the behavior in a subtle way, not caught by the tests.
> 
> To be clear, I don't have enough context here to make a judgment call (we need to wait for @mikedotexe for that), just wanted to raise this consideration!

Yeah, it's totally a fair point, and I couldn't care less if this PR does get closed (and if so hopefully officially marked as deprecated). I just figured I'd PR in the changes since I already did them :)

evgenykuzyakov:
I don't think LegacyTreeMap actually used anywhere. The only project that was using it, has completely changed the code. So we can remove LegacyTreeMap completely in the future versions.

evgenykuzyakov (approved):

austinabell:
Would you like me to mark it as deprecated in this PR then?

PR #395, opened 2021-05-03
refactor: tree map cleanup

austinabell:
Basically same changes as #393

### DIFF
--- a/near-sdk/src/collections/tree_map.rs
+++ b/near-sdk/src/collections/tree_map.rs
@@ -57,7 +57,7 @@ where
     }
 
     pub fn len(&self) -> u64 {
-        self.tree.len() as u64
+        self.tree.len()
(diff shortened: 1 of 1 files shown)


matklad (approved):

evgenykuzyakov (approved):

near-sdk/src/collections/tree_map.rs
```diff
@@ -390,30 +392,25 @@ where
...
-                        }
-                    } else {
-                        match node.lft {
-                            Some(lft) => {
-                                p = node;
-                                at = lft;
-                            }
-                            None => break,
-                        }
+        while let Some(node) = self.node(at) {
```
mikedotexe, line 395: "mo bettah" as we'd say in Hawaii

PR #397, opened 2021-05-03
refactor(vector): remove explicit lifetime and collapse match statements

austinabell:
Just makes code more concise to be more readable, no functional changes

### DIFF
--- a/near-sdk/src/collections/vector.rs
+++ b/near-sdk/src/collections/vector.rs
@@ -52,10 +52,7 @@ impl<T> Vector<T> {
             return None;
         }
         let lookup_key = self.index_to_lookup_key(index);
-        match env::storage_read(&lookup_key) {
-            Some(raw_element) => Some(raw_element),
-            None => env::panic(ERR_INCONSISTENT_STATE),
-        }
+        Some(env::storage_read(&lookup_key).unwrap_or_else(|| env::panic(ERR_INCONSISTENT_STATE)))
     }
 
     /// Removes an element from the vector and returns it in serialized form.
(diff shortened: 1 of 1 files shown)


near-sdk/src/collections/vector.rs
```diff
@@ -52,10 +52,7 @@ impl<T> Vector<T> {
             return None;
         }
         let lookup_key = self.index_to_lookup_key(index);
-        match env::storage_read(&lookup_key) {
-            Some(raw_element) => Some(raw_element),
-            None => env::panic(ERR_INCONSISTENT_STATE),
-        }
+        Some(env::storage_read(&lookup_key).unwrap_or_else(|| env::panic(ERR_INCONSISTENT_STATE)))
```
matklad, line 55: I wonder if, instead of extracting just the message, we should extrct the function itsef:

```
fn expect_storage(x: Option<T>) -> T { x.unwrap_or_else(|| env::panic(""))}
```

not sure what's the best name.

matklad (approved):

near-sdk/src/collections/vector.rs
austinabell: that's a good idea, I'll play around with that now (to narrow down inconsistent state errors)

austinabell: Made the change, named it `expect_consistent_state`, let me know if you think that's a good name (I wanted to tie the name to the error message).

I also missed a couple of match statements to collapse, oops!

evgenykuzyakov (approved):

near-sdk/src/collections/vector.rs
```diff
@@ -126,24 +118,18 @@ impl<T> Vector<T> {
...
+                expect_consistent_state(env::storage_get_evicted())
             } else {
                 env::panic(ERR_INCONSISTENT_STATE);
             }
         }
     }
 
     /// Iterate over raw serialized elements.
-    pub fn iter_raw<'a>(&'a self) -> impl Iterator<Item = Vec<u8>> + 'a {
+    pub fn iter_raw(&self) -> impl Iterator<Item = Vec<u8>> + '_ {
```
mikedotexe, line 129: Interested in hearing what's going on here, but not urgent or blocking

austinabell: ah so with any reference you can specify the lifetime explicitly, but if there is only one possible lifetime that can be used, it can be inferred instead. This just cleans up the usage so that it's more easily read. Anytime there is a `'_` it's an [anonymous lifetime](https://doc.rust-lang.org/edition-guide/rust-2018/ownership-and-lifetimes/the-anonymous-lifetime.html) that just means that lifetime is inferred.

In this case, the lifetime must come from `&self` as it's the only reference as a parameter.

PR #398, opened 2021-05-03
feat: implement FromStr for Base58CryptoHash and cleanup

austinabell:
a similar change to #391 and also cleanup for readability and deserialize type for potential performance increase (`Cow` can avoid an allocation depending on the underlying deserialize buffer).

### DIFF
--- a/near-sdk/src/json_types/hash.rs
+++ b/near-sdk/src/json_types/hash.rs
@@ -1,6 +1,8 @@
 use crate::CryptoHash;
 use borsh::{BorshDeserialize, BorshSerialize};
-use std::convert::{TryFrom, TryInto};
+use serde::{de, Deserialize};
+use std::str::FromStr;
+use std::{borrow::Cow, convert::TryFrom};
 
 #[derive(
     Debug, Copy, Clone, PartialEq, PartialOrd, Ord, Eq, BorshDeserialize, BorshSerialize, Default,
@@ -20,10 +22,7 @@ impl From<CryptoHash> for Base58CryptoHash {
 }
 
 impl serde::Serialize for Base58CryptoHash {
-    fn serialize<S>(
-        &self,
-        serializer: S,
-    ) -> Result<<S as serde::Serializer>::Ok, <S as serde::Serializer>::Error>
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: serde::Serializer,
     {
@@ -31,14 +30,13 @@ impl serde::Serialize for Base58CryptoHash {
     }
 }
 
-impl<'de> serde::Deserialize<'de> for Base58CryptoHash {
-    fn deserialize<D>(deserializer: D) -> Result<Self, <D as serde::Deserializer<'de>>::Error>
+impl<'de> de::Deserialize<'de> for Base58CryptoHash {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
         D: serde::Deserializer<'de>,
     {
-        let s = <String as serde::Deserialize>::deserialize(deserializer)?;
-        s.try_into()
-            .map_err(|err: Box<dyn std::error::Error>| serde::de::Error::custom(err.to_string()))
+        let s: Cow<'de, str> = Deserialize::deserialize(deserializer)?;
+        Self::from_str(s.as_ref()).map_err(|err| de::Error::custom(err.to_string()))
     }
 }
 
@@ -60,6 +58,14 @@ impl TryFrom<&str> for Base58CryptoHash {
     type Error = Box<dyn std::error::Error>;
 
     fn try_from(value: &str) -> Result<Self, Self::Error> {
+        Self::from_str(value)
+    }
+}
+
+impl FromStr for Base58CryptoHash {
+    type Err = Box<dyn std::error::Error>;
+
+    fn from_str(value: &str) -> Result<Self, Self::Err> {
         let mut crypto_hash: CryptoHash = CryptoHash::default();
         let size = bs58::decode(value).into(&mut crypto_hash)?;
         if size != std::mem::size_of::<CryptoHash>() {
--- a/near-sdk/src/json_types/hash.rs
+++ b/near-sdk/src/json_types/hash.rs
@@ -1,8 +1,9 @@
 use crate::CryptoHash;
 use borsh::{BorshDeserialize, BorshSerialize};
 use serde::{de, Deserialize};
+use std::borrow::Cow;
+use std::convert::TryFrom;
 use std::str::FromStr;
-use std::{borrow::Cow, convert::TryFrom};
 
 #[derive(
     Debug, Copy, Clone, PartialEq, PartialOrd, Ord, Eq, BorshDeserialize, BorshSerialize, Default,


near-sdk/src/json_types/hash.rs
```diff
@@ -20,25 +23,21 @@ impl From<CryptoHash> for Base58CryptoHash {
...
-    fn deserialize<D>(deserializer: D) -> Result<Self, <D as serde::Deserializer<'de>>::Error>
+impl<'de> de::Deserialize<'de> for Base58CryptoHash {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
         D: serde::Deserializer<'de>,
     {
-        let s = <String as serde::Deserialize>::deserialize(deserializer)?;
-        s.try_into()
-            .map_err(|err: Box<dyn std::error::Error>| serde::de::Error::custom(err.to_string()))
+        let s: Cow<'de, str> = Deserialize::deserialize(deserializer)?;
```
matklad, line 39: :+1: nice optimization!

near-sdk/src/json_types/hash.rs
```diff
@@ -60,6 +59,14 @@ impl TryFrom<&str> for Base58CryptoHash {
     type Error = Box<dyn std::error::Error>;
 
     fn try_from(value: &str) -> Result<Self, Self::Error> {
+        Self::from_str(value)
+    }
+}
+
+impl FromStr for Base58CryptoHash {
```
matklad, line 66: As an alternative, I sometimes like to write similar impls as 

```suggestion
impl std::str::FromStr for Base58CryptoHash {
```

that way, you avoid importing the trait, so you explicitly show that you are *not* using the trait, and only implement it.

near-sdk/src/json_types/hash.rs
```diff
@@ -60,6 +59,14 @@ impl TryFrom<&str> for Base58CryptoHash {
...
+    type Err = Box<dyn std::error::Error>;
```
matklad, line 67: If this were a general-purpose probject, I'd write

```suggestion
    type Err = Box<dyn std::error::Error + Send + Sync>;
```

but I guess for us it doesn't really make sense

matklad (commented):
Hm, this got me thinking...

Do we perhaps want to introduce `pub struct ParseCrypoHashError` error, while we are adding this API anyway? There's one problem with `Box<dyn Error>` -- it's not trivially compatible with `anyhow` and such. That is, you can't `?` one type-erased error into the other.

near-sdk/src/json_types/hash.rs
austinabell: ah, that's actually a good practice, I didn't think of that. I'll change so it's future proofed!

austinabell:
> Hm, this got me thinking...
> 
> Do we perhaps want to introduce `pub struct ParseCrypoHashError` error, while we are adding this API anyway? There's one problem with `Box<dyn Error>` -- it's not trivially compatible with `anyhow` and such. That is, you can't `?` one type-erased error into the other.

I'm glad you brought it up. I was going to open an issue around changing this as the standard, as this error type is used everywhere. I was going to ask if this would be an acceptable change but was not in a rush because I didn't know if the base type was used to minimize compilation size.

near-sdk/src/json_types/hash.rs
austinabell: actually, it is used in this file (in the deserialize)

near-sdk/src/json_types/hash.rs
austinabell: I didn't want to open up to a breaking change, but I've created an alias to make this error type easier to switch in the future

matklad: This particular impl is new, so, for this impl, we can change the error type, as long as it is convertable to original one via `?`. In particular, changing this (but only this) type to `ParseCryptoHashError` would work.

austinabell: which `ParseCryptoHashError` are you referring to? The box std::error with send/sync or are you referring to something specific/ generating a new error type?

matklad: Sorry for being confusing here. I mean we can do the following, which should is backwards compatible, but, at the same time, is forwards compatible with a nice API of a hypothetical ideal sdk. 

```rust
impl TryFrom<&str> for Base58CryptoHash {
    type Error = Error;

    fn try_from(value: &str) -> Result<Self, Self::Error> {
        let res = value.parse()?; // `?` converts ParseCryptoHashError to Box<dyn std::error::Error>
        Ok(res)
    }
}

#[non_exhaustive]
pub struct ParseCryptoHashError;
impl std::error::Error for ParseCryptoHashError { ... }

impl std::str::FromStr for Base58CryptoHash {
    // Ok to use new type here, as this is a new impl, so we don't care about backcompat. 
    type Err = ParseCryptoHashError;

    fn from_str(value: &str) -> Result<Self, ParseCryptoHashError> {
        ...
    }
}
```

austinabell: Yeah, the only reason I'm hesitant to add this is that it's unclear what the exact Error will be in the future. Using a concrete type like this enforces we keep that type (and trait bounds) in this ideal future API.

I'm happy to make the change and manually implement std::error if someone gives me the go-ahead with a thumbs up on this comment

near-sdk/src/json_types/hash.rs
```diff
@@ -1,6 +1,10 @@
 use crate::CryptoHash;
 use borsh::{BorshDeserialize, BorshSerialize};
-use std::convert::{TryFrom, TryInto};
+use serde::{de, ser, Deserialize};
+use std::borrow::Cow;
```
evgenykuzyakov, line 4: We should try to rebuild some contract that uses `Base58CryptoHash` with the contract builder, to compare the compilation size with using `Cow`. I don't expect anything dramatic, but just to be careful

austinabell: if you give me a workflow or point me in the direction of one that does, I can do that. I'd be curious to see if it actually increases the compiled code (and to what scale).

near-sdk/src/json_types/hash.rs
matklad: > is that it's unclear what the exact Error will be in the future

I think it's probably good to keep errors for well-isolated parts of code specific. IE, even we have something like `type NearError = ...` in the future, it would still make sense to use `ParseCryptoHashError` here. 

I don't think that'll affect the size of the compiled contract (the type is ZST, the implementation is trivial and inlinable, etc), though my confidence level here is not 100%.

near-sdk/src/json_types/hash.rs
matklad: Not sure what's the existing workflow, but I'd suggest adding a smoke test for this.

Something stupid like:

```rust
#[test]
fn test_we_are_smol() {
    std::process::Command::new("cargo")
      .current_dir("examples/locakable-fungible-token")
      .env("RUSTFLAGS", "--link-args=-s")
      .args(&["build", "--release", "--target", "wasm"])
      .status().unwrap()
    let wasm = std::fs::read("examples/locakable-fungible-token/target/wasm").unwrap();
    assert!(100k <= wasm && wasm <= 300k)
}
```

evgenykuzyakov: You can use Base59CryptoHash in `examples/test-contract` and compile it before the change and after the change using contract builder https://github.com/near/near-sdk-rs/tree/master/contact-builder and compare binary sizes

evgenykuzyakov (approved):

austinabell:
before this and #391 are pulled in, should probably come to consensus or I just implement a concrete error type from the new parse method

near-sdk/src/json_types/hash.rs
```diff
@@ -60,11 +59,45 @@ impl TryFrom<&str> for Base58CryptoHash {
...
         let size = bs58::decode(value).into(&mut crypto_hash)?;
         if size != std::mem::size_of::<CryptoHash>() {
-            return Err("Invalid length of the crypto hash (32)".into());
+            return Err(ParseCryptoHashError::InvalidLength(size));
         }
         Ok(Self(crypto_hash))
     }
 }
+
+#[non_exhaustive]
```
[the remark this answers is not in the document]
matklad: pedantry police: `#[non_exhaustive]` is an attribute. Some attributes are macros (like derives or `#[near_bindgen]`), but this one isn't, it's just some meta info for the compiler.

near-sdk/src/json_types/hash.rs
```diff
@@ -60,11 +59,45 @@ impl TryFrom<&str> for Base58CryptoHash {
...
+#[derive(Debug)]
+pub enum ParseCryptoHashError {
+    InvalidLength(usize),
+    Base58(B58Error),
```
matklad, line 83: Note that this makes `B58Error` a public API, so if we chose to switch to a different library to implement this (or even do a major upgrade of Base58), then we'd have to do a major semver bump. I'd personally would try to be concervative with our errors and a) get maximum room for future extensibility b) without exposing internal implementation details. So, something like this:

```rust
pub struct ParseCryptoHashError { kind: ParseCryptoHashErrorKind }

enum ParseCryptoHashErrorKind {
    InvalidLength(usize),
    Base58(B58Error),
}
```

austinabell: yeah probably best to be safe for now, I'll keep the types private for now, and can just expose later if/when solidified

PR #402, opened 2021-05-04
feat(collections): unstable new vector implementation

austinabell:
Opening just as draft to allow anyone to view/comment on the API design as I go through. This API/impl will likely change a decent amount from it's current state.

This impl would live under an `unstable` feature until API stabilized/tested by stakeholders

My plan:
- Migrate core functionality from existing `Vector` implementation while cleaning up/constraining current API to more closely match `std::Vec` (UX main priority in change)
  - I will try to keep the core functionality/initialization the same for easy transition
  - Implement a cached layer on top of the structure to minimize redundant reads as I migrate
  - Migrate existing tests over to ensure no regressions other than API changes
- Explore different patterns/serialization formats to optimize on existing/new contract examples
- Fuzz test implementation to ensure no inconsistencies/bugs
- Explore best caching strategy (not clear if data representation is intended to be consistent)

Open questions:
- Why was u64 used as the index previously? Is this a strict need? Seems like there could be some gains, especially given it's a 32 bit wasm compilation, to have the index as u32 (based on intuition, I hope to benchmark the difference)
- Ideal naming for this structure? Is it to keep as `Vector`, to use `Vec` and leave to users to rename if they want to use a `std::Vec` in the same scope, or to something else?


austinabell:
Okay, the API is close to being finalized in this exploratory state. I'm definitely not happy with how much `unsafe` is needed to cache reads and writes without sharding the data but I think moving in this direction is good to reduce the number of storage reads and writes.

Would love to get some input in tomorrow before I get too deep into testing/auditing unsafe/fuzzing/documenting as to if moving in this direction is a possibility.

To give a high level of the changes, all reads are cached and all writes are written to the cache and only the modified values are persisted when the data structure is dropped. This changes the API to return references (which hit the cached values) instead of moving the type out of the function, which matches the `std::Vec` implementation. This new functionality gets rid of the raw reads/writes as it adds too much complexity given the intermediate cache. Other than that, and potentially larger code size, the API should be more idiomatic and optimize much better for storage costs outside of simple usages.

near-sdk/src/collections/vec/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+mod iter;
+
+use self::iter::{Iter, IterMut};
+use crate::collections::append_slice;
+use crate::{env, CacheCell, CacheEntry, EntryState, IntoStorageKey};
+use borsh::{BorshDeserialize, BorshSerialize};
+use std::collections::{btree_map::Entry, BTreeMap};
+use std::ptr::NonNull;
+
+const ERR_INCONSISTENT_STATE: &[u8] = b"The collection is an inconsistent state. Did previous smart contract execution terminate unexpectedly?";
```
matklad, line 15: std::env::panic should take `str` rather than `u8` argument, the same issue as with log, and withe the same compat concerns.

near-sdk/src/collections/vec/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+
+mod impls;
+mod iter;
+
+use self::iter::{Iter, IterMut};
+use crate::collections::append_slice;
+use crate::{env, CacheCell, CacheEntry, EntryState, IntoStorageKey};
+use borsh::{BorshDeserialize, BorshSerialize};
+use std::collections::{btree_map::Entry, BTreeMap};
+use std::ptr::NonNull;
```
matklad, line 13: I am personally not a huge fan of mashing all the imports together for a new code. I'd written this as 

```rust
use std::collections::{btree_map::Entry, BTreeMap};
use std::ptr::NonNull;

use crate::collections::append_slice;
use crate::{env, CacheCell, CacheEntry, EntryState, IntoStorageKey};
use crate::collections::vec::iter::{Iter, IterMut};
```

Might make sense to add a code style section to the contributing.md, and say "for the new code, we use the following import order / don't care about import order"

near-sdk/src/collections/vec/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+/// An iterable implementation of vector that stores its content on the trie.
+/// Uses the following map: index -> element.
+///
+/// This implementation will cache all changes and loads and only updates values that are changed
+/// in storage after it's dropped through it's [`Drop`] implementation.
+///
+/// TODO examples
+#[derive(BorshSerialize, BorshDeserialize)]
+#[cfg_attr(not(feature = "expensive-debug"), derive(Debug))]
+pub struct Vector<T>
```
matklad, line 33: Yeah, `Vector` seems like a bad name. It's different from Vec, but it's uncleare why. 

My understanding is that collections here are very different from std's collections, in that they are persistent (as in, backed by the try directly). So it's more like an ORM than a collections.

I suggest choosing some perfix here, and use it for all collections. A couple of options would be `Trie` or `Stored`: `StoredVec`, `TrieHashMap`.

Allternatively, *maybe* rather than using `collections` module, it should be part of the `store` module? So that the call-site looks like this:

```
struct MyContrct {
    data: store::Vec<String>
}

fn do_something() {
    store::write(..)
}
```

matklad: Would it be helpful to try to design the whole API holistically holistically? That is, creating a docmuent which lists only signatures and bodies of all the collections?

matklad (commented):
Didn't review all the things yet, :+1: on adding unstable flag, but maybe we need a bit of up-front design here  about the end state we want to end up in?

near-sdk/src/collections/vec/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+    T: BorshSerialize + BorshDeserialize,
+{
+    fn deserialize_element(raw_element: &[u8]) -> T {
+        T::try_from_slice(&raw_element).unwrap_or_else(|_| env::panic(ERR_ELEMENT_DESERIALIZATION))
+    }
+
+    /// Loads value from storage into cache, if it does not already exist.
+    /// This function must be unsafe because it requires modifying the cache with an immutable
+    /// reference.
+    unsafe fn load(&self, index: u32) -> NonNull<CacheEntry<T>> {
```
matklad, line 160: Hm, I missed this unsafety on the first read. I am 89% sure that [miri](https://github.com/rust-lang/miri) will complain that this violates stacked borrows. 

Hm, I guess it's even unsound in a typical way, via re-entrancy. In `Self::deserialize_element(&v)`, we are calling user-supplied code for `T`. That user-suppliend code can call a `load` on the same vec, which will mean that we'd call `.entry` re-entrantly, and, after the first insertion, the second will use dangling memory. 

This is an edge case, where the user goes out of their way to cause UB, but it shows that the overall API is not sound. 

I suggest trying to build the caching layer out of some existing building blocks. I think you need two things here:

* first, a way to allocate slots using only shared refernces. Basically, you need an `fn push(&self, item: T)`. Annoyingly, I don't know what's the goto crate. https://docs.rs/appendlist/1.4.0/appendlist/ is the right API, but I haven't auditied implementation. I feel there's some famous crate which does that, and which I don't remember the name of? 
* second,  when you get a `&Slot<T>`, you need a way to fill it, only if it was empty before. This is fully covered by https://docs.rs/once_cell/1.7.2/once_cell/unsync/struct.OnceCell.html. 

I don't necessary think we should depend on third party crates, but it definitelly makes sense to at least make sure that our home-grown impl follows established patterns. This stuff is tricky to get right.

austinabell:
> Didn't review all the things yet,  on adding the unstable flag, but maybe we need a bit of up-front design here about the end state we want to end up in?

I agree, I've had a late start to my day but will go through and try to add as many docs/tests and audit/limit the unsafe interactions as much as possible this afternoon and hopefully create a table to compare std/old/new interfaces.

near-sdk/src/collections/vec/mod.rs
austinabell: Yeah I agree, I can make an additional API for that and the log under the unstable flag and group it with #366 or open a separate PR for that. This change is a bit more annoying to put under a flag as the interactions don't go through a macro like `log!`

near-sdk/src/collections/vec/mod.rs
austinabell: Yeah, I was thinking about that, I just figured I'd keep it as is (which I didn't see as bad because it's already grouped as its lexicographical) until someone shared an opinion on the style.

near-sdk/src/collections/vec/mod.rs
austinabell: yeah for naming, I don't have an opinion, happy to change to whatever makes sense. I just favoured using the existing name to be as close to drag and drop from what it was (except for some APIs now closer to std).

As for documenting, absolutely will do, this first day was just exploratory to see what's possible before I narrowed down the API

near-sdk/src/collections/vec/mod.rs
austinabell: I'm not sure I completely understand or agree with the concerns you brought up. I totally agree there can be issues with the implementation as it is, but the reference guarantees should be the same as self is borrowed immutably during this and this functionality is only used internally and not exposed, so any unsafe code that a user could do was something they could do before?

I also was unsure about the best path forward while keeping the existing functionality and couldn't think of a crate that handles this use case well. I'll explore alternatives soon, and definitely am very uncomfortable with the unsafe code needed, but just wanted to minimize compilation size and have control over the API

austinabell: So miri doesn't bork on testing using ported tests from the current API. I will try to write out some more intensive and edge case tests to see if I can trigger a failure running tests with miri now.

matklad: Yeah, I might be mistaken as well. Let me try to come up with an "exploit"  here )

matklad: I've spend some time breaking this, and I wasn't able to produce outright crash here -- looks like there's enough boxing here for the compiler to not make use of UB. However, I believe this commit demonstrates unsoundness:

https://github.com/matklad/near-sdk-rs/commit/caeb5693b998c753725b827aaadcaec752e040e4

Before we look at the actual test, here's the printouts I added to 

```diff
diff --git a/near-sdk/src/collections/vec/mod.rs b/near-sdk/src/collections/vec/mod.rs
index a1634e3..9a82d3e 100644
--- a/near-sdk/src/collections/vec/mod.rs
+++ b/near-sdk/src/collections/vec/mod.rs
@@ -159,14 +159,20 @@ where
     /// reference.
     unsafe fn load(&self, index: u32) -> NonNull<CacheEntry<T>> {
         // TODO safety docs
-        match self.cache.get_ptr().as_mut().entry(index) {
+        let mut b_tree_map_ptr = self.cache.get_ptr();
+        let mut b_tree_map = b_tree_map_ptr.as_mut();
+        eprintln!("mut ref active:  {:?}, index: {}", b_tree_map as *mut _, index);
+        let res = match b_tree_map.entry(index) {
             Entry::Occupied(mut occupied) => NonNull::from(&mut **occupied.get_mut()),
             Entry::Vacant(vacant) => {
                 let value = env::storage_read(&self.index_to_lookup_key(index))
                     .map(|v| Self::deserialize_element(&v));
                 NonNull::from(&mut **vacant.insert(Box::new(CacheEntry::new_cached(value))))
             }
-        }
+        };
+        eprintln!("mut ref dropped: {:?}, index: {}", b_tree_map as *mut _, index);
+        eprintln!("keyset: {:?}", self.cache.get_ptr().as_mut().keys().collect::<Vec<_>>());
+        res
     }
```

Here's the printout I get with my test:

```
mut ref active:  0x7fdc05c303a0, index: 0
mut ref active:  0x7fdc05c303a0, index: 0
mut ref dropped: 0x7fdc05c303a0, index: 0
keyset: [0]
mut ref dropped: 0x7fdc05c303a0, index: 0
keyset: [0, 0]
```

That is, what happens here is that we do a recursive lookup into a BTreeMap, and end up in a situation where there are two active vacant entries at the same time, pointing into the same slot in the BTreeMap, and modifications to the one effectively overwrite the other. This causes the map to get really confused and store key zero twice!

Here's how I came up with the test. I've noticed that the load method has the following shape:

```rust
fn load(&self) {
  unsafe {
    let exclusive_ref = some_unsafe_op();
    T::deserialize()
  }
}
```

The smell here is that we are calling user-supplied code (`T::deserialize`) from an unsafe block. This is hard to prove to be correct, as user code can do anything. 

In particular (and that's the bug which every crate that does once-cell-like thing has), user supplied code can call load re-entrantly:

```rust
x.load 
  calls T::deserialize
    calls x.load again
```

This is possible becaues we require only a shared reference for `self`. However, once we did that, we end up with two exclusive references we created unsafely.

For the actual test case, I had to smuggle that shared reference via a thread local variable:

```rust
    #[test]
    fn unsound() {
        test_env::setup();

        #[derive(Debug, Clone, Copy)]
        struct T;

        impl BorshSerialize for T {
            fn serialize<W: std::io::Write>(&self, _writer: &mut W) -> std::io::Result<()> {
                Ok(())
            }
        }

        impl BorshDeserialize for T {
            fn deserialize(_buf: &mut &[u8]) -> std::io::Result<Self> {
                GLOBAL.with(|g| {
                    FLAG.with(|f| {
                        if f.replace(false) {
                            g.get(0);
                        }
                        Ok(T)
                    })
                })
            }
        }

        thread_local! {
            static GLOBAL: Vector<T> = Vector::new(92);
            static FLAG: Cell<bool> = Cell::new(true);
        }

        eprintln!("Phase One");
        let mut v: Vector<T> = Vector::new(92);
        for _ in 0..10 {
            v.push(T);
        }
        drop(v);
        eprintln!("Phase Two");
        GLOBAL.with(|g| {
            g.get(0);
        });
    }
```

matklad: Let me jot down the API I think we need here:

```rust
// This one we write ourselves
pub(crate) struct StableMap<K, V> {
    inner: RefCell<BTreeMap<K, Box<V>>>,
}

impl StableMap<K, V> {
    pub(crate) fn get_or_insert(&self, k: K) -> &V where V: Default {
        let v: std::cell::RefMut<'_, V> = ...
        // Extending lifetime beyound the borrow here should be fine
        let v: &V = unsafe {}
    }
    pub(crate) fn entry(&mut self, k: K) -> Entrly {
        ...
    }

}
// This one we just copy-paste from `unsync` once_cell (or we might just add a dependency to Cargo.toml, no strong opinion either way)
pub(crate) struct OnceCell { .. }

// This one we can now compose using only safe code:

type VecCache<T> = StableMap<u32, OnceCell<T>>;

impl VecCache<T> {
    fn load(&self, index: u32) -> T {
        self.get_or_insert(index).get_or_init(|| T::deserialize())
    }
    fn remove(&mut self, index: u32) {
        self.entry(index).remove()
    }
}
```

austinabell: Thank you so much for going so in detail, this gives me some good direction in ways I can solidify this API. I still am missing how the unsafe can cause complications externally or how this could negatively affect a user who is not including the unsafe to trigger the re-entrancy issue specifically. 

I'll try to investigate your solution or think through how I can move the deserialization outside of the load, as I think your suggestion is to keep the bytes as serialized in the cache, and deserialize each time it's read. The negative to this is that the user would be charged and consume gas to deserialize each time a value is read from cache, which defeats half of the purpose of the cache.

Maybe there is a solution around having the value type just be an enum:
```rust
enum Entry<T> {
	Serialized(Vec<u8>),
	Deserialized(<T>),
}
```

and defer the deserialization outside of the load this way?

austinabell: actually, that probably still is in the category of UB. I think you're right in being cautious about that deserialization call, I'll find a way around

matklad: >or how this could negatively affect a user who is not including the unsafe to trigger the re-entrancy issue specifically.

I mean, the `unsound` test function doesn't use `unsafe` anywhere, it's fully safe code which can be written by contract developers. So this does cause unsoundness via pure safe code, because `unsafe` inside sdk is not correct. 

It definitely is true that contract's author needs to go out of their way to trigger this though, but that's enough to break Rust's safety promise.

austinabell: > > or how this could negatively affect a user who is not including the unsafe to trigger the re-entrancy issue specifically.
> 
> I mean, the `unsound` test function doesn't use `unsafe` anywhere, it's fully safe code which can be written by contract developers. So this does cause unsoundness via pure safe code, because `unsafe` inside sdk is not correct.
> 
> It definitely is true that contract's author needs to go out of their way to trigger this though, but that's enough to break Rust's safety promise.

You're right, let me change the deserialization to happen lazily, and after the load

austinabell: I made what I think is minimal changes to avoid the issue above here [`8fb7324` (#402)](https://github.com/near/near-sdk-rs/pull/402/commits/8fb73248de4b6d094cfd76f0a5e1754a5d9eb5a1) but will work to refactor it to get it to a better place and have the caching layer more reusable (closer to what you suggested above)

austinabell: Hi @matklad I'm having a bit of trouble trying to execute on https://github.com/near/near-sdk-rs/pull/402#discussion_r627451952 as it seems hard to create an API that suits the need here well. The main issue I have with the proposed API here is that for the append-only map to have `get_or_insert`, the insert would have to be a function (to avoid loading storage on get) and the API gets very clunky quickly. There is also the cache entry data that needs to be kept to make sure storage is updated correctly and minimally and I'm having a hard time fitting it into this API.

Do you have any intuition if [`8fb7324` (#402)](https://github.com/near/near-sdk-rs/pull/402/commits/8fb73248de4b6d094cfd76f0a5e1754a5d9eb5a1) is a bad solution? There might be slight duplication with this and HashMap and some harder-to-read code internally, but it solves the functionality needed and can be kept under unstable until it's been heavily tested and/or internals improved.

matklad: Finally got to look into this gain. I can't vouch for the current state of the code -- the amount of unsafe in `CacheCell` and `load` / `load_mut` is big, and I don't see easy-to-describe invariants proving that the unsafe is actually sound. 

Please consider this implementation: it reduces the number of `unsafe` calls to just one, with a somewhat easier to verify invariant. This impl seems to pass existing tests


```diff
diff --git a/near-sdk/src/store/vec/mod.rs b/near-sdk/src/store/vec/mod.rs
index 0c1c908..1b4c265 100644
--- a/near-sdk/src/store/vec/mod.rs
+++ b/near-sdk/src/store/vec/mod.rs
@@ -7,12 +7,12 @@ mod iter;
 
 use borsh::{BorshDeserialize, BorshSerialize};
 use once_cell::unsync::OnceCell;
-use std::collections::{btree_map::Entry, BTreeMap};
-use std::ptr::NonNull;
+use std::cell::RefCell;
+use std::collections::BTreeMap;
 
 use self::iter::{Iter, IterMut};
 use crate::collections::append_slice;
-use crate::{env, CacheCell, CacheEntry, EntryState, IntoStorageKey};
+use crate::{env, CacheEntry, EntryState, IntoStorageKey};
 
 const ERR_INCONSISTENT_STATE: &[u8] = b"The collection is an inconsistent state. Did previous smart contract execution terminate unexpectedly?";
 const ERR_ELEMENT_DESERIALIZATION: &[u8] = b"Cannot deserialize element";
@@ -23,8 +23,6 @@ fn expect_consistent_state<T>(val: Option<T>) -> T {
     val.unwrap_or_else(|| env::panic(ERR_INCONSISTENT_STATE))
 }
 
-type CacheMap<K, V> = CacheCell<BTreeMap<K, Box<V>>>;
-
 /// An iterable implementation of vector that stores its content on the trie.
 /// Uses the following map: index -> element.
 ///
@@ -43,7 +41,44 @@ where
     /// Cache for loads and intermediate changes to the underlying vector.
     /// The cached entries are wrapped in a [`Box`] to avoid existing pointers from being
     /// invalidated.
-    cache: CacheMap<u32, CacheEntry<OnceCell<T>>>,
+    cache: StableMap<u32, OnceCell<CacheEntry<T>>>,
+}
+
+struct StableMap<K, V> {
+    map: RefCell<BTreeMap<K, Box<V>>>,
+}
+
+impl<K: Ord, V> Default for StableMap<K, V> {
+    fn default() -> Self {
+        StableMap { map: Default::default() }
+    }
+}
+
+impl<K, V> StableMap<K, V> {
+    fn get(&self, k: K) -> &V
+    where
+        K: Ord,
+        V: Default,
+    {
+        let mut map = self.map.borrow_mut();
+        let v: &mut Box<V> = map.entry(k).or_default();
+        let v: &V = &*v;
+        // SAFETY: here, we extend the lifetime of `V` from local `RefCell`
+        // borrow to the `&self`. This is valid because we only append to the
+        // map via `&` reference, and the values are boxed, so we have stability
+        // of addresses.
+        unsafe { &*(v as *const V) }
+    }
+    fn get_mut(&mut self, k: K) -> &mut V
+    where
+        K: Ord,
+        V: Default,
+    {
+        &mut *self.map.get_mut().entry(k).or_default()
+    }
+    fn inner(&mut self) -> &mut BTreeMap<K, Box<V>> {
+        self.map.get_mut()
+    }
 }
 
 impl<T> Vector<T>
@@ -80,32 +115,31 @@ where
             env::storage_remove(&lookup_key);
         }
         self.len = 0;
-        self.cache.as_inner_mut().clear();
+        self.cache.inner().clear();
     }
 
     // TODO expose this? Could be useful to not force a user to drop to persist changes
     /// Flushes the cache and writes all modified values to storage.
     fn flush(&mut self) {
-        for (k, v) in self.cache.as_inner_mut().iter_mut() {
-            if v.is_modified() {
-                let key = append_slice(&self.prefix, &k.to_le_bytes()[..]);
-                match v.value().as_ref() {
-                    Some(modified) => {
-                        // Value was modified, write the updated value to storage
-                        env::storage_write(
-                            &key,
-                            &Self::serialize_element(expect_consistent_state(modified.get())),
-                        );
-                    }
-                    None => {
-                        // Element was removed, clear the storage for the value
-                        env::storage_remove(&key);
+        for (k, v) in self.cache.inner().iter_mut() {
+            if let Some(v) = v.get_mut() {
+                if v.is_modified() {
+                    let key = append_slice(&self.prefix, &k.to_le_bytes()[..]);
+                    match v.value().as_ref() {
+                        Some(modified) => {
+                            // Value was modified, write the updated value to storage
+                            env::storage_write(&key, &Self::serialize_element(modified));
+                        }
+                        None => {
+                            // Element was removed, clear the storage for the value
+                            env::storage_remove(&key);
+                        }
                     }
-                }
 
-                // Update state of flushed state as cached, to avoid duplicate writes/removes
-                // while also keeping the cached values in memory.
-                v.replace_state(EntryState::Cached);
+                    // Update state of flushed state as cached, to avoid duplicate writes/removes
+                    // while also keeping the cached values in memory.
+                    v.replace_state(EntryState::Cached);
+                }
             }
         }
     }
@@ -117,12 +151,11 @@ where
             env::panic(ERR_INDEX_OUT_OF_BOUNDS);
         }
 
-        match self.cache.as_inner_mut().entry(index) {
-            Entry::Occupied(mut occupied) => {
-                occupied.get_mut().replace(Some(OnceCell::from(value)));
-            }
-            Entry::Vacant(vacant) => {
-                vacant.insert(Box::new(CacheEntry::new_modified(Some(OnceCell::from(value)))));
+        let entry = self.cache.get_mut(index);
+        match entry.get_mut() {
+            Some(entry) => *entry = CacheEntry::new_modified(Some(value)),
+            None => {
+                let _ = entry.set(CacheEntry::new_modified(Some(value)));
             }
         }
     }
@@ -151,63 +184,43 @@ where
         T::try_from_slice(&raw_element).unwrap_or_else(|_| env::panic(ERR_ELEMENT_DESERIALIZATION))
     }
 
-    /// Loads value from storage into cache, if it does not already exist.
-    /// This function must be unsafe because it requires modifying the cache with an immutable
-    /// reference.
-    unsafe fn load(&self, index: u32) -> NonNull<CacheEntry<OnceCell<T>>> {
-        // * SAFETY: The pointer into the values will be valid if the map is updated from a cache
-        //           load and potential move of values because the values are wrapped in a `Box`.
-        //           There also cannot be a re-entrancy bug with modifying an entry twice because
-        //           the deserialization happens outside of the modification of the map.
-        let (entry, storage_bytes) = match self.cache.get_ptr().as_mut().entry(index) {
-            Entry::Occupied(mut occupied) => (NonNull::from(&mut **occupied.get_mut()), None),
-            Entry::Vacant(vacant) => {
-                let storage_bytes = env::storage_read(&self.index_to_lookup_key(index));
-                let value = if storage_bytes.is_some() { Some(OnceCell::new()) } else { None };
-
-                (
-                    NonNull::from(&mut **vacant.insert(Box::new(CacheEntry::new_cached(value)))),
-                    storage_bytes,
-                )
-            }
-        };
-
-        if let Some(bytes) = storage_bytes {
-            // If storage bytes is `Some`, it is guaranteed that it is an entry with
-            // an empty `OnceCell`. This deserialization must happen outside of the entry load
-            // to avoid re-entrancy issues.
-            let filled = expect_consistent_state(entry.as_ptr().as_ref()).value();
-            expect_consistent_state(filled.as_ref())
-                .get_or_init(|| Self::deserialize_element(&bytes));
-        }
-
-        entry
-    }
-
-    /// Loads value from storage into cache, and returns a mutable reference to the loaded value.
-    /// This function is safe because a mutable reference of self is used.
-    fn load_mut(&mut self, index: u32) -> &mut CacheEntry<OnceCell<T>> {
-        // * SAFETY: A mutable reference can be returned here because it references a value in a
-        //           `Box` and no other references should exist given function takes a mutable
-        //           reference. This has the assumption that other references are not kept around
-        //           past this function call.
-        unsafe { &mut *self.load(index).as_ptr() }
-    }
-
     /// Returns the element by index or `None` if it is not present.
     pub fn get(&self, index: u32) -> Option<&T> {
         // TODO doc safety
-        let cell = unsafe { &*self.load(index).as_ptr() }.value().as_ref()?;
-
-        // Cell is guaranteed to be filled by the `load` call.
-        Some(expect_consistent_state(cell.get()))
+        let entry = self.cache.get(index).get_or_init(|| {
+            let storage_bytes = env::storage_read(&self.index_to_lookup_key(index));
+            let value = if let Some(bytes) = storage_bytes {
+                Some(Self::deserialize_element(&bytes))
+            } else {
+                None
+            };
+            CacheEntry::new_cached(value)
+        });
+        entry.value().as_ref()
     }
 
     /// Returns a mutable reference to the element at the `index` provided.
     pub fn get_mut(&mut self, index: u32) -> Option<&mut T> {
-        let cell = self.load_mut(index).value_mut().as_mut()?;
+        let entry = self.get_mut_inner(index);
+        entry.value_mut().as_mut()
+    }
 
-        cell.get_mut()
+    /// Returns a mutable reference to the element at the `index` provided.
+    fn get_mut_inner(&mut self, index: u32) -> &mut CacheEntry<T> {
+        let index_to_lookup_key = self.index_to_lookup_key(index);
+        let entry = self.cache.get_mut(index);
+        entry.get_or_init(|| {
+            let storage_bytes = env::storage_read(&index_to_lookup_key);
+            let value = if let Some(bytes) = storage_bytes {
+                Some(Self::deserialize_element(&bytes))
+            } else {
+                None
+            };
+            CacheEntry::new_cached(value)
+        });
+        let entry = entry.get_mut().unwrap();
+        entry.mark_modified();
+        entry
     }
 
     fn swap(&mut self, a: u32, b: u32) {
@@ -219,19 +232,9 @@ where
             // Short circuit if indices are the same, also guarantees uniqueness below
             return;
         }
-
-        // * SAFETY: references are guaranteed to be distinct because the indices are checked to not
-        //           be equal above. These mutable references will both be dropped before the end
-        //           of the scope of the swap call.
-        let a_value = unsafe { &mut *self.load(a).as_ptr() };
-        let b_value = unsafe { &mut *self.load(b).as_ptr() };
-
-        if a_value.value().is_none() || b_value.value().is_none() {
-            // Should never be able to swap a filled value with an empty value in a vec.
-            env::panic(ERR_INCONSISTENT_STATE);
-        }
-
-        core::mem::swap(a_value.value_mut(), b_value.value_mut());
+        let val_a = self.get_mut_inner(a).replace(None);
+        let val_b = self.get_mut_inner(b).replace(val_a);
+        self.get_mut_inner(a).replace(val_b);
     }
 
     /// Removes an element from the vector and returns it.
@@ -258,9 +261,8 @@ where
             let last_idx = self.len - 1;
             self.len = last_idx;
 
-            // Replace current value with none, and return the existing value
-            let popped_value = expect_consistent_state(self.load_mut(last_idx).replace(None));
-            popped_value.into_inner()
+            let entry = self.get_mut_inner(last_idx);
+            entry.replace(None)
         }
     }
 
@@ -272,9 +274,7 @@ where
     ///
     /// If `index` is out of bounds.
     pub fn replace(&mut self, index: u32, element: T) -> T {
-        let cache =
-            expect_consistent_state(self.load_mut(index).replace(Some(OnceCell::from(element))));
-        expect_consistent_state(cache.into_inner())
+        self.get_mut_inner(index).replace(Some(element)).unwrap()
     }
 
     /// Returns an iterator over the vector. This iterator will lazily load any values iterated
diff --git a/near-sdk/src/utils/cache_cell.rs b/near-sdk/src/utils/cache_cell.rs
deleted file mode 100644
index 8c82b1a..0000000
--- a/near-sdk/src/utils/cache_cell.rs
+++ /dev/null
@@ -1,60 +0,0 @@
-use core::{cell::UnsafeCell, fmt, fmt::Debug, ptr::NonNull};
-
-pub struct CacheCell<T: ?Sized> {
-    inner: UnsafeCell<T>,
-}
-
-impl<T> CacheCell<T> {
-    pub fn new(value: T) -> Self {
-        Self { inner: UnsafeCell::new(value) }
-    }
-
-    #[allow(dead_code)]
-    pub fn into_inner(self) -> T {
-        self.inner.into_inner()
-    }
-}
-
-impl<T> Debug for CacheCell<T>
-where
-    T: ?Sized + Debug,
-{
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        <T as Debug>::fmt(self.as_inner(), f)
-    }
-}
-
-impl<T> From<T> for CacheCell<T> {
-    fn from(value: T) -> Self {
-        Self::new(value)
-    }
-}
-
-impl<T> Default for CacheCell<T>
-where
-    T: Default,
-{
-    fn default() -> Self {
-        Self::new(<T as Default>::default())
-    }
-}
-
-impl<T> CacheCell<T>
-where
-    T: ?Sized,
-{
-    pub fn as_inner(&self) -> &T {
-        // TODO doc safety
-        unsafe { &*self.inner.get() }
-    }
-
-    pub fn as_inner_mut(&mut self) -> &mut T {
-        // TODO doc safety
-        unsafe { &mut *self.inner.get() }
-    }
-
-    pub fn get_ptr(&self) -> NonNull<T> {
-        // TODO doc safety
-        unsafe { NonNull::new_unchecked(self.inner.get()) }
-    }
-}
diff --git a/near-sdk/src/utils/cache_entry.rs b/near-sdk/src/utils/cache_entry.rs
index 711c779..6455c34 100644
--- a/near-sdk/src/utils/cache_entry.rs
+++ b/near-sdk/src/utils/cache_entry.rs
@@ -54,6 +54,10 @@ impl<T> CacheEntry<T> {
         matches!(self.state, EntryState::Modified)
     }
 
+    pub fn mark_modified(&mut self) {
+        self.state = EntryState::Modified;
+    }
+
     #[allow(dead_code)]
     /// Returns true if the entry state has not been changed.
     pub fn is_cached(&self) -> bool {
diff --git a/near-sdk/src/utils/mod.rs b/near-sdk/src/utils/mod.rs
index 217fd08..2a14d53 100644
--- a/near-sdk/src/utils/mod.rs
+++ b/near-sdk/src/utils/mod.rs
@@ -1,9 +1,5 @@
 pub(crate) mod storage_key_impl;
 
-#[cfg(feature = "unstable")]
-pub(crate) use cache_cell::CacheCell;
-#[cfg(feature = "unstable")]
-mod cache_cell;
 #[cfg(feature = "unstable")]
 mod cache_entry;
 #[cfg(feature = "unstable")]

```

austinabell: Great suggestions! I made these changes with some small tweaks here: [`8d39c57` (#402)](https://github.com/near/near-sdk-rs/pull/402/commits/8d39c57c03077ed946b2ead9ba09dcd2e2f35942) and will cleanup, document, and match functions to std/existing now :)

near-sdk/src/store/vec/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+    }
+}
+
+#[cfg(feature = "expensive-debug")]
+impl<T: std::fmt::Debug + BorshDeserialize> std::fmt::Debug for Vector<T>
+where
+    T: BorshSerialize,
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.iter().collect::<Vec<_>>().fmt(f)
```
matklad, line 286: I am not  fan of calling `.fmt()` using method call syntax -- there's the same method on Debug and Display, and that might cause method resolution ambiguities. IIRC, the `failure` crate once broke a chunk of the ecosystem due to this issue. I don't think this is a problem in this particular case, but still, I suggest spelling this as 

```rust
fmd::Debug::fmt(&self.iter().collect::<Vec<_>>(), f)
```

near-sdk/src/store/vec/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+            for _ in 0..(rng.gen::<u64>() % 20 + 1) {
+                let value = rng.gen::<u64>();
+                vec.push(value);
+            }
+            assert!(!vec.is_empty());
+            vec.clear();
+            assert!(vec.is_empty());
+        }
+    }
+
```
matklad, line 382: Nice tests! 

For testing data structures against a model, a useful (albeit a bit heavy-weight) approach is to randomly generate a sequence of ops.

Ie, instead of a test that does 100 pushes and 100 pops, you can genrate a `Vec<Op>`, where 

```rust
enum Op {
     Push(u16)
     Pop,
}
```

and, that way, you'll also check all the interleavings of the operations. Naturally, the `Op` enum can be extended to cover sets, swap_removes, clears and the like.

near-sdk/src/store/mod.rs
```diff
@@ -1,2 +1,4 @@
 mod lazy;
 pub use lazy::Lazy;
+mod vec;
+pub use vec::Vector;
```
matklad, line 4: I think we need to make the `vec` module itself public, so that the cilents can name `Iter` strcuts.

near-sdk/src/store/vec/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+/// vec.extend([1, 2, 3].iter().copied());
+/// assert!(Iterator::eq(vec.into_iter(), [7, 1, 2, 3].iter()));
+/// ```
+#[derive(BorshSerialize, BorshDeserialize)]
+pub struct Vector<T>
+where
+    T: BorshSerialize,
+{
+    len: u32,
+    prefix: Vec<u8>,
```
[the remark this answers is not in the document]
austinabell: ah, that's a good idea, this would make it harder to misuse as well.

near-sdk/src/store/vec/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+    /// invalidated.
+    cache: StableMap<u32, OnceCell<CacheEntry<T>>>,
+}
+
+impl<T> Vector<T>
+where
+    T: BorshSerialize,
+{
+    /// Returns the number of elements in the vector, also referred to as its size.
+    pub fn len(&self) -> u32 {
```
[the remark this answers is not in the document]
austinabell: Consistency between targets imo, would be a footgun for some functionality to work in tests but then not work on chain. Happy to change to `usize`, but I'm just always hesitant to use `usize` in this type of context

near-sdk/src/store/vec/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+    /// Flushes the cache and writes all modified values to storage.
+    pub fn flush(&mut self) {
+        for (k, v) in self.cache.inner().iter_mut() {
+            if let Some(v) = v.get_mut() {
+                if v.is_modified() {
+                    let key = append_slice(&self.prefix, &k.to_le_bytes()[..]);
+                    match v.value().as_ref() {
+                        Some(modified) => {
+                            // Value was modified, write the updated value to storage
+                            env::storage_write(&key, &Self::serialize_element(modified));
```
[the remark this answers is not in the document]
austinabell: Sorry, I'm a bit confused about what you're suggesting here, do you mean to keep a buffer attached to the `Vector` struct and serialize there and keep it in memory?

near-sdk/src/store/vec/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+    fn get_mut_inner(&mut self, index: u32) -> &mut CacheEntry<T> {
+        let index_to_lookup_key = self.index_to_lookup_key(index);
+        let entry = self.cache.get_mut(index);
+        entry.get_or_init(|| {
+            let storage_bytes = env::storage_read(&index_to_lookup_key);
+            let value = storage_bytes.as_deref().map(Self::deserialize_element);
+            CacheEntry::new_cached(value)
+        });
+        let entry = entry.get_mut().unwrap();
+        entry
```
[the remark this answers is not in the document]
austinabell: The entry is marked as modified when it's borrowed mutably. No need to do it twice and/or prematurely mark as modified in case this entry is never used

near-sdk/src/store/vec/iter.rs
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    fn count(self) -> usize {
+        self.remaining() as usize
+    }
+
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
+        debug_assert!(self.begin <= self.end);
+        let n: u32 = n.try_into().ok()?;
+        if self.begin + n >= self.end {
```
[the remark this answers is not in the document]
austinabell: I'll add this test, but I don't see why not have it fused in this case, do you have a good reason why there could be a panic instead and not have this be a fused iterator?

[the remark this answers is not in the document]
austinabell: ah, caught a bug (or inconsistency with `Vec`) so good call on this

near-sdk/src/store/vec/mod.rs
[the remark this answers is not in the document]
matklad:
```diff
diff --git a/near-sdk/src/store/vec/mod.rs b/near-sdk/src/store/vec/mod.rs
index 7d07e67..d419f43 100644
--- a/near-sdk/src/store/vec/mod.rs
+++ b/near-sdk/src/store/vec/mod.rs
@@ -109,14 +109,17 @@ where
 
     /// Flushes the cache and writes all modified values to storage.
     pub fn flush(&mut self) {
+        let mut buf = Vec::new();
         for (k, v) in self.cache.inner().iter_mut() {
             if let Some(v) = v.get_mut() {
                 if v.is_modified() {
                     let key = append_slice(&self.prefix, &k.to_le_bytes()[..]);
                     match v.value().as_ref() {
                         Some(modified) => {
-                            // Value was modified, write the updated value to storage
-                            env::storage_write(&key, &Self::serialize_element(modified));
+                            buf.clear();
+                            BorshSerialize::serialize(modified, &mut buf)
+                                .unwrap_or_else(|_| env::panic(ERR_ELEMENT_SERIALIZATION));
+                            env::storage_write(&key, &buf);
                         }
                         None => {
                             // Element was removed, clear the storage for the value
@@ -148,10 +151,6 @@ where
         }
     }
 
-    fn serialize_element(element: &T) -> Vec<u8> {
-        element.try_to_vec().unwrap_or_else(|_| env::panic(ERR_ELEMENT_SERIALIZATION))
-    }
-
     /// Appends an element to the back of the collection.
     pub fn push(&mut self, element: T) {
         if self.len() >= u32::MAX {
```

near-sdk/src/store/vec/mod.rs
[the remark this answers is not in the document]
matklad: Yeah, consistency between targets is a good argument. It might make sense to put this into some kind of comment, maybe even a doc comment. As a user, I *would* wonder why we differ from standard rust APIs.

near-sdk/src/store/vec/mod.rs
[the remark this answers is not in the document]
matklad: I'd default to dirtying the entry as early as possible -- that way you can't have a bug. If the entry ends up not being used, I'd (as a rule of thumb) change the calling code to not do mutable lookup in the first place. I haven't look closely at all the calling contexts, so, if you feel the current approach is ok, it's fine to leave it as is.

near-sdk/src/store/vec/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+        }
+    }
+
+    /// Inserts a element at `index`, returns an evicted element.
+    ///
+    /// # Panics
+    ///
+    /// If `index` is out of bounds.
+    pub fn replace(&mut self, index: u32, element: T) -> T {
+        self.get_mut_inner(index).replace(Some(element)).unwrap()
```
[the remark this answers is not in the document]
austinabell: yeah good call, I switched the API from what you suggested previously to return None when out of bounds on the inner function: [`e755735` (#402)](https://github.com/near/near-sdk-rs/pull/402/commits/e75573585bb14d0e0bb409c29b0723a287ad7573). I think this way makes it easier to make sure that the cache map is not modified outside of the bounds of the current vector.

near-sdk/src/store/vec/mod.rs
austinabell: Yeah, I agree it's a bit clunky. I've done the way you suggested before, esp with fuzzing, but in this case I was just porting over existing tests and expanding them to test the functionality I wanted.

near-sdk/src/store/vec/mod.rs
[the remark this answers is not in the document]
austinabell: The value cannot be modified without marking it as dirty, so seems inconsistent to manually do it here and not do other places. 

In the current code, there is no way for this to be not marked as mutable, but I'd like to keep it as is if possible because the `replace` call can avoid marking as dirty if both values are `None`, so it's future-proofed and open to avoid unnecessary state writes.

near-sdk/src/store/vec/iter.rs
```diff
@@ -0,0 +1,158 @@
+use borsh::{BorshDeserialize, BorshSerialize};
+use core::{iter::FusedIterator, ops::Range};
+
+use super::{Vector, ERR_INDEX_OUT_OF_BOUNDS};
+use crate::env;
+
+/// An interator over references to each element in the stored vector.
```
mikedotexe, line 7:
```suggestion
/// An iterator over references to each element in the stored vector.
```

near-sdk/src/store/vec/iter.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    fn nth_back(&mut self, n: usize) -> Option<Self::Item> {
+        let idx = self.range.nth_back(n)?;
+        Some(self.vec.get(idx).unwrap_or_else(|| env::panic_str(ERR_INDEX_OUT_OF_BOUNDS)))
+    }
+}
+
+/// An iterator over exclusive references to each element of a stored vector.
+#[derive(Debug)]
```
mikedotexe, line 76: Do we want the `Debug` here? Genuinely don't know but saw this in `near-sdk/src/collections/vector.rs` and thought I'd mentioned it, seeing if this is a good pattern to follow:
`#[cfg_attr(not(feature = "expensive-debug"), derive(Debug))]`
https://github.com/near/near-sdk-rs/blob/072b47779fcb219f468977bd7e727b4a83d9bd44/near-sdk/src/collections/vector.rs#L22

near-sdk/src/utils/stable_map.rs
```diff
@@ -0,0 +1,39 @@
+use std::cell::RefCell;
+use std::collections::BTreeMap;
+
+pub(crate) struct StableMap<K, V> {
```
ChaoticTempest, line 4: is this struct just for the purpose of providing us with unwrapped/default values? It's internal only, but I feel like whoever has to debug later might find it odd that `get()` modifies state, so feels like a footgun to me

near-sdk/src/store/vec/iter.rs
austinabell: Well, since derived, it would use this pattern for when the feature is enabled. There is nothing in the iterator that would be different here

near-sdk/src/utils/stable_map.rs
austinabell: It acts as an append-only map (which allows you to keep references to values). This is needed for the caching layer.

Yeah, potentially, this should be documented for future developers, but `get()` only inserts a default value if not existing. All usages have a default value that acts as if it doesn't exist but can be changed in place having a real value. There could be alternative API options, but they would increase complexity with no benefit. Do you have a suggestion on what could be better, or are you suggesting documenting better?

ChaoticTempest: yea, just better or more documentation for this

ChaoticTempest (approved):

mikedotexe (approved):

PR #404, opened 2021-05-05
Explicitly set Rust toolchain to 1.51.0

olonho:


evgenykuzyakov (commented):
Why?

EDIT: Here is PR and the discussion that lead to the removal of the rust-toolchain https://github.com/near/near-sdk-rs/pull/376 Does this break tests somewhere?

olonho:
OK, let's abandon it.

Issue #406, opened 2021-05-05
Add Lazy collection type


austinabell:
There is currently a `LazyOption` type, which behaves as an `Option` but it would be beneficial to just have a lazily loaded value where the assertion that a value is always stored is kept (stored on initialization and errors if not found) instead of leaving to developers to check/unwrap in code.

Side benefit of this (and lazy option) is to have the values and changes cached to avoid duplicate reads/writes to storage, similar to what is being explored with #402. This would be a good candidate for using `oncecell` @matklad might have thoughts as to if there is overhead that won't be acceptable in this context (there are some thread blocking components which I'm not sure how it fits into the wasm smart contract env) or re-using the cache primitive from #402.

This caching change can also be applied to `LazyOption` but likely all changes would live under the `unstable` flag.

cc @evgenykuzyakov as this was related to what I asked about in contract review

PR #409, opened 2021-05-07
feat(collections): Lazy storage value

austinabell:
As mentioned in #406 this implements the lazy storage type, which is similar to `LazyOption` except that the assertion that the value exists in storage is maintained by the API to allow easier usage for developers.

This PR re-uses the cache entry type from #402 (with small adjustments) to solidify that API and ensure it can be used for other types. Also for simplicity, this brings in the `once_cell` crate for lazily filling the cache cell under the feature flag, but only specific logic needed can be ported over to minimize compiled size if necessary.


austinabell:
per our discussions, I will be moving this type to the new module. The name is still not specified, so either `storage` or `store`, does anyone have opinions?

near-sdk/Cargo.toml
```diff
@@ -29,6 +29,9 @@ near-primitives-core = "=0.4.0"
 # Export dependencies for contracts
 wee_alloc = { version = "0.4.5", default-features = false, features = [] }
 
+# Used for caching, might be worth porting only functionality needed.
+once_cell = { version = "1.7.2", optional = true }
```
matklad, line 33:
```suggestion
once_cell = { version = "1.7.2", optional = true, default-features = false }
```

I think need only no_std (single threaded ) version.

matklad:
`store` feels better to me -- shorter, and it has a "a specific thing" connotation, which is good -- `store` refers to a specific global store.

evgenykuzyakov (approved):

near-sdk/src/store/lazy/impls.rs
```diff
@@ -0,0 +1,91 @@
+use borsh::{BorshDeserialize, BorshSerialize};
+
+use super::Lazy;
```
mikedotexe, line 3: me this weekend.

near-sdk/src/store/lazy/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+    let serialized = value.try_to_vec().unwrap_or_else(|_| env::panic(ERR_VALUE_SERIALIZATION));
+    env::storage_write(&key, &serialized);
+}
+
+/// An persistent lazily loaded value, that stores a value in the storage.
+///
+/// This will only write to the underlying store if the value has changed, and will only read the
+/// existing value from storage once.
+///
+/// # Examples
```
mikedotexe, line 52: Again appreciate these doc examples

mikedotexe (approved):

PR #412, opened 2021-05-10
feat: add compiled code size tests

austinabell:
cc @matklad as he suggested adding these before making other changes/refactoring.

The tests compile to wasm using release and any other usual flags and check the output wasm binary size.

The test checks to make sure it's no larger than 10% of what it currently is, to accommodate for external/platform discrepancies. Still feels a bit janky to me to test like this personally, but seems like it might be a reasonable check to have.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,5 +24,7 @@ jobs:
       - uses: Swatinem/rust-cache@v1
       - name: Test Format
         run: cargo fmt -- --check
+      - name: Add wasm32 target
+        run: rustup target add wasm32-unknown-unknown
       - name: Test
         run: cargo test --all
--- a/near-sdk/tests/code_size.rs
+++ b/near-sdk/tests/code_size.rs
@@ -1,0 +1,37 @@
+/// Compiles contract to wasm with release configuration and returns the code size.
+fn check_example_size(example: &str) -> usize {
+    let status = std::process::Command::new("cargo")
+        .env("RUSTFLAGS", "-C link-arg=-s")
+        .args(&["build", "--release", "--target", "wasm32-unknown-unknown", "--manifest-path"])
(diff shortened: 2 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -24,5 +24,7 @@ jobs:
       - uses: Swatinem/rust-cache@v1
       - name: Test Format
         run: cargo fmt -- --check
+      - name: Add wasm32 target
+        run: rustup target add wasm32-unknown-unknown
```
matklad, line 28: If I am not mistaken, this should not be necessary? `actions-rs/toolchain@v1` already installs this target, no?

near-sdk/tests/code_size.rs
```diff
@@ -0,0 +1,37 @@
+/// Compiles contract to wasm with release configuration and returns the code size.
+fn check_example_size(example: &str) -> usize {
+    let status = std::process::Command::new("cargo")
```
matklad, line 3: I'd add `Command::new("rustup").args(&["target", "add", "wasm32-unknown-unknown"]).status().unwrap()` here, so that the test *just works* in more cases.

austinabell: hmm, that seems like a bad implication to update a user's toolchain unknowingly through running a test, maybe if this test was ignored by default?

austinabell: yeah, that's a fair point. I'll make this change

austinabell:
reverted the change @matklad as there were interactions through rustup which don't play well in the test, locally and on CI. Let me know if you have a strong opinion on this and I can push the commit again and try to find a way around

matklad:
hey, looks my issue from a while ago about this is seeing some recent progress: https://github.com/rust-lang/rustup/issues/988

evgenykuzyakov (approved):

mikedotexe (approved):

PR #414, opened 2021-05-11
chore: remove public visibility of b58 error variant

austinabell:
As pointed out by @matklad [here](https://github.com/near/near-sdk-rs/pull/398#discussion_r630385963) it's probably best if we don't expose the error variants, to make it easier to make changes without being stuck to specific impls (in this case the base 58 error).

Probably best if this comes in before the next release, sorry for overlooking exposing this!

### DIFF
--- a/near-sdk/src/json_types/public_key.rs
+++ b/near-sdk/src/json_types/public_key.rs
@@ -170,10 +170,10 @@ impl std::fmt::Display for ParsePublicKeyError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self.kind {
             ParsePublicKeyErrorKind::InvalidLength(l) => {
-                write!(f, "Invalid length of the public key, expected 32 got {}", l)
+                write!(f, "invalid length of the public key, expected 32 got {}", l)
             }
-            ParsePublicKeyErrorKind::Base58(e) => write!(f, "Base58 decoding error: {}", e),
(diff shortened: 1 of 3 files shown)


near-sdk/src/json_types/public_key.rs
```diff
@@ -152,29 +154,33 @@ impl std::str::FromStr for Base58PublicKey {
...
-        match self {
-            Self::InvalidLength(l) => {
+        match self.kind {
+            ParsePublicKeyErrorKind::InvalidLength(l) => {
                 write!(f, "Invalid length of the public key, expected 32 got {}", l)
             }
-            Self::Base58(e) => write!(f, "Base58 decoding error: {}", e),
-            Self::UnknownCurve => write!(f, "Unknown curve kind"),
+            ParsePublicKeyErrorKind::Base58(e) => write!(f, "Base58 decoding error: {}", e),
+            ParsePublicKeyErrorKind::UnknownCurve => write!(f, "Unknown curve kind"),
```
matklad, line 176: Unrelated to the PR, but let's switch error messages to start with the lowercase letter, as recommended by the API guidelines and stdlib: https://doc.rust-lang.org/beta/std/error/trait.Error.html.

matklad (approved):

matklad:
> sorry for overlooking exposing this!

No worries, that's what reviews are for! I guess, a useful action item here would be, before the next release, for someone to walk through the generated rustdoc documentation, to see if the API accidentlally exposes stuff it shouldn't.

evgenykuzyakov (approved):

Issue #372, opened 2021-04-21
Untangle dependencies between nearcore and near-sdk (-sim)


matklad:
My understanding is that there's a bunch of accidental complexity around the interactions between near-core and near-sdk-rs code basses. Specifically, I see the following pains:

*   code organization in the near-core depends on near-sdk needs (primitives / primitives-core split)
*   changes to nearcore can break simulator implementation
*   changes to nearcore can break sdk public APIs
*   nearcore needs to think about publishing and versioning crates
*   contracts pull in a lot of dependencies from nearcore (we complile near_vm_logic to wasm only to have the liker dead-code-eliminate it)
*   overall, coordination between core and sdk is required, which slow down development of both.
*   overall, core and sdk have opposing goals: sdk has to provide stable API for our users to confidently build on, core wants to have nimble and cheap to change code.

I see the following fundamental "points of coupling" between the core and the sdk:

*   types used by WASM contracts at runtime (example: core contract uses <span class="error">[PublicKey]</span>([https://github.com/near/core-contracts/blob/39e029619041b564ec08d33a8d65325b653b121b/staking-pool/src/lib.rs#L9](https://github.com/near/core-contracts/blob/39e029619041b564ec08d33a8d65325b653b121b/staking-pool/src/lib.rs#L9)) defined in near-vm-logic)
*   vm logic used by `MockedBlockchain` for "unit-tests" inside the contract itself (<span class="error">[example]</span>([https://github.com/near/near-sdk-rs/blob/31cba6213919ec4866bc0fab7f6963de241cc43a/near-sdk/src/environment/mocked_blockchain.rs#L16](https://github.com/near/near-sdk-rs/blob/31cba6213919ec4866bc0fab7f6963de241cc43a/near-sdk/src/environment/mocked_blockchain.rs#L16)))
*   runtime logic used by the simulator (<span class="error">[example]</span>([https://github.com/near/near-sdk-rs/blob/31cba6213919ec4866bc0fab7f6963de241cc43a/near-sdk-sim/src/runtime.rs#L24](https://github.com/near/near-sdk-rs/blob/31cba6213919ec4866bc0fab7f6963de241cc43a/near-sdk-sim/src/runtime.rs#L24)))

Here's what I think the dependencies should look like.

For the ***first category***, I think it's reasonable to require that sdk just never uses any types from core. Almost all types are just type-defs at the moment, re-defining them shouldn't be problematic. re-definition would allow us to evolve types flexibly. 

[...]

Issue #387, opened 2021-04-29
List of odd APIs


matklad:
This issue is for me to brain dump all API papaercuts I find, just so that I don't forget about them

* `env::log` uses `[u8]` rather then `str`
*  `near_sdk::CryptoHash` and `near_sdk::MerkleHash` which are similar, but not the same
* `type Duration = u64` while we have `core::time::Duration`
* `Gas` is not a newtype (along with a bunch of other types)
* `ValidAccountId` -- just `AccountId` should always be valid
* `impl TryFrom<String> for CurveType` should be `FromStr`
* the whole `json_types` module is suspicious. If we newtype most of the things, the need for this module should go away
* "Exporting common crates" feels a bit odd, I think that maybe some of those crates want to be opt-in? In particular, serde_json looks like something that not every contrat needs
* `VMConfig` and `VMContext` are available when `target = wasm32`
* `init` methods require implementing paniking defaults yourselft
* `&mut self` methods can overwrite manualy changed contract storage, footgun when writing migrations.

austinabell:
The most prevalent one I've been thinking about has been around the collections API, specifically the map types.

It would be ideal if the API matches `BTreeMap`/`HashMap` closely. Specifically, the annoying point was for accessor methods, where the functions accepted `&K` where you really only need any generic type `Q` (in std for example) where `K: Borrow<Q>` which might seem like a small difference, but if you have a `&str` and the key type is `String`, you need to convert it to a string and take a reference to it, which is unergonomic and less performant. This does open the compiled code size to be larger due to added generics use, but I think it's worth exploring.

I think it would be also awesome if we added the entry API to these types, which would come with some overhead of having to cache all values (which I think is a thing to strive for anyway to avoid redundant storage reads), but it allows for some very clean usage.

If these data structures are modified to cache loaded values from storage, which would take some careful checking of the safety to make sure the API couldn't be misused, we can have APIs for these collections that more closely match std and also potentially save on gas usages from redundant reads.

I'll edit this comment with other ideas to list in the same form as your list, I just needed to add some additional context for this example because it's larger.

matklad:
Right, there's also a minor point that `UnorderedMap` is unlikely to be the best name -- it feels like a carry over from C++, where this name was chosen due to backwards compatibility concerns: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2003/n1456.html

evgenykuzyakov:
[...]

PR #417, opened 2021-05-13
refactor: replace syscall interactions in wasm32 environment

### LINKED ISSUE #373: Consider alternative, simpler and faster way to mock runtime in sdk
TL;DR: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=833aa3518e2173f47b25508400b2934f

At the moment, we use a rather sophisticated blockchain interface to facilitate using both real wasm-based blockchain and a mock blockchaing for tests.

We store a dynamically dispatch blockchain interface in a thread local:

https://github.com/near/near-sdk-rs/blob/97fc632fcc58eb7ff7faad0be54ea8ec91dbf694/near-sdk/src/environment/env.rs#L17-L22

For the real blockchaing, if we are compiling for wasm, we include near's ABI here:

https://github.com/near/near-sdk-rs/blob/97fc632fcc58eb7ff7faad0be54ea8ec91dbf694/near-sdk-macros/res/sys.rs#L1-L9

And implement that dyn interface for it:

https://github.com/near/near-sdk-rs/blob/97fc632fcc58eb7ff7faad0be54ea8ec91dbf694/near-sdk-macros/res/near_blockchain.rs#L1-L16

Similarly, there's a mocked blockchain here:

https://github.com/near/near-sdk-rs/blob/97fc632fcc58eb7ff7faad0be54ea8ec91dbf694/near-sdk/src/environment/mocked_blockchain.rs#L1

The result is that even wasm contracts need to go through thread-local access and dynamic dispatch, which hurts runtime.

https://github.com/near/near-sdk-rs/blob/97fc632fcc58eb7ff7faad0be54ea8ec91dbf694/near-sdk/src/environment/env.rs#L693-L703

On the complexity front, the implementation of sdk needs to explicitly care about existence of the mock blockchain, it can't just directly use ABI exposed by near.

There's a simple way to do this! Instead of using `&dyn` to inject a mock blockchain, we can just *define* the `extern "C"` interface ourselves:

```rust
extern "C" {
    pub fn log_utf8(len: u64, ptr: u64);
}

#[cfg(not(target_arch = "wasm32"))]
mod mock_chain {
    #[no_mangle]
    extern "C" fn log_utf8(len: u64, ptr: u64) {
        unsafe {
            let s = std::slice::from_raw_parts(ptr as *const u8, len as usize);
            let s = std::str::from_utf8_unchecked(s);
            println!("{}", s);
        }
    }
}

pub fn log(s: &str) {
    unsafe {
        log_utf8(s.len() as u64, s.as_ptr() as *const u8 as u64)
    }
}

fn main() {
    log("Hello, world");
}
```

The `extern "C" {` block corresponds to sys.rs file. Note that we include it unconditionally, for all tragets. 

The `log` function corresponds to the SDK API. It is codded against near ABI and, when compiled to wasm, produces the optimal code without any extra logic to handle "is this a mock blockchan?"

The `mock_chain` module is included if we are not compiling for wasm. It *defines* the functions of the near ABI. When the code is compiled for host, the linker links these mock definitions with declarations in the `extern "C"` block. Note that mock_chain may even live in a separate crate. 

To clarify, the mock chain would probably still want to use therad-local storage for a blockthain struct, to allow the user to set VMContext and such. However, there's no dynamic dispatch anywhere, the SDK proper doesn't need to think about mocking, and the compiled wasm code is unaware of mocks as well. 

I've learned about this linker-based dependency injection technique from the rust standard library initially: https://github.com/rust-lang/rust/blob/88b99dec2a4d103b8153ca9300bf0fbebdf65bda/library/std/src/panicking.rs#L33-L52

### LINKED ISSUE #416: Refactor to remove vm logic from wasm32 target
The vm_logic dependencies and logic should not be compiled for wasm32 environments and can be hidden behind config flags on the dep and code.

The first part to organize the dependencies is done with #415 and the goal for this is to remove the need to compile testing specific code when compiling contracts with the `wasm32-unknown-unknown` target.

austinabell:
- Removes need to go through thread local, refcell, dynamic dispatch just to call syscalls in wasm32 environments through their host functions
  - I also hid a lot of the related logic under the `wasm32` config (and might add more) which might be breaking if currently misused externally

Depends on #415 (not necessarily directly, but makes a lot of this a lot cleaner)

I tested with all examples and they all work, will go on to text external things now to check for breakages

Closes #373 
Closes #416

Progress on #372


near-sdk-macros/src/lib.rs
```diff
@@ -12,14 +12,10 @@ use syn::{File, ItemEnum, ItemImpl, ItemStruct, ItemTrait};
 #[proc_macro_attribute]
 pub fn near_bindgen(_attr: TokenStream, item: TokenStream) -> TokenStream {
     if let Ok(input) = syn::parse::<ItemStruct>(item.clone()) {
-        let sys_file = rust_file(include_str!("../res/sys.rs"));
```
matklad, line 15: :tada:

near-sdk/src/environment/sys.rs
```diff
@@ -0,0 +1,492 @@
...
+    pub fn storage_has_key(key_len: u64, key_ptr: u64) -> u64;
+    // ###############
+    // # Validator API #
+    // ###############
+    pub fn validator_stake(account_id_len: u64, account_id_ptr: u64, stake_ptr: u64);
+    pub fn validator_total_stake(stake_ptr: u64);
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+mod mock_chain {
```
matklad, line 146: Perhaps we can move this to the existing `mocked_blockchain.rs` file? I think there's a benefit in having `sys.rs` containing *only* the `extern "C"` block.

near-sdk/src/types/vm_types.rs
```diff
@@ -1,9 +1,26 @@
...
+#[derive(Debug, PartialEq)]
+pub enum PromiseResult {
+    /// Current version of the protocol never returns `PromiseResult::NotReady`.
+    NotReady,
+    Successful(Vec<u8>),
+    Failed,
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+pub use near_vm_logic::types::PromiseResult;
```
matklad, line 26: What if we remove `#[cfg(target_arch = "wasm32")]` and just always define our own `PromiseResult`, and use that in the `MockedBlockchain` interfaces, but, inside `MockedBlockchain`, convert `near_sdk::PromiseResult` into `near_vm_logic::types::PromiseResult`?

matklad (commented):
-200 lines look nice!

near-sdk/src/environment/sys.rs
austinabell: Yeah, I'll do that! I was deciding between the two, but I put it there because I figured if any API changed, it would be easy to see that this mock implementation needs to be updated.

near-sdk/src/types/vm_types.rs
austinabell: I didn't because the only interface when it is used, the parameter is a `&[PromiseResult]` and I wanted to avoid creating a new vector just to pass it as a parameter. I'll change to this now though, because I do agree it could cause some jankyness, but I figured this would be abstracted away sometime soon.

austinabell:
Think we need to make the `sys` module public to allow for contracts, like ref-finance, to retain low level controls that were previously going through the `BLOCKCHAIN_INTERFACE`.

For now, I'll just make it public, but I'd love opinions on how this should be namespaced (within `env` or exposed at root level) and if this should live under the `unstable` feature flag

austinabell:
I put the `sys` interface (syscalls) under an unstable API for now. There are still some necessary interactions that were previously using the `BLOCKCHAIN_INTERFACE` directly in the smart contract, but ideally, this would have some safe abstraction on top of it to avoid exposing the raw syscalls to a developer, which could change in the future.

This will absolutely be a breaking change but would be very good to get in if working towards having a stable contract upgradability pattern to use this updated interface.

matklad:
>Think we need to make the sys module public to allow for contracts, like ref-finance, to retain low level controls that were previously going through the BLOCKCHAIN_INTERFACE.

Oh dear, so we actually publicly exposed the whole edifice, with thread_local and such? 

I think the solution of exposing the naked `sys` module is a good one! I'd probably keep it at the top-level `near_sdk::sys` and not in  `environment::sys`, but that's a minor point. The important one is that, from the outside, *only* `near_sdk::sys` is visible, and I think we are good on that front. We might also consider publishing a `near-sys` crate in the future, so that it's easy to build alternative sdks, but I don't think necessary near-term (squatted the name non the less). If we publish the sys-crate, we can backwards compatible re-export it as a sys module. 

The question is, what do we do with version numbers though?

* we can publish this as a new major version, but that will add needless churn
* we can publish this in a minor version, pulling a semver trick and saying that this wasn't a documented public API anyway,

matklad:
>Yeah, this is definitely a good idea, how about we keep the interface under unstable for a bit and see what exactly will be needed/can be abstracted, then move it into this other crate after some concrete use?

:+1: 

>Yeah, this is definitely a good idea, how about we keep the interface under unstable for a bit and see what exactly will be needed/can be abstracted, then move it into this other crate after some concrete use?

I think we need to explicitly decide what we do with respect to version numbers & publishing. Thinking about this some more, the choices are:

* make a minor release ASAP, partners who use low-level interface will be seriously annoyed, other partners wont notice
* collect any other super major changes we want to make, and which we can't make later, and release a major ASAP. People who use low-level API will be annoyed by churn, but would love simpler sys calls. People who don't use low-level API will be slightly annoyed by a major revision which doesn't actually improve anything for them. We loose some "stability" marketing points
* fold other API cleanups (like strongly type AccountIds and such) into this change, and release a new major much later. 
* bite the bullet and just preserve API compat. That is, leave thread_local and dyn Trait there, but implement the trait in terms of sys module

matklad:
To me it seems that the least controversial option is 2, burn yet another major version number. It has the fewest unknown unknowns.

austinabell:
> I think we need to explicitly decide what we do with respect to version numbers & publishing. Thinking about this some more, the choices are:
> 
> * make a minor release ASAP, partners who use low-level interface will be seriously annoyed, other partners wont notice
> * collect any other super major changes we want to make, and which we can't make later, and release a major ASAP. People who use low-level API will be annoyed by churn, but would love simpler sys calls. People who don't use low-level API will be slightly annoyed by a major revision which doesn't actually improve anything for them. We loose some "stability" marketing points
> * fold other API cleanups (like strongly type AccountIds and such) into this change, and release a new major much later.
> * bite the bullet and just preserve API compat. That is, leave thread_local and dyn Trait there, but implement the trait in terms of sys module

Yeah maybe option 2 is the best option, and I'm having a tough time making a point for another. 

I would really like to avoid keeping the `BLOCKCHAIN_INTERFACE` as it has literally no benefit and creates some really ugly code patterns to interact with directly.

Let's create a list of all things that should change for a major release and try to work to that ASAP (while limiting as many breaking changes), and can always decide to defer making these changes if the scoped out list is not suitable for external/internal parties. Maybe we can use #387 for coordination on this?

near-sdk/src/environment/env.rs
```diff
@@ -80,6 +77,7 @@ macro_rules! method_into_register {
 ///       );
 /// near_sdk::env::set_blockchain_interface(Box::new(mocked_blockchain));
 /// ```
+#[cfg(not(target_arch = "wasm32"))]
```
matklad, line 80: If we are making a breaking change here, perhaps we should move these public methods elsewhere? 

I think, ideally, we'll have one single public cfg module, so as not to duplicate cfgs

```
#[cfg(not(target_arch = "wasm32"))]
pub mock {
}
```

austinabell: Trying to reduce the amount of breakage for this release. If I were to do this, I would still like to keep this method here and just mark it as deprecated. Do you think this is a good idea?

Honestly seems like there are a lot of breaking changes in next major release and I don't want it to be annoying to migrate.

matklad: Yeah, that's reasonable, better to keep it here *without* adding a deprecation.

A deprecation can be added in a minor release, and it's nice to be able to switch to new major without seeing a bunch of new deprecations from the get-go.

mikedotexe (approved):
Thanks for the discussion in our meeting about this today. 👍🏼

matklad (approved):

PR #419, opened 2021-05-18
chore: remove release notes from README and add unreleased changes to CHANGELOG

austinabell:
- Removes release notes from README to remove clutter
- Adds unreleased changes to changelog
  - This only includes changes that would be relevant to SDK users, but I am open to updating any format or including all changes
- Adds date to releases.
  - this can be useful for users to easily check what has changed since they last used it or know how mature a release is

Please share any options on how this should be structured, as it would be best if this has a format we're all happy with to keep this up to date. I considered having crate level `CHANGELOG`s but decided against it because there wasn't a need for that granular of changes it seems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,20 @@
 # Changelog
 
-## `3.1.0`
+## [unreleased]
+
+* Move type aliases and core types to near-sdk to avoid coupling. [PR 415](https://github.com/near/near-sdk-rs/pull/415).
+* Implements new `Lazy` type under the new `unstable` feature which is a lazily loaded storage value. [PR 409](https://github.com/near/near-sdk-rs/pull/409).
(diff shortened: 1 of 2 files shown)


mikedotexe:
I dig this approach. Let's keep the README as simple as possible.
I like the way Rust has it laid out here, if we're talking about structure:
https://github.com/rust-lang/rust/blob/master/RELEASES.md

I'd say having the dates is also great (Rust RELEASES file does that, too) and if we can give examples of old way vs. new way, that would be extra credit.

mikedotexe (approved):
good eye, I hadn't thought of this in a while. appreciate your fresh eyes on this subject

austinabell:
> I dig this approach. Let's keep the README as simple as possible.
> I like the way Rust has it laid out here, if we're talking about structure:
> https://github.com/rust-lang/rust/blob/master/RELEASES.md
> 
> I'd say having the dates is also great (Rust RELEASES file does that, too) and if we can give examples of old way vs. new way, that would be extra credit.

Yeah it's very subjective, so I didn't change anything major. Do you think there is any information or format that should be changed/added?

PR #427, opened 2021-05-26
Modify ci-test script so it can be run on OS X

mikedotexe:
When attempting to run the `ci-test.sh` script, the use of `sed` has an issue on OS X.
After reading this article:
https://ed.gs/2016/01/26/macos-sed-invalid-command-code
found that this small modification will allow it to run locally. It looks like this change is only additive and linux systems (used for CI) shouldn't have an issue with it.

### DIFF
--- a/ci-test.sh
+++ b/ci-test.sh
@@ -11,7 +11,7 @@ else
     cp Cargo.toml{,.bak}
     cp Cargo.lock{,.bak}
 
-    sed -i "s|###||g" Cargo.toml
+    sed -i "" "s|###||g" Cargo.toml
     
     set +e
     cargo test --all


austinabell (approved):

matklad:
JFYI, with cargo it's actually possible to write such ad-hoc build scripts in rust. So that

```
$ git clone git@github.com:near/near-sdk-rs.git && cd near-sdk-rs
$ cargo ci-test
```

just works without having to install anything besides Cargo. In my experience, over the long run, that's beneficial over using bash scripts, as you don't have inconsistencies between OSes like this one, and keep everything in a single language. 

It probably doesn't really matter that much for this scale of project, but a useful trick to keep in mind: https://github.com/matklad/cargo-xtask

PR #428, opened 2021-05-26
feat: Implement optimized TreeMap iterator functions

### LINKED ISSUE #426: Extend TreeMap iterator implementation
In order to optimize computation and storage reads, there are some iterator methods that can be implemented to avoid using the naive implementations (which just call `next` n times and loads unnecessary values from storage).

The implementations that can be added are:
- `size_hint`: can use the tree `Vector`'s length for upper/lower bound if the amount of iterated elements is being counted. This one isn't relied on but might avoid reallocations when collecting the iterator.
- `count`: Can just return the `number of elements in the tree - the number of items iterated` and avoid all storage reads, instead of performing two storage reads per element left.
- `last`: can use the `max` function currently implemented and avoid iterating through the treemap
- `nth`: Can just find the `nth` element by iterating through the tree vector, and avoid loading all values. There still will be redundant reads until the vector implementation caches reads (ie #402), but will still be much better than what exists

austinabell:
Closes #426 

This PR just makes the implementation more efficient, to avoid unnecessary storage reads

This should be functionally the same externally, but since it will reduce storage reads it's a breaking change in a smart contract context


near-sdk/src/collections/tree_map.rs
```diff
@@ -600,17 +600,61 @@ where
...
+        if self.asc && matches!(self.hi, Bound::Unbounded) {
+            self.map.max().and_then(|k| self.map.get(&k).map(|v| (k, v)))
+        } else if !self.asc && matches!(self.lo, Bound::Unbounded) {
+            self.map.min().and_then(|k| self.map.get(&k).map(|v| (k, v)))
+        } else {
+            // Cannot guarantee what the last is within the range, must load keys until last.
+            let mut key: Option<K> = None;
+            while self.key.is_some() {
+                key = self.progress_key();
+            }
```
matklad, line 647:
```suggestion
            // Cannot guarantee what the last is within the range, must load keys until last.
            let key= std::iter::from_fn(|| self.progress_key()).last()
```

matklad:
Probably makes sense to add some smoke tests here. There was a funny bug in std when a combination of nth and rev flat out gave wrong results in some cases, and nobody noticed )

matklad:
>Probably makes sense to add some smoke tests here.

not related to the PR per-se, but it'd be cool to look into setting up property-based/fuzz testing for collections. std collections give us a good, free oracle we can compare against. The two relevant crates in the rust ecosystem are (from most important to least importat)

* https://lib.rs/crates/arbitrary
* https://lib.rs/crates/quickcheck
* https://lib.rs/crates/proptest

Which reminds me that I should dig into https://github.com/graydon/photesthesis at some point...

austinabell:
> > Probably makes sense to add some smoke tests here.
> 
> not related to the PR per-se, but it'd be cool to look into setting up property-based/fuzz testing for collections. std collections give us a good, free oracle we can compare against. The two relevant crates in the rust ecosystem are (from most important to least importat)
> 
> * [lib.rs/crates/arbitrary](https://lib.rs/crates/arbitrary)
> * [lib.rs/crates/quickcheck](https://lib.rs/crates/quickcheck)
> * [lib.rs/crates/proptest](https://lib.rs/crates/proptest)
> 
> Which reminds me that I should dig into [graydon/photesthesis](https://github.com/graydon/photesthesis) at some point...

Yeah, I added some basic tests, but yeah fuzzing probably is out of the scope of this. I'd love to tackle this when I have time.

I assume when you mention arbitrary you mean to be used with [cargo-fuzz](https://github.com/rust-fuzz/cargo-fuzz)? quickcheck is already used in the sdk for tree map tests, so probably least cost to add those to be run as tests, but definitely would be nice to extensively fuzz all data structures and interactions.

austinabell:
> Not necessary -- I believe that the arbitrary crate is setup in such a way that is useful even without coverage-guided fuzzers, so it _should_ be possible to use it the same way `quickcheck` is used, via simple `#[test]`. I don't think the ecosystem is quite there yet, so it might be that we'll have to do some ground-work here ourselves. `arbitrary` does seem to be a very well-designed crate that captures the essense of randomized testing, so I'd like to find a way to marry convenient tests with it.

And the benefit of doing this over using `quickcheck`? More control?

matklad:
>And the benefit of doing this over using quickcheck? More control?

The `arbitrary` crate is a good ecosystem-wide interface. I expect that, over time, a lot of other crates would implement `arbitrary::Arbitrary`. So, if we make our stuff `Adbitrary`, then we'll be more inter-operable. 

That's I am betting on `arbitrary`: it's a 1.0 crate which defines just the narrow interface for randomization, _without_ being opinionated about how actual property testing is commenced.

austinabell:
> The `arbitrary` crate is a good ecosystem-wide interface. I expect that, over time, a lot of other crates would implement `arbitrary::Arbitrary`. So, if we make our stuff `Adbitrary`, then we'll be more inter-operable.
> 
> That's I am betting on `arbitrary`: it's a 1.0 crate which defines just the narrow interface for randomization, _without_ being opinionated about how actual property testing is commenced.

You okay with this being done in a separate PR? I can create an issue later.

Rather not increase the scope of this PR if possible

matklad:
>You okay with this being done in a separate PR? I can create an issue later.

Sure, I am always for splitting up PR, and for non-blocking PRs!

mikedotexe (approved):

PR #433, opened 2021-06-01
example e2e test written with near-api-js run on near-sandbox

ailisp:
This was part of near/docs#706, @thisisjoshford points out the testing script is over sized for doc, so move it here for a better organization.


thisisjoshford (approved):
Awesome! My suggestion wasn't so much that the code snippet is too long for the doc but more easily accessible and maintainable if it is a part of the SDK. Now if something changes that breaks the example, the test will fail forcing us to update it. 

Also, the tutorial is reliant on a wasm file thats part of this SDK so it will be easier to clone and run without having to copy/paste wasm files and code snippets as the original tutorial guided you through.

austinabell:
> Awesome! My suggestion wasn't so much that the code snippet is too long for the doc but more easily accessible and maintainable if it is a part of the SDK. Now if something changes that breaks the example, the test will fail forcing us to update it.
> 
> Also, the tutorial is reliant on a wasm file thats part of this SDK so it will be easier to clone and run without having to copy/paste wasm files and code snippets as the original tutorial guided you through.

But this test wouldn't be run by the CI, right? It doesn't seem any more forcing to keep up to date if put here, but I do agree it could be nice to see an example like this attached to a Rust example.

It does seem like it couples these things together and I don't see the direct connection to the Rust SDK, but also seems like a benefit to have here, so I don't really have a strong opinion on this.

examples/status-message/package.json
```diff
@@ -0,0 +1,15 @@
...
+  "description": "Records the status messages of the accounts that call this contract.",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "borsh": "^0.4.0",
+    "near-api-js": "near/near-api-js#bace1ee"
```
austinabell, line 13: do we want to use a specific version or is there a reason this commit is used?

austinabell (commented):
LGTM, I'll leave to others judgement for whether this is a good location for this example

examples/status-message/package.json
ailisp: used to need this commit after current release (May 7 0.41.0): 
image

because near-api-js didn't handle `{"result":null}` jsonrpc response correctly. I think I don't need this commit now because I also changed in nearcore side to not return empty response

ailisp: Changed to 0.41.0

ailisp:
@evgenykuzyakov @mikedotexe Please take a look :)

examples/status-message/package.json
ailisp: @austinabell Hmm actually nearcore still returns null, because serde serialize this to null:
```
struct Foo;
```
It serialize `struct Bar {}` to `{}`, so i made this PR: https://github.com/near/nearcore/pull/4387, but before that or near-api-js have a new release, I need to pin near-api-js with this commit.

mikedotexe:
Apologies for the delay @ailisp 
I am confused on this PR, and also think it's an odd place to put it.
Like many folks, we are trying to guard how much we add to the CI tests. Not that I'm against it, but it seems like it must be a particularly good reason to add more.
This does seem like the wrong repo for this kind of test unless I'm missing something?
I think I might encourage this logic to live somewhere like the Guest Book example: https://github.com/near-examples/guest-book
Especially now that the Sandbox is in early stages.
I do think that wherever we put it, let's make sure there are instructions on how to correctly run this, or else have CI set everything up so it can be run.

What I tried:
Went into the `near-sdk-rs` repository, 

```bash
cd examples/status-message
npm i
```

then I tried:

`node test.js`

I got an error:

>(node:33133) UnhandledPromiseRejectionWarning: Error: Cannot find module '/tmp/near-sandbox/validator_key.json'

Realized I must need to run this locally, and saw that CI isn't running this, which Austin mentioned.
Went into my local `nearcore` repository, 

```bash
git pull
make release
```

Then:

`nearup run localnet --binary-path /Users/mike/near/nearcore/target/release`

Then, in another tab:

```bash
mkdir -p /tmp/near-sandbox
ln -s ~/.near/localnet/node0/validator_key.json /tmp/near-sandbox/validator_key.json
```

Then back in the `near-sdk-rs/examples/status-message`

Tried running:

`node test.js`

and now I get:

>Finish init NEAR
>(node:49147) UnhandledPromiseRejectionWarning: Error: [-32000] Server error: The node does not track the shard ID 0

Not sure what this means, but I don't think this PR is ready, and should probably be placed in Guest Book.

mikedotexe:
Okay, I now see that the docs from: https://github.com/near/docs/pull/706
are meant to accompany this. Sorry, from the discussion I thought everything was moved over here.
I do still think we should put this in Guest Book, but now I understand how to run it:

```bash
mike@n  ~/near/near-sdk-rs/examples/status-message   sandbox-e2e ± node test.js
Finish init NEAR
Finish deploy contracts and create test accounts
Receipt: pPG9wCtCsHg7TZoHN15XDaznXXZvW7jeJcnKQ59cBbh
	Log [status-message.test.near]: alice.test.near set_status with message hello
Log [status-message.test.near]: get_status for account_id alice.test.near
Log [status-message.test.near]: get_status for account_id bob.test.near
Receipt: GjUBqYfLQUjrmXYcHWR51uqyhrkrf9GogWhKQZ9H6Nxu
	Log [status-message.test.near]: bob.test.near set_status with message world
Log [status-message.test.near]: get_status for account_id bob.test.near
Log [status-message.test.near]: get_status for account_id alice.test.near
```

ailisp:
@mikedotexe Yep Seems you get it run! I agree it's better to be in a guest book example. Also I discovered this repo: https://github.com/near-examples/rust-status-message. Can we put test there so there's no need to adjust test code since guest book is similar but still a little different from status-message

PR #435, opened 2021-06-04
Restructuring of exports, add basic comments

mikedotexe:
Hey Austin, this is what we paired on. I am tempted to go hog wild adding comments, but I am eager to get the NFT contract standard merged. Another pass for documentation will have to do. Thanks for all the suggestions. Hoping we can just give this a quick perusal and merge into `standards/nft` and go from there.


near-contract-standards/src/non_fungible_token/approval/approval_impl.rs
```diff
@@ -1,10 +1,12 @@
+/// Common implementation of the [approval management standard](https://nomicon.io/Standards/NonFungibleToken/ApprovalManagement.html) for NFTs.
```
austinabell, line 1: So small point on these but you only use `///` for when you're documenting the thing below it. You can use `//!` at the top of the file like this to document the module itself

near-contract-standards/src/non_fungible_token/mod.rs
```diff
@@ -1,37 +1,24 @@
...
 pub mod metadata;
-/// Trait for the receiver of an `nft_transfer_call` method call. This method is
-/// implemented on a receiving account separate from the NFT contract. This is
-/// part of the core standard.
-pub mod receiver;
-/// Trait for the resolver of an `nft_transfer_call` method call. This method is
-/// implemented on the NFT contract and used to resolve how the transfer-and-call
-/// process went. This is part of the core standard.
-pub mod resolver;
 /// The Token struct for the non-fungible token.
```
austinabell, line 15: Not sure if intentional but this is just documenting the private module (doesn't matter, can leave, just noting it won't show up anywhere in docs)

near-contract-standards/src/non_fungible_token/token.rs
```diff
@@ -3,8 +3,10 @@ use near_sdk::serde::{Deserialize, Serialize};
 use near_sdk::AccountId;
 use std::collections::HashMap;
 
+/// Note that token IDs for NFTs are strings on NEAR. It's still fine to use autoincrementing numbers as unique IDs if desired, but they should be stringified. This is to make IDs more future-proof as chain-agnostic conventions and standards arise, and allows for more flexibility with considerations like bridging NFTs across chains, etc.
 pub type TokenId = String;
 
+/// In this implementation, the Token struct takes two extensions standards (metadata and approval) as optional fields, as they are frequently used in modern NFTs.
```
austinabell, line 9: Might be helpful to mention that it's an NFT token early in the comment (if trying to get info on an imported type could be a better devx)

near-contract-standards/src/non_fungible_token/token.rs
```diff
@@ -3,8 +3,10 @@ use near_sdk::serde::{Deserialize, Serialize};
 use near_sdk::AccountId;
 use std::collections::HashMap;
 
+/// Note that token IDs for NFTs are strings on NEAR. It's still fine to use autoincrementing numbers as unique IDs if desired, but they should be stringified. This is to make IDs more future-proof as chain-agnostic conventions and standards arise, and allows for more flexibility with considerations like bridging NFTs across chains, etc.
```
austinabell, line 6: This comment seems too long for one line, can mess with rust fmt or be hard to read (esp on a phone which I am now). Maybe worth splitting up?

near-contract-standards/src/non_fungible_token/mod.rs
mikedotexe: yeah, I considered removing it when I saw it wasn't in the docs, but what the heck, I'll let someone find it when browsing the codebase

austinabell (approved):

Issue #312, opened 2021-03-05
Unsafe AccountId to ValidAccountId conversion


ilblackdragon:
Can we add unsafe AccountId to ValidAccountId conversion so it's easier to pass them between each other in tests?
It can be pragma tests, for example.

evgenykuzyakov:
I think `#cfg(test)` doesn't propagate into other test modules. So the only option is to add into the `test_utils`

Should be fixed with strict types later

austinabell:
No need for this since `AccountId` is now validated by default, with `new_unchecked` as an option for initializing without checking in release

ilblackdragon:
I thought we created ValidAccountId to not check all the time validity when AccountId is just passed around?

austinabell:
The change was initially proposed in https://github.com/near/near-sdk-rs/issues/387. (PR for reference https://github.com/near/near-sdk-rs/pull/448)

Having the two types was confusing for developers and led to a clunky API. Yes, there can be some slight overhead, but I think that the benefits outweigh this. It seems reasonable to have it validated by default, and if anyone wants to hyper optimize their contract, they can use `String` as the type and create their alias if they wish to and/or use `AccountId::new_unchecked`.

cc @mikedotexe @ChaoticTempest

PR #448, opened 2021-06-24
refactor(types): change account id to newtype and deprecate ValidAccountId

### LINKED ISSUE #446: Remove `ValidAccountId` and make `AccountId` a newtype
The second wrapper type, `ValidAccountId` is only used for JSON serialization and validating the account ID is valid. This should all happen within one type, instead of `AccountId` being an alias to `String`

austinabell:
Definitely more involved than I would have liked and it's unclear what is the best way to initialize an `AccountId` but here is what changes:
- Changes `AccountId` from a `String` alias to a newtype which performs the validation, JSON/borsh serialization, and other necessary interactions
- Removes `ValidAccountId`
  - I was planning on aliasing `AccountId` for this, but it caused a lot of confusing errors linking to it instead of `AccountId`, and should probably be removed sooner than later

This is quite a large breaking change, because using `String` is quite nested within everything, and interactions from `ValidAccountId` specifically will cause breakages with existing contracts.

closes #446


near-sdk/src/types/account_id.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+impl std::str::FromStr for AccountId {
+    type Err = ParseAccountIdError;
+
+    fn from_str(value: &str) -> Result<Self, Self::Err> {
+        validate_account_id(value)?;
+        Ok(Self(value.to_string()))
+    }
+}
```
matklad, line 125: I think these two are the only two conversions we need. I think it's better to remove the rest of the conversions, as they can already be expressed by the user.

near-sdk/src/types/account_id.rs
```diff
@@ -0,0 +1,174 @@
...
+)]
+pub struct AccountId(String);
+
+impl AccountId {
+    /// Returns reference to the account ID bytes.
+    pub fn as_bytes(&self) -> &[u8] {
+        self.0.as_bytes()
+    }
+    /// Returns reference to the account ID string.
+    pub fn as_str(&self) -> &str {
```
matklad, line 30: I am debating if we should just `impl Deref<str> for AccountId`.

examples/status-message/src/lib.rs
```diff
@@ -21,11 +21,7 @@ impl StatusMessage {
         self.records.insert(account_id, message);
     }
 
-    pub fn get_status(&self, account_id: String) -> Option::<String> {
-        assert!(
-            env::is_valid_account_id(account_id.as_bytes()),
-            "Given account ID is invalid"
-        );
+    pub fn get_status(&self, account_id: AccountId) -> Option::<String> {
```
matklad, line 24:
```suggestion
    pub fn get_status(&self, account_id: AccountId) -> Option<String> {
```

Or is the turbofish actually required by the macro?

austinabell: It is required, I haven't dug into why specifically it's needed yet

near-sdk/src/types/account_id.rs
austinabell: I was thinking the same, but I wanted to explore without it first

near-sdk/src/types/account_id.rs
austinabell: yeah I was just trying to reduce the amount of breakages, since the types used previously have these conversions explicitly and would require changes on their end

examples/status-message/src/lib.rs
austinabell: it's needed because of the metadata macro generation, because it calls `Option::<String>::schema_container()`

near-sdk-sim/src/user.rs
```diff
@@ -35,10 +33,10 @@ type Runtime = Rc<RefCell<RuntimeStandalone>>;
 /// use near_sdk_sim::{to_yocto, account::AccessKey};
 /// use near_crypto::{InMemorySigner, KeyType, Signer};
 /// let master_account = near_sdk_sim::init_simulator(None);
-/// let account_id = "alice".to_string();
-/// let transaction = master_account.create_transaction(account_id.clone());
+/// let account_id = "alice";
+/// let transaction = master_account.create_transaction(account_id.parse().unwrap());
```
mikedotexe, line 37: Good attention to detail catching this in comment blocks 👍🏼

near-sdk/src/types/account_id.rs
```diff
@@ -0,0 +1,174 @@
...
+        self.0.as_str()
+    }
+    /// Consumes `self` to return inner [`String`] representation.
+    pub fn into_string(self) -> String {
+        self.0
+    }
+    // TODO this should probably be marked as unstable or crate scoped
```
mikedotexe, line 37: how do we want to handle this, with the unstable feature flag we've discussed at times?

near-sdk/src/types/account_id.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+#[derive(Debug)]
+pub struct ParseAccountIdError {
+    kind: ParseAccountIdErrorKind,
+}
+
+#[derive(Debug)]
+enum ParseAccountIdErrorKind {
+    InAccountId,
```
mikedotexe, line 134: `InvalidAccountId` too long?

mikedotexe (approved):
Looks quite thorough. And I believe you'd addressed what Kladov has brought up. My only silly suggestion is to have the one error begin with `Invalid` instead of `In` but 🤷🏼

near-sdk/src/types/account_id.rs
austinabell: Yeah I'd like to get others input about this. The conversion before was also not so certain (related #312) so it'd be nice to have others input to make sure we make a good long term API decision

near-sdk/src/types/account_id.rs
austinabell: oops I don't know how that happened in the migration

near-sdk-sim/src/user.rs
austinabell: I mean, rust doc tests catch this one, but I think I changed the ones outside of doc tests too, I'll double-check tomorrow

near-sdk/src/types/account_id.rs
austinabell: Just to update on thoughts, I think that adding the Deref to string opens up for a much more difficult API to maintain and really doesn't benefit in this case. I was originally considering `Deref<String>` because of how ubiquitously `String` is used for the account ID type, but I think it might be best to just move away from this pattern.

Also not sure if trying to move away from `String` inner representation altogether in favour of just `Vec<u8>` to minimize contract size in the future, and I'm just trying to be careful

frol:
Hey-hey! Node Interfaces team is working on a similar thing in nearcore (@miraclx has it in a branch state for now: https://github.com/near/nearcore/compare/master...miraclx:strict-account-id-validity) and even decided to create a separate crate out of it (`near-account-id`), so we can merge our efforts here.

austinabell:
> Hey-hey! Node Interfaces team is working on a similar thing in nearcore (@miraclx has it in a branch state for now: [near/nearcore@master...miraclx:strict-account-id-validity](https://github.com/near/nearcore/compare/master...miraclx:strict-account-id-validity)) and even decided to create a separate crate out of it (`near-account-id`), so we can merge our efforts here.

We can't use that implementation in its current state because it pulls in a few dependencies and uses regex for validation which is too bulky for the SDK (especially since an OKR for the last and upcoming quarter is minimizing contract size)

austinabell:
Because this is such a breaking change, it'd be great to get input from others.

Also, the primary remaining design decision is if we change the underlying type to `Box<[u8]>` which on one hand could allow for smaller code sizes but also does not hold up the compiler guarantee that it will always be utf8 bytes. It would also require calling `from_utf8` on the bytes for every reference, and I'm not sure how that gets optimized.

frol:
We can definitely onboard this implementation as I also don't feel very good about pulling the whole regex machinery to validate the account ID format.

Have you measured the difference of `Box<[u8]>` vs `Box<str>`?

austinabell:
> We can definitely onboard this implementation as I also don't feel very good about pulling the whole regex machinery to validate the account ID format.
> 
> Have you measured the difference of `Box<[u8]>` vs `Box<str>`?

The size of memory is the same, not sure about the compiled code. I am also noticing when switching to the Boxed versions that `BorshSchema` is broken for them. I can look into fixing this if it's possible.

austinabell:
@frol wrt above https://github.com/near/borsh-rs/pull/36 which is just a benefit to borsh, but would allow these alternatives to be used here (I checked)

near-sdk/src/types/account_id.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+}
+
+impl TryFrom<&[u8]> for AccountId {
+    type Error = ParseAccountIdError;
+
+    fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
+        core::str::from_utf8(value)
+            .map_err(|_| ParseAccountIdError { kind: ParseAccountIdErrorKind::InvalidUtf8 })?
+            .parse()
```
matklad, line 112: `?`-less alternative: 

```suggestion
        core::str::from_utf8(value).map_err(|_| ParseAccountIdError { kind: ParseAccountIdErrorKind::InvalidUtf8 })
            .and_then(str::parse)
```

near-sdk/src/types/account_id.rs
matklad: Not sure how I feel about this. The problem is that, if we don't take away this conversions now, we won't be able to remove them in the future. OTOH, it's not they are actively harmful. 

I guess, I'd comment them out, see what the breakage looks in practice, and then decide, biasing to remove. 

No strong opinion here though!

near-sdk/src/types/account_id.rs
```diff
@@ -0,0 +1,200 @@
...
+#[derive(Debug)]
+pub struct ParseAccountIdError {
+    kind: ParseAccountIdErrorKind,
+}
+
+#[derive(Debug)]
+enum ParseAccountIdErrorKind {
+    InvalidAccountId,
+    InvalidUtf8,
+}
```
matklad, line 162: Let's derive at least `Clone` as well, as that's impossible to provide at call-site. We might also derive PartialEq/Eq

near-sdk/src/types/mod.rs
```diff
@@ -4,6 +4,9 @@ pub use self::vm_types::*;
 mod primitives;
 pub use self::primitives::*;
 
+mod account_id;
+pub use self::account_id::*;
```
matklad, line 8: Style preference -- for public re-exports, I like to explicitlely list all things exported and avoid `*` -- that way, if the user learns the library by reading the source, they get a view of the public API at a glance. 

This is unrealated to the PR, and better be changed (if we do want to change this) in a follow-up.

near-sdk/src/types/account_id.rs
```diff
@@ -0,0 +1,200 @@
...
+
+impl std::fmt::Display for ParseAccountIdError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self.kind {
+            ParseAccountIdErrorKind::InvalidAccountId => write!(f, "the account ID is invalid"),
+            ParseAccountIdErrorKind::InvalidUtf8 => write!(f, "bytes are not valid utf-8"),
+        }
+    }
+}
```
matklad, line 171:
```suggestion
impl fmt::Display for ParseAccountIdError {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        match self.kind {
            ParseAccountIdErrorKind::InvalidAccountId => write!(f, "the account ID is invalid"),
            ParseAccountIdErrorKind::InvalidUtf8 => write!(f, "bytes are not valid utf-8"),
        }
    }
}
```

matklad (commented):
Looks good to me, left a whole bunch of nits, but only three serious comments:

* deriving BorshDeserialize for AccountId is wrong
* still not sure that extra conversions are worth it, but don't see anything particularly problematic there
* i don't undrestand the reason for AsRef-ing promise API

HELP.md
```diff
@@ -320,15 +320,15 @@ impl Contract {
...
         self.owner_id = new_owner_id.into();
     }
 
     /// View method that "modifies" state, for code structure or computational
     /// efficiency reasons. Changes state in-memory, but does NOT save the new
     /// state. If called internally by a change method, WILL result in updated
     /// contract state.
-    pub fn update_stats(&self, account_id: ValidAccountId, score: U64) -> Account {
+    pub fn update_stats(&self, account_id: AccountId, score: U64) -> Account {
         let account = self.accounts.get(account_id).expect("account not found");
```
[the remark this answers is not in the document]
austinabell: > Actually, meta: why do we need `HELP.md` at all? Seems to be a rather non-intuitive place for this. Ideally, code examples should run as a part of test-suite. One way to do this is to move the docs into a separate doc-only `guide` module in the SDK (as a bonus, the docs are now available on doc.rs and offline via cargo doc) or into an mdbook (I _think_ mdbook can run the examples). Completely unrealated to the PR obviously.

I agree, would like to migrate away from this pattern for sure. The mdbook style docs are almost ready to be public so can move any still relevant docs over and remove once it is

near-sdk/src/environment/env.rs
```diff
@@ -393,8 +394,8 @@ pub fn promise_and(promise_indices: &[PromiseIndex]) -> PromiseIndex {
     }
 }
 
-pub fn promise_batch_create<A: Borrow<AccountId>>(account_id: A) -> PromiseIndex {
-    let account_id = account_id.borrow();
+pub fn promise_batch_create<A: AsRef<str>>(account_id: A) -> PromiseIndex {
```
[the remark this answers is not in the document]
austinabell: yeah, I was just trying to minimize breakages and validation required to be compiled in. Let me give this another look though

near-sdk/src/test_utils/test_env.rs
```diff
@@ -1,15 +1,15 @@
 use crate::test_utils::VMContextBuilder;
-use crate::{env, AccountId, MockedBlockchain, VMConfig};
+use crate::{env, MockedBlockchain, VMConfig};
 
-pub fn alice() -> AccountId {
+pub fn alice() -> String {
```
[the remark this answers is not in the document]
austinabell: I think it would make sense to deprecate these eventually, but was trying to avoid changing the type for this to make migration easier. This is just testing utils but I suppose this should be updated

near-sdk/src/types/account_id.rs
austinabell: Just for readability?

matklad: More for minimality, the `?` does early return and `Into` conversion, and we don't really need those here. 

There's also https://github.com/rust-lang/rust/issues/37939, which make me cautious about using `?` where I can cleanly avoid it.

examples/cross-contract-low-level/tests/general.rs
```diff
@@ -39,7 +39,7 @@ fn check_promise() {
     let (master_account, contract, _alice) = init(to_yocto("10000"));
     let res = view!(contract.promise_checked());
     assert_eq!(res.unwrap_json::<bool>(), false);
-    let status_id = "status".to_string();
+    let status_id: near_sdk::AccountId = "status".parse().unwrap();
```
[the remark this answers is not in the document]
austinabell: I'll check around the code size in a bit. Committing the wasm files was a mistake and how I built them also causes a difference on `master`. Definitely want to look into making it part of the CI in some way to check differences

examples/fungible-token/tests/sim/utils.rs
```diff
@@ -19,10 +20,10 @@ const DEFI_ID: &str = "defi";
 // Register the given `user` with FT contract
 pub fn register_user(user: &near_sdk_sim::UserAccount) {
     user.call(
-        FT_ID.to_string(),
+        FT_ID.parse().unwrap(),
```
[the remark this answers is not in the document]
austinabell: what would be the benefit to using that over `AccountId::new_unchecked`? How would the extra bit be relevant? Seems clunky to have both so we should think through what's best long term.

near-sdk/src/environment/env.rs
```diff
@@ -149,13 +150,13 @@ pub fn register_len(register_id: u64) -> Option<u64> {
 // ###############
 /// The id of the account that owns the current contract.
 pub fn current_account_id() -> AccountId {
-    String::from_utf8(method_into_register!(current_account_id)).unwrap()
+    AccountId::try_from(method_into_register!(current_account_id)).unwrap()
 }
```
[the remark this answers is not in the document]
austinabell: hmm, the unreachable definitely feels uncertain to me, since this account id validation is custom to the SDK and if there ever was a discrepancy, this would be a tough error to debug. I will change to this now though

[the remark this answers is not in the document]
austinabell: Also, why do we want to remove `TryFrom<Vec<u8>>`? I would have thought the opposite with the trait being part of the core prelude with the next Rust edition

near-sdk/src/types/mod.rs
austinabell: Yeah, I also usually prefer, forgot to cycle back

austinabell:
> * deriving BorshDeserialize for AccountId is wrong

Think I've addressed everything else, but curious why you think this is wrong? Any contract using the old `AccountId` will have the ID borsh serialized as a string, and it would be a state-breaking change when they update the SDK. What would you suggest is a better format? I don't see how string would be a bad borsh format either, or am I missing something about borsh storing more metadata for strings over byte slices?

examples/fungible-token/tests/sim/utils.rs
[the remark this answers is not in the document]
matklad: Sorry for channeling  unfiltered stream of consciousness :)

From looking at the usages here, it seems that "hard-coded" account names are a somewhat frequent use-case, at least for tests. So, it would be more user-friendly if the API allowed to directly express constant accounts, such that runtime validation and unwrapping is not required. 

This is doable, we can have a String-like type which is both const-constructable and supports dynamic allocation, the simplest version would be just `Cow<'static, str>`. The cow would add some extra space overhead though. 

It is possible to implement a string type which has the same size_of as `std::String`, but which is const-constructable and avoids allocation for short strings. https://github.com/rust-analyzer/smol_str/ is a half-baked attempt at that, it can be improved quite a bit.

near-sdk/src/environment/env.rs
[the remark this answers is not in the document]
matklad: >hmm, the unreachable definitely feels uncertain to me, since this account id validation is custom to the SDK and if there ever was a discrepancy, this would be a tough error to debug.

So, the `.unreachable` is more or less equivalent to `.unwrap()`, the difference is that it doesn't invoke formatting machinery. Unwrap is defined as

image

and `E: Debug` is what's problematic.


>Also, why do we want to remove TryFrom<Vec<u8>>?

The problem with that is that it causes us to have `InvalidUtf8` as separate error condition in the ErrorKind enum. Today, that's just a field-less variant, but that's suboptimal, as there's a perfectly fine type for that in the standard library: https://doc.rust-lang.org/stable/std/string/struct.FromUtf8Error.html. Not that it allows recovering the original bytes, and also allows to know what exactly failed during utf8 validation (which byte is faulty). If we don't expose this functionality, we provide half-baked API. If we do expose this functionality, we increase API surface and design space a bit. 

A simpler solution is to let the caller handle "invalid ut8" and "invalid account id" errors separately: `String::from_utf8(bytes).and_then(AccountId::try_from)`. This gives the maximal flexibility to the user to handle utf8 errors however they like, and allows us to not think about this error case at all. One drawback here is that this is potentially less optimal -- in theory, you can validate both utf8 validity and account id validity in a single pass over data, but that's a very small benefit for high implementation complexity (and potentially some code bloat), so I don't think we'll ever do that. 


Separately, it seems that the main beneficiary of this impl is actually the SDK itself? It's SDK who gets raw bytes from the extern "C" functions. The contracts themselves seem to work with utf8 directly.

matklad:
>why you think this is wrong?

It allows creating `AccoundId` which is not valid

```rust
    #[test]
    fn deser_checks_validity() {
        let r1 = "💩".parse::<AccountId>();
        assert!(r1.is_err()); // correct, 💩 is not a valid account name

        let bytes = "💩".try_to_vec().unwrap();
        let r2 = AccountId::try_from_slice(&bytes);
        let acc = r2.unwrap();
        assert_eq!(acc.as_str(), "💩") // BUG, 💩 is treated as a valid account
    }
```


As a rule of thumb, type invariants should be checked in every constructor, and deserialization provides an extra constructor.

austinabell:
> As a rule of thumb, type's invariants should be checked in every constructor, and deserialization provides an extra constructor.
> 
> We shouldn't change serialization format, but we should add an extra validation to deserialization.

Ah, of course, oops

near-sdk/src/types/account_id.rs
austinabell: I don't see which would be preferred to be removed. I can see the `TryFrom<&str>` but I see a benefit to keeping both of these you linked to in the comment because `FromStr` allows for the easiest conversion and `TryFrom<String>` allows keeping the allocated `String`. Is your point that we should force the re-allocation just because the internal type might be switched from String in the future?

near-sdk/src/types/account_id.rs
austinabell: That's what I had thought, but apparently, it had caused bloat in the past, unless I misunderstood some reasoning.

examples/fungible-token/tests/sim/utils.rs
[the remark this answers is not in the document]
austinabell: Seems like a bit of a micro-optimization that might be out of the scope of this PR. I like the idea but should probably have better infrastructure for testing contract size and performance before trying to eagerly optimize.

As for skipping runtime validation, that's what `AccoundId::new_unchecked` currently services. Is there a better API for this? I see some need to have a runtime string not be validated, so having two methods would be confusing and would limit us from switching the internal type

near-sdk/src/types/account_id.rs
matklad: I mean the two linked conversions should stay, the *others* (tryFrom &str, &[u8], Vec<u8>) could go.

near-sdk/src/types/account_id.rs
matklad: >but apparently, it had caused bloat in the past, unless I misunderstood some reasoning.

I think this is a miscommunication / misunderstanding (but maybe misunderstanding on my side :) ). I believe that neither `String` nor `str` causes bloat in comparison to `Vec<u8>` or `[u8]`. What causes bloat is string *formatting* (eg, unwrapping results) and accidentally using unicodo-aware APIs (#458).  Switching to `Vec<u8>` won't help with the first (`Vec<u8>` implements Debug). It might margianaly help with the second, but the cost would be that you'd have to re-implement ascii `lo_lowercase` on bytes directly yourself.

near-sdk/src/types/account_id.rs
austinabell: hmm, I suppose that's the case, I just see the benefit of coming from bytes because it's the internal representation when accessing from the runtime. Yeah we could remove it, but if someone wanted to do low-level things, it's just a much harder API to use

I'm fine with removing `&str` but I'd like to keep the bytes ones if possible to make it easier to work directly with registers

matklad: > I just see the benefit of coming from bytes because it's the internal representation when accessing from the runtime

Note that this is just a limitation of wasm/near API: there are `str` on the other side, we just can't express this. 

But yeah, you point that if someone builds directly against `sys::current_account_id` and friends, they'll use the `TryFrom<Vec<u8>>` API is a good one. 

I guess, overall my personal preference would still be to let the caller to utf8 validation in this case: if you are using low-level things anyway, you probably are not doing this for ergonomics, but for control. And you might want to do things like `Account::new_unchecked(String::from_utf8_unchecked)`.

But not strong opinion here. Having to carry `Utf8Error` is unfortunate, but it's not a big deal

austinabell: I see the point, let me try to trim this down more, I suppose the breakages will be worth this in the long term

examples/fungible-token/tests/sim/utils.rs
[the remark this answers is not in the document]
matklad: Yeah, this is def out of scope, something for the time when all the base APIs are done and future-proofed.

near-sdk/src/types/account_id.rs
```diff
@@ -0,0 +1,176 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ParseAccountIdError {
+    kind: ParseAccountIdErrorKind,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum ParseAccountIdErrorKind {
+    InvalidAccountId,
+}
```
matklad, line 130: Might remove the `Kind`, as we have only one anyway.

near-sdk/src/utils/mod.rs
```diff
@@ -86,8 +91,13 @@ impl PendingContractTx {
...
 
-    pub fn new_from_bytes(receiver_id: &str, method: &str, args: Vec<u8>, is_view: bool) -> Self {
-        Self { receiver_id: receiver_id.to_string(), method: method.to_string(), args, is_view }
+    pub fn new_from_bytes(
+        receiver_id: AccountId,
+        method: &str,
+        args: Vec<u8>,
+        is_view: bool,
+    ) -> Self {
+        Self { receiver_id: receiver_id, method: method.to_string(), args, is_view }
```
matklad, line 100: nit: :bulb: use field shorthand syntax

matklad (approved):

near-sdk/src/utils/mod.rs
austinabell: ah, yeah definitely going to fix clippy and add to CI before next release

examples/cross-contract-low-level/tests/general.rs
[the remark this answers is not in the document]
mikedotexe: Agreed, and as we discussed in this morning's meeting, this is capture in #465

Issue #440, opened 2021-06-15
Default blockchain interface in non-wasm environments and update interface


austinabell:
Somewhat opinionated so open to discussion, but it makes more sense to me to have the `BlockchainInterface` set with some reasonable defaults so it doesn't require a user which doesn't override anything to build and set up the context and initialize the interface through `testing_env!`. If this is defaulted, the `BLOCKCHAIN_INTERFACE` could not be wrapped in an `Option` and thus remove a lot of confusing and unnecessary API usages.

Also, the current API only allows for replacing the interface completely, when it would be better to be able to just modify specific parts of the context on the initialized one. To keep changes somewhat minimal, can just expose a function that returns a mutable reference to the context initialized or just passing a function to apply a change on the context mutably.

One other thing that can possibly be done after this is changing the `BLOCKCHAIN_INTERFACE` from a `Box<dyn BlockchainInterface>` to just `MockedBlockchain` because after #417 there should be no need to have multiple interface types. This would simplify a lot of things, but is a breaking change.

Probably easiest if this is done after #417 comes in, as there will be conflicting changes and it would be simpler to do after.

ilblackdragon:
May be instead just removing all this complexity via #373?

austinabell:
> May be instead just removing all this complexity via #373?

Different problem, this issue is just relating to the testing framework and setting the `BLOCKCHAIN_INTERFACE` or whatever it is named to in the future is needed to setup the testing environment.

This issue is just relating to the unnecessary boilerplate needed for tests right now

austinabell:
> [gov.near.org/t/revisiting-contract-testing-framework/2920](https://gov.near.org/t/revisiting-contract-testing-framework/2920) is related. If we are to disrupt our users with a new shiny testing framework, it probably makes sense to not also change existing ones.

Can you explain what specifically you are suggesting? What this issue suggests is not a breaking change for existing users, it just removes the *need* to have the boilerplate to be able to write unit tests.

I don't see this testing framework replacing unit tests, as it requires additional processes, so I'm wondering if you're saying to drop this issue or just to make sure there are no breaking changes?

matklad:
Nothing specific, just the general observation that it looks like sandbox will eat a large part of "churn budget", so we need to be more conservative with test-related changes elsewhere. Even if sandbox supplements, rather than supplants, unit-testing, it's still will be perceived as "change to the way you write tests". 

Non-breaking changes are probably fine though!

austinabell:
[...]

PR #450, opened 2021-06-25
feat: Initialize blockchain interface by default

austinabell:
Made changes based on #417 because it removes `BLOCKCHAIN_INTERFACE` from non-wasm32 environments, and this default only makes sense outside anyway.

One part of #440, going to split up changes because other ideas I have are more opinionated/breaking.

This change is completely backwards compatible from #417 and I changed only some of the examples to show this.

Ideally, in the future, it would be possible to modify the context of the mocked blockchain in place instead of still rigidly constructing a new context and replacing the current one, which will come with more QOL improvements.

### DIFF
--- a/examples/mission-control/src/mission_control.rs
+++ b/examples/mission-control/src/mission_control.rs
@@ -87,35 +87,11 @@ fn rates_default() -> HashMap<Exchange, Rate> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use near_sdk::MockedBlockchain;
-    use near_sdk::{testing_env, VMContext};
-
-    fn get_context(input: Vec<u8>, is_view: bool) -> VMContext {
-        VMContext {
-            current_account_id: "alice_near".to_string(),
-            signer_account_id: "bob_near".to_string(),
-            signer_account_pk: vec![0, 1, 2],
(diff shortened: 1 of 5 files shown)


examples/mission-control/src/mission_control.rs
```diff
@@ -87,35 +87,11 @@ fn rates_default() -> HashMap<Exchange, Rate> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use near_sdk::MockedBlockchain;
-    use near_sdk::{testing_env, VMContext};
-
-    fn get_context(input: Vec<u8>, is_view: bool) -> VMContext {
```
ChaoticTempest, line 93: So by having a default mocked blockchain, we don't have to go thru the shenanigans of setting up the context here and also explicitly using `testing_env!` right? This is pretty nice for setting up new tests 👍.

austinabell: yeah, you still need it if you need to configure anything, but if not this saves some boilerplate.

Ideally, I think it might be nice to just be able to modify the context in place rather than replacing it, but there are a lot of connected pieces rn.

ChaoticTempest (approved):
👍

mikedotexe (approved):

PR #451, opened 2021-06-25
refactor!: make blockchain interface concrete type and remove interface trait

austinabell:
Definitely should not come in before the next release as there are some more extensive breaking changes, but I'm just hacking around to minimize the overhead and try to clean up the APIs

Related: #440


austinabell:
Opening for review, but we may want to wait to merge until after the next major release. The breaking change is just that `BLOCKCHAIN_INTERFACE` is no longer exposed (no longer needed) and `env::take_blockchain_interface` is no longer available because the interface is not an `Option`.

@matklad or @mikedotexe maybe you have a gut check on whether this may be adding too many breaking changes into the next release, depending on how this currently gets used.

matklad:
Can't estimate the amount of de-facto breakages. Although, if we do change this, I wonder if it also makes sense to hide nearcore types at the same change? Like, `MockedBlockchain::new` is public, and it depends on all kinds of stuff we don't actually want to be the part of our public API.

austinabell:
> Can't estimate the amount of de-facto breakages. Although, if we do change this, I wonder if it also makes sense to hide nearcore types at the same change? Like, `MockedBlockchain::new` is public, and it depends on all kinds of stuff we don't actually want to be the part of our public API.

I mean, right now those dependencies are already hidden for `wasm32` architectures, and I'm not sure if they can be hidden given the API currently. Removing would definitely increase the breakages as they are used quite commonly in tests. The reason I think hiding `BLOCKCHAIN_INTERFACE` is okay is few usages from what I've seen. 

I just also like the idea of not having to maintain that interface so that changes can be done without requiring at least a minor version bump.

matklad (approved):

ChaoticTempest (approved):

CHANGELOG.md
```diff
@@ -24,6 +24,10 @@
   * This type will have `ValidAccountId`'s JSON (de)serialization and the borsh serialization will be equivalent to what it was previously
 * Initializes default for `BLOCKCHAIN_INTERFACE` to avoid requiring to initialize testing environment for tests that don't require custom blockchain interface configuration
   * This default only affects outside of `wasm32` environments and is optional/backwards compatible
+* Makes `BLOCKCHAIN_INTERFACE` a concrete type and no longer exports it.
+  * If for testing you need this mocked blockchain, `near_sdk::mock::with_mocked_blockchain` can be used
```
mikedotexe, line 28: These changelog notes are very helpful, thanks

mikedotexe (approved):

Issue #447, opened 2021-06-23
Make `Base58PublicKey` an enum or struct which does not erase curve type and expose underlying bytes


austinabell:
Currently, the curve type is deserialized and the curve byte is put as the first byte and the underlying `Vec<u8>` is exposed. There is no need to store the serialized format internally, instead of just keeping the curve and converting it to a vec for borsh serialization only.

mikedotexe:
Recommended to go to @ChaoticTempest when there's bandwidth soon

PR #453, opened 2021-06-29
feat: Make CurveType accessible within Base58PublicKey

ChaoticTempest:
Addresses #447. This makes it possible to access the curve type without directly accessing the `Vec<u8>` itself


near-sdk/src/json_types/public_key.rs
```diff
@@ -67,17 +80,19 @@ impl Base58PublicKey {
 
 impl From<Base58PublicKey> for Vec<u8> {
     fn from(v: Base58PublicKey) -> Vec<u8> {
-        v.0
+        let mut out = vec![v.curve as u8];
```
austinabell, line 83: I don't think casting the enum to u8 is safe for all platforms unless the enum is `repr(u8)`, or am I mistaken?

Probably best to have some conversion to u8

near-sdk/src/json_types/public_key.rs
```diff
@@ -67,17 +80,19 @@ impl Base58PublicKey {
...
+        out.extend(v.data);
+        out
```
austinabell, line 85:
```suggestion
        v.data.insert(0, v.curve as u8);
        v.data
```

Aside from the other comment, can re-use the allocation since this consumes the public key. (also have to make v `mut`)

near-sdk/src/json_types/public_key.rs
```diff
@@ -67,17 +80,19 @@ impl Base58PublicKey {
...
 impl TryFrom<Vec<u8>> for Base58PublicKey {
     type Error = Box<dyn std::error::Error>;
 
     fn try_from(v: Vec<u8>) -> Result<Self, Self::Error> {
-        match v.len() {
-            33 if v[0] == 0 => Ok(Self(v)),
-            65 if v[0] == 1 => Ok(Self(v)),
+        match &*v {
+            [0, other @ ..] => Ok(Self { curve: CurveType::ED25519, data: other.into() }),
+            [1, other @ ..] => Ok(Self { curve: CurveType::SECP256K1, data: other.into() }),
```
austinabell, line 95: prob best to remove the first element since you can just use that allocation, as the `into()` here is sneakily re-allocating.

near-sdk/src/json_types/public_key.rs
```diff
@@ -104,10 +119,13 @@ impl<'de> serde::Deserialize<'de> for Base58PublicKey {
...
-            0 => "ed25519:".to_string() + &bs58::encode(&str_public_key.0[1..]).into_string(),
-            1 => "secp256k1:".to_string() + &bs58::encode(&str_public_key.0[1..]).into_string(),
-            _ => panic!("Unexpected curve"),
+        match str_public_key.curve {
+            CurveType::ED25519 => {
+                "ed25519:".to_string() + &bs58::encode(&str_public_key.data).into_string()
+            }
+            CurveType::SECP256K1 => {
+                "secp256k1:".to_string() + &bs58::encode(&str_public_key.data).into_string()
+            }
```
austinabell, line 128:
```suggestion
        match str_public_key.curve {
            CurveType::ED25519 => {
                ["ed25519:", &bs58::encode(&str_public_key.data).into_string()].concat()
            }
            CurveType::SECP256K1 => {
                ["secp256k1:", &bs58::encode(&str_public_key.data).into_string()].concat()
            }
        }
```

This one is a bit optional because I'm actually not sure about the internals if it can avoid the reallocation on adding the other string (since you're not specifying a capacity) and I think concat does that, but I could be wrong.

near-sdk/src/json_types/public_key.rs
```diff
@@ -191,7 +203,7 @@ mod tests {
     use std::convert::TryInto;
 
     fn binary_key() -> Vec<u8> {
-        let mut binary_key = vec![0];
+        let mut binary_key = vec![CurveType::ED25519 as u8];
         binary_key.extend(
             bs58::decode("6E8sCci9badyRkXb3JoRpBj5p8C6Tw41ELDZoiihKEtp").into_vec().unwrap(),
```
austinabell, line 208: so since these tests are essentially checking themselves, this probably could just return the actual `Base58PublicKey` to make the tests cleaner. More of an opinion though, but with this change, the tests read a bit weird.

near-sdk/src/json_types/public_key.rs
```diff
@@ -5,7 +5,17 @@ use std::convert::TryFrom;
...
+    Clone,
+    Copy,
+    Serialize,
+    Deserialize,
+    PartialOrd,
+    Ord,
+    Eq,
+    PartialEq,
+    BorshDeserialize,
+    BorshSerialize,
```
austinabell, line 18: I'm actually not sure why there are serialize impls on this, it probably shouldn't as it's inconsistent and would be a foot gun. For the `Base58PublicKey` it should have a custom borsh implementation so these shouldn't be used at all

near-sdk/src/json_types/public_key.rs
```diff
@@ -51,7 +61,10 @@ impl std::str::FromStr for CurveType {
 ///             .unwrap();
 /// ```
 #[derive(Debug, Clone, PartialEq, PartialOrd, Ord, Eq, BorshDeserialize, BorshSerialize)]
-pub struct Base58PublicKey(pub Vec<u8>);
+pub struct Base58PublicKey {
+    curve: CurveType,
+    data: Vec<u8>,
+}
```
austinabell, line 67: extension from previous, this is actually a breaking change of borsh format, so you need to implement borsh (de)serialization methods manually

austinabell (changes requested):

austinabell:
I think also we should decide if the `PublicKey` type should be replaced with this, to avoid having multiple of the same type, similar to #448 changes. This can always be done in a separate issue I suppose. Also since we are making breaking changes would be nice to change error type to a concrete one

near-sdk/src/json_types/public_key.rs
ChaoticTempest: hmm, the compiler doesn't complain about it though. And I can't seem to find anything on it being unsafe to cast. https://doc.rust-lang.org/nomicon/casts.html says `field-less enums as integer` are considered to be true casts.

near-sdk/src/json_types/public_key.rs
ChaoticTempest: nice -- both will copy data, but your suggestion will avoid the extra allocation

near-sdk/src/json_types/public_key.rs
austinabell: ah, that seems to be correct. Also seems the compiler is already having the repr being one byte so the explicit `repr(u8)` probably doesn't matter, although I'm not sure if it's best practice when casts like this are used.

near-sdk/src/json_types/public_key.rs
ChaoticTempest: Maybe they were there before borsh was a thing? In any case, I'll try and remove them

near-sdk/src/json_types/public_key.rs
austinabell: but yeah make sure somehow that the length of the data is `>0` as my suggestion can panic if it's empty

near-sdk/src/json_types/public_key.rs
ChaoticTempest: true, I always favor explicit over implicit. Marking it as `repr(u8)` shouldn't be an issue

near-sdk/src/json_types/public_key.rs
austinabell: I mean, the type is exported publicly now, so probably best to either make it only visible within the crate or have a function to return a reference to the curve type on the public key type. Either way I don't see the need for serialize impls (either) to be implemented for it. There _shouldn't_ be a need to serialize just the curve type.

ChaoticTempest: mmm, I'd imagine the developer potentially wanting to access the curve type, so I think having the function to a reference of it would be best. Well, this is pretty much going to be a breaking change already, so getting rid Serialize impls for both `CurveType` and `Base58PublicKey` should be fine, right?

austinabell: no keep for `Base58PublicKey`, that is the whole point of this type is serialization. Get rid of it for `CurveType` though imo.

near-sdk/src/json_types/public_key.rs
ChaoticTempest: welp, I went down the rabbithole for this one. `str.to_string()` allocates a `string` with a capacity, so the `+` operator will indeed do a second allocation. The concat version is better here

ChaoticTempest:
> I think also we should decide if the `PublicKey` type should be replaced with this, to avoid having multiple of the same type, similar to #448 changes. This can always be done in a separate issue I suppose.

Yea let's do that in a separate issue/PR. Don't wanna snowball this one

near-sdk/src/json_types/public_key.rs
```diff
@@ -217,13 +253,21 @@ mod tests {
...
         assert_eq!(actual, "ed25519:6E8sCci9badyRkXb3JoRpBj5p8C6Tw41ELDZoiihKEtp");
     }
+
+    #[test]
+    fn test_public_key_borsh() {
+        let key: Base58PublicKey = expected_key();
+        let encoded_key = key.clone().try_to_vec().unwrap();
+        let decoded_key = Base58PublicKey::try_from_slice(&encoded_key).unwrap();
+        assert_eq!(key, decoded_key);
+    }
```
austinabell, line 272: note: this test would not catch the format breaking change from this PR, would be nice to have serialized data (can just be in base58 format since those utils will be available in this module). Checking if it serializes symmetrically doesn't check much

near-sdk/src/json_types/public_key.rs
```diff
@@ -132,24 +170,22 @@ impl std::str::FromStr for Base58PublicKey {
...
+        Ok(Self { curve: CurveType::deserialize(buf)?, data: BorshDeserialize::deserialize(buf)? })
+    }
+}
+
+impl BorshSerialize for Base58PublicKey {
+    fn serialize<W: io::Write>(&self, writer: &mut W) -> io::Result<()> {
+        self.curve.serialize(writer)?;
+        BorshSerialize::serialize(&self.data, writer)
     }
 }
```
austinabell, line 190: unless I'm mistaken about borsh internals, this is just doing what the derive is doing and will be in the wrong format still. Can double-check in a bit, but I don't believe this will be correct.

near-sdk/src/json_types/public_key.rs
```diff
@@ -1,19 +1,29 @@
-use borsh::{BorshDeserialize, BorshSerialize};
+use borsh::{maybestd::io, BorshDeserialize, BorshSerialize};
```
austinabell, line 1: prob not best to use `maybestd` from borsh, as how borsh handles nostd might be very different than us in the future and ties us to that dependency

near-sdk/src/json_types/public_key.rs
ChaoticTempest: What would you recommend to change it to? I'm not sure what the right format would be

near-sdk/src/json_types/public_key.rs
ChaoticTempest: I see, that's a good point. Maybe unwrapping the results and re-wrapping it in our own Result type would do the trick?

near-sdk/src/json_types/public_key.rs
ChaoticTempest: true, this was to just check if the manually implemented borsh (de)serialization worked as intended. But definitely having the serialized data to check for newer formats would be nice

austinabell: well it can work but not be in the same format as previously, which is a very breaking change and would break storage interactions when people migrate, which we can't have happen

near-sdk/src/json_types/public_key.rs
austinabell: the format previously would just be serializing as a single `Vec<u8>` of bytes. Try creating a serialized vector from what exists now with borsh and match with what's changed now.

I'm 90% sure borsh derive just serializes as the inner newtype, which previously was `Vec<u8>`, but best to double check.

ChaoticTempest: Oh I see what you mean. I'm not too familiar with how borsh/serde does (de)serialization. I'm guessing there's not an equivalent to versioning like how protobuf does it, since we gotta manually serialize it the same way as a previous version

austinabell: Yeah, we wouldn't want it versioned because then more data than necessary would need to be stored

near-sdk/src/json_types/public_key.rs
ChaoticTempest: The newest version should have this where I created the original struct and check for the serialization between it and the current version to see if they're the same:
```Rust
    fn test_public_key_borsh_format_change() {
          // Original struct to reference Borsh serialization from
          struct Base58PublicKeyRef(Vec<u8>);
```
What do you think?

mikedotexe (approved):
Seems like there's a small amount of conversation to wrap up and bump from comments, but looking good to me

near-sdk/src/json_types/public_key.rs
```diff
@@ -1,19 +1,29 @@
...
+
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
+        match value {
+            0 => Ok(Self::ED25519),
+            1 => Ok(Self::SECP256K1),
+            _ => Err(ParsePublicKeyError { kind: ParsePublicKeyErrorKind::UnknownCurve }),
+        }
+    }
+}
+
```
matklad, line 24: This is a bit of an odd *public* API, getting a structured error here which also describes all other possible parse errors feels strange. Let's not add this impl, and instead add a private inherent `from_u8()` method (or just inline at the call-site).

near-sdk/src/json_types/public_key.rs
```diff
@@ -63,23 +76,45 @@ impl Base58PublicKey {
...
+            return Err(ParsePublicKeyError {
+                kind: ParsePublicKeyErrorKind::InvalidLength(data.len()),
+            });
+        }
+
+        Ok(Self { curve, data })
+    }
+
+    /// Get info about the CurveType for this public key
+    pub fn curve(&self) -> CurveType {
```
matklad, line 95:
```suggestion
    pub fn curve_type(&self) -> CurveType {
```

near-sdk/src/json_types/public_key.rs
```diff
@@ -97,31 +132,34 @@ impl<'de> serde::Deserialize<'de> for Base58PublicKey {
...
-            1 => "secp256k1:".to_string() + &bs58::encode(&str_public_key.0[1..]).into_string(),
-            _ => panic!("Unexpected curve"),
+        match str_public_key.curve {
+            CurveType::ED25519 => {
+                ["ed25519:", &bs58::encode(&str_public_key.data).into_string()].concat()
+            }
+            CurveType::SECP256K1 => {
+                ["secp256k1:", &bs58::encode(&str_public_key.data).into_string()].concat()
+            }
         }
```
matklad, line 149: That's much better that what was here before, but it still creates a bunch of intermediate strings. As a rule of thumb, all good APIs in Rust are organized in a way that allows for avoiding intermediate allocations, so usually there's a way to do only a single alloc. For this case, this would be something like

```rust
let mut buf = String::with_capacity(cap);
buf.push_str("ed25519");
bs58::encode(&str_public_key.data).into(&mut buf);
```

matklad (commented):
Yeah, I think @austinabell is correct in that we don't want to have to different `PublicKey` types -- there should be only one key we use throughout. 

So, something like

```rust
#[derive(Debug, Clone, Copy PartialEq)]
pub enum CurveType { ED25519, SECP256K1 }

pub struct PublicKey { data: Vec<u8> }

impl PublicKey {
        match self.data[0] {
            0 => CurveType::ED25519,
            1 => CurveType::SECP256K1,
            _ => unreachable!(),
        }
    }

    pub fn key_bytes(&self) -> &[u8] { &self.data[1..] }

    /// Don't necessary want to expose the fact that at the ABI level we prepend
    /// curve type to the key bytes. Might consider making this public later.
    pub(crate) fn as_bytes(&self) -> &[u8] { self.data.as_slice() }
}

impl FromStr for PublicKey { ... }
impl BorshSerialize for PublicKey { ... }
impl BorshDeserialize for PublicKey { ... }
impl serde::Serialize for PublicKey { ... }
impl serde::Deserialize for PublicKey { ... }
```

Note that we probably do want to use a single `Vec` as an internal repr, as that's what required by `promise_batch_action_add_key_with_full_access` and friends. We just don't want to expose this fact for the user. 


This PR moves in the generally right direction, but don't know what would be faster -- merge this and then send a follow up, or to make this existing thing mode ambitious.

matklad:
Found while reviewing: https://github.com/near/near-sdk-rs/issues/458

near-sdk/src/json_types/public_key.rs
austinabell: Ideally, the vector would just be the serialized bytes, which can be generated from `master` currently. This would test the correct format, it's just less strict.

near-sdk/src/json_types/public_key.rs
ChaoticTempest: The tricky part about this is that we need to know the capacity ahead of time. Digging into the internals of `bs58` crate it does resize and truncate the buffer if it's not exactly that capacity, but it does use some upper bound we can make use of to avoid reallocating. We just gotta be careful if that upper bound ever gets changed from under us whenever we go to upgrade `bs58`.

Also, trying to implement this right now but the API is kinda unwieldy since `.into(&mut buf)` overwrites the initial `push_str("ed25519")`. Grabbing a slice to allocate into without a known capacity ahead of time makes it look pretty ugly. @matklad Would you happen to know of a better way than this:
```Rust
let mut buf = vec![0u8; 8 + upper_bound];
buf.splice(..8, "ed25519:".as_bytes().iter().cloned());
let size = bs58::encode(&str_public_key.data).into(&mut buf[8..]).unwrap();
buf.truncate(8 + size);
String::from_utf8(buf).unwrap()
```

near-sdk/src/json_types/public_key.rs
```diff
@@ -97,31 +128,38 @@ impl<'de> serde::Deserialize<'de> for Base58PublicKey {
...
 }
 
 impl From<&Base58PublicKey> for String {
     fn from(str_public_key: &Base58PublicKey) -> Self {
-        match str_public_key.0[0] {
-            0 => "ed25519:".to_string() + &bs58::encode(&str_public_key.0[1..]).into_string(),
-            1 => "secp256k1:".to_string() + &bs58::encode(&str_public_key.0[1..]).into_string(),
-            _ => panic!("Unexpected curve"),
-        }
+        let upper_bound = (str_public_key.data.len() / 5 + 1) * 8;
```
austinabell, line 138: I'll look more into this tomorrow, but where are the values for this coming from? This seems like it might be above the amount needed for base58

near-sdk/src/json_types/public_key.rs
```diff
@@ -97,31 +128,38 @@ impl<'de> serde::Deserialize<'de> for Base58PublicKey {
...
+            CurveType::ED25519 => (8, "ed25519:"),
+            CurveType::SECP256K1 => (10, "secp256k1:"),
+        };
+
+        // Efficient concat of "curve_type:" + bs58 encoded bytes:
+        let mut buf = vec![0u8; pad + upper_bound];
+        buf.splice(..pad, curve_as_str.as_bytes().iter().cloned());
+        let size = bs58::encode(&str_public_key.data).into(&mut buf[pad..]).unwrap();
+        buf.truncate(pad + size);
+        String::from_utf8(buf).unwrap()
```
austinabell, line 149: why not just encode the bs58 to a `&mut str` buffer? Can do `buf.get_mut(pad..)` if it's a string buffer and can do the same as above. The unwrap might bring in code not needed

near-sdk/src/json_types/public_key.rs
ChaoticTempest: This should be in `bs58/src/encode.rs::318`. You can additionally try to get to it via going to the definition of `into` within `bs58::encode(&str_public_key.data).into(&mut buf[pad..]).unwrap();`

near-sdk/src/json_types/public_key.rs
ChaoticTempest: The thing is, `get_mut` returns an optional which is none when it's out of bounds. I don't know if there's an easy way to create a buffer string similar to `vec![0u8; size]` where we fill the whole capacity at the start.

near-sdk/src/json_types/public_key.rs
matklad: Oh wow, looks like I was confused by b58 API and confused you as well @ChaoticTempest , sorry for that!

>As a rule of thumb, all good APIs in Rust are organized in a way that allows for avoiding intermediate allocations

Seems like base58 API isn't setup to do that in a nice way, see https://github.com/mycorrhiza/bs58-rs/issues/77.

Calculating the lengths manually and truncating *can* work, but the complexity is not worth it (and we have to pay extra for duplicate utf8 validation), I suggest sticking with the original impl then!

ChaoticTempest: Ahh, no worries. Best to keep it simple then

ChaoticTempest:
@austinabell I added the changelog in #464. I think it be best to merge it into here first, then merge this completely

near-sdk/src/json_types/mod.rs
```diff
@@ -2,10 +2,26 @@
...
+
+// This deprecated attribute doesn't work for the current rust version (1.52)
+// but will likely work in the future. Also included just to note that it is
+// indeed deprecated and not just a random export.
+#[deprecated(
+    since = "4.0.0",
+    note = "This import is deprecated. Best to import directly from near_sdk"
+)]
+pub use crate::types::CurveType;
+use crate::types::PublicKey;
```
austinabell, line 19: you might want to move this above the other to not have it grouped or just reference it directly below. Opinionated though as it was just confusing for me to read

near-sdk/src/types/public_key.rs
```diff
@@ -0,0 +1,282 @@
...
+    }
+}
+
+impl TryFrom<String> for CurveType {
+    type Error = ParsePublicKeyError;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        Ok(value.parse::<Self>()?)
+    }
+}
```
austinabell, line 29: I don't think we need this?

near-sdk/src/types/public_key.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    fn try_from(mut data: Vec<u8>) -> Result<Self, Self::Error> {
+        if data.len() == 0 {
+            return Err(ParsePublicKeyError {
+                kind: ParsePublicKeyErrorKind::InvalidLength(data.len()),
+            });
+        }
+
+        let curve = CurveType::from_u8(data.remove(0))?;
+        Self::from_parts(curve, data)
```
austinabell, line 129: This is removing the first byte to create the curve type, then adding it back (essentially copying over the data twice), can we not just validate the curve type and initialize it here now that we are just storing the vector?

near-sdk/src/types/public_key.rs
```diff
@@ -0,0 +1,282 @@
...
+    }
+}
+
+impl TryFrom<&str> for PublicKey {
+    type Error = ParsePublicKeyError;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        Ok(value.parse::<Self>()?)
+    }
+}
```
austinabell, line 179: We probably don't need either of these, as I'm sure @matklad would point out, but I see some benefit in keeping so I don't have a strong opinion

near-sdk/src/json_types/mod.rs
ChaoticTempest: ahh for sure, I'll move it out of the way

near-sdk/src/types/public_key.rs
ChaoticTempest: Good catch. I thought I forgot about something when I was transitioning back to an internal `Vec<u8>`

near-sdk/src/types/public_key.rs
ChaoticTempest: Converting from a string into a `PublicKey` is still used in `environment/env.rs` and testing, so I think we should still keep these

near-sdk/src/types/public_key.rs
ChaoticTempest: Yeah, this shouldn't be needed. Also shouldn't break anything since CurveType was never exposed before

near-sdk/src/types/public_key.rs
austinabell: These are all using `FromStr` at its base, though, so they probably should just use `FromStr` directly imo

austinabell (approved):

near-sdk/src/types/public_key.rs
ChaoticTempest: I see, I'll switch to that then

PR #454, opened 2021-06-29
Fix typo in dir, add dependencies to Dockerfile, cleanup

telezhnaya:
It is required for https://github.com/near/core-contracts/pull/151
This configuration of container works fine for https://github.com/near/core-contracts/lockup

------

### General question

Are we sure that it should be here and not in https://github.com/near/core-contracts ?
I think it's better to delete the whole directory `contact-builder` here and to put it into `core-contracts` repository.


austinabell:
> Are we sure that it should be here and not in [near/core-contracts](https://github.com/near/core-contracts) ?

I don't have all context around the contract builder, but I assume that it wouldn't make sense to only include for core-contracts because it's a tool that can be generally used for any contracts. I think it's reasonable to have a custom copy for now within core-contracts, but I think it still has a use for here currently.

Ideally in the future, this should happen through some binary or cargo extension imo so this doesn't have to be explicitly known and used manually.

austinabell (approved):

chefsale (approved):
LgTm

ailisp (approved):
lgtm!

ailisp:
>  I think it's better to delete the whole directory contact-builder here and to put it into core-contracts repository.

Don't have a strong opinion, put it here is to give a "standard setup" for near-sdk-rs users to build their contract, if there local native rust doesn't work. They can test build with this docker before submitting an issue I guess

evgenykuzyakov:
Does it affect the resulting binaries? E.g. if you rebuild `examples/fungible-token`, doesn't it produce different wasm?

telezhnaya:
> Does it affect the resulting binaries? E.g. if you rebuild `examples/fungible-token`, doesn't it produce different wasm?

@evgenykuzyakov thank you for the question. The binaries are the same if you build `examples/fungible-token` with old Dockerfile or new Dockerfile, but you need to know that it is modified anyway. It's different with what we have now.

I built it in both containers and it gives the same result, but different with the commited one.

telezhnaya:
Created the issue about that: https://github.com/near/near-sdk-rs/issues/456

evgenykuzyakov (approved):

PR #459, opened 2021-07-05
chore(examples): remove deprecated setup_alloc from examples

austinabell:
Should have removed with #429, but the examples now remove this to avoid confusion or deprecation warnings in the future if copied.

Also removes documentation around the macro as it's irrelevant and overriding a custom allocator can just be a section of the mdbook (issue https://github.com/near/sdk-docs/issues/29)


ChaoticTempest (approved):
nice!

Cargo.toml
```diff
@@ -7,15 +7,7 @@ members = [
...
-    "examples/cross-contract-high-level",
-    "examples/fungible-token",
-    "examples/non-fungible-token",
-    "examples/lockable-fungible-token",
-    "examples/cross-contract-low-level",
-    "examples/mission-control",
-    "examples/status-message",
-    "examples/status-message-collections",
-    "examples/test-contract",
+    "examples/",
```
mikedotexe, line 10: 🤯

mikedotexe (approved):

PR #461, opened 2021-07-07
fix(examples): rebuild all after updating Dockerfile

telezhnaya:
I just ran `./examples/build_all.sh` inside the container and put all the diff here.

@chefsale could you please add the check that the binaries correspond with the source code for `examples` folder here?
The binaries changed before the last update, I just caught it now.


austinabell (approved):
thanks, this should definitely be part of the CI or at least a script to run within the docker image to make this more replicable in the future so this can be more easily tracked which changes have an effect on the wasm blob sizes

austinabell:
Closed because #509 replaces it

Issue #456, opened 2021-07-02
Wasm binary does not correspond to the current version of the source code


telezhnaya:
While working on https://github.com/near/near-sdk-rs/pull/454, I found out that the binary for `examples/fungible-token` changes if I run `build.sh` in [the container](https://github.com/near/near-sdk-rs/tree/master/contact-builder).

I think we should check and rebuild all the binaries here and in `core-contracts` (maybe somewhere else)?

It would be also cool to set up CI that will build the binaries on the server and prevent merging the code to master if the binary does not correspond to the source code.

ilblackdragon:
Let's include the build script inside container together with the repo to avoid issues like this. see https://github.com/ref-finance/ref-contracts/blob/main/ref-exchange/build_docker.sh for example

austinabell:
wasm binaries are now tied to the code through CI check

PR #464, opened 2021-07-09
Use Base58PublicKey for PublicKey instead of Vec<u8>

ChaoticTempest:
This somewhat goes in a bit of a different direction than #453, by making the internal representation of `Base58PublicKey` back into a `Vec<u8>`, but with this there won't be any breaking changes with `PublicKey` since we can deref into `Vec<u8>`. And CurveType is still accessible via `curve_type()`.


near-sdk/src/json_types/public_key.rs
```diff
@@ -77,19 +87,29 @@ impl Base58PublicKey {
...
+    }
+}
+
+impl Deref for Base58PublicKey {
+    type Target = Vec<u8>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.data
     }
 }
```
austinabell, line 109: This impl can be difficult to maintain, as it removes any flexibility to change the underlying type later. We need to be sure this will be the final type if doing this. `Deref` impls can be very tricky to maintain in a backwards compatible way

austinabell (commented):
I think there is definitely a benefit in avoiding logic on borsh (de)serialize impls, but not sure how much of an effect it would have on the compiled code. 

This would be a great time to have the contracts binary/size checked by the CI or manually to see implications (#456), maybe we should try to get this done sooner than later

near-sdk/src/json_types/public_key.rs
ChaoticTempest: That's a good point. Maybe `as_vec() -> &Vec<u8>` would be better for now, even though this would add on to the list of breaking changes

ChaoticTempest:
> This would be a great time to have the contracts binary/size checked by the CI or manually to see implications (#456), maybe we should try to get this done sooner than later

Definitely, since more PRs roll out, it would be great to have this kind of metric/test

near-sdk/src/json_types/public_key.rs
austinabell: That function would be just as hard to maintain, because it still requires the underlying type to be `Vec` to reference it. Could expose as_bytes with `&[u8]` to allow `Box<[u8]>` or some type that allows partial stack allocation.

Also pretty sure clippy warns against `&Vec<u8>` because there is no benefit to this

near-sdk/src/environment/env.rs
```diff
@@ -152,7 +153,7 @@ pub fn signer_account_id() -> AccountId {
 
 /// The public key of the account that did the signing.
 pub fn signer_account_pk() -> PublicKey {
-    method_into_register!(signer_account_pk)
+    PublicKey::try_from(method_into_register!(signer_account_pk)).unwrap()
```
matklad, line 156: This might want to do `.unwrap_or_else(|_| unreachable())`, as `.unwrap()` includes formatting machinery and associated bloat.

near-sdk/src/json_types/public_key.rs
```diff
@@ -77,19 +87,29 @@ impl Base58PublicKey {
...
+        bytes.extend(data);
 
-        Ok(Self { curve, data })
+        Ok(Self { data: bytes })
     }
 
     /// Get info about the CurveType for this public key
     pub fn curve_type(&self) -> CurveType {
-        self.curve
+        CurveType::from_u8(self.data[0]).unwrap()
```
matklad, line 99: Same, better to avoid unwraps on results which shouldn't be possible to avoid formatting related code bloat.

near-sdk/src/json_types/public_key.rs
matklad: I'd say `From<PublicKey> for Vec<u8>`, `PublicKey::into_bytes`  and `PublicKey::as_bytes` would be the right API here (that's how `std::String` exposes its internal impl)

near-sdk/src/json_types/public_key.rs
```diff
@@ -135,12 +150,12 @@ impl<'de> serde::Deserialize<'de> for Base58PublicKey {
 
 impl From<&Base58PublicKey> for String {
```
matklad, line 151: `From<X> for String` is an anti-pattern (unless the internal impl is a `String`), better to implement `fmt::Display`

matklad (commented):
Moves into the right direction. I suggest renaming the type to just `PublicKey`, and adding `#[deprecated] Base58PublicKey = PublicKey;` if we want to reduce incompat.

austinabell:
> Moves into the right direction. I suggest renaming the type to just `PublicKey`, and adding `#[deprecated] Base58PublicKey = PublicKey;` if we want to reduce incompat.

I don't think the deprecated tag works on type aliases, or do I have it wrong?

matklad:
Never tried it before, but seems to work: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=97005656c2fb5b9281777680949cda48

austinabell:
> Never tried it before, but seems to work: [play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=97005656c2fb5b9281777680949cda48](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=97005656c2fb5b9281777680949cda48)

Oops, I was wrong then, last time I tried it didn't. Also keep in mind that keeping the alias may cause worse error messages depending on which artifact is found first, but maybe you have a better idea for this. Probably worth keeping consistent and adding one (in another PR) for `ValidAccountId` as it's the same case as this

near-sdk/src/environment/env.rs
ChaoticTempest: That's a good catch 👍. I've also noticed that there's a lot of similar `unwrap`s in `env.rs`. We can evaluate those in a separate issue/PR but I'll change this one to make use of `unwrap_or_else` for now

near-sdk/src/json_types/public_key.rs
ChaoticTempest: Wouldn't this force us to use `format!` which potentially adds to the code size? The impl for `serde::Serialize` makes use of into String so we would end having to call `format!` instead there

near-sdk/src/json_types/public_key.rs
```diff
@@ -1,7 +1,15 @@
-use borsh::{maybestd::io, BorshDeserialize, BorshSerialize};
+use borsh::{BorshDeserialize, BorshSerialize};
 use bs58::decode::Error as B58Error;
 use std::convert::TryFrom;
 
+#[deprecated(
+    since = "4.0.0",
+    note = "PublicKey type is now unified with Base58PublicKey. It is \
```
mikedotexe, line 7: 🙏🏼

near-sdk/src/json_types/public_key.rs
```diff
@@ -39,24 +57,23 @@ impl std::str::FromStr for CurveType {
...
+/// use near_sdk::json_types::PublicKey;
 ///
 /// // Compressed ed25519 key
-/// let ed: Base58PublicKey = "ed25519:6E8sCci9badyRkXb3JoRpBj5p8C6Tw41ELDZoiihKEtp".parse()
+/// let ed: PublicKey = "ed25519:6E8sCci9badyRkXb3JoRpBj5p8C6Tw41ELDZoiihKEtp".parse()
 ///             .unwrap();
 ///
 /// // Uncompressed secp256k1 key
-/// let secp256k1: Base58PublicKey  = "secp256k1:qMoRgcoXai4mBPsdbHi1wfyxF9TdbPCF4qSDQTRP3TfescSRoUdSx6nmeQoN3aiwGzwMyGXAb1gUjBTv5AY8DXj"
+/// let secp256k1: PublicKey = "secp256k1:qMoRgcoXai4mBPsdbHi1wfyxF9TdbPCF4qSDQTRP3TfescSRoUdSx6nmeQoN3aiwGzwMyGXAb1gUjBTv5AY8DXj"
```
mikedotexe, line 67: good attention to details 🙏🏼 again

mikedotexe (approved):

near-sdk/compilation_tests/bad_argument.stderr
```diff
@@ -1,3 +1,13 @@
+warning: use of deprecated type alias `json_types::public_key::Base58PublicKey`: PublicKey type is now unified with Base58PublicKey. It is recommended to use PublicKey going forward to avoid using similar sounding types for the same thing.
+  --> $DIR/mod.rs:12:22
+   |
+12 | pub use public_key::{Base58PublicKey, CurveType, PublicKey};
+   |                      ^^^^^^^^^^^^^^^
+   |
+   = note: `#[warn(deprecated)]` on by default
+
+warning: 1 warning emitted
+
```
austinabell, line 10: just update the tests to use the new type, the deprecation should not be accounted for in tests

near-sdk/compilation_tests/impl_generic.stderr
```diff
@@ -1,3 +1,13 @@
+warning: use of deprecated type alias `json_types::public_key::Base58PublicKey`: PublicKey type is now unified with Base58PublicKey. It is recommended to use PublicKey going forward to avoid using similar sounding types for the same thing.
+  --> $DIR/mod.rs:12:22
+   |
+12 | pub use public_key::{Base58PublicKey, CurveType, PublicKey};
+   |                      ^^^^^^^^^^^^^^^
+   |
+   = note: `#[warn(deprecated)]` on by default
+
+warning: 1 warning emitted
+
```
austinabell, line 10: ditto for the other files too, I won't comment on all

near-sdk/src/json_types/public_key.rs
```diff
@@ -39,24 +57,23 @@ impl std::str::FromStr for CurveType {
...
 ///             .parse()
 ///             .unwrap();
 /// ```
-#[derive(Debug, Clone, PartialEq, PartialOrd, Ord, Eq)]
-pub struct Base58PublicKey {
-    curve: CurveType,
+#[derive(Debug, Clone, PartialEq, PartialOrd, Ord, Eq, BorshDeserialize, BorshSerialize)]
+pub struct PublicKey {
```
austinabell, line 72: This should not be exported as `near_sdk::json_types::PublicKey` and should not live in this module at all. Just need the `Base58PublicKey` alias to live here as deprecated until we remove it. The goal of this is to move away from the `json_types` module

austinabell (changes requested):

near-sdk/compilation_tests/bad_argument.stderr
ChaoticTempest: AFAIK the test's don't actually use the `Base58PublicKey` type itself. This is just the export and it fails the compilations tests if I don't include it

near-sdk/src/json_types/public_key.rs
ChaoticTempest: I see, I'll just move to the whole entire thing to `types/public_key.rs`. I was thinking of re-exporting CurveType from there too, and adding a deprecated attribute, but deprecated doesn't work on exports just yet

near-sdk/compilation_tests/bad_argument.stderr
ChaoticTempest: I guess manually exporting it causes this warning to appear in the tests. After moving the deprecated type into `json_types/mod.rs`, it's fine

near-sdk/src/types/vm_types.rs
```diff
@@ -2,8 +2,7 @@
 pub use near_vm_logic::types::{PromiseResult as VmPromiseResult, ReturnData};
 
 //* Types from near_vm_logic
-
-pub type PublicKey = Vec<u8>;
+// pub use super::public_key::PublicKey;
```
austinabell, line 5:
```suggestion
```

ChaoticTempest: oh dang, good catch

ChaoticTempest:
@mikedotexe Let's have this PR be merged into #453 first, then we can merge #453

austinabell (approved):

PR #469, opened 2021-07-12
Add deprecated ValidAccountId type mentioned in #464

ChaoticTempest:
Just to stay consistent with #464, added the deprecated attribute to `ValidAccountId`


austinabell (approved):

ChaoticTempest:
Simple enough change, so merging it in

PR #470, opened 2021-07-13
chore: fix clippy warnings

austinabell:
- Fixes clippy warnings for everything
- Adds clippy check to CI so these don't pile up again


near-sdk-sim/src/outcome.rs
```diff
@@ -100,10 +96,10 @@ impl ExecutionResult {
     }
 
     fn get_outcome(&self, hash: &CryptoHash) -> Option<ExecutionResult> {
-        match (*self.runtime).borrow().outcome(hash) {
-            Some(out) => Some(ExecutionResult::new(out, &self.runtime, hash.clone())),
-            None => None,
-        }
+        (*self.runtime)
+            .borrow()
+            .outcome(hash)
```
ChaoticTempest, line 101: hmm, I guess clippy didn't complain about this unnecessary explicit `deref`

.github/workflows/test.yml
```diff
@@ -11,20 +11,32 @@ jobs:
...
-          components: rustfmt
           target: wasm32-unknown-unknown
       - uses: Swatinem/rust-cache@v1
-      - name: Test Format
-        run: cargo fmt -- --check
       - name: Add wasm32 target
         run: rustup target add wasm32-unknown-unknown
       - name: Test
         run: cargo test --all --features unstable
+  lint:
```
mikedotexe, line 28: 👌🏼

near-contract-standards/src/non_fungible_token/approval/approval_impl.rs
```diff
@@ -64,19 +62,17 @@ impl NonFungibleTokenApproval for NonFungibleToken {
         refund_deposit(storage_used);
 
         // if given `msg`, schedule call to `nft_on_approve` and return it. Else, return None.
-        if let Some(msg) = msg {
-            Some(ext_approval_receiver::nft_on_approve(
+        msg.map(|msg| {
```
mikedotexe, line 65: This is still so confusing to me. :D

mikedotexe (approved):

ChaoticTempest (approved):

near-sdk-sim/src/outcome.rs
austinabell: it is needed here though, to indicate that it's borrowing through `RefCell` and not the `Rc`. There are other ways to indicate this, but I don't want to make any changes that can be potentially breaking if possible in this PR.

near-contract-standards/src/non_fungible_token/approval/approval_impl.rs
austinabell: Instead of manually constructing both the `Some` and `None` variants, it just transforms the inner value

Issue #408, opened 2021-05-07
Handle u64 and u128 at the serializer layer rather via wrapper types?


matklad:
We currently format `u64` and `u128` numbers as string when using JSON. That's because formatting them as numbers causes silent truncation errors for JavaScript clients.

At the moment, we do this via dedicated wrapper types, so the users need to take care to use wrapper for all of their types. This is tedious and error prone. Given that we want to just uniformly enforce that all `u64` are rendered as strings, perhaps we can do this at the level of serializer?

From the cursory look, this should be doable:

```rust
pub fn to_json_with_safe_numbers(obj: impl serde::Serialize) -> String {
    let mut res = Vec::new();
    let ser = SafeNumbersSerializer(serde_json::Serializer(&mut res));
    obj.serialize(ser);
    String::from_utf8(res).unwrap()
}

struct SafeNumbersSerializer(serde_json::ser::Serializer);

impl serde::Serializer for SafeNumbersSerializer {
    fn serialize_u64(self, v: u64) -> Result<Self::Ok, Self::Error> {
        self.0.serialize_str(&v.to_string())
    }
}
```

Actually doing this requires quite a bit of boilerplate (delegation of seq and map serializers is also required), but it seems doable, and the user-visible API would much friendlier than the current one.

If we are to do this, we probably should publish this as a crate, to be re-used for nearcore rpc as well.

austinabell:
related #286. 

I was exploring some alternatives at the proc macro level to allow customizing the serialization through annotations rather than having to wrap types, but I think I'll abandon because it ties the API too closely with serde, and I know we want to have the flexibility to change in the future to minimize code size.

This might be a better solution, but I was a bit uncomfortable with forcing the serialization of these integer types, which might be hard to debug if a user expects this to not be the case.

felixniemeyer:
I have the same situation here. 
I have some structs for initialization of contract parts and it seems I have to duplicate them: 
- one with `u128` for contract storage and calculations
- one with `U128` for use in function parameters for parsing from JSON.
That does not seem so nice. 

I also would wish, I could simply use u128 and `near_bindgen` would take care of parsing number strings fine.

PR #471, opened 2021-07-13
refactor(types)!: create newtype for gas

austinabell:
- Creates newtype for gas (part of https://github.com/near/near-sdk-rs/issues/387)
- Implements string serialization for this newtype
  - I'm assuming this is what we want, to move away from needing the `json_types` wrapper, but this could cause unexpected bugs if people migrating don't notice this has changed from the `u64` alias default of serializing as integer

I'm not convinced this should come in the next release, but let me know your thoughts!


near-sdk/src/types/gas.rs
```diff
@@ -0,0 +1,115 @@
...
+        Self(amount)
+    }
+}
+
+impl Serialize for Gas {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        serializer.serialize_str(&self.0.to_string())
```
ChaoticTempest, line 31: I wonder if we could use [smallstr](https://crates.io/crates/smallstr) for these kinds of fixed-size strings to avoid the `to_string` allocation underneath, but might not be necessary

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -14,8 +14,9 @@ use near_sdk::{
 };
 use std::collections::HashMap;
 
-const GAS_FOR_RESOLVE_TRANSFER: Gas = 5_000_000_000_000;
-const GAS_FOR_FT_TRANSFER_CALL: Gas = 25_000_000_000_000 + GAS_FOR_RESOLVE_TRANSFER;
+const GAS_FOR_RESOLVE_TRANSFER_INT: u64 = 5_000_000_000_000;
+const GAS_FOR_RESOLVE_TRANSFER: Gas = Gas::new(GAS_FOR_RESOLVE_TRANSFER_INT);
+const GAS_FOR_FT_TRANSFER_CALL: Gas = Gas::new(25_000_000_000_000 + GAS_FOR_RESOLVE_TRANSFER_INT);
```
ChaoticTempest, line 19: if only we had `#![feature(const_trait_impl)]`

ChaoticTempest (approved):
LGTM and nicely done

ChaoticTempest:
> I'm not convinced this should come in the next release, but let me know your thoughts!

it doesn't seem to break anything, besides requiring developers to use a `Gas` type instead of `u64` directly, but I might be missing something here

near-sdk/src/types/gas.rs
matklad: We don't even need smallstr here, the length is bounded (not sure what the actual bound is), so we can just use stack-allocated `[u8]`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=132595fc106b94493a5a240512ec9063

near-sdk/src/types/gas.rs
```diff
@@ -0,0 +1,115 @@
...
+    PartialEq,
+    PartialOrd,
+    Ord,
+    Eq,
+    BorshSerialize,
+    BorshDeserialize,
+    Hash,
+    BorshSchema,
+)]
+pub struct Gas(u64);
```
matklad, line 18:
```suggestion
pub struct Gas(pub u64);
```

We won't be changing the internal repr here, and the type doesn't have any invariants, so we might as well just expose this fact.

near-sdk/src/types/gas.rs
matklad: I am conflicted about this  impl via `to_string`. I think it's the best we can do, but we must be aware of the drawbacks.

The problem we want to solve here is that we want to serialize this as string *in json*. The current impl uses strings for every serde format. So if someone wants to use, eg, `bincode` for their contract, they'll be surprised to see that gas is encoded as string into the binary format. 

I feel that https://github.com/near/near-sdk-rs/issues/408 might be the "proper" soulution here, but it's clear that we aren't going to do that short-medium term. 

A possible hack here is using https://docs.serde.rs/serde/trait.Serializer.html#method.is_human_readable as a proxy for "is this JSON", but I am not sure if that's actually better. 

So, yeah, nothing to improve here, but maybe a good place for "// NOTA BENE: " comment.

near-contract-standards/src/non_fungible_token/core/core_impl.rs
matklad: We might provide inherent const methods: 

```rust
const GAS_FOR_RESOLVE_TRANSFER Gas = Gas(5_000_000_000_000);
const GAS_FOR_FT_TRANSFER_CALL: Gas = Gas::new(25_000_000_000_000).add(GAS_FOR_RESOLVE_TRANSFER_INT);
```

where `add` is defined as

```rust
impl Gas {
    pub const fn add(self, other: Gas) -> Gas { Gas(self.0 + other.0)}
}

impl ops::Addfor Gas {
    fn add(self, other: Gas) -> Gas { self.add(other) }
}
```

near-sdk/src/types/gas.rs
```diff
@@ -0,0 +1,115 @@
...
+
+impl core::ops::Add for Gas {
+    type Output = Self;
+
+    fn add(self, other: Self) -> Self {
+        Self(self.0 + other.0)
+    }
+}
+
+impl core::ops::AddAssign for Gas {
```
matklad, line 65: Nit: I'd `impl ops::AddAssign for Gas`, looks neater that way.

near-sdk/src/types/gas.rs
```diff
@@ -0,0 +1,115 @@
...
+        Self(self.0 * other)
+    }
+}
+
+impl core::ops::Div<u64> for Gas {
+    type Output = Self;
+
+    fn div(self, other: u64) -> Self {
+        Self(self.0 / other)
+    }
```
matklad, line 114: Feels a bit strange to provide `div` without also providing `div_rem` -- division is a lossy operation (it forgets remined), for orthogonal API there shoud be a way to recover the rem.

near-sdk/src/types/gas.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+impl core::ops::Sub<u64> for Gas {
+    type Output = Self;
+
+    fn sub(self, other: u64) -> Self {
+        Self(self.0 - other)
+    }
+}
```
matklad, line 99: It feels like we'd have more type safety without Add<u64>/Sub<u64> impls.

near-sdk/src/types/gas.rs
```diff
@@ -0,0 +1,115 @@
...
+    Clone,
+    Copy,
+    PartialEq,
+    PartialOrd,
+    Ord,
+    Eq,
+    BorshSerialize,
+    BorshDeserialize,
+    Hash,
+    BorshSchema,
```
matklad, line 16: `Default` is missing

near-sdk/src/types/gas.rs
austinabell: hmm thought it would have been better to avoid exposing the inner type, but yeah I guess we can assume it will never change

near-contract-standards/src/non_fungible_token/core/core_impl.rs
austinabell: I was considering that, but seemed like a smell to have the trait impl and const fn have the same name, and if it wasn't it would still be an unintuitive usage for most developers. I think if we are exposing the inner type, and I'll do that, it somewhat removes the need for this

near-sdk/src/types/gas.rs
austinabell: Do you mean just implementing `Rem` or implementing some trait from `num_integer` or something similar. I'm not sure what `div_rem` you are referring to, unless you mean as a core impl

near-sdk/src/types/gas.rs
austinabell: okay I'll take these ones out, just trying to have a convenient API

near-sdk/src/types/gas.rs
matklad: Hm, I was sure std had `div_rem` function, but that's not the case actually. Yeah, in this case just `Rem` would do the trick.

near-sdk/src/types/gas.rs
austinabell: Yeah, I think we can just ideate what is the best solution wrt #408 after this next release, as swapping the integer serialization at the serialization layer would also add some unintended behaviour that users may not expect and may want to eject from, and this release already has a lot of breaking changes.

near-sdk/src/types/gas.rs
austinabell: Added, just couldn't think of a reason that `Rem` would be needed for usage with Gas.

Also, just to confirm, you also think it makes the most sense to have the `Mul`, `Div`, and `Rem` operators be on `u64` rather than gas? The asymmetry is a bit weird to me, but in practice it seems to make more sense

austinabell:
Also, anyone have thoughts about adding a tera gas helper? Can always come after the major release, and seems like there is no clear best API, but just seeing if anyone has opinions.

near-sdk/src/types/gas.rs
matklad: There's no assymetry here -- for mul, div and rem, the multiplier is [dimensionless](https://en.wikipedia.org/wiki/Dimensionless_quantity). Had we used `Gas` there, we'd have to return `Gas²`.

>Added, just couldn't think of a reason that Rem would be needed for usage with Gas.

What's the use-case for Div? Div is somewhat rarely useful without Rem -- if you want to devide something in parts, you have to do something about the remainder.

austinabell:
> it doesn't seem to break anything, besides requiring developers to use a `Gas` type instead of `u64` directly, but I might be missing something here

I mean to break in the context of a breaking change, since the `Gas` type was u64 before and is now the wrapper type, so there are definitely cases (even in our own codebase) where this is breaking. It's definitely a small update if using specifically, but I'm just being cautious since this release already has a bunch of changes.

matklad:
>Also, does anyone have thoughts about adding a tera gas helper?


```rust
impl Gas {
    const TERA_GAS = Gas(1_000_000_000_000);
}
let amount = 42 * Gas::TERA_GAS;
```

austinabell:
> > Also, does anyone have thoughts about adding a tera gas helper?
> 
> ```rust
> impl Gas {
>     const TERA_GAS = Gas(1_000_000_000_000);
> }
> let amount = 42 * Gas::TERA_GAS;
> ```

I don't have an opinion, either this or some constructor function I'd be okay with adding, or we can wait to do in another PR so we can have specific buy-in for what's needed. cc @mikedotexe maybe you have an idea of the importance of having this or if there would be any other helper that would be useful?

examples/cross-contract-low-level/src/lib.rs
```diff
@@ -1,10 +1,10 @@
 use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
 use near_sdk::json_types::U128;
 use near_sdk::serde_json::{self, json};
-use near_sdk::{env, near_bindgen, AccountId, PromiseResult};
+use near_sdk::{env, Gas, near_bindgen, AccountId, PromiseResult};
 
 // Prepaid gas for making a single simple call.
-const SINGLE_CALL_GAS: u64 = 200000000000000;
+const SINGLE_CALL_GAS: Gas = Gas::new(200000000000000);
```
mikedotexe, line 7: Whoa, cool

mikedotexe (approved):

examples/cross-contract-low-level/src/lib.rs
matklad, line 7: I am undecided between options

* status quo (using `new` everywhere)
* using `Gas(200000000000000)` (without `new`) here and eslwhere in the sdk, to promote this as idiomatic
* not having `Gas::new` at all and requring usage of `Gas(amount)`

Don't think this is an important thing to decide (either option is fine), but I still don't know which one is the best. I have just so slightly bigger pref for the last one, getting rid of `new` entirely.

near-sdk/src/types/gas.rs
```diff
@@ -0,0 +1,140 @@
...
+)]
+pub struct Gas(pub u64);
```
matklad, line 20: I think we might also want to slap `#[repr(transparent)]` here, but we also can just wait until someone asks for it.

near-sdk/src/types/gas.rs
```diff
@@ -0,0 +1,140 @@
...
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut buf = [0u8; 20];
+        let remainder = {
+            use std::io::Write;
+
+            let mut w: &mut [u8] = &mut buf;
+            write!(w, "{}", self.0).ok().unwrap();
```
matklad, line 38: potentially code-bloating `.unwrap()`.

near-sdk/src/types/gas.rs
```diff
@@ -0,0 +1,140 @@
...
+        Self(amount)
+    }
+}
+
+impl Serialize for Gas {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut buf = [0u8; 20];
```
matklad, line 33: Let's add a test that we can round-trip `Gas(u64::max)`, to make sure the buffer size is enough.

examples/cross-contract-low-level/src/lib.rs
austinabell: yeah, fair point, probably safer to remove and then can always add back if needed. This was just needed before the inner type was exposed

near-sdk/src/types/gas.rs
austinabell: I already did, it's in this PR (checked with max, `0` and arbitrary `8`)

matklad: ah, I looked at that test and somehow missed that :)

PR #474, opened 2021-07-15
chore: deprecate block_index and replace with block_height

### LINKED ISSUE #460: Update `env::block_index` to `env::block_height`
This issue is just to create a new function `block_height` within the `env` module and deprecate `block_index` function as of version `4.0`.

The `sys` interface cannot be changed and it doesn't make sense to migrate the function on the `BlockchainInterface` trait because it will be removed soon.

Ref: https://github.com/near/nearcore/issues/1903#issuecomment-874926362

Edit: also potentially worth creating a `sys::block_height` function, which is not part of the extern "C" interface, but that calls `block_index` to match the high-level (`env`) interface.

austinabell:
- Deprecates `env::block_index`
- Adds `env::block_height` and `sys::block_height`

closes #460

### DIFF
--- a/near-sdk/src/environment/env.rs
+++ b/near-sdk/src/environment/env.rs
@@ -176,8 +176,14 @@ pub fn input() -> Option<Vec<u8>> {
 }
 
 /// Current block index.
+#[deprecated(since = "4.0.0", note = "Use block_height instead")]
 pub fn block_index() -> BlockHeight {
-    unsafe { sys::block_index() }
+    block_height()
+}
+
+/// Returns the height of the block the transaction is being executed in.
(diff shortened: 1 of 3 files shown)


ChaoticTempest (approved):

near-sdk/src/environment/env.rs
```diff
@@ -152,8 +152,14 @@ pub fn input() -> Option<Vec<u8>> {
 }
 
 /// Current block index.
+#[deprecated(since = "4.0.0", note = "Use block_height instead")]
```
mikedotexe, line 155: perfect

mikedotexe (approved):

PR #475, opened 2021-07-15
fix(nft): switch nft traits to not require moving underlying struct

austinabell:
- Pointed out by @mattlockyer as an API with friction when customizing, there was no need to move the struct in the first place
- Also switched enumeration standard to not require moving `self` either

~I also bumped the version so that I can make the minor release after it comes in~ can't actually release this change unless I cherry-pick on top of the previous version, this contains other changes not released


ChaoticTempest:
We'll likely need to update the `near-examples/NFT` example after this, since users will probably be using the example as a baseline. So better to update it sooner than later I believe

ChaoticTempest (approved):

austinabell:
> We'll likely need to update the `near-examples/NFT` example after this, since users will probably be using the example as a baseline. So better to update it sooner than later I believe

I'm not certain, but I think all of these functions are generated with the impl macro so won't need to change anything unless you override methods. I could be wrong though

ChaoticTempest:
> I'm not certain, but I think all of these functions are generated with the impl macro so won't need to change anything unless you override methods. I could be wrong though

Oh if its in the macro, then it should be fine

mikedotexe (approved):
I wasn't quite sure of the motivation to not use `&` and prefer this approach. I think perhaps there was an idea of having `self` be captured, but think this is in line with what we have elsewhere. Thanks

PR #476, opened 2021-07-16
chore: remove lockfile, bump versions

austinabell:
Prep for next release
- Removes lockfile (idk why it would be needed for a lib like this)
- ~~Bumps borsh to 0.9~~ 
  - we might want to revert this one if other near repos are not planning on updating to this. This will bring in multiple versions of borsh for testing (because of near-vm-logic)
- Bumps dev-dependencies


austinabell:
So the buildkite fails because it's using an old stable version that is not compatible with a newer version of a dependency. I don't get the context of why this old version is used and versions are pinned because this is used as a lib, not a binary. I don't have access to change the toolchain to just `stable` though or understand why this specific version is used.

ci-test.sh
```diff
@@ -8,6 +8,8 @@ if [[ "${NEAR_RELEASE}" == "true" ]]; then
 else
     echo "Test with git version of borsh and near-vm-logic"
 
+    cargo generate-lockfile
```
mikedotexe, line 11: cool, didn't know about this command

mikedotexe (approved):

PR #477, opened 2021-07-16
feat: merge and delete near-sdk-core

### LINKED ISSUE #369: move near-sdk-core crate into near-sdk-macros
`near-sdk-core` and `near-sdk-macros` are two crate which look like they should be one. The `-core` crate just defines the proc-macro machinery that is invoked from (and only from) the `-macros` crate.

Merging the two has the following benefits:

* one less semver version to manage / crate to publish
* one less entry in Cargo.lock for contract authors (fighting perception that Cargo is like npm)
* less confusion about what `near-sdk-core` is -- the name sounds like this crate defines the core runtime types of the SDK for the purposes of greater stability (like rand-core or rayon-core). But in fact its just the opposite -- it's the private impl details of a proc macro

While we are at it, it makes sense to fix nonsensical cfgs in this crate:

https://github.com/near/near-sdk-rs/blob/31cba6213919ec4866bc0fab7f6963de241cc43a/near-sdk-core/src/code_generator/item_impl_info.rs#L17-L23

This is a proc macro, so `target_arch` is *never* wasm32

austinabell:
closes #369 
- Deletes near-sdk-core
- Fixes cfg attributes


near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
```diff
@@ -151,7 +151,7 @@ impl ImplItemMethodInfo {
             #non_bindgen_attrs
             #[cfg(target_arch = "wasm32")]
             #[no_mangle]
-            pub extern "C" fn #ident() {
```
matklad, line 154: I think `extern "C"` should stay, it protects you from accidently exposing the undercommunted and explicitly unstable details of the ABI of the partciular compiler version:

```
14:59:07|~/tmp
λ rustc lib.rs --crate-type cdylib --target wasm32-unknown-unknown -C link-arg=-s
warning: `extern` fn uses type `S`, which is not FFI-safe
 --> lib.rs:4:24
  |
4 | pub extern "C" fn f(_: S) {}
  |                        ^ not FFI-safe
  |
  = note: `#[warn(improper_ctypes_definitions)]` on by default
  = help: consider adding a `#[repr(C)]` or `#[repr(transparent)]` attribute to this struct
  = note: this struct has unspecified layout
note: the type is defined here
 --> lib.rs:1:1
  |
1 | pub struct S(u64, u64);
  | ^^^^^^^^^^^^^^^^^^^^^^^

warning: 1 warning emitted


14:59:09|~/tmp
λ micro lib.rs

14:59:17|~/tmp
λ bat -p lib.rs
pub struct S(u64, u64);

#[no_mangle]
pub fn f(_: S) {}

14:59:23|~/tmp
λ rustc lib.rs --crate-type cdylib --target wasm32-unknown-unknown -C link-arg=-s
```

austinabell: ah, interesting, was not aware. Will add back

ChaoticTempest (approved):

mikedotexe (approved):

PR #480, opened 2021-07-20
Hotfix/455 local paths

### LINKED ISSUE #455: Configure build to not include local paths into wasm
Currently when developers compile wasm file, the local paths are included into the wasm file in the asserts and other errors.

This is very suboptimal for reproducibility and also leaks private information about developer, which they are not always aware.

There should be a way to remove these absolute paths from the resulting file. If not, we need a tool that will clean this up after compilation and as default part of SDK.

mikedotexe:
Fixes #455 

Fairly straightforward


austinabell (commented):
your cargo path and near-sdk-rs path are still included in these built binaries.

`/Users/mike/.cargo`

`/Users/mike/near/near-sdk-rs`

So I'm a bit confused at what the point of this is

austinabell:
Closing because assuming you are not working on this anymore. Feel free to open again if you'd like to continue :)

PR #487, opened 2021-07-21
feat: caching LookupMap implementation

austinabell:
The interface closely matches a `std::collections::HashMap` without the functions that require being aware of all keys (as this does not store keys, another data structure will be built on this for that) and adding on functions that are useful in this context such as:
- `set`: update value without loading the existing value from state and returning replaced
- `flush`: To write cached modifications to storage manually (done automatically on `Drop`, but this could be useful for if measuring storage usage before and after and not requiring drop)


near-sdk/src/store/lookup_map/mod.rs
```diff
@@ -0,0 +1,533 @@
...
+            let _ = self.cache.get(k.to_owned()).set(CacheEntry::new_cached(None));
+        }
+        contains
+    }
+
+    /// Removes a key from the map, returning the value at the key if the key
+    /// was previously in the map.
+    ///
+    /// The key may be any borrowed form of the map's key type, but
+    /// [`BorshSerialize`] and [`ToOwned<Owned = K>`](ToOwned)on the borrowed form *must* match those for
```
ChaoticTempest, line 284: idk how this looks like when rendered, but might want a space here?
```suggestion
    /// [`BorshSerialize`] and [`ToOwned<Owned = K>`](ToOwned) on the borrowed form *must* match those for
```

ChaoticTempest (changes requested):
just some comments about naming, but overall looking good

near-sdk/src/utils/stable_map.rs
```diff
@@ -40,4 +41,12 @@ impl<K, V> StableMap<K, V> {
     pub(crate) fn inner(&mut self) -> &mut BTreeMap<K, Box<V>> {
         self.map.get_mut()
     }
+    pub(crate) fn with_value<Q: ?Sized, F, T>(&self, k: &Q, f: F) -> Option<T>
```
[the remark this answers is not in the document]
austinabell: My thinking was that this is not actually mapping the value since it's just calling a function on the reference to the value, not moving the owned value to be mapped. This might just be semantics, but map would be `FnOnce(V) -> T` when this is actually `FnOnce(&V) -> T`

If you feel strongly about this I'm happy to change since I don't really have an opinion here and it's a private API

near-sdk/src/store/lookup_map/mod.rs
```diff
@@ -0,0 +1,521 @@
...
+    /// Initialize a [`LookupMap`] with a custom hash function.
+    ///
+    /// # Example
+    /// ```
+    /// use near_sdk::hash::Keccak256;
+    /// use near_sdk::store::LookupMap;
+    ///
+    /// let map = LookupMap::<String, String, Keccak256>::new_with_hasher(b"m");
+    /// ```
+    pub fn new_with_hasher<S>(prefix: S) -> Self
```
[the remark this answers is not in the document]
austinabell: Yeah, I think I avoided making the same because it has a different API, this one doesn't take an owned hasher into the structure, but I'll switch.

near-sdk/src/store/lookup_map/entry.rs
```diff
@@ -0,0 +1,379 @@
...
+    ///
+    /// let mut map: LookupMap<String, String> = LookupMap::new(b"m");
+    /// let s = "hoho".to_string();
+    ///
+    /// map.entry("poneyland".to_string()).or_insert_with(|| s);
+    ///
+    /// assert_eq!(map["poneyland"], "hoho".to_string());
+    /// ```
+    pub fn or_insert_with<F: FnOnce() -> V>(self, default: F) -> &'a mut V {
+        match self {
```
matklad, line 69: this could delegate to `or_insert_with_key`

near-sdk/src/store/lookup_map/entry.rs
```diff
@@ -0,0 +1,379 @@
...
+    /// assert_eq!(map["poneyland"], 3);
+    ///
+    /// *map.entry("poneyland".to_string()).or_insert(10) *= 2;
+    /// assert_eq!(map["poneyland"], 6);
+    /// ```
+    pub fn or_insert(self, default: V) -> &'a mut V {
+        match self {
+            Self::Occupied(entry) => entry.into_mut(),
+            Self::Vacant(entry) => entry.insert(default),
+        }
```
matklad, line 50: This can delegate to `or_insert_with`

```suggestion
        self.or_insert_with(|| default)
```

That way, there's a single place which contains the match.

near-sdk/src/store/lookup_map/mod.rs
```diff
@@ -0,0 +1,533 @@
...
+
+    /// Returns `true` if the map contains a value for the specified key.
+    ///
+    /// The key may be any borrowed form of the map's key type, but
+    /// [`BorshSerialize`] and [`ToOwned<Owned = K>`](ToOwned) on the borrowed form *must* match those for
+    /// the key type.
+    pub fn contains_key<Q: ?Sized>(&self, k: &Q) -> bool
+    where
+        K: Borrow<Q>,
+        Q: BorshSerialize + ToOwned<Owned = K> + Ord,
```
matklad, line 260: Huh, that's subtle! 

I think here we have an extra requrenment that `Q: BorshSerialize` and `K: BorshSerialize` use the same serialization format. IE, this extends the usual `Borrow` contract to cover serialization format as well. 

I don't see any immediate problems here.

matklad (commented):
Did a very quick look, will continue tomorrow!

near-sdk/src/store/lookup_map/mod.rs
austinabell: yeah, it's not ideal. I couldn't think of a cleaner way to handle this, though. Maybe we can document just to make sure this assumption is clear. Assuming `K` and `Borrow<K>` have the same serialization seems in line with how `Borrow` is defined in some places, but others indicate that it only covers `Hash`, `Eq`, `Ord`

near-sdk/src/store/lookup_map/entry.rs
austinabell: Does this get compiled away? I just noticed that `std::collections::HashMap` [does not do this](https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#2791) and decided to follow their pattern.

near-sdk/src/store/lookup_map/entry.rs
austinabell: same here as above, I'm following std::HashMap pattern. I'll try to dig into comparing what the two compile to exactly to make sure there is no difference

near-sdk/src/store/lookup_map/entry.rs
matklad: Yes, according to my mental model, the compiler should trivially compile that away. This pattern is used by `Option` and once cell: https://github.com/rust-lang/rust/blob/cdeba02ff71416e014f7130f75166890688be986/library/core/src/option.rs#L1228

It's interesting that maps don't do that. I suppose that's historical -- this is the form the API was added before 1.0, and the form it was copy-pasted into hash brown: https://github.com/rust-lang/rust/pull/22930. 

No strong opinion here though -- if you'll feel more confident with the current impl, that's totally fine!

near-sdk/src/store/lookup_map/mod.rs
```diff
@@ -0,0 +1,533 @@
...
+    /// the key type.
+    pub fn get<Q: ?Sized>(&self, k: &Q) -> Option<&V>
+    where
+        K: Borrow<Q>,
+        Q: BorshSerialize + ToOwned<Owned = K>,
+    {
+        //* ToOwned bound, which forces a clone, is required to be able to keep the key in the cache
+        let entry = self
+            .cache
+            .get(k.to_owned())
```
matklad, line 207: thought: it's a bit sad that we `to_owned` on every access (instead of only on the *first* access), but this needs (unstable) raw_entry API to get fixed in a nice way.

near-sdk/src/store/lookup_map/mod.rs
```diff
@@ -0,0 +1,533 @@
...
+    /// Inserts a key-value pair into the map.
+    ///
+    /// If the map did not have this key present, [`None`] is returned.
+    ///
+    /// If the map did have this key present, the value is updated, and the old
+    /// value is returned. The key is not updated, though; this matters for
+    /// types that can be `==` without being identical.
+    pub fn insert(&mut self, k: K, v: V) -> Option<V>
+    where
+        K: Clone,
```
matklad, line 247: suggestion(blocking for stabilization): the `K: Clone` bound doesn't make sense -- we already have an owned `K` here. Can we write 

```rust
impl StableMap {
  fn get_key_value_mut(&mut self, k: K) -> (&mut K, &mut V) { ... }
}
```

sounds doable in theory?

near-sdk/src/store/lookup_map/mod.rs
```diff
@@ -0,0 +1,533 @@
...
+{
+    /// Flushes the intermediate values of the map before this is called when the structure is
+    /// [`Drop`]ed. This will write all modified values to storage but keep all cached values
+    /// in memory.
+    pub fn flush(&mut self) {
+        let mut buf = Vec::new();
+        for (k, v) in self.cache.inner().iter_mut() {
+            if let Some(v) = v.get_mut() {
+                if v.is_modified() {
+                    let key = Self::lookup_key(&self.prefix, k);
```
matklad, line 339: thought: looking at this, I don't like it for two reasons:

_First_, we could re-use the buffer inside `lookup_key`

_Second_, I think this means that we hash the data twice. When doing `.insert`, you first hash to check if the data is there, and then do a second `hash` on Drop. 

Does it make sense to always store a hash in a modified entry? Seems that can allow us to remove K: Serialize bound from drop, which might be nice. 

(I don't think we can remove `V: Serilalize` in the same way -- it's important that you can do several in-memory modifications of the value without re-serializing)

near-sdk/src/store/lookup_map/mod.rs
```diff
@@ -0,0 +1,533 @@
...
+    #[test]
+    fn test_insert() {
+        let mut map = LookupMap::new(b"m");
+        let mut rng = rand_xorshift::XorShiftRng::seed_from_u64(0);
+        for _ in 0..500 {
+            let key = rng.gen::<u64>();
+            let value = rng.gen::<u64>();
+            map.insert(key, value);
+        }
+    }
```
matklad, line 381: issue: randomized tests don't seem to be thorough. Specific problems: 

* `gen::<u64>` is guaranteed to generate different numbers, this'll never check the code path when the item is already there 
* we don't randomize the sequence of operation itself, it's the mixture of different operations that's interesting
* I think we want to keep these kinds of test as on of the main tool for checking sdk, so using `rand` directly feels like under-engineering. 

I've tried to come up with a better test here, and it found some problem. I haven't investigated the issue yet (it very well might be that my test is no good). Here's how I'd do this (using the astonishing `arbitrary` crate): 

```rust

use arbitrary::{Arbitrary, Unstructured};
use rand::seq::SliceRandom;
use rand::{Rng, RngCore, SeedableRng};

use crate::test_utils::test_env::setup_free;

#[derive(arbitrary::Arbitrary)]
enum Op {
    Add(u8, u8),
    Remove(u8),
    Flush,
    Restore,
    Get(u8),
}

#[test]
fn arbitrary() {
    setup_free();

    let mut rng = rand_xorshift::XorShiftRng::seed_from_u64(0);
    let mut buf = vec![0; 4096];
    for _ in 0..4096 {
        rng.fill_bytes(&mut buf);

        let mut lm = LookupMap::new(b"l");
        let mut hm = HashMap::new();
        let u = Unstructured::new(&buf);
        if let Ok(ops) = Vec::<Op>::arbitrary_take_rest(u) {
            for op in ops {
                match op {
                    Op::Add(k, v) => {
                        let r1 = lm.insert(k, v);
                        let r2 = hm.insert(k, v);
                        assert_eq!(r1, r2)
                    }
                    Op::Remove(k) => {
                        let r1 = lm.remove(&k);
                        let r2 = hm.remove(&k);
                        assert_eq!(r1, r2)
                    }
                    Op::Flush => {
                        lm.flush();
                    }
                    Op::Restore => {
                        lm = LookupMap::new(b"l");
                    }
                    Op::Get(k) => {
                        let r1 = lm.get(&k);
                        let r2 = hm.get(&k);
                        assert_eq!(r1, r2)
                    }
                }
            }
        }
    }
}
```

this currently gives

```
λ cargo test --release --package near-sdk --lib --features unstable -- store::lookup_map::tests::arbitrary --exact --nocapture
   Compiling near-sdk v4.0.0-pre.2 (/home/matklad/projects/near-sdk-rs/near-sdk)
    Finished release [optimized] target(s) in 2.75s
     Running unittests (target/release/deps/near_sdk-1734f3cd20b4f681)

running 1 test
thread 'store::lookup_map::tests::arbitrary' panicked at 'assertion failed: `(left == right)`
  left: `Some(147)`,
 right: `None`', near-sdk/src/store/lookup_map/mod.rs:409:29
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test store::lookup_map::tests::arbitrary ... FAILED
```

near-sdk/src/environment/hash.rs
```diff
@@ -0,0 +1,50 @@
+use crate::sys;
+
+const ATOMIC_OP_REGISTER: u64 = u64::MAX - 2;
+
+fn read_register_fixed(register_id: u64, buf: &mut [u8]) {
```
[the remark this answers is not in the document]
austinabell: Yeah, I just didn't because it's inline with the assumptions made about the host functions inside of `env`. Happy to switch over, though

near-sdk/src/environment/hash.rs
```diff
@@ -0,0 +1,50 @@
...
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum Sha256 {}
+
+impl CryptoHasher for Sha256 {
+    type Digest = [u8; 32];
+
+    fn hash(ingest: &[u8]) -> Self::Digest {
+        unsafe { sys::sha256(ingest.len() as _, ingest.as_ptr() as _, ATOMIC_OP_REGISTER) };
+
+        let mut hash = [0u8; 32];
```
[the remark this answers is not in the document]
austinabell: For some reason I assumed that would be more overhead for code size, but yeah probably should be this

near-sdk/src/store/lookup_map/mod.rs
austinabell: Yeah these are just ported tests from the existing lookup map, I'll make sure this (and have plans to do the same for others) are fuzzed correctly

near-sdk/src/store/lookup_map/entry.rs
austinabell: No, that's fine, I'll switch. I'm just noting the reason I had was that I assumed because it was the pattern in std::.::HashMap that there was a benefit

near-sdk/src/lib.rs
```diff
@@ -9,6 +9,9 @@ pub use near_sdk_macros::{
 #[cfg(feature = "unstable")]
 pub mod store;
 
+#[cfg(feature = "unstable")]
+pub use environment::hash;
```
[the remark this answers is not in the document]
austinabell: hmm yeah I did want it module scoped because I didn't want these things to be accidentally used. Possibly I need to think this through more before before this comes in or definitely before becoming stable as to how this will fit in

near-sdk/src/store/lookup_map/mod.rs
```diff
@@ -0,0 +1,533 @@
...
+const ERR_NOT_EXIST: &str = "Key does not exist in map";
+
+type LookupKey = [u8; 32];
+
+/// A non-iterable, lazily loaded storage map that stores its content directly on the storage trie.
+///
+/// This map stores the values under a hash of the map's `prefix` and [`BorshSerialize`] of the key
+/// using the map's [`CryptoHasher`] implementation.
+///
+/// The default hash function for [`LookupMap`] is [`Sha256`] which uses a syscall to hash the
```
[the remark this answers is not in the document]
austinabell: I'll improve the docs here, but nothing really that great to link to right now. I'll keep this in mind

near-sdk/src/utils/stable_map.rs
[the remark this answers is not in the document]
ChaoticTempest: ahh I see. I don't have too strong of an opinion what this should be either. And huh, [VaListImpl::with_copy](https://doc.rust-lang.org/std/ffi/struct.VaListImpl.html#method.with_copy) has a similar signature, so it's not too bad for it to have a prefix of `with_` but certainly very uncommon

near-sdk/src/store/lookup_map/mod.rs
austinabell: Intuitively, it makes sense, but I don't know how you're getting the `&mut K`, let alone a `&K` through the entry API like this.

I'll come back to this to see if there is anything that can be done, but at best with some code duplication all I can come up with is removing one clone, which I'm not certain is very helpful

near-sdk/src/store/lookup_map/mod.rs
austinabell: So the error in the test shown above was due to the mocked storage not being cleared in-between iterations. Modifications of what's shown here added in [`be564e8` (#487)](https://github.com/near/near-sdk-rs/pull/487/commits/be564e8fafb26b59c49a2d1883776d25fc14ae73)

near-sdk/src/store/lookup_map/mod.rs
austinabell: addr in [`62ca86a` (#487)](https://github.com/near/near-sdk-rs/pull/487/commits/62ca86a10301515c0c3b1eeb74044739eff90645) and [`1f525c8` (#487)](https://github.com/near/near-sdk-rs/pull/487/commits/1f525c8ec0ca9b874853be48e458e3575fef318d). 

Was a little tricky to come up with a solution that fit all cases because it's not always the same. For example, `set` doesn't do any storage reads, so the prefix doesn't need to be calculated there at all, so this solution needed to handle a few permutations for each situation, and it was hard to write this with very simple code. I'll adjust the arbitrary test to try to make sure all cases are tested

near-sdk/src/utils/stable_map.rs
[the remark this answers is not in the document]
austinabell: renamed to `map_value_ref`. I don't really think there is a great name, but it's internal and this might be the clearest. Sound good to you?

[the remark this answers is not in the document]
ChaoticTempest: Yup that sounds good to me. Thanks

matklad (approved):
Didn't do a second pass through, but this certainly feel ready as an *unstable* feature.

ChaoticTempest (approved):

PR #491, opened 2021-07-23
fix: private annotation codegen to include near_sdk prefix

austinabell:
noticed doing https://github.com/ref-finance/ref-contracts/pull/34 that this dep reference is incomplete and requires a user to import `near_sdk::env` when any function is private.


willemneal (approved):

ChaoticTempest (approved):
Macros are fun

mikedotexe:
Interesting, this doesn't change anything from the perspective of a contract developer, does it?

austinabell:
> Interesting, this doesn't change anything from the perspective of a contract developer, does it?

Not at all, just means they won't be required to import `env` if they aren't already using `env` (see https://github.com/near/core-contracts/blob/54f8267436058567b66efdaba3e790394f7f1529/w-near/src/lib.rs#L19 for one example of this)

mikedotexe (approved):

PR #492, opened 2021-07-25
feat(env): Replace panic with panic_str and add abort method

austinabell:
I wanted to avoid doing this to increase the number of changes necessary, but since it doesn't break anything I think it might be okay. This makes `panic` inline with `log` where it correctly takes a `&str` because the bytes must be utf8, symmetric with #362. It seems weird to update one and not the other before the stable release.

Also adds `abort` which just panics without a custom message, as this could be useful for some. This second part is optional as likely if someone is trying to optimize this much for contract size, it might be fine to require going through unsafe sys interface directly, to discourage this usage by default.


ChaoticTempest (approved):

near-sdk/src/environment/env.rs
```diff
@@ -488,9 +488,23 @@ pub fn value_return(value: &[u8]) {
...
     unsafe { sys::panic_utf8(message.len() as _, message.as_ptr() as _) }
 }
+
+/// Terminates the execution of the program with the UTF-8 encoded message.
+pub fn panic_str(message: &str) -> ! {
+    unsafe { sys::panic_utf8(message.len() as _, message.as_ptr() as _) }
+}
+
+/// Aborts the current contract execution without a custom message.
+/// To include a message, use [`panic_str`].
```
mikedotexe, line 503: I believe you said this is to really minimize overhead. Should we put something in the comments for folks to know when they might want to use this? This might be wrong, but I was imagining something like, "This can be a useful approach for smart contracts that avoid using UTF-8 for optimization reasons, something something…"

austinabell: Has nothing to do with optimizing, as this has the same internal type as bytes, this just makes sure at compile time that the "bytes" are utf8 and is a more ergonomic type (easier to get to `&str` for logging messages and bytes gives no benefit since non-utf8 disallowed).

mikedotexe (approved):

PR #495, opened 2021-07-26
refactor(env)!: remove public key generics from promise API

austinabell:
No need for this generic. Can potentially lead to a larger code size if misused without any noticeable benefit.


ChaoticTempest:
Since this seems to be a breaking change, should we update the changelog too?

ChaoticTempest (approved):

mikedotexe (approved):

PR #496, opened 2021-07-26
refactor: move BLOCKCHAIN_INTERFACE to mock and update usages

austinabell:
Logically makes more sense to live here, and makes references to it cleaner.

This does not functionally change anything


ChaoticTempest (approved):
Nice, looks a lot cleaner now without directly needing to call `borrow`/`borrow_mut` each time

PR #499, opened 2021-07-27
feat: add Eq/PartialEq to Receipt and actions

austinabell:
Esp since this is just used for testing, I noticed a contract I am migrating would benefit a lot from this.


ChaoticTempest (approved):

mikedotexe (approved):
yes. good idea.

PR #502, opened 2021-07-27
fix: pubkey deserialize validation and add hash trait derive

### LINKED ISSUE #498: PublicKey type ignores validation on deserialization
Noticed while doing something related, `PublicKey` just uses the borsh derive deserialize, which ignores validation and accepts any bytes.

Also, it doesn't implement hash, which was used by the `multisig` contract. Although it's generally better to avoid std hash collections in wasm contracts, probably worth it to have for tests. Thoughts @matklad if it's worth leaving out or adding `Hash` derive to public key?

austinabell:
- Invalid public keys can currently be constructed, my bad for missing this on review
- Adds `Hash` trait derive
  - This is generally not ideal to use within a contract, but can be useful for tests and good to support in case absolutely needed

closes #498


ChaoticTempest (approved):
Nice catch!

near-sdk/src/types/public_key.rs
```diff
@@ -60,7 +60,7 @@ impl std::str::FromStr for CurveType {
 ///             .parse()
 ///             .unwrap();
 /// ```
-#[derive(Debug, Clone, PartialEq, PartialOrd, Ord, Eq, BorshDeserialize, BorshSerialize)]
+#[derive(Debug, Clone, PartialEq, PartialOrd, Ord, Eq, BorshSerialize, Hash)]
```
matklad, line 63: I've recently came up with a rule-of-thumb: "don't derive serialize/deserialize for newtype structs". 

For newtypes, implementing serialization by hand is easy -- just delegate to the field.  And the benefit can be large -- the "derive" part of serde/borsh is heavier than the runtime. 

For the SDK, it might makes sense to:

* check if we can get rid of all derives
* if we can, consider not depending on the corresponding proc-macros

Though, I guess for the sdk this doesn't make too much sense --- the sdk is heavy proc-macro-based itself, so there won't be any extra savings.

austinabell: yes, this can come in a future change, not going to group with this PR because it's unrelated

Issue #463, opened 2021-07-09
Consider splitting SDK into separately publishable layers


matklad:
Our current SDK is pretty hight-level and opinionated – is presupposes borsh serialization, strongly-typed wrappers for types, etc. While it is a great fit for the average use-case (or at least aspires to be one), one size doesn't fit all!

It might be useful to publish unopinionated subsets of the SDK, which just give bindings to the system APIs, but don't prescribe any particular style of contracts. 

There's not huge **specific** benefits here, but this unlocks the ecosystem for serendipitous experimentation. If we publish low-level sdk bindings, someone from the ecosystem might decide to publish a better hight-level SDK than ours, and that can have a huge positive effect.

I can see three possible layers we might want to split out:

***near-sys*** our <span class="error">[`sys.rs`]</span>([https://github.com/near/near-sdk-rs/blob/master/near-sdk/src/environment/sys.rs](https://github.com/near/near-sdk-rs/blob/master/near-sdk/src/environment/sys.rs)) published as a crate. 

***near-api*** thin, no_std, no_alloc, safe wrapper over `near-sys`. <span class="error">[nesdie]</span>([https://github.com/austinabell/nesdie](https://github.com/austinabell/nesdie)) is roughly this layer, but I think we also should get rid of alloc here, and use the APIs like:

```rust
 pub fn register_len(register_id: u64) -> Option<u64> {
 let len = unsafe 

 { sys::register_len(register_id) } 

;
 if len == core::u64::MAX 

 { None } 

 else 

 { Some(len) } 

}

pub fn read_register(register_id: u64, buf: &mut <span class="error">[u8]</span>) -> Result<usize, ()>{
 let len = register_len(register_id).ok_or(())?;
 if buf.len() < len 

 { return Err(()); } 

 unsafe 

 { sys::read_register(register_id, buf.as_ptr() as u64) } 

;
 Ok(len)
 }
 ```

***near-sdk*** opinionated, alloc wrapper around `near-api`.

This depends on specifics, but, if it is actually possible to provide zero-overhead fully safe wrapper around `sys`, we might merge `near-sys` and `near-api` crates into one.

matklad:
cc https://github.com/near/nearcore/issues/4485 <- this is the kind of problem which might prevent us from shipping 100% unopinonated **near-api** layer.

austinabell:
[...]

PR #507, opened 2021-07-29
feat: Add new syscalls and move to separate crate

austinabell:
- Adds new syscalls
- Moves sys to own crate, to be more easily shared and used externally

There was one function I didn't add, which is `gas(u64)` which just consumes gas, as I didn't see any benefit. Happy to add if anyone can think of a reason, but I don't see any usage of it anywhere.

I won't release this crate until this is approved, just to be safe


austinabell:
new clippy version causes warnings, will fix in a separate PR

mikedotexe:
Is this part of the broader goal to have an efficient "under layer" to the SDK so folks can optionally build no_std contracts?

austinabell:
> Is this part of the broader goal to have an efficient "under layer" to the SDK so folks can optionally build no_std contracts?

Yes, it is. Let me link the issue here: #463

sys/src/lib.rs
```diff
@@ -31,16 +34,25 @@ extern "C" {
     pub fn sha256(value_len: u64, value_ptr: u64, register_id: u64);
     pub fn keccak256(value_len: u64, value_ptr: u64, register_id: u64);
     pub fn keccak512(value_len: u64, value_ptr: u64, register_id: u64);
+    pub fn ripemd160(value_len: u64, value_ptr: u64, register_id: u64);
+    pub fn ecrecover(
```
ChaoticTempest, line 38: I wonder if we can't just link the docs from nearcore directly into these sys calls. I feel like it would be a painpoint eventually for someone who's trying to optimize their contracts but not directly know what all these syscalls are, and have to indirectly discover [near-vm-logic/logic.rs](https://github.com/near/nearcore/blob/master/runtime/near-vm-logic/src/logic.rs) to understand them

ChaoticTempest (approved):
LGTM: didn't see anything out of the ordinary when cross referencing them with nearcore

PR #509, opened 2021-07-29
feat: setup docker builds for contracts and add WASM check to CI

### LINKED ISSUE #465: Update the container on DockerHub, add checks to CI
- [x] We need to update the container on DockerHub https://hub.docker.com/r/nearprotocol/contract-builder
New Dockerfile is here https://github.com/near/near-sdk-rs/blob/master/contract-builder/Dockerfile
This is required for https://github.com/near/core-contracts/pull/156 and https://github.com/near/core-contracts/pull/159
Please check that we will not break anything else by this change.

- [ ] We also need to add the check that the binaries correspond to code sources, at least in `examples` folder #461

austinabell:
- Sets up docker build scripts for examples
- Updates buildkite script to use them and check for modified wasm blobs
  - This makes it so that any change that affects the binary must be committed with the change
- Removes size check script, there is a bug in twiggy, which is being used

Closes #465 
Supersedes #480 (kinda, this just switches the path to the docker image's path but is reproducible now)
Supersedes #461 (These were built before some other changes were done so this should update to current)


chefsale (approved):
Thanks for doing this.

.github/workflows/test_examples.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest]
-        toolchain: [stable-2020-10-08]
+        toolchain: [stable]
```
mikedotexe, line 14: thank you, why did we even have the other version. didn't notice 👍🏼

austinabell: idk it was the version used across the board before, I assume for consistency, but is counter-intuitive for a lib like the SDK

mikedotexe:
Awesome. at some point (doesn't have to be in this PR) we can probably let folks know (in the "Contributing" section of the README or something) that they don't need to revision Wasm binaries anymore.

mikedotexe (approved):

PR #514, opened 2021-08-03
refactor: replace assert statements with require 

austinabell:
- Uses #493, was curious to see the change in wasm files from switching over. 

The error messages will be affected. No line numbers and path included which is more of an opinionated change so I put as separate PR


HELP.md
```diff
@@ -100,8 +100,8 @@ impl Contract {
         // Verify that the migration can only be done by the owner.
         // This is not necessary, if the upgrade is done internally.
         assert_eq!(
```
ChaoticTempest, line 102: Should we also be switching over the doc `assert!`s to `require!`s? Might be more beneficial to recommend `require` this way so we can have a singular de facto way of making assertions

austinabell: ah, yeah, this might have swapped back from a merge resolution. Let me switch

austinabell: this will be deleted once https://github.com/near/sdk-docs/pull/37 comes in, but possibly worth adding a point about `require` cc: @mikedotexe

ChaoticTempest (approved):
Sweet, that's a good amount of reduction from the binary size just from removing the formatting from asserts

Issue #494, opened 2021-07-26
Rust SDK 4.0 stabilization tracker


austinabell:
Probably a good idea to keep track of things we want to be sure of before the 4.0 release:
- [x] #491 
- [x] #492 
- [x] Borsh to 0.9, which would require releases of VM logic crates (to avoid duplicate deps for tests) #476 #500 https://github.com/near/nearcore/issues/4603 #639 
- [x] Resolution on #263 so this doesn't keep getting pushed back?
- [x] Switch storage keys to `Box<[u8]>`?
- [x] #495 
- [x] #497 
- [x] #498 
- [x] ~~Resolution on #526~~ (API will be changed for next major release)
- [x] #527 
- [x] #521 
- [x] #520
- [x] https://github.com/near/nearcore/pull/4732 (license updates, remove gpl3 requirement from SDK)
- [x] #740 #742 
- [x] https://github.com/near/near-sdk-rs/issues/813
- [x] https://github.com/near/near-sdk-rs/issues/802 https://github.com/near/near-sdk-rs/pull/817

Feel free to add anything else relevant or thoughts on these!

robert-zaremba:
Any ETA for the release?

austinabell:
> Any ETA for the release?

Hard to give a good estimate for ETA, as is it dependant on a few things out of our control. Will try to update here when we have a better idea.

jberrytech:
Sprint planning update: Awaiting resolution for NEP. @austinabell will provide more detail.

jberrytech:
@jberrytech organizing a formal tech review.

PR #515, opened 2021-08-04
fix(env): Update method name parameters in env to string equivalents

austinabell:
- Updates the function signatures from raw bytes for method names to String
  - These method names _must_ be utf8 so this signature will be more ergonomic and harder to misuse.

Was related to something I was working on, and decided to break apart and PR separately

This is a breaking change, so could leave out of 4.0 release (#494) but I think this is incorrect and better to change sooner than later.


HELP.md
```diff
@@ -458,7 +458,7 @@ mod ext_calculator {
         Promise::new(receiver_id.clone())
             .function_call(
                 b"mult",
-                json!({ "a": a, "b": b }).to_string().as_bytes(),
+                json!({ "a": a, "b": b }).to_string(),
```
mikedotexe, line 461: yes, so nice

mikedotexe:
Interesting, any idea why we would do this the other way in the first place?

mikedotexe (approved):

austinabell:
> Interesting, any idea why we would do this the other way in the first place?

My guess, as similar from the `log` and `panic` signatures, that it was done cautiously to avoid any potential code size increases but this is not the case.

Also, going to think through alternatives and consider deprecating and creating a new function tomorrow morning, which is probably the better route than this breaking change.

PR #518, opened 2021-08-05
fix: update ext contract codegen function signature

austinabell:
Another thing I noticed could be fixed from what I was working on and decided to backport because it seems important to get out before the next pre-release. This signature uses `ToString` to take any of the account ids and converts them to `AccountId`s, which at best clones the type, and adds unnecessary generics which can be misused and opens the API to pass in invalid account ids.

Also fixes some doc tests


ChaoticTempest (approved):
Very nice! This looks like it also removes some unnecessary conversions into `String`, then to `&str` and then back to `String` as well

PR #520, opened 2021-08-07
fix: metadata codegen warning

austinabell:
So the metadata macro derives serialize for a duplicate struct to be used to derive metadata and never uses the field which emits this warning:

```bash
    Checking status-message v0.1.0 (/home/austinabell/development/github.com/near/near-sdk-rs/examples/status-message)
warning: field is never read: `message`
  --> src/lib.rs:16:34
   |
16 |     pub fn set_status(&mut self, message: String) {
   |                                  ^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: field is never read: `account_id`
  --> src/lib.rs:22:30
   |
22 |     pub fn get_status(&self, account_id: AccountId) -> Option::<String> {
   |                              ^^^^^^^^^^^^^^^^^^^^^

warning: `status-message` (lib) generated 2 warnings
    Finished dev [unoptimized + debuginfo] target(s) in 0.16s
```

And I finally got annoyed by it so adding this so developers using the metadata API don't get confused with their usage (as the warning is placed incorrectly).

I feel like there should be a better solution that involves not redefining the type and deriving the serialize impls twice, but I just don't have the time to look at it yet, especially since the motivation/value for metadata is unclear to me yet.


mikedotexe:
This is _super_ weird. I've spent a lot of time writing contracts and don't remember ever seeing this before. Poking around now

mikedotexe:
Ohh, it's the `metadata` macro. I seem to remember Eugene briefly saying we may not need this. I wonder if we should include this PR and consider removing it. @evgenykuzyakov thoughts?

mikedotexe (approved):

PR #525, opened 2021-08-10
fix(contract-standards): remove `mint` function from exposed nft standard and trait

### LINKED ISSUE #522: NFT mint not payable
hey, I'm wondering why is NFT `mint` not payable? Since it requires some deposit, why not make it payable?
https://github.com/near/near-sdk-rs/blob/master/near-contract-standards/src/non_fungible_token/core/core_impl.rs#L342

austinabell:
~~Seems like the nft example uses a wrapper definition of mint which is marked as payable, meaning the base impl function is broken. Going to dig a bit deeper into this, but opening for expediency to make sure if wrong that it makes it in for the next release.~~

Hides `mint` function from being exposed and removes it from the `NonFungibleTokenCore` trait, as it is not part of the standard.

Closes #522


austinabell:
Actually, this is somewhat of a bigger issue, the mint function is exposed as part of the API of any contract which uses the standard, when it really shouldn't be. I'm wondering why mint is part of the trait.

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -280,6 +280,65 @@ impl NonFungibleToken {
...
+    ///
+    /// Requirements:
+    /// * Caller must be the `owner_id` set during contract initialization.
+    /// * Caller of the method must attach a deposit of 1 yoctoⓃ for security purposes.
+    /// * If contract is using Metadata extension (by having provided `metadata_prefix` during
+    ///   contract initialization), `token_metadata` must be given.
+    /// * token_id must be unique
+    ///
+    /// Returns the newly minted token
+    pub fn mint(
```
danielwpz, line 295: still need to mark it as payable i think

danielwpz:
Hey, moving mint function out of the core trait looks good to me!
However we still need to mark it as payable to make it callable.

near-contract-standards/src/non_fungible_token/core/core_impl.rs
austinabell: actually, this is doing the opposite, which is not exposing this function for external calls at all. 

Looking at this again, I think this function can just be removed altogether, the mint functionality can still be called through the `NonFungibleToken` as the [NFT example does](https://github.com/near/near-sdk-rs/blob/cf3ef24ee51449b48a17246a5744a714e13f660b/examples/non-fungible-token/nft/src/lib.rs#L98). I was just trying to mitigate breaking changes but I think this just causes more issues for no benefit.

Any objections from removing this function from the NFT trait which is [not part of the standard](https://nomicon.io/Standards/NonFungibleToken/Core.html) and this no longer by default exposes the `mint` function which is currently broken anyway on its current version? cc @mikedotexe

danielwpz: got your point.
still prefer to keep this function, mark it as payable to prevent any breaking changes. But if you guys are ok to remove this function totally I'm good with it.

austinabell: > got your point.
> still prefer to keep this function, mark it as payable (other mechanism shall be used to make it internal), in order to avoid any breaking changes. But if you guys are ok to remove this function totally I'm good with it.

well, the only breaking change is no longer externally exposing a function that was broken and should not have been included in the standard in the first place. You can still use the mint functionality as the example does, too. I'll wait on @mikedotexe's take or maybe @chadoh since he worked on the standard also.

mikedotexe (approved):
Thanks for catching this, I had no idea it was in the `NonFungibleTokenCore`. My memory was that `mint` was only in `lib.rs` and never in the `near-contract-standards` crate at all. Good catch and thanks for the patience.

PR #527, opened 2021-08-10
refactor: remove rust panic calls from examples and SDK code

austinabell:
- Attempts to replace all `expect`, `unwrap`, or `panic!` usage to try to avoid bringing in parts of `std` and including local path data into binary
  - Not all cases can be removed, and some are beneficial for now (to avoid having to format errors and include a message
  - No large gains because the formatting utils are still being pulled in otherwise, but moves in that direction
- Updates `nft_is_approved` signature to not move `self`
- Cleans up some logic with nft standard, because there was a lot of unnecessary unwraps


near-contract-standards/src/fungible_token/core_impl.rs
```diff
@@ -105,8 +105,10 @@ impl FungibleToken {
         let balance = self.internal_unwrap_balance_of(account_id);
         if let Some(new_balance) = balance.checked_add(amount) {
             self.accounts.insert(account_id, &new_balance);
-            self.total_supply =
-                self.total_supply.checked_add(amount).expect("Total supply overflow");
```
matklad, line 109: Needs double checking, but I *think* `.expect` should be pretty cheap, as it doesn't involve formatting machinery.

near-sdk/src/store/lazy_option/mod.rs
```diff
@@ -115,7 +115,7 @@ where
     /// be reloaded.
     pub fn get_mut(&mut self) -> &mut Option<T> {
         self.cache.get_or_init(|| load_and_deserialize(&self.storage_key));
-        let entry = self.cache.get_mut().expect("cell should be filled above");
+        let entry = self.cache.get_mut().unwrap_or_else(|| unreachable!());
```
matklad, line 118: this feels odd: I'd expect `.expect` and `.unwrap_or_esle(|| unreachable!())` to be equivalent in terms of code size: both panic with a specific string.

near-contract-standards/src/fungible_token/core_impl.rs
austinabell: Seems like it does, looking at the [rust src](https://github.com/rust-lang/rust/blob/e8e1b32a7840c07f30c04b252c379a044a73902d/library/core/src/option.rs#L1618). Calls panic which I assume adds line information and also uses formatting syntax (is fmt not used when formatting a `&str`?)

matklad: Ah, right, `.expect` is `#[track_caller]`, so it will include line info!

near-sdk/src/store/lazy_option/mod.rs
austinabell: hmm, I think I was under the false assumption that unreachable in wasm doesn't call `panic!`. I guess in this case it does save some data from not having new custom strings from being put into the wasm binary.

Perhaps better to just have all these calls just call `sys::panic()` internally to remove file info and the static string in most cases?

matklad: >hmm, I think I was under the false assumption that unreachable in wasm doesn't call panic!.

were you thinking about `std::arch::wasm32::unreachable`? `sys::panic()` is also fine.

README.md
```diff
@@ -109,7 +109,7 @@ want to disable default initialization then you can prohibit it like this:
 ```rust
 impl Default for StatusMessage {
     fn default() -> Self {
-        panic!("Contract should be initialized before the usage.")
+        near_sdk::env::panic_str("Contract should be initialized before the usage.")
```
mikedotexe, line 112: good catch, I don't know how I missed this

near-contract-standards/src/fungible_token/core_impl.rs
```diff
@@ -105,8 +105,10 @@ impl FungibleToken {
...
+            self.total_supply = self
+                .total_supply
+                .checked_add(amount)
+                .unwrap_or_else(|| env::panic_str("Total supply overflow"));
```
mikedotexe, line 111: Good pattern for us to highlight in the best practices, I'm making a note

near-contract-standards/src/fungible_token/core_impl.rs
```diff
@@ -209,7 +213,7 @@ impl FungibleToken {
 
         // Get the unused amount from the `ft_on_transfer` call result.
         let unused_amount = match env::promise_result(0) {
-            PromiseResult::NotReady => unreachable!(),
+            PromiseResult::NotReady => env::abort(),
```
mikedotexe, line 216: @evgenykuzyakov I remember you defended how this is supposed to be `unreachable` before. What do you think?

near-sdk/src/environment/env.rs
```diff
@@ -486,7 +489,15 @@ pub fn panic_str(message: &str) -> ! {
 /// Aborts the current contract execution without a custom message.
 /// To include a message, use [`panic_str`].
 pub fn abort() -> ! {
-    unsafe { sys::panic() }
+    // Use wasm32 unreachable call to avoid including the `panic` external function in Wasm.
```
mikedotexe, line 492: okay now I get it

near-contract-standards/src/fungible_token/core_impl.rs
mikedotexe: disregard, i didn't see this farther down: https://github.com/near/near-sdk-rs/pull/527/files#diff-28f92d462ad3e2a8554f6a34935fbf5ba6f7dcfbc615546eff3741d4356adb1bR492

near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
```diff
@@ -563,13 +563,13 @@ mod tests {
...
                     near_sdk::PromiseResult::Successful(x) => x,
-                    _ => panic!("Callback computation {} was not successful", 0u64)
+                    _ => near_sdk::env::panic_str("Callback computation 0 was not successful")
                 };
                 let mut x: u64 = near_sdk::borsh::BorshDeserialize::try_from_slice(&data)
                     .expect("Failed to deserialize callback using Borsh");
                 let data: Vec<u8> = match near_sdk::env::promise_result(1u64) {
                     near_sdk::PromiseResult::Successful(x) => x,
-                    _ => panic!("Callback computation {} was not successful", 1u64)
+                    _ => near_sdk::env::panic_str("Callback computation 1 was not successful")
```
ChaoticTempest, line 572: nice

ChaoticTempest (approved):

mikedotexe (approved):

PR #543, opened 2021-08-18
chore: remove flags script for cargo config

austinabell:
The only part that can't be removed is copying the wasm binary to the `res` directory, which can't be done with cargo config because the `out-dir` option is nightly only. Not quite sure why this copying pattern is necessary to be a standard outside of this repo, though.

Related to https://github.com/near/sdk-docs/issues/13#issuecomment-899645186


willemneal:
Looks good to me and feel free to merge; I have some questions, which could be addressed in a future PR.  Is there a reason that the examples are not part of the workspace? Or their own workspace?  It seems it would make more sense to just have one build script.

Also is there a reason for using the `res` directory? It seems that the minifier could just look `examples/target/wasm32-unknown-unknown/*.wasm`.  I know the sim tests use this directory too, but again seems like an easy edit.

Another issue I've had in the past working with the examples is that since they are not part of the top level `Cargo.toml`, `rust-analyzer` can't find them.  So I have to add their folders individually to my VSCode workspace.

willemneal (approved):

mikedotexe (approved):
yesss! Thank you so much. This also helps our Windows users

austinabell:
> Looks good to me and feel free to merge; I have some questions, which could be addressed in a future PR. Is there a reason that the examples are not part of the workspace? Or their own workspace? It seems it would make more sense to just have one build script.

Don't want to require rebuilding all examples when making changes, these are long build times

> 
> Also is there a reason for using the `res` directory? It seems that the minifier could just look `examples/target/wasm32-unknown-unknown/*.wasm`. I know the sim tests use this directory too, but again seems like an easy edit.

If in the target dir, it will be ignored by git and there was a reason unknown to me as to why the wasm blobs were included in git but now we are using them to track the affected changes for the SDK on the examples

> 
> Another issue I've had in the past working with the examples is that since they are not part of the top level `Cargo.toml`, `rust-analyzer` can't find them. So I have to add their folders individually to my VSCode workspace.

Yeah, it's not ideal, but worst case you can just open a specific example in a new window if you're making non-simple changes and everything works fine

willemneal:
If we can just use cargo then `cargo build -p ...` should allow only building the contract you want to work on. Again this doesn't need to be addressed in this PR, just curious why we don't take advantage of cargo's workspace. Could still copy over the wasm bins.

PR #547, opened 2021-08-20
fix(xcc): return Promise; fix README; make child

chadoh:
Parallel change: https://github.com/near-examples/rust-high-level-cross-contract/pull/73

Previous version relied on caller to correctly format child account name, even though the README did not specify this need. New version automatically creates child account.

Previous version also incorrectly marked all calls to `deploy_status_message` as successful, even if the actual account creation failed (see https://explorer.testnet.near.org/transactions/4kex5jTUHhVkRKXBkwhfuHrBtaWMMJgabsXSUoqkao9h). Returning a Promise fixes this.

This also fixes two docs issues:

* contract accepts `U128`, but README instructed to use a `number`
* README was specifying parent account in `--accountId`


examples/cross-contract-high-level/src/lib.rs
```diff
@@ -41,14 +41,15 @@ pub trait ExtStatusMessage {
 
 #[near_bindgen]
 impl CrossContract {
-    pub fn deploy_status_message(&self, account_id: AccountId, amount: U128) {
-        Promise::new(account_id)
+    pub fn deploy_status_message(&self, account_id: AccountId, amount: U128) -> Promise {
+        let new_account_id = format!("{}.{}", account_id, env::current_account_id())
+        Promise::new(new_account_id)
```
austinabell, line 46:
```suggestion
    pub fn deploy_status_message(&self, mut account_id: AccountId, amount: U128) -> Promise {
        let new_account_id = format!("{}.{}", account_id, env::current_account_id());
        Promise::new(AccountId::new_unchecked(new_account_id))
```

I wonder if it's worth adding convenience functions to prepend or append `AccountId`s separated by a `.` so it doesn't require calling formatting like this.

austinabell (commented):
You'll also need to rebuild this contract if changing like this, I'll add contributing docs, but you'd just need to run `./examples/build_all_docker.sh` when the build is fixed (doesn't currently compile)

examples/cross-contract-high-level/README.md
```diff
@@ -50,10 +50,13 @@ near deploy --accountId=cross_contract --wasmFile=../examples/cross-contract-hig
 
 ### Deploying another contract
 Let's deploy another contract using `cross-contract`, factory-style.
+
 ```bash
-near call cross_contract deploy_status_message "{\"account_id\": \"status_message\", \"amount\":1000000000000000}" --accountId=test_near 
+near call cross_contract deploy_status_message "{\"account_id\": \"status_message\", \"amount\":"10000000000000000000000000"}" --accountId=cross_contract
```
ChaoticTempest, line 55: I believe the added quotes also need a backslack:
```suggestion
near call cross_contract deploy_status_message "{\"account_id\": \"status_message\", \"amount\":\"10000000000000000000000000\"}" --accountId=cross_contract
```

examples/cross-contract-high-level/src/lib.rs
```diff
@@ -41,14 +41,15 @@ pub trait ExtStatusMessage {
 
 #[near_bindgen]
 impl CrossContract {
-    pub fn deploy_status_message(&self, account_id: AccountId, amount: U128) {
-        Promise::new(account_id)
+    pub fn deploy_status_message(&self, account_id: AccountId, amount: U128) -> Promise {
+        let new_account_id = format!("{}.{}", account_id, env::current_account_id())
```
ChaoticTempest, line 45: this line should throw a compiler error since there's no semicolon at the end

examples/cross-contract-high-level/README.md
chadoh: Oops, thanks, I'm too used to having the whole JSON in single quotes

examples/cross-contract-high-level/src/lib.rs
chadoh:
```suggestion
        let new_account_id = format!("{}.{}", account_id, env::current_account_id());
```

examples/cross-contract-high-level/src/lib.rs
chadoh: Hmm, I didn't need `AccountId::new_unchecked` here: https://github.com/near-examples/rust-high-level-cross-contract/pull/73/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759

I compiled and deployed the contract at that link and everything worked.

chadoh:
@austinabell I remove docker from my machine when I run out of disc space – do you mind compiling this for me? I think it will compile now.

ChaoticTempest:
> I remove docker from my machine when I run out of disc space – do you mind compiling this for me? I think it will compile now.

Oh dang. I was trying to add in a separate service to automate this, instead of forcing the dev to manually have to do this every time blobs change. Hopefully I'll get some spare cycles soon to get that up

examples/cross-contract-high-level/src/lib.rs
austinabell: how did you build? a string isn't an `AccountId` and the build fails on my machine. I'll fix and push the wasm change

examples/cross-contract-high-level/src/lib.rs
```diff
@@ -41,14 +41,15 @@ pub trait ExtStatusMessage {
 
 #[near_bindgen]
 impl CrossContract {
-    pub fn deploy_status_message(&self, account_id: AccountId, amount: U128) {
-        Promise::new(account_id)
+    pub fn deploy_status_message(&self, account_id: AccountId, amount: U128) -> Promise {
+        let new_account_id = format!("{}.{}", account_id, env::current_account_id());
```
willemneal, line 45: Seems like a good sdk function.

examples/cross-contract-high-level/src/lib.rs
```diff
@@ -41,14 +41,15 @@ pub trait ExtStatusMessage {
...
+        Promise::new(AccountId::new_unchecked(new_account_id))
```
willemneal, line 46: Why unchecked?

examples/cross-contract-high-level/README.md
```diff
@@ -50,10 +50,13 @@ near deploy --accountId=cross_contract --wasmFile=../examples/cross-contract-hig
 
 ### Deploying another contract
 Let's deploy another contract using `cross-contract`, factory-style.
+
 ```bash
-near call cross_contract deploy_status_message "{\"account_id\": \"status_message\", \"amount\":1000000000000000}" --accountId=test_near 
+near call cross_contract deploy_status_message "{\"account_id\": \"status_message\", \"amount\":\"10000000000000000000000000\"}" --accountId=cross_contract
 ```
 
+This creates a child account, `status_message.cross_contract`, and transfers 10 $NEAR to it (that very long number above is a 10 followed by 24 zeroes, which is the number of yoctoNEAR to transfer; it's a number so big it goes beyond the JSON number serialization limit of 2^53 and must use strings).
```
willemneal, line 58:
```suggestion
This creates a sub-account, `status_message.cross_contract`, and transfers 10 $NEAR to it (the long number above is a 1e25, which is the number of yoctoNEAR to transfer; it's larger than JS's JSON number max size of 2^53 and so it must serialized as a string).
```

austinabell:
Outdated, this has since been updated. @itegulov you might want to double-check if I've missed something

PR #549, opened 2021-08-20
chore: add contributing docs

austinabell:
Intentionally keeping it simple and we can add any other info when necessary. Does anyone have suggestions on anything I missed or important to note?


CONTRIBUTING.md
```diff
@@ -0,0 +1,29 @@
+# Contributing to near-sdk-rs
```
ChaoticTempest, line 1: maybe we can also add a comment saying to look at `good first issues` to get people's feet wet and learn about the codebase?

ChaoticTempest (approved):
Awesome

CONTRIBUTING.md
austinabell: I personally think that this is intuitive and from other contributing files I looked at I never saw this, but feel free to add a point in another PR, just going to pull this in so we have some info on contrib asap

PR #552, opened 2021-08-26
Approval should pass U64 in callback to external contract

mattlockyer:
This fixes a breaking implementation where u64 was passed instead of U64.

There are already marketplaces and NFT contracts that routinely pass U64 for `nft_on_approve` so this must be fixed in the standard implementation.

While the initial intent was to have approval IDs always be numbers/u64 because

they're never expected to exceed the JSON limit of 2^53 and
we may as well therefore save some (minimal?) serialization/deserialization costs,
at this point many applications in the wild expect to receive approval_id: U64 [also, please link to such deployed apps]. Rather than attempting to change all of these deployed contracts, we believe we should instead update the standard because:

it is the more backward-compatible change – it is possible to go from strings to numbers but not numbers to strings;
it is more consistent – in general, we could always suggest using U64/U128 for arguments, even if they theoretically will never be that large, so that developers don't have to think about which one to use;
the gas savings [can we quantify this?] do not justify the costs mentioned in points 1-2 above.

TODO please test.

TODO might want to check `nft_is_approved` as well, consider U64. Not sure.

### DIFF
--- a/near-contract-standards/src/non_fungible_token/approval/approval_impl.rs
+++ b/near-contract-standards/src/non_fungible_token/approval/approval_impl.rs
@@ -73,7 +73,7 @@ impl NonFungibleTokenApproval for NonFungibleToken {
             ext_approval_receiver::nft_on_approve(
                 token_id,
                 owner_id,
-                approval_id: approval_id.into(),
+                approval_id.into(),
                 msg,
                 account_id,
                 NO_DEPOSIT,
--- a/near-contract-standards/src/non_fungible_token/approval/approval_impl.rs
+++ b/near-contract-standards/src/non_fungible_token/approval/approval_impl.rs
@@ -8,6 +8,7 @@ use crate::non_fungible_token::utils::{
 };
 use crate::non_fungible_token::NonFungibleToken;
 use near_sdk::{assert_one_yocto, env, ext_contract, require, AccountId, Balance, Gas, Promise};
+use near_sdk::json_types::{U64};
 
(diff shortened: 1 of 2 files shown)


mattlockyer:
Now that Mintbase doesn't care about migrations I'm fine with the current standard. We'll have to talk to other partners like Collab.Land and have them redeploy all NFT contracts to work with current market implementations.

Pluminite will also need everything redeployed.

mikedotexe:
> Now that Mintbase doesn't care about migrations I'm fine with the current standard. We'll have to talk to other partners like Collab.Land and have them redeploy all NFT contracts to work with current market implementations.
> 
> Pluminite will also need everything redeployed.

Thanks for the update. Yes, all projects that implemented the standard incorrectly will have fix their mistake, unfortunately.
We can consider a major version release of a standard, but I do not see any evidence of that approach, so will have to assume there are reasons not to do that. Seeing as this thread may become lengthy, there is a special link in Discord where we can have a more free-form conversation about this.

mattlockyer:
I'm seeing some promising results for redeployment here: https://github.com/near-apps/nft-market/commit/2814e5f5d899006471abc2c1aa12fa63b09ef07e

With this diff, I've tested re-deploying both NFT and Market contracts to the same accounts and running my test suite on testnet (many approvals and sales). There appears to not be any adverse behaviour.

I'll do a deeper dive tomorow.

mikedotexe:
Issue moved here: https://github.com/near/NEPs/pull/257

PR #554, opened 2021-08-27
feat: implement callback_result annotation

### LINKED ISSUE #155: Differentiate successful and failed callback
We need for a function to differentiate result of a successful and failed callback. Currently, when we declare function that expects a callback like this:
```rust
pub fn callback_receiver(&self, #[callback] arg0: A, arg1: B, #[callback] arg2: C) {
}
```
It will fail if execution of whatever produced `arg0` or `arg2` failed. Instead it should be:
```rust
pub fn callback_receiver(&self, #[callback_result] arg0: Result<A>, arg1: B, #[callback_result] arg2: Result<C>) {
// Code that handles cases of arg0 and arg2 errors.
}
```

austinabell:
- Implements a `callback_result` annotation which allows handling errors of promise callbacks
- Adds `callback_unwrap` annotation to replace `callback`
  - Given the way that the code is set up, I don't think there is a way to deprecate this `callback` annotation, but would be happy if someone proves me wrong
- Adds a new example that uses this and handles some use cases not covered in other examples (like stateless methods)

Closes #155 

Optionally can add a `CallbackResult` type which is just
```rust
type CallbackResult<T> = Result<T, PromiseError>;
```

But it's a bit awkward because it really should be `PromiseResult`, but that type is already built into the current API and is a custom enum


ChaoticTempest:
just making sure, but would we need to add in `callback_unwrap` and `callback_result` to `near-sdk-macros/lib.rs`? I see there's `callback` and `callback_vec` defined there as marker attributes, but I'm not how these relate at all since everything should be going thru the `near_bindgen` macro itself

austinabell:
> just making sure, but would we need to add in `callback_unwrap` and `callback_result` to `near-sdk-macros/lib.rs`? I see there's `callback` and `callback_vec` defined there as marker attributes, but I'm not how these relate at all since everything should be going thru the `near_bindgen` macro itself

No, that was an artifact of a previous version. These are no longer used anymore, as the annotations now are filtered out of the macro output. I just didn't remove the old ones yet since I'm not sure the implications and didn't want to group that change with this.

ChaoticTempest (approved):
this looks good to me so far, but might want another set of eyes on this just in case I overlooked anything

austinabell:
> this looks good to me so far, but might want another set of eyes on this just in case I overlooked anything

Yeah, I'm in no rush to get this in, so will leave until it has another review. It is a little more involved of a PR so this is definitely a good candidate for more reviews

mikedotexe (approved):

PR #556, opened 2021-08-31
setup build image for contract builder

ilyar:
example result https://github.com/ilyar/near-sdk-rs/pkgs/container/near-sdk-rs%2Fnear-contract-builder

### DIFF
--- a/.github/workflows/contract-builder.yml
+++ b/.github/workflows/contract-builder.yml
@@ -1,0 +1,29 @@
+name: Contract builder
+on:
+  pull_request:
+    paths:
+      - 'contract-builder/**'
+    branches:
+      - master
+  push:
+    paths:
+      - 'contract-builder/**'
+    branches:
+      - master
+
+jobs:
+  image:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: docker/login-action@v1
+        with:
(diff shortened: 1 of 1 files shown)


austinabell:
I'll leave it to others to approve if this is helpful, as it's unclear how exactly this helps. By the way, there is an image set up already which is used in this repo (https://hub.docker.com/r/nearprotocol/contract-builder) not sure if it was not visible. 

Not sure in the value in having this connected to CI as this never really changes and we might want the control.

ilyar:
@austinabell I didn't know that this already existed, I can't even imagine that you are using manual or private deployment processes

austinabell:
> @austinabell I didn't know that this already existed, I can't even imagine that you are using manual or private deployment processes

I haven't touched any of the docker setup other than using the image, but @chefsale input on this would be appreciated.

ilyar:
I see https://github.com/near/near-sdk-rs/tree/master/contract-builder and https://github.com/skyward-finance/contracts/blob/master/build_all.sh

but it should be something like this https://gist.github.com/ilyar/6d4fa5afd432532bd5170dc4de40e92c

austinabell:
@rtsai123 @anshal-savla relevant discussion to updating the docker image. Something to consider

austinabell:
#751

PR #557, opened 2021-08-31
fix: update atomic register to u64::MAX-2 to avoid collisions

austinabell:
When using `env` and `sys` calls within the same method, it feels like a foot gun to have the register used for temporary operations to be 0, as most examples and usages I've seen of syscalls use this register.


ChaoticTempest:
How does this register work usually? I only see it being used for reads so I'm not sure if changing it has any effect besides reading from a different location. Or does the read do something special like create a lock?

austinabell:
> How does this register work usually? I only see it being used for reads so I'm not sure if changing it has any effect besides reading from a different location. Or does the read do something special like create a lock?

Nah, this register is just used as a temporary buffer to be used within the same method call. This changes nothing functionally. The only negative is that someone could potentially be using `u64::MAX-2` register with one of these calls sandwiched in between and it would break, but I think this is unreasonably unlikely, and much more of a common pattern for people to use register 0

ChaoticTempest (approved):
LGTM

PR #565, opened 2021-09-10
feat: add helper macro to prepare arguments

ilyar:
Helper macro to prepare arguments `env::arguments`.

 ```rust
use near_sdk::Gas;
use near_sdk::env;
use near_sdk::arguments;
use near_sdk::AccountId;

fn main() {
   env::promise_create(
      AccountId::new_unchecked("ft.near".to_string()),
      "storage_deposit",
      arguments!({
         "account_id": AccountId::new_unchecked("alice.near".to_string()),
      }),
      1_250_000_000_000_000_000_000,
      Gas(5_000_000_000_000),
   );
}
```

### DIFF
--- a/near-sdk/src/utils/mod.rs
+++ b/near-sdk/src/utils/mod.rs
@@ -11,6 +11,37 @@ pub(crate) use cache_entry::{CacheEntry, EntryState};
 
 use crate::{env, AccountId, PromiseResult};
 
+/// Helper macro to prepare arguments [`env::arguments`].
+///
+/// # Example use
+///
+/// ```no_run
+/// use near_sdk::Gas;
+/// use near_sdk::env;
+/// use near_sdk::arguments;
+/// use near_sdk::AccountId;
+///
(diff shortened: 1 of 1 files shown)


austinabell (changes requested):
I'm very hesitant to add this to the public API, and I'll provide some reasons:
- arguments is a very ambiguous name, as it doesn't specify that this is JSON (contracts do not need JSON parameters)
- This macro does not save much code and seems to be less intuitive of what's going on
- We want to move away from dependence on `serde` and `serde_json` and this would just make migrating harder in the future
- Probably better to make the `ext_contract` API more friendly if it doesn't handle any use cases, so this is not needed at all

I'd love to get others' thoughts on this though @mikedotexe @ChaoticTempest

austinabell:
Sorry, this isn't an API that makes a lot of sense to maintain within the SDK for the reasons listed above.

If you have an idea of how this can be progressed, please feel free to re-open :)

PR #566, opened 2021-09-10
chore: add docs for insert method of collections

ilyar:


austinabell (approved):
Not sure these tests increase coverage, but good to have nonetheless. Thanks!

ilyar:
> Not sure these tests increase coverage, but good to have nonetheless. Thanks!

these tests are not for increasing coverage, but for documentation, for me the behavior of these methods was not obvious, I think the reason is inattentive reading of the documentation. For me tests, an alternative version of the documentation is the code that makes itself known.

PR #569, opened 2021-09-16
feat: add near-workspaces-ava in non-intrusive way

chadoh:
Keeps [near-sdk-sim](https://github.com/near/near-sdk-rs/tree/master/near-sdk-sim) tests, but also adds equivalent [near-workspaces-ava](https://github.com/near/runner-js) tests.

Yes, this means adding some JavaScript to this codebase. But this is done in a way so that no one with the codebase on their machine is required to run it or have Node installed:

- The `test_all.sh` script has a new `--with-js` flag for people who want to opt-in to the JS tests
- The JS setup is minimal, consisting of only a couple new files in `examples` and new `__tests__` directories which can be easily ignored
- These new tests will be run in a new GitHub Action test, so even if individual contributors don't want to run them locally, they will still be guaranteed to be up-to-date

### Rust version coming soon

[near-workspaces-rs](https://github.com/near/runner-rs) is under active development, but has been moving more slowly since it did not already have an equivalent of near-api-js to build from. 

Once it exists, we will be able to fully deprecate the old simulation tests. At that point, we may also decide to get rid of the JS tests, though it could still be useful to keep them for newbies to learn from, since the examples here become the front door for NEAR newbies at [examples.near.org](https://examples.near.org) / [near.dev](https://near.dev).

### Why should we do this?

Why pollute a Rust library with JS tests?

Because the examples are the front door to NEAR. New developers use them all the time to learn about NEAR. They get copied to the github.com/near-examples org, which then show up at [near.dev](https://near.dev), such as https://near.dev/NFT

It is useful for developers to see all testing options.

Soon enough, we will deprecate simulation tests. If people want to stick to Rust-only, they can still use the sim tests to do so. If they don't care so much, they may prefer to use JS. They will appreciate being able to compare both approaches.


ci-test.sh
```diff
@@ -31,8 +31,8 @@ else
 
     ./examples/build_all_docker.sh --check
     echo "Testing all examples"
-    ./examples/test_all.sh
-    # TODO re-enable when parsing unknown sections is fixed in twiggy 
+    ./examples/test_all.sh --no-js
```
chadoh, line 34: If we add JS support to buildkite, we can run near-runner tests there, too. I think this is probably low priority, given that I added a dedicated GitHub action to run the near-runner tests.

austinabell: Should this maybe be the opposite flag, to require enabling it to run js tests? Would be best to not require contributors to have node setup or know about this flag to disable

examples/cross-contract-high-level/src/lib.rs
```diff
@@ -1,23 +1,24 @@
...
-    json_types::U128,
-    log,
-    near_bindgen,
-    AccountId,
-    Promise,
-    PromiseOrValue,
+    env, ext_contract, json_types::U128, log, near_bindgen, AccountId, Gas, Promise, PromiseOrValue,
 };
 
+const TGAS: u64 = 1_000_000_000_000;
```
austinabell, line 6: optionally this can be `Gas` type, and it has a `Mul<u64>` impl so usage below won't break

examples/cross-contract-high-level/src/lib.rs
```diff
@@ -1,23 +1,24 @@
...
 #[near_bindgen]
 #[derive(Default, BorshDeserialize, BorshSerialize)]
 pub struct CrossContract {}
 
 // One can provide a name, e.g. `ext` to use for generated methods.
 #[ext_contract(ext)]
 pub trait ExtCrossContract {
-    fn merge_sort(&self, arr: Vec<u8>) -> PromiseOrValue<Vec<u8>>;
+    fn internal_merge_sort(&self, arr: Vec<u8>) -> PromiseOrValue<Vec<u8>>;
+    fn finalize_merge_sort(
```
austinabell, line 16: Can you explain why the change? is it so that you can call it with JSON params? Can you still not use borsh with runner? This example was already pretty unreadable, and this makes it even harder to follow imo

examples/cross-contract-low-level/src/lib.rs
```diff
@@ -1,10 +1,11 @@
 use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
 use near_sdk::json_types::U128;
 use near_sdk::serde_json::{self, json};
-use near_sdk::{env, near_bindgen, require, AccountId, Gas, PromiseResult};
+use near_sdk::{env, log, near_bindgen, require, AccountId, Gas, PromiseResult};
 
 // Prepaid gas for making a single simple call.
-const SINGLE_CALL_GAS: Gas = Gas(200000000000000);
+const SINGLE_CALL_GAS: Gas = Gas(10_000_000_000_000);
+const TGAS: u64 = 1_000_000_000_000;
```
austinabell, line 8: Same here as above, this can be `Gas` type

examples/cross-contract-low-level/src/lib.rs
```diff
@@ -42,24 +46,34 @@ impl CrossContract {
         let arr0 = arr[..pivot].to_vec();
         let arr1 = arr[pivot..].to_vec();
         let account_id = env::current_account_id();
-        let prepaid_gas = env::prepaid_gas();
+        let gas_to_pass = Gas(match pivot {
+            1 => 1 * TGAS,
+            2 => 40 * TGAS,
+            // TODO: make work with input arrays of length 5, maybe 6
```
austinabell, line 52: Why changing to static gas for these? Will it not work if you make it proportional to the amount of prepaid gas (like it was before) and just passing in more gas for each call? Or does it hit some other limit?

austinabell (commented):
you can also remove `rlib` from all the crates you removed sim tests for! I don't know why some of the wasm files increased in size (other than the ones you changed), did you build without docker?

examples/cross-contract-low-level/src/lib.rs
chadoh: It took some work to get this contract to work with real gas values. If you look at the old simulation tests, you'll see that they only passed because the set gas price to 0 and max attached gas to ~10<sup>19</sup>. 

While figuring out realistic gas values, it became clear that the last steps in the recursion, arrays of length 1, require barely any gas. We were reserving too much gas for this step, using the simplistic previous formula, which cause earlier steps in the recursion to easily run into gas limits. Walking through a concrete example:

* With array of length 4 and 160Tgas, 
* you end up with 40Tgas passed to the depth-2 calls to `merge_sort`,
* which then try to pass 10Tgas each to the last calls to `merge_sort` and `merge`, when these only needed 1Tgas, 
* and just queueing all those cross-contract calls in the depth-2 calls to `merge_sort` requires 29Tgas, 
* so the whole thing fails

You can [see these gas values in Explorer](https://explorer.testnet.near.org/transactions/431CmS7pXjPtbvHD8shPSS5QH59rQhsV9LKsqzGP1V7S).

If we had a third number here, we might be able to determine a better formula. But with only two numbers in the sequence, it's hard to predict what the formula should be.

examples/cross-contract-high-level/src/lib.rs
chadoh: I started with a `Gas` type, but the arithmetic got somewhat complicated. I ended up with calculations like:

```rs
Gas(2 * TGAS.0)
```

Is there a better way to do that?

ci-test.sh
chadoh: I was thinking that `test_all` should run all the tests by default, but I'm ok making it run only Rust tests unless told otherwise.

examples/cross-contract-high-level/src/lib.rs
willemneal: The low level uses JSON and so with this change they could both have the same API and thus use the same tests.  We also discussed making a future PR that breaks up the cross contract example contracts into multiple ones that focus on just one aspect and not all in one.

chadoh: runner-js CAN work with borsh, but it's very ugly, and will be until we have an Application Contract Interface. 

Since this contract had been using borsh serialization, we wanted to keep that, but for the usability of both the tests and as a general example of the pattern, Willem and I thought it might be best to make it use JSON for external interfaces while saving gas by using Borsh for internal calls.

However, after completing this, we noticed that `cross-contract-low-level` does not use any Borsh, and I copy-pasted the same tests. So I know at this point that we could just remove Borsh from this example and have it work.

Honestly, I want to do a lot of rewriting of these examples:

* Make a "Factory Pattern" example, with a high-level and low-level implementation, that does the `status_message.wasm` deploy to subaccount. The same exact tests can run on each implementation, as a stronger guarantee of feature parity between the implementations (no borsh-on-one, not-on-the-other).
* Make a separate "Merge Sort" example, with a high-level and low-level implementation. Same thing with using the same tests for each implementation.
* Add good documentation for each of these, explaining what they demonstrate and why it's cool, rather than assuming that this is self-evident.

Given that we plan to give some love to all of these examples in the near future, we thought it was best to not change them too much right now. And the JSON-external, Borsh-internal pattern is a useful one, as is the `#[callback_unwrap] #[serializer(borsh)]`. We will probably want to keep these features in the reworked examples and just explain them better.

austinabell: All I'm saying is that it's very uncomfortable to change the example just to fit the test, especially since it just adds complexity.

I'm fine with this if we open an issue to make sure fixing and updating this doesn't slip under the rug though

examples/cross-contract-high-level/src/lib.rs
austinabell: https://github.com/near/near-sdk-rs/commit/6bba3327c2e0439a3a2444819c35074c98dbe126 (it's a little finicky because of how the multiplication trait is setup in rust so the ordering matters)

chadoh: Thanks!

examples/cross-contract-high-level/src/lib.rs
chadoh: I removed all Borsh from cross-contract-high-level in dbc265a4f79232703d3225282ff7326f486f9eca and added two tracking issues for the examples reorganization, #589 & #590.

ci-test.sh
austinabell: as long as the js tests still run in the CI, yeah. Can make a note in `CONTRIBUTING.md` if you'd like, doesn't really matter though

chadoh: I added help output to the `test_all.sh` script. It's right at the top of the file to make it discoverable whether people try to run the file with `-h`/`--help` or if they read the file. Do you think it's worth also adding info about it to CONTRIBUTING.md?

nearmax (changes requested):
Rust developers should be able to develop&test Rust smart contracts without learning JS. Similarly to how web-developers should be able to develop&test AS smart contracts without learning Rust.

Currently, we have community members and partners who have Rust contracts and the corresponding tests written in Rust. Which includes repositories like https://github.com/near/core-contracts This change will force them to migrate to JS tests. One of the goals set for the testing framework was to make transition for existing projects as smooth as possible. Forcing them to rewrite Rust tests into JS is not a smooth transition.

Please, deprecate near-sdk-sim with https://github.com/near/runner-rs instead and keep this repo to Rust only. Read our long-term product strategy here: https://near.zulipchat.com/#narrow/stream/295801-dev-developer-platform/topic/Separation.20of.20languages/near/255031261

frol (commented):
Even though this PR only cleans up the examples, I feel it is a premature change. If we feel runner-js is a great way to write the tests, lets create a separate example and feature it, compare it to sim testing, and tune people for runner-rs :pray:

examples/cross-contract-high-level/src/lib.rs
```diff
@@ -20,12 +15,8 @@ pub trait ExtCrossContract {
     fn merge_sort(&self, arr: Vec<u8>) -> PromiseOrValue<Vec<u8>>;
     fn merge(
         &self,
-        #[callback_unwrap]
-        #[serializer(borsh)]
-        data0: Vec<u8>,
-        #[callback_unwrap]
-        #[serializer(borsh)]
-        data1: Vec<u8>,
```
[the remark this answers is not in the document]
willemneal: That was our thought too.  It also doesn't have parity with the other example since it uses borsh for the external interface.  We had initial rewritten to use a JSON method that then called the inner merge, but found that to be way to confusing.

These two examples in general try to combine too many concepts: Factory pattern, cross contract calls, callback_unwrap, and borsh.  They should be combined into one project that provides examples for each at higher and lower levels.  https://github.com/near-examples/cross-contract-calls tries to do this for assemblyscript with sim tests included.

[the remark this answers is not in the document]
austinabell: related https://github.com/near/near-sdk-rs/issues/589 https://github.com/near/near-sdk-rs/issues/590

examples/cross-contract-low-level/src/lib.rs
nearmax: @austinabell FYI. I don't think we should be modifying merge sort example like that for several reasons:
* An example that only works for array of length 4 is not really a good example. It undermines the whole point that shows that NEAR can work with deeply nested and branching function calls. We use this example in the workshops to impress people, e.g. when we give them assignment to write MapReduce on NEAR;
* This new added logic makes the code more complex which is not good, since the examples are supposed to be as simple as possible. Besides, I am not sure the following does anything if we already limit array size to 4:
```rust
1 => TGAS * 1,
2 => TGAS * 40,
```

Base function call cost is ~5Tgas. Callback cost is ~10Tgas (once @Longarithm 's protocol change gets into Mainnet it will be negligible).

So we should be able to do arrays of size ~16 even with the current costs, if Wasm operation cost is negligible (5+20+20*2+20*4+20*8 ~=300Tgas). We can significantly reduce the cost of Wasm execution itself by using Borsh serialization for all arguments and return types, and if we need to have JSON interface to call the contract from near-cli then we can separate `merge_sort` into `merge_sort` and `internal_merge_sort`.

chadoh:
@austinabell I've reassigned to you and unsubscribed from notifications ✌🏼

austinabell:
Closing because we are replacing all with workspaces-rs

PR #572, opened 2021-09-18
feat: NFT mint unguardedly

danielwpz:
The existing `mint` function requires that the caller must be the owner.

However this is just an assumption of how the users might use this contract, in fact there might be (and already has) some cases in which other users can also mint NFTs as long as some requirements are met.

The existing `mint` function is kept to make this contract backwards compatible. I added one `internal_mint_unguarded` function for derived contracts to bypass the owner restriction.


willemneal (approved):
Thank you! This a great first step in making the minting process more customizable and standardizable.

austinabell (approved):
Awesome, was planning on doing exactly this tomorrow. Thanks for the contribution

austinabell:
cc @mikedotexe since you're most aware of NFT standard and how it will be changing.

As for CI failing, `cargo fmt` is failing and the NFT example needs to be rebuilt (see `CONTRIBUTING.md` for info on this).

If you'd like I can fix this if I have permissions to push to this PR branch

ilblackdragon:
I suggest we use `internal_` as indicator of everything that doesn't do any checks. This is how it is in Fungible Token.
So we don't need extra "_unguaded" then.

I would then remove `mint` completely. It doesn't belong in the library anyway.

austinabell:
> I would then remove `mint` completely. It doesn't belong in the library anyway.

Can just mark the function as deprecated in this PR, and this will be removed later (avoid hard breaking change)

mikedotexe (approved):

danielwpz:
@ilblackdragon @mikedotexe updated

Issue #559, opened 2021-09-01
The hardcoded `GAS_FOR_FT_TRANSFER_CALL` and `GAS_FOR_RESOLVE_TRANSFER ` in NFT contract doesn't work well sometimes


think-in-universe:
The default GAS configuration seems not working well sometimes for `nft_transfer_call`.

https://github.com/near/near-sdk-rs/blob/bcdf5babf9cea4aef94457767735dc9195d84798/near-contract-standards/src/non_fungible_token/core/core_impl.rs#L17-L18

For example, this project failed when calling `nft_transfer_call`: https://explorer.testnet.near.org/transactions/bvTaJHMDYpmvLwDNeXEYEg8WnJaVLQGTGvCLcy4PRes 

So they updated the configs to make it work for their project.

https://github.com/web3gamesofficial/near-sdk-rs/blob/master/near-contract-standards/src/non_fungible_token/core/core_impl.rs#L18-L19

Questions: 

1. How are the default gas configuration defined? 
2. Does it work in most cases? 
3. Should we update the constants?

PR #573, opened 2021-09-18
fix: Increase GAS for NFT transfer call

danielwpz:
current value is too low to even run in testnet


think-in-universe:
the related issue is #559

danielwpz:
@mikedotexe please help review this as well

mikedotexe:
Thanks @danielwpz seems like there is more discussion happening in #593 do you want to keep this issue? I've approved the other one and moved it into the "Awaiting Review" column for Dev Platform

danielwpz:
@mikedotexe hi mike, this one and #593 are very similar, I think both places need to be updated

danielwpz:
sure! will do @mikedotexe

mikedotexe:
@danielwpz does my comment in the other issue make sense? Do you think this is still valid to keep open?

austinabell:
Closing just because this is a stop-gap solution and we should soon come up with a better pattern for these.

cc @jlogelin input on this and the discussion on #593 would be appreciated to spark this issue again, as it's a real problem

PR #576, opened 2021-09-20
test(vector): Add fuzz test for Vector

austinabell:
I'm in progress with refactoring the internals of this but figured I could PR this separately to make the transition cleaner

### DIFF
--- a/near-sdk/src/store/vec/mod.rs
+++ b/near-sdk/src/store/vec/mod.rs
@@ -295,6 +295,7 @@ where
 #[cfg(test)]
 mod tests {
     use arbitrary::{Arbitrary, Unstructured};
+    use borsh::{BorshDeserialize, BorshSerialize};
     use rand::{Rng, RngCore, SeedableRng};
 
     use super::Vector;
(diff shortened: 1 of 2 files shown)


near-sdk/src/store/vec/mod.rs
```diff
@@ -253,6 +257,7 @@ where
     /// # Panics
     ///
     /// If `index` is out of bounds.
+    // TODO determine if this should be stabilized, included for backwards compat with old version
     pub fn replace(&mut self, index: u32, element: T) -> T {
```
ChaoticTempest, line 261: I'm surprised `Vec` doesn't have a `replace` function. It doesn't seem like there's any caveats to having such a thing, but forces people to do something like:
```rust
let old_val = std::mem::replace(&mut v[3], new_val);
```

ChaoticTempest (approved):

near-sdk/src/store/vec/mod.rs
austinabell: yeah, the convenience is why I left it, but the method isn't needed

PR #577, opened 2021-09-21
refactor(store): Move internal vec functionality into index map

austinabell:
Refactoring out this functionality to build other primitives upon (want something other than `Vector` for `UnorderedMap` to optimize the number of storage reads/writes)

### DIFF
--- a/near-sdk/src/store/index_map.rs
+++ b/near-sdk/src/store/index_map.rs
@@ -1,0 +1,156 @@
+use std::fmt;
+
+use borsh::{BorshDeserialize, BorshSerialize};
+use once_cell::unsync::OnceCell;
+
+use crate::collections::append_slice;
+use crate::utils::StableMap;
+use crate::{env, CacheEntry, EntryState, IntoStorageKey};
+
+const ERR_ELEMENT_DESERIALIZATION: &str = "Cannot deserialize element";
+const ERR_ELEMENT_SERIALIZATION: &str = "Cannot serialize element";
+
+#[derive(BorshSerialize, BorshDeserialize)]
+pub(crate) struct IndexMap<T>
+where
+    T: BorshSerialize,
+{
+    pub(crate) prefix: Box<[u8]>,
+    /// Cache for loads and intermediate changes to the underlying index map.
+    /// The cached entries are wrapped in a [`Box`] to avoid existing pointers from being
+    /// invalidated.
+    #[borsh_skip]
+    pub(crate) cache: StableMap<u32, OnceCell<CacheEntry<T>>>,
+}
+
+impl<T> IndexMap<T>
+where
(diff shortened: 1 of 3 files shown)


austinabell:
If anyone is looking at this, hold off on reviewing. I'm second-guessing the need to create a parallel structure to `Vector` and might just make this fix the other things mentioned in the PR if so.

austinabell:
There isn't any negative I can think of with having this logic separate, but it ended up not being necessary. Getting thoughts from others, does anyone have an opinion if this should be condensed again, or is the possibility of using the core index map functionality outside of this enough to keep it like this?

near-sdk/src/store/index_map.rs
```diff
@@ -0,0 +1,152 @@
...
+}
+
+impl<T> fmt::Debug for IndexMap<T>
+where
+    T: BorshSerialize + BorshDeserialize + fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("IndexMap").field("prefix", &self.prefix).finish()
+    }
+}
```
ChaoticTempest, line 152: Probably best to have some tests for `IndexMap`, even if `Vector` already covers it at a higher level, just so we can see some of the basic features it can cover

near-sdk/src/store/index_map.rs
```diff
@@ -0,0 +1,152 @@
...
+pub(crate) struct IndexMap<T>
+where
+    T: BorshSerialize,
+{
+    pub(crate) prefix: Box<[u8]>,
+    /// Cache for loads and intermediate changes to the underlying index map.
+    /// The cached entries are wrapped in a [`Box`] to avoid existing pointers from being
+    /// invalidated.
+    #[borsh_skip]
+    pub(crate) cache: StableMap<u32, OnceCell<CacheEntry<T>>>,
```
ChaoticTempest, line 22: I think its best to also mention why we're using `u32` in the comments somewhere over `usize`. I remember it being addressed in one of the older PRs but don't know where it was placed

ChaoticTempest (approved):
So far, I don't see any cons to having this be separate from `Vector`. It makes `Vector`'s code a lot clearer having this be a separate abstraction, so I think its a plus to have it.

Actually, one downside is having too many of these collections just be a very thin wrapper on top of another one. But in this case, I think this one is fine since it alleviates some hassle with handling prefixes. `StableMap` on the other hand is probably the thin wrapper

near-sdk/src/store/index_map.rs
austinabell: yeah, you're definitely right I should, I was just being lazy because it's an internal type. Adding now

PR #581, opened 2021-09-22
Push forward #572

mikedotexe:
Addresses updates to #572 to help move that along.

Perhaps we can improve our contributing guide to include the new building of contracts using the `./build_all_docker.sh` script and using `cargo fmt` to correct things.


austinabell:
can you also mark the mint function as deprecated as of 4.0? As indicated we want to move away from this being in the standard, but allow users to still use it for now

danielwpz:
hey, @mikedotexe could please rebase my branch? I made some updates according to @austinabell comment.
https://github.com/near/near-sdk-rs/pull/572

ilblackdragon:
> can you also mark the mint function as deprecated as of 4.0? As indicated we want to move away from this being in the standard, but allow users to still use it for now

Not related to the PR, but I would really ask to be clear when talking about these things:
 - standards live in https://nomicon.io/Standards/README.html
 - in SDK you have library of components that allow people to build compatible with standard contracts. There can be many libraries that provide this in different ways. The library just provides building blocks.

mikedotexe:
This PR was an attempt to move things along, but then I didn't address everything and Daniel buttoned up the original PR (https://github.com/near/near-sdk-rs/pull/572) very quickly. That has been merged and hence I'm going to close this PR

PR #591, opened 2021-09-24
Add AccessControl library for NEAR (similar with OpenZeppelin for Solidity)

ctindogaru:
This PR aims to add a new library for restricting access to certain functions in the contract.


austinabell:
this also seems stale, if this is in progress, feel free to re-open. Also not clear where this standard is coming from. Is this one being iterated on now?

PR #592, opened 2021-09-24
feat(vector): Implement drain iterator for Vector

austinabell:
motivation: https://github.com/near/near-sdk-rs/pull/584#issuecomment-926146144

As I started to get into it, I was thinking about abandoning for a simpler solution outside of the vector, but the more I thought about it the more I realized this could be very useful for users. There isn't really a clean way to do this otherwise because we only have the ability to swap remove, which if used for multiple elements will not preserve ordering.

This PR is built on #577 to handle merge conflicts better

The only difference from std::vec drain functionality is that they panic on an invalid range, we will just cap the max at length or have an empty iterator if invalid. We can match if we want, I just didn't see a reason to panic explicitly like this.


near-sdk/src/store/vec/iter.rs
```diff
@@ -156,3 +156,113 @@ where
         Some(self.get_mut(idx).unwrap_or_else(|| env::panic_str(ERR_INDEX_OUT_OF_BOUNDS)))
     }
 }
+
+/// A draining iterator for [`Vector<T>`].
+#[derive(Debug)]
+pub struct Drain<'a, T>
+where
+    T: BorshSerialize + BorshDeserialize,
```
ChaoticTempest, line 164: just to note, but what's your opinion on getting rid of these trait bound declarations on the struct itself? I see that `Vector` and subsequently other types have trait bounds underneath for `Borsh*`, but it seems like they're really a hard requirement until the `impl` block. I just feel like it'd be cleaner if the struct itself didn't have these, but also would be fine if they stayed

ChaoticTempest (approved):
nothing out of the ordinary, so looking good to me

near-sdk/src/store/vec/iter.rs
austinabell: I believe it was because of the drop impl on the vec reference, but I can't think of why deserialize would be needed, I'll try to remove it now to check

PR #593, opened 2021-09-27
fix: bump Tgas on nft_approve

mattlockyer:
TLDR here: https://github.com/near-apps/nft-series/issues/1

The first approval succeeds, but subsequent approvals exceed the 10 Tgas.

This is causing several issues for users of:
https://github.com/near-examples/NFT

And in other repositories relying on:
https://github.com/near/near-sdk-rs/tree/master/near-contract-standards/src/non_fungible_token

e.g
https://github.com/near-apps/nft-series/issues/1

### DIFF
--- a/near-contract-standards/src/non_fungible_token/approval/approval_impl.rs
+++ b/near-contract-standards/src/non_fungible_token/approval/approval_impl.rs
@@ -9,7 +9,7 @@ use crate::non_fungible_token::utils::{
 use crate::non_fungible_token::NonFungibleToken;
 use near_sdk::{assert_one_yocto, env, ext_contract, require, AccountId, Balance, Gas, Promise};
 
-const GAS_FOR_NFT_APPROVE: Gas = Gas(10_000_000_000_000);
+const GAS_FOR_NFT_APPROVE: Gas = Gas(11_000_000_000_000);
 const NO_DEPOSIT: Balance = 0;
 
 fn expect_token_found<T>(option: Option<T>) -> T {


mattlockyer:
I suggest testing against testnet and trying to call `nft_approve` more than once in your tests. This is a common scenario for NFTs.

Otherwise these NFTs are broken since they can only be approved once.

Further scanning the code for `nft_approve`

Is the unwrapping of existing values is more costly than defaults?

https://github.com/near/near-sdk-rs/blob/d99f30175720d06331d7747dbc1a59cc0e66b3ff/near-contract-standards/src/non_fungible_token/approval/approval_impl.rs#L53-L55

**Most likely** it is the extra call to refund_deposit required by subsequent approvals:

https://github.com/near/near-sdk-rs/blob/d99f30175720d06331d7747dbc1a59cc0e66b3ff/near-contract-standards/src/non_fungible_token/approval/approval_impl.rs#L66-L68

Because of:
https://github.com/near/near-sdk-rs/blob/d99f30175720d06331d7747dbc1a59cc0e66b3ff/near-contract-standards/src/non_fungible_token/utils.rs#L39-L41

austinabell:
> Is the unwrapping of existing values is more costly than defaults?
> 
> https://github.com/near/near-sdk-rs/blob/d99f30175720d06331d7747dbc1a59cc0e66b3ff/near-contract-standards/src/non_fungible_token/approval/approval_impl.rs#L53-L55

What do you mean, this is defaulting? Did you mean to link to something different?


> **Most likely** it is the extra call to refund_deposit required by subsequent approvals:
> 
> https://github.com/near/near-sdk-rs/blob/d99f30175720d06331d7747dbc1a59cc0e66b3ff/near-contract-standards/src/non_fungible_token/approval/approval_impl.rs#L66-L68
> 
> Because of:
> https://github.com/near/near-sdk-rs/blob/d99f30175720d06331d7747dbc1a59cc0e66b3ff/near-contract-standards/src/non_fungible_token/utils.rs#L39-L41

I don't really have the context for this, but this solution doesn't really seem that great because it still leaves the contract open to be broken if there is some other case like this that is hit or for some reason gas costs change, for whatever reason, when using the contract on any network. I don't understand why these values are constant and not a ratio based on how much relative gas they cost so that they are not this rigid. https://github.com/near/near-sdk-rs/pull/573 is related to this type of change.

austinabell:
> > What do you mean, this is defaulting? Did you mean to link to something different?
> 
> No. This was a question. What is causing increased gas on subsequent calls?

Oh, I mean there will be slightly more gas used when a value does exist rather than using a default, to handle deserializing the value itself. Not sure if the difference is that granular, though.

mattlockyer:
> Oh, I mean there will be slightly more gas used when a value does exist rather than using a default, to handle deserializing the value itself. Not sure if the difference is that granular, though.

Hmmm then this may be the culprit...

I was a bit off on the `refund_deposit` comment, promise likely happens in most cases since refund amount will be > 1 yocto.

Either way, the only real way to tell is to run it against testnet...

Feel free to try this out:
https://github.com/near-apps/nft-series/

All you need is `yarn && yarn test:deploy`. See tests in `test/api.test.js`.

If you want to see it fail update `contract/Cargo.toml` to the standard and _not_ my fork which has 11 Tgas.

Double check it's doing a clean build after you switch crates.

danielwpz:
actually I had a similar issue.. seems the entire gas usage stuff for NFT needs to be redesigned

mikedotexe (dismissed):

mikedotexe:
Awesome, thanks @mattlockyer

mikedotexe:
No protocol changes.

Screen Shot 2021-10-08 at 2 30 13 PM

austinabell:
> That being said, I can't reproduce the issue. Though I'm still going to compile contracts with this gas bump applied because I don't want users to run into this issue in the wild. 1-2 Tgas seems trivial.
> 
> @austinabell where's the extra gas usage coming from? I read the above and can't find a definitive answer. Could it be something like the serialization/deserialization of the msg parameter?? This would indeed be bad as they are supplied by the client and subject to change as receiving contracts get more sophisticated in the future and received larger payloads.

Yeah, that's definitely a smaller change but also consider if the approved account ids increase. This leads to loading, deserializing, serializing, and storing more data which could bump up the amount of gas needed for the function. If this case is ever hit, the contract would just still be in a deadlock state where you can't perform that operation because not enough gas is left for this function. If the gas is based on a fraction of the prepaid gas, then a caller can just attach more gas to allow the tx to succeed.

Please let me know if any of these considerations are invalid or if I have a bad intuition on the scale of the differences. I don't have a good grasp of the magnitude of these differences yet, and also don't have a great understanding of the internals of the NFT standard

PR #595, opened 2021-09-30
Changed comment for the owner_id field in the NFT struct

### LINKED ISSUE #594: Comment for owner ID in NonFungibleTokenStruct outdated.
https://github.com/near/near-sdk-rs/blob/aa1cc4744b8d94bd03ed40ce49ed2cf46c90f60d/near-contract-standards/src/non_fungible_token/core/core_impl.rs#L56

Currently, the owner field in the NFT struct says `// owner of contract; this is the only account allowed to call 'mint'`. This is outdated since the mint function no longer asserts the owner is calling the method.

BenKurrek:
Fixes #594 

Removed the comment that stated only the owner could mint in the NFT struct.


austinabell (approved):
LGTM, also might be useful to update the deprecation warning on mint to indicate that using `internal_mint` doesn't check this anymore, so let me know if you want to group with this change or it can be done separately

PR #596, opened 2021-09-30
feat: update LookupMap key to be prefixed key serialized bytes

austinabell:
I realize this detail or tradeoff might have been overlooked or not noticed in the change from v1 to v2 of the `LookupMap`. In this PR, I change the key lookup so that it is strictly compatible with the old format (and thus drop-in replacement)

Here is what I see as the tradeoff from the changes in this PR:

Pros:
- Drop-in replacement of the old structure doesn't require migration
- Does not use hash syscall at all
- Prefixed storage to make it easier to delete (can also have this with the hash with only https://github.com/near/near-sdk-rs/commit/39a6e81fe8f776f507372254ecebb5e9a82b6907, but this defeats the purpose of the hash not allocating)

Cons:
- Requires a lot more allocations because each key is variable length
- Keys persisted to storage is full serialized bytes, which is unnecessary and can be potentially expensive
- Much more data is kept in memory because it keeps serialized keys as well as keys themselves (can mitigate by not caching the lookup key but would force more serializations)


austinabell:
When checking gas usage with https://github.com/austinabell/storage_collections_bench this alternative definitely used more gas. The change was slight (~3% more gas) but hard to filter out operations not relating to the map itself and get a good mock of how users would actually use this.

near-sdk/src/store/lookup_map/mod.rs
```diff
@@ -190,19 +154,21 @@ where
...
         K: Borrow<Q>,
     {
-        // Concat the prefix with serialized key and hash the bytes for the lookup key.
-        buffer.extend(prefix);
+        // Hash the serialized key
         key.serialize(buffer).unwrap_or_else(|_| env::panic_str(ERR_ELEMENT_SERIALIZATION));
 
-        H::hash(buffer)
+        // Prefix the hash for lookup
+        let mut key = Vec::with_capacity(buffer.len() + prefix.len());
```
ChaoticTempest, line 161: I'm confused a little by these comments. Is it still doing hashing somewhere? Or is one of the prefix, key or buffer expected to be already hashed?

near-sdk/src/store/lookup_map/mod.rs
```diff
@@ -18,17 +16,12 @@ const ERR_ELEMENT_DESERIALIZATION: &str = "Cannot deserialize element";
 const ERR_ELEMENT_SERIALIZATION: &str = "Cannot serialize element";
 const ERR_NOT_EXIST: &str = "Key does not exist in map";
 
-type LookupKey = [u8; 32];
+type LookupKey = Vec<u8>;
```
ChaoticTempest, line 19: > Prefixed storage to make it easier to delete (can also have this with the hash with only 39a6e81, but this defeats the purpose of the hash not allocating)

Maybe we can allow the `LookupKey` to be parameterized in `LookupMap` to allow users to either choose this current lookup key or the hashing version, but that might be a bit unwieldy since I could see it adding a lot more type parameters

austinabell:
Closing in favour of #674 because it's just a better option overall

Issue #194, opened 2020-06-24
[Feature Request] An 'owner' macro to ensure methods are called by owner


willemneal:
This should be the first function called.
```rust
fn assert_owner(&self) {
  assert_eq!(
            env::predecessor_account_id(),
            self.owner_id,
            "Can only be called by the owner"
        );
    }
```

willemneal:
https://github.com/near/core-contracts/issues/73#issuecomment-649091675

KaiserKarel:
Something like a proc-macro perhaps?

```
#[restricted(owner)]
fn kill() { ... }
```

austinabell:
Not sure this is great to add to the API, since owner is very application-specific and not any concept tied to the SDK. Perhaps this would be better if we could create some system of modifiers for function calls so that this could be better expressed.

The issue with doing this is that you would need to infer the `owner_id` field name on state, which is too loose of a connection. Perhaps there could be some connection if you also add an annotation on the struct itself, but it seems like a more general solution would be much more maintainable. I'm going to close this, but feel free to re-open it if you have an idea of how this could be done without assumptions

willemneal:
Really this is asserting a caller.  So perhaps an easy solution is to use @KaiserKarel's example

```rs
#[assert_caller(owner_id)]
fn kill() { ... }
```

austinabell:
> Really this is asserting a caller. So perhaps an easy solution is to use @KaiserKarel's example
> 
> ```rust
> #[assert_caller(owner_id)]
> fn kill() { ... }
> ```

This isn't intuitive that `owner_id` is coming from state, though. I think if we can come up with a more general solution it would be better than handling it for every specific case (in this one, asserting predecessor) since we already have `#[private]` to handle a very specific sub-case of this.

willemneal:
Could add `self.owner_id` then.  The macro is already on a method of the struct so that should make it clear.  Could just even add it to private.

PR #598, opened 2021-10-01
feat: introduce modifier attribute to near_bindgen

austinabell:
Was an idea I had that I wanted to try out to see if it was feasible and what the API would look like. This is intended to be like Solidity function modifiers https://www.tutorialspoint.com/solidity/solidity_function_modifiers.htm. Currently, this is a stateless version of this. This allows creating assertions that happen before or after the method logic (as opposed to only being able to create assertions 

Very rough, so just feel free to share high-level ideas about what this could look like. Here are my ideas:

- Allow passing variables into the modifiers to make them stateful (maybe this should be done automatically and just force your macro to have the correct signature? This comes with negative that they might not be easily re-usable)
- Allow passing in multiple modifiers ie `#[modifiers(one_modifier, other_modifier)]` possibly?
- Maybe apply the modifiers within the current function instead of just on wasm32 codegen
  - This might lead to a better unit testing experience, but might come at the cost of poor formatting that comes with wrapping code with a macro, not sure.
- Possibly also include the serialization tokens within another variable, so you can add assertions before anything is done for stateless things like if the state is already set
  - This not only provides a more standard way of checking, but also would save gas when able to short circuit on the failed criteria

I can see some serious benefits for DevX but also functional benefits for developers that don't want to hand-write methods outside the SDK to achieve these same things. The tricky part is this API is very unclear what the best direction is or the semantics of how it would be applied. If expanded on, this eliminates the need for adding modifiers for every use case, for example https://github.com/near/near-sdk-rs/issues/194. This also means that these modifiers could be composable and can be initiated by the community.

I probably won't touch this unless just hacking on or having fun with, unless others think there might be a value add here.


austinabell:
This was just exploratory to replicate solidity function modifiers. Unfinished with no plans to continue and opinionated, so closing

PR #605, opened 2021-10-10
remove external private modules

ghost:
Your `near-sdk-macros` depends on private modules is `syn`.

It's not good for code to depend on private modules in other packages.

This was fixed by `syn` https://github.com/dtolnay/syn/commit/957840eba2c7f11c95e0227760d78dc481a91de7

Please merge this. Causing conflicts when using the fixed `syn` versions.


austinabell:
Will add this in the next 4.0 pre-release. Would it be beneficial if I release this ASAP or is it fine to wait a few days?

PR #616, opened 2021-10-28
fix: code generation for mut self methods

### LINKED ISSUE #615: Require `#[init]` methods to return `Self` / `mut self` methods fail to compile
It would be useful to require `#[init]` methods to return `Self` so the type system can catch this sort of mistake, otherwise (I believe) a `()` type value could be stored as the state.

Example using `near-sdk = "4.0.0-pre.4"`:
```rust
use near_sdk::{
    borsh::{self, BorshDeserialize, BorshSerialize},
    near_bindgen, PanicOnDefault,
};
#[near_bindgen]
#[derive(BorshDeserialize, BorshSerialize, PanicOnDefault, Clone)]
pub struct Struct {
    a: u8,
}
trait Trait {
    fn f();
}
#[near_bindgen]
impl Trait for Struct {
    #[init]
    fn f() {
        todo!()
    }
}

// generates the following, where `contract` has type `()`
//
// #[cfg(target_arch = "wasm32")]
// #[no_mangle]
// pub extern "C" fn f() {
//     near_sdk::env::setup_panic_hook();
//     if near_sdk::env::attached_deposit() != 0 {
//         near_sdk::env::panic_str("Method f_stateless doesn\'t accept deposit");
//     }
//     if near_sdk::env::state_exists() {
//         near_sdk::env::panic_str("The contract has already been initialized");
//     }
//     let contract = Struct::f();
//     near_sdk::env::state_write(&contract);
// }
```

---

An owned `mut self` method suffers a similar problem, but I think that the move semantics doesn't allow the value to be stored as the state (as it has been moved) :
```
trait Trait: Sized {
    fn g(self);
}

#[near_bindgen]
impl Trait for Struct {
    fn g(mut self) {
        todo!()
    }
}

// generates the following, where the last line correctly doesn't compiles:
//
// #[cfg(target_arch = "wasm32")]
// #[no_mangle]
// pub extern "C" fn g() {
//     near_sdk::env::setup_panic_hook();
//     if near_sdk::env::attached_deposit() != 0 {
//         near_sdk::env::panic_str("Method f_owned_mut doesn\'t accept deposit");
//     }
//     let mut contract: Struct = near_sdk::env::state_read().unwrap_or_default();
//     contract.g();
//     near_sdk::env::state_write(&contract);
// }
```

austinabell:
Closes #615 

Probably rare someone would want to modify this structure in place for a view method, which is probably how this wasn't hit before, but good to have this functional

### DIFF
--- a/near-sdk-macros/src/core_impl/info_extractor/attr_sig_info.rs
+++ b/near-sdk-macros/src/core_impl/info_extractor/attr_sig_info.rs
@@ -105,7 +105,7 @@ impl AttrSigInfo {
 
         if let Some(ref receiver) = receiver {
             if matches!(method_type, MethodType::Regular) {
-                if receiver.mutability.is_none() {
+                if receiver.mutability.is_none() || receiver.reference.is_none() {
                     method_type = MethodType::View;
                 }
             } else {
--- a/near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
+++ b/near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
@@ -82,6 +82,43 @@ mod tests {
     }
 
     #[test]
+    fn owned_no_args_no_return_no_mut() {
+        let impl_type: Type = syn::parse_str("Hello").unwrap();
+        let mut method: ImplItemMethod = syn::parse_str("pub fn method(self) { }").unwrap();
+        let method_info = ImplItemMethodInfo::new(&mut method, impl_type).unwrap();
+        let actual = method_info.method_wrapper();
+        let expected = quote!(
+            #[cfg(target_arch = "wasm32")]
+            #[no_mangle]
+            pub extern "C" fn method() {
+                near_sdk::env::setup_panic_hook();
+                let contract: Hello = near_sdk::env::state_read().unwrap_or_default();
+                contract.method();
+            }
+        );
+        assert_eq!(expected.to_string(), actual.to_string());
+    }
+
+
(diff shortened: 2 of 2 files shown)


near-sdk-macros/src/core_impl/info_extractor/attr_sig_info.rs
```diff
@@ -105,7 +105,7 @@ impl AttrSigInfo {
 
         if let Some(ref receiver) = receiver {
             if matches!(method_type, MethodType::Regular) {
-                if receiver.mutability.is_none() {
+                if receiver.mutability.is_none() || receiver.reference.is_none() {
```
ChaoticTempest, line 108: I'm not sure why we can't just have
```rust
if receiver.reference.is_none() {
```
but `compilation_tests/payable_view.rs` successfully compiles even though its expected not to.

This conditional should just only be checking for `mut self` and `self` signatures, where checking for mutability shouldn't matter at all. But somehow also related to `pay(&self)` signature in `payable_vew.rs`? Maybe there's some sort of complexity somewhere we're not noticing or I'm probably missing something here.

austinabell: Nah this is intentional. This was the point of the PR that `mut self` was treated as mutating state and tries to write the data after. If you check only for reference, you don't differentiate between mutable and immutable references

ChaoticTempest (approved):

PR #618, opened 2021-10-28
feat(NFT_standard): Add option to use refund with `internal_mint`

willemneal:
This removes the assumption that an internal mint will refund the `predecessor_id` at the end.  This isn't needed if minting more than token since the refund should happen after all tokens have been minted.  Furthermore, since minting is no longer guarded the refund should typically go to the owner_id, though authors of NFTs might want to specify another.

### DIFF
--- a/near-contract-standards/src/non_fungible_token/core/core_impl.rs
+++ b/near-contract-standards/src/non_fungible_token/core/core_impl.rs
@@ -305,7 +305,7 @@ impl NonFungibleToken {
 
     /// Mint a new token without checking:
     /// * Whether the caller id is equal to the `owner_id`
-    /// * Assumes there will be a refund to the owner after covering the storage costs
+    /// * Assumes there will be a refund to the predecessor after covering the storage costs
     ///
     /// Returns the newly minted token
     pub fn internal_mint(


willemneal:
@austinabell @ChaoticTempest Thoughts?

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -300,18 +300,23 @@ impl NonFungibleToken {
...
+    ///
     pub fn internal_mint(
         &mut self,
         token_id: TokenId,
         token_owner_id: AccountId,
         token_metadata: Option<TokenMetadata>,
+        refund_id: Option<AccountId>,
     ) -> Token {
-        let initial_storage_usage = env::storage_usage();
+        let initial_storage_usage = if refund_id.is_some() { env::storage_usage() } else { 0 };
```
austinabell, line 319: Can you put this in the if block below? Doesn't really make sense to do this check twice when only used below if Some

near-contract-standards/src/non_fungible_token/utils.rs
```diff
@@ -26,7 +26,7 @@ pub fn refund_approved_account_ids(
     refund_approved_account_ids_iter(account_id, approved_account_ids.keys())
 }
 
-pub fn refund_deposit(storage_used: u64) {
+pub fn refund_deposit(storage_used: u64, account_id: AccountId) {
```
austinabell, line 29: to avoid breaking changes, this should probably be a new function, and just have `refund_deposit` call this with the `predecessor_account_id`. Can think about deprecating this old function too if accepted

austinabell (commented):
I don't know enough about the token economics of this contract to give an approval. @mikedotexe, who would be the one to decide if this is a better pattern?

Doesn't feel right to always refund owner because they may not be the ones that are increasing the storage usage, it seems like it should be as is because if the caller frees up storage they should be refunded for it? Idk just my thoughts

willemneal:
@austinabell, the guarded mint method already asserts that the owner_id is the predecessor.  In the case that the mint is unguarded, e.g. the contract author allows other users to mint, then is better to be explicit with who gets the refund instead of the predecessor.  I'm sure owner wouldn't want other accounts to mint for only the cost of storage.

near-contract-standards/src/non_fungible_token/core/core_impl.rs
willemneal: This was to avoid calling `env::storage_usage` if not refunding.  I just updated it to be a little more rusty, by using map to make this an option.

willemneal:
@mattlockyer  @mikedotexe Thoughts?

willemneal:
@mikedotexe

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -346,8 +374,10 @@ impl NonFungibleToken {
         let approved_account_ids =
             if self.approvals_by_id.is_some() { Some(HashMap::new()) } else { None };
 
+        if let Some((id, storage_usage)) = initial_storage_usage {
```
danielwpz, line 377: hi, why not refund to the signer if no refund id is provided?

willemneal: The idea is that if you pass a `None`, no refund should happen.  The point is that you want to have the ability to mint multiple tokens in one transaction and then do a refund after adding them all to storage.

danielwpz: > The point is that you want to have the ability to mint multiple tokens in one transaction and then do a refund after adding them all to storage.           

yeah this makes sense to me.

But I would prefer to refund to the txn signer *by default* in case no refund id is given, rather than keeping all the extra deposit in the contract

willemneal: That's what happens in the guarded mint. How else would you specify no refund? Add another argument?

willemneal:
@mikedotexe seems to no longer be a code owner so @austinabell does my explanation make sense? @ChaoticTempest Could you take a look.  It's a simple issue really, but without this fix the code only allows you to mint one token in a transaction.

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -305,13 +305,41 @@ impl NonFungibleToken {
...
+            Some(self.owner_id.clone()),
+        )
+    }
+
+    /// Mint a new token without checking:
+    /// * Whether the caller id is equal to the `owner_id`
+    /// * `refund_id` will transfer the left over balance after storage costs are calculated to the provided account.
+    ///   Typically the account will be the owner. If `None`, will not refund. This is useful for delaying refunding
+    ///   until multiple tokens have been minted.
+    ///
```
austinabell, line 332:
```suggestion
```

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -305,13 +305,41 @@ impl NonFungibleToken {
 
     /// Mint a new token without checking:
     /// * Whether the caller id is equal to the `owner_id`
+    /// * `refund_id` will transfer the left over balance after storage costs are calculated to the provided account.
```
austinabell, line 308: You might want to change this comment, since there is no `refund_id` anymore as a param

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -305,13 +305,41 @@ impl NonFungibleToken {
...
+    ///   Typically the account will be the owner. If `None`, will not refund. This is useful for delaying refunding
+    ///   until multiple tokens have been minted.
+    /// * Assumes there will be a refund to the owner after covering the storage costs
+    ///
```
austinabell, line 312:
```suggestion
```

austinabell (commented):
This all LGTM, I just can't give approval until I'm sure that changing the refund to the owner instead of the predecessor is a valid change since I don't have the context currently. It's a change that might silently break the token economics of some contracts

willemneal:
Thanks! 

@ChaoticTempest For some context

The default mint function checks if the owner is the `predecessor`,
```rs
        assert_eq!(env::predecessor_account_id(), self.owner_id, "Unauthorized");
        self.internal_mint(token_id, token_owner_id, token_metadata) // <-- Assumes that owner should be refunded.
```

ChaoticTempest (approved):
LGTM: everything here looks to work the same w/ just some internal/helper functions if somebody else wants to make use of them

ChaoticTempest:
> Keep in mind the subtle change that affects the existing method. The account id which is refunded has changed

There was a guard for this when it gets used within the `mint` call:
```
assert_eq!(env::predecessor_account_id(), self.owner_id, "Unauthorized");
```
So thought that would be fine which is what @willemneal mentioned before.

`internal_mint` is `pub` so that could a breaking change but thought that was fine since this was still within the prerelease I think

austinabell:
> > Keep in mind the subtle change that affects the existing method. The account id which is refunded has changed
> 
> There was a guard for this when it gets used within the `mint` call:
> 
> ```
> assert_eq!(env::predecessor_account_id(), self.owner_id, "Unauthorized");
> ```
> 
> So thought that would be fine which is what @willemneal mentioned before.
> 
> `internal_mint` is `pub` so that could a breaking change but thought that was fine since this was still within the prerelease I think

Yeah, all I'm suggesting is that it makes sense but lets just make sure before we bring it in because we will forget about a detail like this surely :D

mattlockyer:
> @austinabell, the guarded mint method already asserts that the owner_id is the predecessor. In the case that the mint is unguarded, e.g. the contract author allows other users to mint, then is better to be explicit with who gets the refund instead of the predecessor. In most cases the owner wouldn't want other accounts to mint for only the cost of storage.

I disagree. If the owner wants to make money per mint then it should be explicit.

There's many examples of "free minting" where the user only pays the storage for their token.

Otherwise, we are stuck with a scenario of having to estimate storage outside the runtime and send exactly the right amount. This can be tricky.

What's the harm in leaving predecessor as refund recipient?

austinabell:
> I disagree. If the owner wants to make money per mint then it should be explicit.
> 
> There's many examples of "free minting" where the user only pays the storage for their token.
> 
> Otherwise, we are stuck with a scenario of having to estimate storage outside the runtime and send exactly the right amount. This can be tricky.
> 
> What's the harm in leaving predecessor as refund recipient?

Thanks for your input, Matt. I think the best way forward is to keep the refund as predecessor and this new function makes it easy to override this if someone would like it to be the owner refunded, but it would have to be intentionally done.

The other main changes in this PR are good and will be a benefit to have more flexibility of usage :)

willemneal:
@mattlockyer Thanks for the review.  My final argument for keeping the owner as the default, is it is safer for the contract author.  If we default to predecessor and then the author uses the unguarded mint and thinking the refund went to the owner, then they will lose the funds on a mint.  Whereas if it is the case of the opposite and they intend on users minting for free, they could still return funds to the signer after the fact since it was sent to the owner.

willemneal:
@austinabell @ChaoticTempest Switched to predecessor to close this PR.

austinabell (approved):
LGTM! Thanks for the contribution :) Feel free to continue the discussion of predecessor vs owner in an issue or a separate PR

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -305,13 +305,40 @@ impl NonFungibleToken {
 
     /// Mint a new token without checking:
     /// * Whether the caller id is equal to the `owner_id`
+    /// * Assumes there will be a refund to the owner after covering the storage costs
```
austinabell, line 308: Ah do you mind updating the comment to the predecessor, if not nbd I can do after this comes in

willemneal:
```suggestion
    /// * Assumes there will be a refund to the predecessor after covering the storage costs
```

PR #624, opened 2021-11-02
Allow refunding a specific account when calling storage_deposit

ctindogaru:


austinabell:
This interface can't be updated like this for this use case as it is breaking. I'd suggest proposing a change in https://github.com/near/NEPs 

Going to close since it seems like it isn't in progress, but feel free to re-open :)

PR #627, opened 2021-11-04
feat(NFT): add events to standard

willemneal:
This borrows the types for [NFT event standard](https://github.com/near/NEPs/blob/master/specs/Standards/NonFungibleToken/Event.md) used by indexer.  The only real change is that serializers are implemented for these types and `serde_with` is used to not serialize optional fields.

Perhaps these types should live in a separate repo that is shared between the sdk and the indexer, but I thought I'd put them here for now.


willemneal:
@austinabell @ChaoticTempest This is a rough start.  I am a fan of this method since it seems to add a lot of code, so I would love some feedback.

ChaoticTempest:
yeah, would be nice to have a common crate for all these types. Let me see if I can't get one up, but let me workshop around a name. `near-primitives` is already taken. `near-commons` could be one but might be confusing with `near-primitives`

willemneal:
@ChaoticTempest How about `near-standards-primitives`?

willemneal:
@ChaoticTempest any luck?

ChaoticTempest:
@willemneal so was talking to @austinabell about this, but it seems like this would be better still living here since most of this is just the NFT related types. If anything, we'd surface `NearEvent` into `lib.rs` where we can potentially populate it with other events too like FT. What do you think?

willemneal:
Sounds good to me.  I'll go ahead and update it.  @frol @telezhnaya What are your thoughts for adding `near-contract-standards` as a dep to the indexer so that we can share the types?

near-contract-standards/src/non_fungible_token/event.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    pub fn log_nft_burn(
+        owner_id: String,
+        authorized_id: Option<String>,
+        token_ids: Vec<String>,
+        memo: Option<String>,
+    ) {
+        NearEvent::nft_burn(vec![NftBurnData { owner_id, authorized_id, token_ids, memo }]).log();
+    }
```
frol, line 120: This API has a few problems:

1. The usage of these methods will be error-prone to use and hard to reason about:
    ```rust
    NearEvent::log_nft_transfer("frol.near", "willem.near", vec!["dacha".to_string()], Some("qq".to_string()), Some("alex.near"))
    NearEvent::log_nft_burn("willem.near", Some("alex.near"), vec!["dacha".to_string()], Some("qq".to_string()))
    ```
  
    (You already mixed up the order of `authorized_id` in transfer and burn methods)
2. These methods promote non-batched events and thus we promote using more gas on events than necessary

near-contract-standards/src/non_fungible_token/event.rs
```diff
@@ -0,0 +1,174 @@
...
+
+use serde::{Deserialize, Serialize};
+use serde_with::skip_serializing_none;
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(tag = "standard")]
+#[serde(rename_all = "snake_case")]
+pub enum NearEvent {
+    Nep171(Nep171Event),
+}
```
frol, line 11: You don't need `NearEvent` type in the non-fungible token module. Go straight with `Nep171Event`

near-contract-standards/src/non_fungible_token/event.rs
```diff
@@ -0,0 +1,174 @@
...
+
+#[skip_serializing_none]
+#[derive(Serialize, Deserialize, Debug)]
+pub struct NftBurnData {
+    pub authorized_id: Option<String>,
+    pub owner_id: String,
+    pub token_ids: Vec<String>,
+    pub memo: Option<String>,
+}
```
frol, line 55: We may consider using [`Cow<'a, str>`](https://deterministic.space/secret-life-of-cows.html) instead of Strings to avoid unnecessary clones when users would need to generate the events without giving the ownership over the strings. Yet, that might make the API harder to follow.

near-contract-standards/src/non_fungible_token/event.rs
```diff
@@ -0,0 +1,174 @@
...
+#[serde(rename_all = "snake_case")]
+#[allow(clippy::enum_variant_names)]
+pub enum Nep171EventKind {
+    NftMint(Vec<NftMintData>),
+    NftTransfer(Vec<NftTransferData>),
+    NftBurn(Vec<NftBurnData>),
+}
+
+#[skip_serializing_none]
+#[derive(Serialize, Deserialize, Debug)]
```
frol, line 31: Flip the order of these lines, so `derive` is before any use of the derived macro attributes

near-contract-standards/src/non_fungible_token/event.rs
```diff
@@ -0,0 +1,174 @@
...
+pub struct Nep171Event {
+    pub version: String,
+    #[serde(flatten)]
+    pub event_kind: Nep171EventKind,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(tag = "event", content = "data")]
+#[serde(rename_all = "snake_case")]
+#[allow(clippy::enum_variant_names)]
```
frol, line 23: Why?

frol (commented):
> What are your thoughts for adding `near-contract-standards` as a dep to the indexer so that we can share the types?

We would need to evaluate the dependencies tree before making this decision, but I think it should be fine. It is definitely not a priority at the moment, though. We will get back to it after FT events are implemented.

near-contract-standards/src/non_fungible_token/event.rs
willemneal: My thoughts here was that we would collect all events in this enum.

near-contract-standards/src/non_fungible_token/event.rs
willemneal: 1. Yeah oops, this was from just following the ordering from the struct.

2. I agree, my inital thoughts were that they were the common case so it should be easy to use.

near-contract-standards/src/non_fungible_token/event.rs
willemneal: Copy pasted this from the indexer's type definitions.

near-contract-standards/src/non_fungible_token/event.rs
willemneal: Yeah again this is a copy/paste and I agree that we should try to minimize the cost of these logs.

near-contract-standards/src/non_fungible_token/event.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub memo: Option<String>,
+}
+
+impl Display for NearEvent {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(&format!("EVENT_JSON:{}", self.to_json_string()))
+    }
+}
+
+use Nep171EventKind::*;
```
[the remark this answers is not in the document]
willemneal: Yeah this is my distain for the full qualifications in patterns.  But I'm fine adding them back.

near-contract-standards/src/non_fungible_token/event.rs
frol: I see, it slipped through my review there, but I caught it here :smile:

near-contract-standards/src/non_fungible_token/event.rs
frol: I know you copy-pasted it from indexer, but we discussed that going forward we want to move this structure up the hierarchy since it does not make sense to live under NFT module since we expect to have support for other standards as well. However, I am not sure it makes sense to have a single enum with all sorts of NEP events under a single umbrella in near-contract-standards.

near-contract-standards/src/non_fungible_token/event.rs
telezhnaya: Because it should be just `mint`, `transfer`, `burn` (clip warns about duplicated prefix)
But we use these names in json strings
So, we need to choose: 
- This line
- `@name("nft_smth")` for each line in the enum

I preferred first option

danielwpz:
guys any luck with this PR?

willemneal:
@danielwpz About to push up some changes!

near-contract-standards/src/non_fungible_token/event.rs
willemneal: I tried and the code failed to compile.

near-contract-standards/src/non_fungible_token/event.rs
willemneal: I moved the file to the top of the standards crate.  The benefit to this method is that it is taking advantage of the type name when generating the JSON, e.g. ` {"standard: "nep171"}`, and will be able to handle more as the are created.  I do think it makes sense to have a since umbrella for the standards.  

Eventually I see the standards as moving toward WebAssembly Interface types, which will need to be defined in one place.

near-contract-standards/src/non_fungible_token/event.rs
willemneal: I fixed `1` and added methods to log data vectors.

willemneal:
@ChaoticTempest @frol @austinabell @BenKurrek 

I moved everything up a level and added three new methods for logging batches of events.  However, I'm still not a fan of `String` types everywhere.  Do you think it's worth changing the API to `Cow` as @frol suggested?

austinabell:
> @ChaoticTempest @frol @austinabell @BenKurrek
> 
> I moved everything up a level and added three new methods for logging batches of events. However, I'm still not a fan of `String` types everywhere. Do you think it's worth changing the API to `Cow` as @frol suggested?

Might be worth at least switching and seeing how it affects code size. Keep in mind, by default Cow will deserialize as an owned string and you don't get the performance of referencing the deserialized buffer so this just saves an allocation on initialization. 

As for this PR and the other, I think it might have been assumed this was stale as it didn't have the connection to standards, not sure. Hopefully, we can merge the efforts somehow to get the best of both :)

willemneal:
What do you mean ` as it didn't have the connection to standards`?

austinabell:
> What do you mean ` as it didn't have the connection to standards`?

it isn't used in the NFT standard, this just adds the types, which is fine

near-contract-standards/src/event.rs
```diff
@@ -0,0 +1,255 @@
...
+        near_sdk::env::log_str(&self.to_string());
+    }
+
+    pub fn log_nft_mint(owner_id: String, token_ids: Vec<String>, memo: Option<String>) {
+        NearEvent::log_nft_mints(vec![NftMintData { owner_id, token_ids, memo }]);
+    }
+
+    pub fn log_nft_mints(data: Vec<NftMintData>) {
+        NearEvent::nft_mint(data).log();
+    }
```
frol, line 97: Why do you need two way to achieve the same thing? I like the `.log()` API, though I would consider renaming it to `.emit()`

near-contract-standards/src/event.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    pub fn log_nft_transfer(
+        old_owner_id: String,
+        new_owner_id: String,
+        token_ids: Vec<String>,
+        memo: Option<String>,
+        authorized_id: Option<String>,
+    ) {
```
frol, line 105: I am still not sold on having functions with too many arguments, especially when the type is almost the same, and it is hard to reason about the order.

willemneal:
@frol @ChaoticTempest @austinabell @BenKurrek 

I've updated it to use `Cow<'a, str>` as Frol suggested.  It makes the API a bit more ergonomic. However, currently it uses a single generic type, which must have `Into<Cow<'a, str>>`.  This can cause issues when the types are mixed, e.g. `&str`, `String`. This can be solved by being explicit.

```rust
let old_owner_id = "bob".to_string();
let new_owner_id = "alice";
let token_ids = vec!["0", "1"];
let log = NearEvent::nft_transfer(vec![NftTransferData::new::<&str>( // explicit type 
        &old_owner_id,
        new_owner_id,
        token_ids,
        None,
        None,
)])
```


 This also means that it requires calling a constructor for each type, which is what Frol wanted to avoid.  But I think that it is worth it since we prevent copying.


The other noteworthy change is that even with the `#[serde(borrow)` the top level type `NearEvent` wouldn't compile with `#[derive(Deserialize)]`.  This isn't an issue for the sdk, but definitely for the indexer.  Still not sure why the compiler is mad about this.

frol:
> This can cause issues when the types are mixed

It should not cause any issues if you create a separate type of each argument:

```rust
    pub fn new<O, T, M>(owner_id: O, token_ids: Vec<T>, memo: Option<M>) -> NftMintData<'a>
    where
        O: Into<Cow<'a, str>>,
        T: Into<Cow<'a, str>>,
        M: Into<Cow<'a, str>>,
```

near-contract-standards/src/event.rs
```diff
@@ -0,0 +1,299 @@
...
+        token_ids: Vec<&str>,
+        authorized_id: Option<&str>,
+        memo: Option<&str>,
+    ) {
+        NearEvent::emit_nft_burns(vec![NftBurnData::new(owner_id, token_ids, authorized_id, memo)]);
+    }
+
+    pub fn emit_nft_burns(data: Vec<NftBurnData<'a>>) {
+        NearEvent::nft_burn(data).emit()
+    }
```
frol, line 198: I feel these `emit_` wrappers will only confuse and unnecessarily introduce complexity and the number of methods a user should learn about.

Let's remove these.

near-contract-standards/src/event.rs
```diff
@@ -0,0 +1,299 @@
...
+        NearEvent::new_171_v1(Nep171EventKind::NftMint(data))
+    }
+
+    pub(crate) fn to_json_string(&self) -> String {
+        serde_json::to_string(self).unwrap()
+    }
+
+    pub fn to_json_event_string(&self) -> String {
+        format!("EVENT_JSON:{}", self.to_json_string())
+    }
```
frol, line 153: Let's add [`#[must_use]`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute) attribute to these methods to make sure users don't forget about emitting the events or at least doing something with the function result.

Consider also adding some text comment `#[must_use = "don't forget to .emit() the event"]` to improve the DX.

frol (approved):

willemneal:
Tried that and it had trouble inferring the types.

frol:
> Tried that and it had trouble inferring the types.

That is strange. Can you share the error message?

willemneal:
@frol It's due to the `Option` type not providing the underlying type.

```
error[E0283]: type annotations needed
   --> near-contract-standards/src/event.rs:211:24
    |
211 |         let mint_log = NftMintData::new(owner_id, token_ids, None);
    |                        ^^^^^^^^^^^^^^^^ cannot infer type for type parameter `M` declared on the associated function `new`
    |
    = note: cannot satisfy `_: Into<Cow<'_, str>>`
note: required by a bound in `event::NftMintData::<'a>::new`
   --> near-contract-standards/src/event.rs:51:12
    |
47  |     pub fn new<O,T,M>(owner_id: O, token_ids: Vec<T>, memo: Option<M>) -> NftMintData<'a>
    |            --- required by a bound in this
...
51  |         M: Into<Cow<'a, str>>,
    |            ^^^^^^^^^^^^^^^^^^ required by this bound in `event::NftMintData::<'a>::new`
help: consider specifying the type arguments in the function call
    |
211 |         let mint_log = NftMintData::new::<O, T, M>(owner_id, token_ids, None);
    |                                  
    ```

**Comment by willemneal** [comment]
@austinabell @frol @ChaoticTempest 

Updated again to use generics.  In most cases the types will be provided and it definitely is worth it for the flexibility.

I also added a `From` implementation for AccountId so that a reference can be passed as well.

Lastly I added the logs to the standard's mint and transfer methods.

**Comment by willemneal** [comment]
@austinabell @ChaoticTempest @mikedotexe @frol @BenKurrek Any thoughts or blockers?

File: near-contract-standards/src/event.rs [hunk 11]
```diff
@@ -0,0 +1,283 @@
...
+        ])
+        .to_json_string();
+        assert_eq!(
+            log,
+            r#"{"standard":"nep171","version":"1.0.0","event":"nft_burn","data":[{"owner_id":"alice","token_ids":["2","3"],"authorized_id":"4","memo":"has memo"},{"owner_id":"bob","token_ids":["0","1"]}]}"#
+        );
+    }
+
+    #[test]
+    fn nft_transfer() {
```
telezhnaya, line 238: Could you please also add a test with callbacks? (`nft_transfer_call`)
We need to be sure that the log was produced in the positive case, and there was no log in the negative one.

willemneal: @telezhnaya @frol @BenKurrek 

Looking into this, I see two solutions.

1) Emit the transfer after the initial transfer,  then when reverting emit the reverse transfer. (This is currently the case since `internal_transfer` is used in both cases.

2) Refrain from emitting the transfer until the end.  That is don't emit in the internal transfer, but have the caller emit the event. This seems like the more reasonable approach, however, it has a draw back with the current implementation of the standard.

Currently `nft_transfer_call` takes a `memo` parameter, to be emitted with the event.  However, if the event is emitted at the end of the chain of calls, then the final `nft_reslove_transfer` needs this and with it's current API it doesn't.

So really the choices are 1) emitted an extra log on a failure, or 2) change the standard to allow the final callback to have the `memo` argument.

Personally I think the first option is the easiest, but will end up adding 2 extra logs.  So perhaps go with the approach for now and then go through the NEP process to pass the `memo`?

willemneal:
@frol @BenKurrek @telezhnaya @mikedotexe 

Here is the PR to update `nft_transfer_resolve` needed to allow it to emit the memo.  Should we wait until this is merged or go with option one above, emit the extra transfer event when `nft_transfer_resolve` is rolling the transaction back.

near-contract-standards/src/event.rs
BenKurrek: How are you going to deal with logging the `authorized_id`? See my explanation of my implementation found in [docs](https://docs.near.org/docs/tutorials/contracts/nfts/events#logging-transfers) and my code [here](https://github.com/near-examples/nft-tutorial/blob/e8f85847f93c73ac56f926315a3b6abc7290d7d1/nft-contract/src/nft_core.rs#L126)

BenKurrek: How are you going to deal with logging the `authorized_id`? See my implementation explained in [docs](https://docs.near.org/docs/tutorials/contracts/nfts/events#logging-transfers) and my code found [here](https://github.com/near-examples/nft-tutorial/blob/main/nft-contract/src/nft_core.rs#L126)

willemneal: Good point.  Your change isn't part of the standard either.  So adde it to my PR linked below.

willemneal:
@frol @telezhnaya @BenKurrek @austinabell @ChaoticTempest 

It's finally ready! Went with logging twice if a failure in `nft_transfer_call`. And as @telezhnaya asked I added some simulation tests for it.

BenKurrek:
@willemneal I see that we're checking if the attached GAS is greater than GAS for transfer call and resolve transfer as seen [here](https://github.com/near/near-sdk-rs/blob/83a3b7515d014ffa51996f49d5e10750634c7fe6/near-contract-standards/src/non_fungible_token/core/core_impl.rs#L435). In the case where the user attaches 1 GAS more so `GAS_FOR_NFT_TRANSFER_CALL + GAS_FOR_RESOLVE_TRANSFER + 1` the `internal_transfer` logic will fire but there won't be enough GAS for the cross contract calls and so the logic in `internal_transfer` will revert. The log, however, will still be fired. Does the indexer only populate the DB on successful receipts? Or will this be a problem.

willemneal:
@BenKurrek in the scenario you describe the check would still guarantee that the final `nft_resolve_transfer` has gas.

For simplicity consider:

`X = 5 Tgas`, `Y = X + 25 Tgas`, prepaid gas = `Y + X + 1`

```rust
        require!(
            env::prepaid_gas() > Y + X,
            "More gas is required"
        );
       //..
        ext_receiver::nft_on_transfer(
            sender_id,
            old_owner.clone(),
            token_id.clone(),
            msg,
            receiver_id.clone(),
            NO_DEPOSIT,
            (X + 1),            // (Y + X + 1) - Y <-- This might not be enough but the failure will be caught in the cb
        )
        .then(ext_self::nft_resolve_transfer(
            old_owner,
            receiver_id,
            token_id,
            old_approvals,
            env::current_account_id(),
            NO_DEPOSIT,
            X,
        ))
        .into()
```

So in total the contract has `prepaid - CX Gas = (X + Y + 1) - (2X + 1) =  X + X + 25 Tgas + 1 - (2x + 1) = 25 Tgas` left over.  So the only issue is if this `X = 5 Tgas` is enough gas for the `nft_resolve_transfer` and given that it has been working I'd say that it is, but we could increase it if we wanted to be safe.

willemneal:
@austinabell @ChaoticTempest It is failing clippy with an unrelated line:

```rust
 Checking near-sdk-macros v4.0.0-pre.5 (/home/runner/work/near-sdk-rs/near-sdk-rs/near-sdk-macros)
error: `to_string` applied to a type that implements `Display` in `format!` args
  --> near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs:63:74
   |
63 |             let error = format!("Method {} doesn't accept deposit", ident.to_string());
   |                                                                          ^^^^^^^^^^^^ help: remove this
   |
   = note: `-D clippy::to-string-in-format-args` implied by `-D clippy::all`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#to_string_in_format_args

error: `to_string` applied to a type that implements `Display` in `format!` args
  --> near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs:71:62
   |
71 |             let error = format!("Method {} is private", ident.to_string());
   |                                                              ^^^^^^^^^^^^ help: remove this
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#to_string_in_format_args

error: could not compile `near-sdk-macros` due to 2 previous errors
```

Should I fix it in this PR or make another?

near-contract-standards/src/event.rs
```diff
@@ -0,0 +1,283 @@
...
+    NftTransfer(Vec<NftTransferData<'a>>),
+    #[serde(borrow)]
+    NftBurn(Vec<NftBurnData<'a>>),
+}
+
+#[skip_serializing_none]
+#[derive(Serialize, Deserialize, Debug)]
+pub struct NftMintData<'a> {
+    #[serde(borrow)]
+    pub owner_id: Cow<'a, str>,
```
ChaoticTempest, line 39: Maybe worth putting in the comments somewhere in this file about why the optimization with `Cow`'s are needed, or linking back to this PR with the discussions could work too

ChaoticTempest (commented):
I'll leave it up to @austinabell to have the final say, but overall this PR looks good to me. The sim tests will have to be moved over to workspaces in the future, but for now we can leave that for a later PR. Think we need a CHANGELOG entry as well

near-contract-standards/src/event.rs
```diff
@@ -0,0 +1,298 @@
+
+/// This file makes heavy use of Copy-on-write references to strings.
+/// Since the logs should be light weight they should not clone any string.
+/// Using generic arguments bound to `Into<Cow<'a, str>>` any reference passed that uses
+/// can provide a reference to a string can be used. E.g. `&String, String, &str, &AccountId`
+/// This also makes it more ergonomic since not only one type can be passed.
```
austinabell, line 6: Okay so I'm not seeing why you need `Cow` for this. The negative for this is primarily the added code size for introducing a new part of std but also some unnecessary compute cycles. Also, for these logs, there is never any need to deserialize them within the contract.

You can remove the deserialize implementation altogether because these logs will only be parsed off-chain, and just use references for these things, no?

near-contract-standards/src/event.rs
```diff
@@ -0,0 +1,298 @@
...
+        token_ids: Vec<T>,
+        authorized_id: Option<A>,
+        memo: Option<M>,
+    ) -> NftTransferData<'a>
+    where
+        O: Into<Cow<'a, str>>,
+        N: Into<Cow<'a, str>>,
+        T: Into<Cow<'a, str>>,
+        A: Into<Cow<'a, str>>,
+        M: Into<Cow<'a, str>>,
```
austinabell, line 104: these types of generics will make it harder to understand for users, as an extension of my previous comment

near-contract-standards/src/event.rs
```diff
@@ -0,0 +1,298 @@
...
+    #[serde(borrow)]
+    pub old_owner_id: Cow<'a, str>,
+    #[serde(borrow)]
+    pub new_owner_id: Cow<'a, str>,
+    #[serde(borrow)]
+    pub token_ids: Vec<Cow<'a, str>>,
+    #[serde(borrow)]
+    pub authorized_id: Option<Cow<'a, str>>,
+    #[serde(borrow)]
+    pub memo: Option<Cow<'a, str>>,
```
austinabell, line 88: Is there a benefit to exposing all of these? Seems like we are just adding too much surface area to maintain

near-contract-standards/src/event.rs
```diff
@@ -0,0 +1,298 @@
...
+            owner_id: owner_id.into(),
+            token_ids: token_ids.into_iter().map(Into::into).collect(),
+            memo: memo.map(Into::into),
+        }
+    }
+}
+
+#[skip_serializing_none]
+#[derive(Serialize, Deserialize, Debug)]
+pub struct NftTransferData<'a> {
```
austinabell, line 78: Would be nice if there were docs on some of these, but that can always be added in follow up PR

near-contract-standards/src/event.rs
```diff
@@ -0,0 +1,298 @@
...
+        NearEvent::new_171_v1(Nep171EventKind::NftTransfer(data))
+    }
+
+    #[must_use = "don't forget to .emit() the event"]
+    pub fn nft_mint(data: Vec<NftMintData<'a>>) -> Self {
+        NearEvent::new_171_v1(Nep171EventKind::NftMint(data))
+    }
+
+    pub(crate) fn to_json_string(&self) -> String {
+        serde_json::to_string(self).unwrap()
```
austinabell, line 176: this should never fail, so would be slightly more ideal to just panic through host function to avoid file-specific data being included. Fine to have this done in a following PR though.

near-contract-standards/src/event.rs
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    pub fn to_json_event_string(&self) -> String {
+        format!("EVENT_JSON:{}", self.to_json_string())
+    }
+
+    pub fn emit(&self) {
```
austinabell, line 183: Maybe the others are intuitive, but at least this one should have a doc comment explaining that it is logging

near-contract-standards/src/event.rs
```diff
@@ -0,0 +1,298 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use near_sdk::AccountId;
+
+    const AUTHORIZED_ID_NONE: Option<String> = None;
+    const MEMO_NONE: Option<&String> = None;
```
austinabell, line 195: you don't need consts for this, esp if you remove generic from function

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -329,7 +354,7 @@ impl NonFungibleToken {
     ///   Typically the account will be the owner. If `None`, will not refund. This is useful for delaying refunding
     ///   until multiple tokens have been minted.
     ///
-    /// Returns the newly minted token
+    /// Returns the newly minted token and does not emit the mint event
```
austinabell, line 357: Why not?

austinabell (commented):
Sorry for long time to review, think there is a slight miscommunication about who owns standards

near-contract-standards/src/non_fungible_token/core/core_impl.rs
willemneal: Because in you could want emit an event for multiple mints.

near-contract-standards/src/event.rs
willemneal: This was from the original code from the indexer.  So no real reason to expose them.

near-contract-standards/src/event.rs
willemneal: This is just for readability of the tests.  But yes removing the generics would remove the need for this.

near-contract-standards/src/event.rs
willemneal: @frol @telezhnaya Thoughts?

near-contract-standards/src/event.rs
willemneal: @austinabell  I just redid it to use `&'a str` directly since the lifetimes of the logs are shorter than the caller's.

near-contract-standards/src/event.rs
```diff
@@ -51,64 +35,52 @@ pub enum Nep171EventKind<'a> {
...
-        O: Into<Cow<'a, str>>,
-        N: Into<Cow<'a, str>>,
-        T: Into<Cow<'a, str>>,
-        A: Into<Cow<'a, str>>,
-        M: Into<Cow<'a, str>>,
-    {
+    pub fn new(
+        old_owner_id: &'a AccountId,
+        new_owner_id: &'a AccountId,
+        token_ids: Vec<&'a str>,
```
austinabell, line 74: Why are these not `AccountId`s?

near-contract-standards/src/event.rs
```diff
@@ -0,0 +1,265 @@
...
+    use near_sdk::AccountId;
+
+    fn bob() -> AccountId {
+        AccountId::new_unchecked("bob".to_string())
+    }
+
+    fn alice() -> AccountId {
+        AccountId::new_unchecked("alice".to_string())
+    }
+    // const MEMO_NONE: Option<String> = None;
```
austinabell, line 168:
```suggestion
```

willemneal:
@austinabell I addressed most of your concerns.  I didn't change the visibility of the inner fields.

I also improved the API so that AccountId's are used directly ensuring that only valid ones are written to the event (well unless they do `new_unchecked`.

Removing `cow` reduced the size of the final contract by around 3K, so not too much code bloat, but still an improvement.

PTAL and let me know.

near-contract-standards/src/event.rs
willemneal: `token_id`'s?

austinabell: oh I didn't realize they are strings in internal impl. Can disregard this as it wouldn't come with this change

willemneal:
@austinabell Could really use this in the wild!  Also the FT events have been added to the standard to so I'll add that in a future PR.

near-contract-standards/src/event.rs
```diff
@@ -0,0 +1,264 @@
+use near_sdk::AccountId;
+use serde::{Deserialize, Serialize};
+use serde_with::skip_serializing_none;
+
+#[derive(Serialize, Debug)]
+#[serde(tag = "standard")]
+#[serde(rename_all = "snake_case")]
+pub enum NearEvent<'a> {
+    #[serde(borrow)]
```
austinabell, line 9: can remove all of these, only needed when using `Cow`

near-contract-standards/src/event.rs
```diff
@@ -0,0 +1,264 @@
...
+    }
+}
+
+#[skip_serializing_none]
+#[derive(Serialize, Deserialize, Debug)]
+pub struct NftTransferData<'a> {
+    #[serde(borrow)]
+    pub old_owner_id: &'a str,
+    #[serde(borrow)]
+    pub new_owner_id: &'a str,
```
austinabell, line 61: why not these as account ids?

austinabell (approved):
Feel free to change the things I've commented on. Going to follow this to make some other changes wrt what of the API is exposed to make it easier to maintain anyway so nbd if not

near-contract-standards/src/event.rs
willemneal: It is because of the introduction of the lifetime.  For example, when removed:

```rust
cannot infer an appropriate lifetime for lifetime parameter `'de` due to conflicting requirements
expected `MapAccess<'_>`
   found `MapAccess<'de>`
expected `<NftMintData<'a> as Deserialize<'_>>`
   found `<NftMintData<'_> as Deserialize<'_>>`
   ```

near-contract-standards/src/event.rs
willemneal: The AccountId references are used in the constructors.  If I add them here I get:

```rust
the trait bound `&'a AccountId: Deserialize<'_>` is not satisfied
the following implementations were found:
  <AccountId as Deserialize<'de>>
  ```

willemneal:
@austinabell @ChaoticTempest Thanks for reviewing, but it seems I have lost my ability to hit the final merge.

near-contract-standards/src/event.rs
austinabell: hm okay interesting, assumed the compiler could infer

near-contract-standards/src/event.rs
austinabell: yeah, the point before is that the deserialize isn't needed for anything. nbd can leave as is

austinabell:
> @austinabell @ChaoticTempest Thanks for reviewing, but it seems I have lost my ability to hit the final merge.

Because of the conflicts? Was just waiting for that

willemneal:
Literally not an option to merge for me.  I guess you have to be in the NEAR org to get that power.

austinabell:
> Literally not an option to merge for me. I guess you have to be in the NEAR org to get that power.

Nope, just needed to be merged manually. I couldn't have pulled it in either. Hope you don't mind me pushing to your branch to fix

willemneal:
No issues just glad to see it wrapping up.

frol:
@willemneal I am so happy to see how this turned out to be a nice API! I want to thank you for driving this! It is really tiresome work to get to perfection. I also apologize that I failed to provide the feedback in timely manner 🙏 

@miraclx Could you maybe take a look into why `the trait bound `&'a AccountId: Deserialize<'_>` is not satisfied`: https://github.com/near/near-sdk-rs/pull/627#discussion_r791252412 ?

austinabell:
> @willemneal I am so happy to see how this turned out to be a nice API! I want to thank you for driving this! It is really tiresome work to get to perfection. I also apologize that I failed to provide the feedback in timely manner 🙏
> 
> @miraclx Could you maybe take a look into why `the trait bound &'a AccountId: Deserialize<'_> is not satisfied`: [#627 (comment)](https://github.com/near/near-sdk-rs/pull/627#discussion_r791252412) ?

No worries, the API was updated in https://github.com/near/near-sdk-rs/commit/dce4bd8671e1dbb5f9cf74dfeef0a56e2667b2ab so there is no need for this. Deserializing into an AccountId reference is intended given our current `AccountId` API, so no need for @miraclx to look into anything here I don't think

PR #629, opened 2021-11-10
Add method in StorageManagement to check if given account is registered

ctindogaru:


mikedotexe (commented):
Unfortunately, I don't think we can modify these contract standards like this. There's been a codified standard for Storage Management that lives here:
https://nomicon.io/Standards/StorageManagement.html

This underwent a lot of scrutiny and community discussion, and I'd hate to return to NEP discussions for this change if we can avoid it.

I believe we can gain the same benefit if we use this function which you can see from the link above:

```
// Returns the StorageBalance structure of the valid `account_id`
// provided. Must panic if `account_id` is invalid.
//
// If `account_id` is not registered, must return `null`.
function storage_balance_of(account_id: string): StorageBalance|null {}
```

ctindogaru:
Closing this one. As Mike suggested, we can use `storage_balance_of ` instead. Thank you for the suggestion Mike!

PR #630, opened 2021-11-12
Allow deserialization for StorageBalance and StorageBalanceBounds

ctindogaru:
Currently, only serialization is allowed for `StorageBalance` and `StorageBalanceBounds`. This means that we cannot use the result returned from `storage_deposit`, `storage_withdraw`, `storage_balance_bounds` and `storage_balance_of` since we cannot deserialize it.

Ex: the following will result in a compilation error
`near_sdk::serde_json::from_slice::<StorageBalance>(&result).unwrap();` where `result` is the result returned by `storage_balance_of`.

### DIFF
--- a/near-contract-standards/src/storage_management/mod.rs
+++ b/near-contract-standards/src/storage_management/mod.rs
@@ -3,8 +3,6 @@ use near_sdk::json_types::U128;
 use near_sdk::serde::{Deserialize, Serialize};
 use near_sdk::AccountId;
 
-/// Implements both `serde` and `borsh` serialization.
-/// `serde` is typically useful when returning a struct in JSON format for a frontend.
 #[derive(BorshDeserialize, BorshSerialize, Serialize, Deserialize)]
 #[serde(crate = "near_sdk::serde")]
 pub struct StorageBalance {
--- a/near-contract-standards/src/storage_management/mod.rs
+++ b/near-contract-standards/src/storage_management/mod.rs
@@ -1,15 +1,20 @@
+use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
 use near_sdk::json_types::U128;
-use near_sdk::serde::Serialize;
+use near_sdk::serde::{Deserialize, Serialize};
 use near_sdk::AccountId;
 
-#[derive(Serialize)]
+/// Implements both `serde` and `borsh` serialization.
+/// `serde` is typically useful when returning a struct in JSON format for a frontend.
+#[derive(BorshDeserialize, BorshSerialize, Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


mikedotexe (approved):
This makes sense to me for users that are using the near_contract_standards crate. Seems like if it were me I might just add the struct to the contract I'm dealing with, but I could see someone wanting to do this.

near-contract-standards/src/storage_management/mod.rs
```diff
@@ -1,15 +1,20 @@
+use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
 use near_sdk::json_types::U128;
-use near_sdk::serde::Serialize;
+use near_sdk::serde::{Deserialize, Serialize};
 use near_sdk::AccountId;
 
-#[derive(Serialize)]
+/// Implements both `serde` and `borsh` serialization.
+/// `serde` is typically useful when returning a struct in JSON format for a frontend.
```
ChaoticTempest, line 7: We typically don't want these kind of docs to be public (denoted by the `///` comments). They're more of of side note when you look at the code. Public docs should be explaining what the struct should be doing instead, providing a few examples if necessary. If anything just remove these and add a link to the near docs explaining them at the very top:
```
// serialization differences between Borsh & Serde:
// https://www.near-sdk.io/contract-interface/serialization-interface
```

```suggestion
```

near-contract-standards/src/storage_management/mod.rs
```diff
@@ -1,15 +1,20 @@
...
+#[derive(BorshDeserialize, BorshSerialize, Serialize, Deserialize)]
 #[serde(crate = "near_sdk::serde")]
 pub struct StorageBalance {
     pub total: U128,
     pub available: U128,
 }
 
-#[derive(Serialize)]
+/// Implements both `serde` and `borsh` serialization.
+/// `serde` is typically useful when returning a struct in JSON format for a frontend.
```
ChaoticTempest, line 16:
```suggestion
```

ChaoticTempest (changes requested):

near-contract-standards/src/storage_management/mod.rs
mikedotexe: Unresolving conversation we we can add the snippet Phuong has here to the top of the file as suggested. Thanks!

austinabell (approved):
If you need this before this gets released, you can simply just create your own types which impl anything you need and move the inner data into it.

ctindogaru:
> If you need this before this gets released, you can simply just create your own types which impl anything you need and move the inner data into it.

Will use this as a workaround, thank you! Meanwhile, can we merge this?

ChaoticTempest (approved):

PR #632, opened 2021-11-17
refactor: Remove helper function and add safety docs

austinabell:
Helper function was only needed when there were more steps, not needed now as it somewhat hides the unsafety constraints of the operations. Added the safety docs as well to describe why this is safe.

### DIFF
--- a/near-sdk/src/environment/hash.rs
+++ b/near-sdk/src/environment/hash.rs
@@ -4,10 +4,6 @@ use crate::sys;
 
 const ATOMIC_OP_REGISTER: u64 = u64::MAX - 2;
 
-fn read_register_fixed(register_id: u64, buf: &mut [MaybeUninit<u8>]) {
-    unsafe { sys::read_register(register_id, buf.as_ptr() as _) }
-}
-
 mod private {
     /// Seal `CryptoHasher` implementations to limit usage to the builtin implementations
     pub trait Sealed {}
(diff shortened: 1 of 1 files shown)


near-sdk/src/environment/hash.rs
```diff
@@ -36,9 +32,14 @@ impl CryptoHasher for Sha256 {
     fn hash(ingest: &[u8]) -> Self::Digest {
         unsafe { sys::sha256(ingest.len() as _, ingest.as_ptr() as _, ATOMIC_OP_REGISTER) };
 
-        let mut hash = [MaybeUninit::uninit(); 32];
-        read_register_fixed(ATOMIC_OP_REGISTER, &mut hash);
-        unsafe { std::mem::transmute(hash) }
+        let mut hash = [MaybeUninit::<u8>::uninit(); 32];
```
matklad, line 35: Super ultra minor, but i'd move `let mut hash` inside `unsafe` block, as it's the part of overall "read 32 bytes" operation

near-sdk/src/environment/hash.rs
```diff
@@ -36,9 +32,14 @@ impl CryptoHasher for Sha256 {
...
+        //* SAFETY: sha256 syscall will always generate 32 bytes inside of the atomic op register
```
matklad, line 36: :+1: for adding a `SAFETY` section. Not sure I've seen `//*` sytnax before, usually it's done without it:

https://github.com/matklad/rust/blob/6063405c1b502f1c72262278e3b710854c10be25/library/alloc/src/alloc.rs#L206-L211

no strong opinion here.

matklad (approved):

near-sdk/src/environment/hash.rs
austinabell: I generally try to minimize the amount of code inside of an `unsafe` block, but I'm fine with making this change. I might as well put the whole impl in the same block for readability if I do this.

near-sdk/src/environment/hash.rs
austinabell: Screen Shot 2021-11-17 at 11 30 03 AM

My preference comes from seeing it in some codebases and it stands out more than regular docs, especially if you have an extension that changes the colour of different doc types, as shown in the screenshot above. I've used this style across other examples in the SDK and am happy to change, but everything should probably be consistent.

near-sdk/src/environment/hash.rs
matklad: Ah, yeah, now I see why the code written the way it is!

ChaoticTempest (approved):
👍

PR #633, opened 2021-11-17
fix: consistent naming to only using function over method

austinabell:
inline with https://github.com/near/docs/issues/905, naming within the SDK is mixed, even within the same function, so this changes method terminology to function.

This should only be a breaking change for mock receipts, others are all positional and I don't think the promise enums are exposed in any way, which I will follow up on now. (They were private)

### DIFF
--- a/near-sdk-sim/src/runtime.rs
+++ b/near-sdk-sim/src/runtime.rs
@@ -347,7 +347,12 @@ impl RuntimeStandalone {
     }
 
     /// Returns a ViewResult containing the value or error and any logs
-    pub fn view_method_call(&self, account_id: &str, method_name: &str, args: &[u8]) -> ViewResult {
+    pub fn view_method_call(
+        &self,
+        account_id: &str,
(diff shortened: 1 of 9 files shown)


near-sdk-sim/src/runtime.rs
```diff
@@ -347,7 +347,12 @@ impl RuntimeStandalone {
     }
 
     /// Returns a ViewResult containing the value or error and any logs
-    pub fn view_method_call(&self, account_id: &str, method_name: &str, args: &[u8]) -> ViewResult {
+    pub fn view_method_call(
```
ChaoticTempest, line 350: should we change the name here to `view_function_call`? Or is this PR only for args

austinabell: Ah I meant to mention this in the description. I didn't change anything in sdk-sim because it's being deprecated and no reason to create api breakages for something like this

ChaoticTempest (approved):

PR #636, opened 2021-11-19
add must_use; updated examples

### LINKED ISSUE #635: Add `#[must_use]` for `call!()` and `view!()` returns
Usages of `call!()` and `view!()` likely are being used in testing which can signal success and some failures. That signal likely should always be checked against, so a `#<span class="error">[must_use]</span>` attribute on those structures could help in identifying calls that are missing the check. 
 This would be a similar behaviour to the <span class="error">[`Result`]</span>([https://doc.rust-lang.org/src/core/result.rs.html#500](https://doc.rust-lang.org/src/core/result.rs.html#500)) type.

swfsql:
* Closes #635 
* Adds `#[must_use]` on `near_sdk_sim::outcome::ExecutionResult` and `ViewResult`.
* Update `examples/`.

### DIFF
--- a/examples/fungible-token/tests/sim/with_macros.rs
+++ b/examples/fungible-token/tests/sim/with_macros.rs
@@ -22,12 +22,8 @@ fn simulate_simple_transfer() {
 
     // Transfer from root to alice.
     // Uses default gas amount, `near_sdk_sim::DEFAULT_GAS`
-    call!(
-        root,
-        ft.ft_transfer(alice.account_id(), transfer_amount.into(), None),
-        deposit = 1
-    )
-    .assert_success();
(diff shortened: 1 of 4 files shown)


austinabell (commented):
I think this is a reasonable idea, I'm just not certain that the result should be forced to be used. The main concern I have with this is that this is a pretty breaking change and we are in the process of deprecating sim tests.

cc @ChaoticTempest Wdyt? Also something to consider for workspaces API if we aren't using std `Result`

ChaoticTempest:
Yeah, I agree with @austinabell here with the breaking changes. It'll soon replace these tests, but I think we can keep the example changes but remove the `#[must_use]`

austinabell:
Seems like this PR is stale, feel free to re-open if you'd like to update it to allow it to come in :)

PR #639, opened 2021-11-22
chore: update borsh and nearcore crate versions in SDK

austinabell:
- Updates borsh to 0.9
- Updates `near-vm-logic` and `near-primitives-core` to new version `0.10`

Did not update near-sdk-sim because the dependencies needed were not released and this will be deprecated soon.

There seems to be a lot of new dependencies added for these, which isn't a huge issue but will increase compile times for tests. cc @miraclx if you're curious the diff on the SDK side from the releases :)


austinabell:
buildkite failing because rust version is too low to handle 2021 edition from new nearcore crate releases

near-sdk/src/test_utils/context.rs
```diff
@@ -29,10 +31,10 @@ impl VMContextBuilder {
     pub fn new() -> Self {
         Self {
             context: VMContext {
-                current_account_id: alice().into(),
-                signer_account_id: bob().into(),
+                current_account_id: alice().as_str().parse().unwrap(),
+                signer_account_id: bob().as_str().parse().unwrap(),
                 signer_account_pk: vec![0u8; 32],
-                predecessor_account_id: bob().into(),
+                predecessor_account_id: bob().as_str().parse().unwrap(),
```
itegulov, line 37: Wouldn't it make more sense to implement two-way `From`s for both `AccountId`s? Something like this:
```rust
impl From<near_vm_logic::types::AccountId> for AccountId {
    fn from(id: near_vm_logic::types::AccountId) -> Self {
        id.into()
    }
}

impl From<AccountId> for near_vm_logic::types::AccountId {
    fn from(id: AccountId) -> Self {
        id.into()
    }
}
```

austinabell: Because there is no direct conversion between the two. There is no ability to create a vm_logic AccountId without error checking. Can add a `TryFrom` impl for these, but that didn't seem like it would be worth it, given that it's just used internally and is basically the same amount of code.

near-sdk/src/environment/mock/external.rs
```diff
@@ -64,14 +65,18 @@ impl External for SdkExternal {
...
+        receiver_id: VmAccountId,
+    ) -> Result<u64> {
         if let Some(index) = receipt_indices.iter().find(|&&el| el >= self.receipts.len() as u64) {
             return Err(HostError::InvalidReceiptIndex { receipt_index: *index }.into());
         }
         let res = self.receipts.len() as u64;
         self.receipts.push(Receipt {
             receipt_indices,
-            receiver_id: AccountId::new_unchecked(receiver_id),
+            receiver_id: AccountId::new_unchecked(String::from(receiver_id)),
```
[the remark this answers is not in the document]
austinabell: Yeah I'll make the change, but this isn't a public-facing API, no one uses the vm_logic AccountId, and it's generally bad practice to have trait impls for only specific architectures (we cannot have the impl for `wasm32-unknown-unknown`)

austinabell:
commit for suggested change @itegulov [`9e77d35` (#639)](https://github.com/near/near-sdk-rs/pull/639/commits/9e77d35842b31c59ba26c3dd5a9d264a96eb891d). Not convinced this is better since just used internally, but I don't have a strong opinion so I'll just keep the change.

near-sdk/src/environment/mock/external.rs
```diff
@@ -192,11 +197,13 @@ impl External for SdkExternal {
...
         receipt_index: u64,
-        beneficiary_id: String,
+        beneficiary_id: VmAccountId,
     ) -> Result<()> {
-        self.receipts.get_mut(receipt_index as usize).unwrap().actions.push(
-            VmAction::DeleteAccount { beneficiary_id: AccountId::new_unchecked(beneficiary_id) },
-        );
+        self.receipts
+            .get_mut(receipt_index as usize)
+            .unwrap()
```
ChaoticTempest, line 204: I think we can just change this unwrap to yield an invalid receipt index instead for None
```suggestion
            .ok_or_else(|| HostError::InvalidReceiptIndex { receipt_index })?
```

austinabell: Fair point! I didn't change this in the update and this is just used for testing, but this might yield a better error

ChaoticTempest (approved):
LGTM

PR #646, opened 2021-11-25
feat: Implement fixed length hash functions and optimize existing imlementations

### LINKED ISSUE #645: Switch API for env hash functions to return array
Currently, they all return Vec<u8> and do an unnecessary call to `sys::register_len` even though the length is static and always known.

Probably the best API strategy would be to add new functions under new names (I actually don't know what would be a good name to change to, maybe `<algo>_hash`?) and deprecate the current ones.

Reading into fixed-length array logic is done in https://github.com/near/near-sdk-rs/blob/master/near-sdk/src/environment/hash.rs. Probably the logic to read into a fixed-length array can be abstracted into a function with a const generic for hash size to avoid code duplication.

austinabell:
- Adds fixed length array versions of `env` hash functions (no reason to allocate if we know the length)
- Optimizes the existing impl to use this to avoid an extra syscall to `register_len`
  - I am assuming the compiler can optimize the creation on stack and copy to heap, but if not the cost should be trivial and these functions will probably be deprecated in future
- Added helper functions for reading to fixed-length arrays for re-use

I don't necessarily like suffixing `_hash`, but ideally this is a smooth transition to avoid hard-breaking changes if users expect a `Vec<u8>` over the array, even though this comes with a major version bump. Name suggestions welcome!

closes #645


ChaoticTempest:
how about something like `_unboxed` instead of `_hash`? I think at least that would convey the difference, and even if we deprecate the one without the postfix, I think it still makes sense

austinabell:
> how about something like `_unboxed` instead of `_hash`? I think at least that would convey the difference, and even if we deprecate the one without the postfix, I think it still makes sense

I personally am confused with this wording, because I don't see what boxed refers to here or what `unboxed` means. What do you mean by this?

ChaoticTempest:
@austinabell ahh my bad, I meant to say that `_unboxed` would show that you're returning something that isn't boxed/allocated inside something like a `Vec<u8>`, which would show the difference between the two functions. If you're confused by it, then it probably isn't good naming either then

ChaoticTempest (approved):

ChaoticTempest:
> Could also suffix something like _array, to be descriptive, but I just thought that _hash would be most clear because at least it's explaining that these are hashes which might not be intuitive for those unfamiliar.

But aren't they both hashes though? I think `_array` works since it should distinguish what's the different between the two return types

austinabell:
> > Could also suffix something like _array, to be descriptive, but I just thought that _hash would be most clear because at least it's explaining that these are hashes which might not be intuitive for those unfamiliar.
> 
> But aren't they both hashes though? I think `_array` works since it should distinguish what's the different between the two return types

Was just giving a benefit to using hash suffix. I'll switch to array suffix since it might be more clear :)

ChaoticTempest (approved):
New changes LGTM

PR #647, opened 2021-11-29
Implement `Debug` for collection types

### LINKED ISSUE #564: Implement debug for collection types
`UnorderedMap`, `UnorderedSet`, `LookupMap`, `LookupSet`,  `TreeMap` (everything but `Vector`) do not have a debug implementation when they could. The info from debug wouldn't be that useful, so could even just omit most data from debug with a custom implementation, but is a benefit when developing to have this.

Also `store::LazyOption` debug can be added

itegulov:
Closes #564 

This PR implements `Debug` trait in a uniform fashion for `UnorderedMap`, `UnorderedSet`, `LookupMap`, `LookupSet`, `TreeMap`, `LegacyTreeMap` and `LazyOption`.

Where possible I tried to provide a human-readable list of elements as if it was just a plain Rust collection, but for some of them (`LookupMap` and `LookupSet`) it is impossible to do by design, so I think the best we can do is to show the known internal stuff. Implementations for `TreeMap` and `LegacyTreeMap` were actually already present, I just had to move them outside of test scope. I made one compromise in tests: in order not to complicate the test logic too much, most of them assume that the values/pairs are going to be displayed in the order of their insertion.

Also, this is just a side thought, but while implementing the traits I noticed that we could split the implementations for [`UnorderedSet`](https://github.com/near/near-sdk-rs/blob/master/near-sdk/src/collections/unordered_set.rs#L116), [`UnorderedMap`](https://github.com/near/near-sdk-rs/blob/master/near-sdk/src/collections/unordered_map.rs#L150-L151) and others into two separate set of functions when `K` and `V` are either `BorshSerialize` or `BorshDeserialize` (the way it is done in Vector: see [[1]](https://github.com/near/near-sdk-rs/blob/master/near-sdk/src/collections/vector.rs#L156), [[2]](https://github.com/near/near-sdk-rs/blob/master/near-sdk/src/collections/vector.rs#L178)).


near-sdk/src/collections/legacy_tree_map.rs
```diff
@@ -584,6 +584,36 @@ where
...
+            .field("key", &self.key)
+            .field("lft", &self.lft)
+            .field("rgt", &self.rgt)
+            .field("ht", &self.ht)
+            .finish()
+    }
+}
+
+#[cfg(feature = "expensive-debug")]
+impl<K, V> std::fmt::Debug for LegacyTreeMap<K, V>
```
austinabell, line 604: Actually, we can just leave out the impl for `LegacyTreeMap`. It is deprecated and should probably be marked as such so we should not be making any functional changes to it

near-sdk/src/collections/lookup_map.rs
```diff
@@ -284,4 +284,22 @@ mod tests {
...
+        for _ in 0..10 {
+            let key = rng.gen::<u64>();
+            let value = rng.gen::<u64>();
+            baseline.push((key, value));
+            map.insert(&key, &value);
+        }
+
+        assert_eq!(
+            format!("{:?}", map),
+            format!("LookupMap {{ key_prefix: {:?}, el: PhantomData }}", map.key_prefix)
```
austinabell, line 302: This is maybe one you want to have a custom impl so you leave out the `PhantomData`

near-sdk/src/collections/lookup_map.rs
```diff
@@ -284,4 +284,22 @@ mod tests {
...
+        );
```
austinabell, line 303: It doesn't make sense to me why you would add random values to test the debug impl? Esp in this case since the values aren't included in the debug output.

near-sdk/src/collections/lookup_set.rs
```diff
@@ -204,4 +204,21 @@ mod tests {
...
+        for _ in 0..10 {
+            let value = rng.gen::<u64>();
+            baseline.push(value);
+            set.insert(&value);
+        }
+
+        assert_eq!(
+            format!("{:?}", set),
+            format!("LookupSet {{ element_prefix: {:?}, el: PhantomData }}", set.element_prefix)
+        );
```
austinabell, line 222: Same comments as the previous two

near-sdk/src/collections/tree_map.rs
```diff
@@ -580,6 +580,36 @@ where
...
+    K: std::fmt::Debug + Ord + Clone + BorshSerialize + BorshDeserialize,
+    V: std::fmt::Debug + BorshSerialize + BorshDeserialize,
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("TreeMap")
+            .field("root", &self.root)
+            .field("tree", &self.tree.iter().collect::<Vec<Node<K>>>())
+            .finish()
+    }
+}
```
austinabell, line 595:
```suggestion
impl<K, V> std::fmt::Debug for TreeMap<K, V>
where
    K: std::fmt::Debug + Ord + Clone + BorshSerialize + BorshDeserialize,
    V: std::fmt::Debug + BorshSerialize + BorshDeserialize,
{
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.debug_struct("TreeMap")
            .field("root", &self.root)
            .field("tree", &self.tree)
            .finish()
    }
}
```

You need the debug impl for when the feature is not `expensive-debug`. There is no need to manually collect the nodes, the `Vector` debug implementation will do this in `expensive-debug` and do the lighter impl when it is not.

near-sdk/src/collections/tree_map.rs
```diff
@@ -580,6 +580,36 @@ where
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Node")
+            .field("id", &self.id)
+            .field("key", &self.key)
+            .field("lft", &self.lft)
+            .field("rgt", &self.rgt)
+            .field("ht", &self.ht)
+            .finish()
+    }
+}
```
austinabell, line 611: This can just be derived

near-sdk/src/collections/tree_map.rs
```diff
@@ -1777,4 +1777,23 @@ mod tests {
 
         QuickCheck::new().tests(300).quickcheck(prop as Prop);
     }
+
+    #[test]
+    #[cfg(feature = "expensive-debug")]
```
austinabell, line 1782:
```suggestion
```

If you make previous change

near-sdk/src/collections/tree_map.rs
```diff
@@ -1777,4 +1777,23 @@ mod tests {
...
+            let key = rng.gen::<u64>();
+            let value = rng.gen::<u64>();
+            baseline.push((key, value));
+            map.insert(&key, &value);
+        }
+
+        assert_eq!(
+            format!("{:?}", map),
+            format!("TreeMap {{ root: {:?}, tree: {:?} }}", map.root, map.tree)
+        );
```
austinabell, line 1797: Again, seems more useful if a few values are added and the actual output is verified, instead of relying on the inner impls which could change and break this format without knowing

near-sdk/src/collections/unordered_set.rs
```diff
@@ -171,6 +171,16 @@ where
...
 
+#[cfg(feature = "expensive-debug")]
+impl<T> std::fmt::Debug for UnorderedSet<T>
+where
+    T: std::fmt::Debug + BorshSerialize + BorshDeserialize,
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.to_vec().fmt(f)
+    }
+}
```
austinabell, line 182: Same comment as before, where you want this impl for when the expensive debug isn't set, and just let the `Vector` handle this. Also, you should include the prefix in the debug impl

near-sdk/src/collections/unordered_set.rs
```diff
@@ -330,4 +340,19 @@ mod tests {
...
+        let mut rng = rand_xorshift::XorShiftRng::seed_from_u64(4);
+        let mut set = UnorderedSet::new(b"m");
+        let mut baseline = vec![];
+        for _ in 0..10 {
+            let value = rng.gen::<u64>();
+            baseline.push(value);
+            set.insert(&value);
+        }
+
+        assert_eq!(format!("{:#?}", set), format!("{:#?}", baseline));
```
austinabell, line 356: same comment as before

near-sdk/src/store/lazy_option/impls.rs
```diff
@@ -30,3 +30,13 @@ where
...
+
+#[cfg(feature = "expensive-debug")]
+impl<T> std::fmt::Debug for LazyOption<T>
+where
+    T: std::fmt::Debug + BorshSerialize + BorshDeserialize,
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.get().fmt(f)
+    }
+}
```
austinabell, line 42:
```suggestion
impl<T> std::fmt::Debug for LazyOption<T>
where
    T: std::fmt::Debug + BorshSerialize + BorshDeserialize,
{
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        if cfg!(feature = "expensive-debug") {
            self.get().fmt(f)
        } else {
            // Format just the metadata
        }
    }
}
```

Maybe there is a misconception, this feature should be used to distinguish between an expensive but more understandable debug output. For example, `Vector` expensive debug will load all values, allocate them in a vector, and then print that when otherwise it would just print metadata. Debug impls shouldn't be expensive by default, like in this, but expensive debug can incur storage costs for debugging reasons to see the data in testing.

austinabell (changes requested):

near-sdk/src/collections/lookup_map.rs
itegulov: I left it in because the existing Vector implementation had it in as well. I will delete it from both places then.

near-sdk/src/collections/tree_map.rs
itegulov: Hmm, the problem with this is that the tree elements won't be displayed in a sorted order (which I personally would expect from the Debug impl for TreeMap). Alternatively, I could do something like this:
```suggestion
#[cfg(feature = "expensive-debug")]
impl<K, V> std::fmt::Debug for TreeMap<K, V>
where
    K: std::fmt::Debug + Ord + Clone + BorshSerialize + BorshDeserialize,
    V: std::fmt::Debug + BorshSerialize + BorshDeserialize,
{
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.debug_struct("TreeMap")
            .field("root", &self.root)
            .field("elements", &self.to_vec())
            .finish()
    }
}
```

And derive a default Debug impl when expensive-debug is not enabled. What do you think? The downside is that we are losing information about tree's structure (as the actual nodes won't get printed anymore).

near-sdk/src/store/lazy_option/impls.rs
itegulov: Thanks for the clarification. Indeed, I didn't realize that we want both expensive and non-expensive Debug implementations.

near-sdk/src/collections/lookup_map.rs
austinabell: yeah, feel free to update the old `Vector` to remove this, but it's not a big deal as this impl is being deprecated in favour of the one in `near_sdk::store`

near-sdk/src/collections/tree_map.rs
austinabell: I mean, if you wanted to construct an in memory data structure that actually represents the data, you would collect a map (`BTreeMap<K, V>`) from the `TreeMap` iterator. You could optimize this by either iterating all keys unordered since BTreeMap enforces ordering or using some collection which keeps insertion order. This is all for expensive debug so I don't think it matters much the performance.

There aren't any strict rules around how the data is formatted, and I assume expensive debug was just added for very specific use cases, so this doesn't matter that much.

itegulov: Ok, I see your point and I see now that it probably makes more sense to format everything in a human-readable way that is close to the collections' internal structure rather than trying to come up with something more general.

itegulov:
@austinabell I have addressed your comments, should be better now. PTAL whenever you can.

austinabell (approved):

PR #648, opened 2021-11-29
Implement safe API for RIPEMD-160

### LINKED ISSUE #643: Implement safe API for ripemd160
This could have use cases, so we should support this interface. Unlike the APIs that exist within the SDK for other hash functions, an array of 20 bytes should be returned (`[u8; 20])`, since it is guaranteed this is always the length and fixed.

API should look roughly like:
```
pub fn ripemd(value: &[u8]) -> [u8; 20] {
 // ...
}
```

and live in `near_sdk::env`

itegulov:
Closes #643

Based on top of https://github.com/near/near-sdk-rs/pull/646.


austinabell (approved):
LGTM, we have to wait on #646 anyway, and this might change naming of the function (`_hash` suffix, which isn't really needed here)

Issue #649, opened 2021-11-29
Implement the Events Standard in the Core Standards


BenKurrek:
The new [Events Standard](https://nomicon.io/Standards/NonFungibleToken/Event.html) has been released and the core standards need to implement `NftMintLog` and `NftTransferLog`

PR #652, opened 2021-11-30
NFT Events Standard

BenKurrek:
Implements the events standard for the non-fungible-token core standards. 

Fixes (https://github.com/near/near-sdk-rs/issues/649)

### DIFF
--- a/near-contract-standards/src/non_fungible_token/core/core_impl.rs
+++ b/near-contract-standards/src/non_fungible_token/core/core_impl.rs
@@ -360,13 +360,10 @@ impl NonFungibleToken {
             }])
         };
 
-        let mint_log = NftMintData { owner_id, token_ids, memo: None };
-        let event_log = NearEvent::nft_mint(vec![mint_log]);
-
         //convert the struct of type EventLogJson to actual json so we can log it
         let serialized_nft_mint_log = serde_json::to_string(&nft_mint_log).expect("Unable to convert NFT mint log to JSON");
         //log the serialized json
-        env::log_str(&format!("EVENT_JSON:{:?}", serialized_nft_mint_log.to_json_string())); 
+        env::log_str(&format!("EVENT_JSON:{:?}", serialized_nft_mint_log)); 
         env::log_str(&serialized_nft_mint_log); 
         // Return any extra attached deposit not used for storage
         refund_deposit(env::storage_usage() - initial_storage_usage);
--- a/near-contract-standards/src/non_fungible_token/core/core_impl.rs
+++ b/near-contract-standards/src/non_fungible_token/core/core_impl.rs
@@ -1,7 +1,7 @@
 use super::resolver::NonFungibleTokenResolver;
 use crate::non_fungible_token::core::NonFungibleTokenCore;
-use crate::non_fungible_token::events::{EventLog, EventLogVariant, NftTransferLog};
 use crate::non_fungible_token::events::NftMintLog;
+use crate::non_fungible_token::events::{EventLog, EventLogVariant, NftTransferLog};
 use crate::non_fungible_token::metadata::{TokenMetadata, NFT_METADATA_SPEC, NFT_STANDARD_NAME};
 use crate::non_fungible_token::token::{Token, TokenId};
 use crate::non_fungible_token::utils::{
@@ -279,27 +279,28 @@ impl NonFungibleToken {
         }
 
         //default the authorized ID to be None for the logs
-        let mut authorized_id = None; 
+        let mut authorized_id = None;
         //if the approval ID was provided, set the authorized ID equal to the sender
-        if !approval_id.is_none() {
-            authorized_id = Some(sender_id.to_string()); 
+        if approval_id.is_some() {
+            authorized_id = Some(sender_id.to_string());
         }
 
         //construct the mint log as per the events standard
         let nft_transfer_log: EventLog = EventLog {
             standard: NFT_STANDARD_NAME.to_string(), //standard name ("nep171")
-            version: NFT_METADATA_SPEC.to_string(), //version of the standard ("nft-1.0.0")
-            event: EventLogVariant::NftTransfer( vec![ NftTransferLog { //the data related with the event stored in a vector
-                authorized_id: authorized_id,
+            version: NFT_METADATA_SPEC.to_string(),  //version of the standard ("nft-1.0.0")
+            event: EventLogVariant::NftTransfer(vec![NftTransferLog {
+                //the data related with the event stored in a vector
+                authorized_id,
                 old_owner_id: owner_id.to_string(),
                 new_owner_id: receiver_id.to_string(),
                 token_ids: vec![token_id.to_string()],
-                memo: memo,
-            }])
+                memo,
+            }]),
         };
 
         //log the serialized json
-        env::log_str(&nft_transfer_log.to_string()); 
+        env::log_str(&nft_transfer_log.to_string());
 
         // return previous owner & approvals
         (owner_id, approved_account_ids)
@@ -370,20 +371,21 @@ impl NonFungibleToken {
         // Approval Management extension: return empty HashMap as part of Token
         let approved_account_ids =
             if self.approvals_by_id.is_some() { Some(HashMap::new()) } else { None };
-        
+
         //construct the mint log as per the events standard
         let nft_mint_log: EventLog = EventLog {
             standard: NFT_STANDARD_NAME.to_string(), //standard name ("nep171")
-            version: NFT_METADATA_SPEC.to_string(), //version of the standard ("nft-1.0.0")
-            event: EventLogVariant::NftMint( vec![ NftMintLog { //the data related with the event stored in a vector
+            version: NFT_METADATA_SPEC.to_string(),  //version of the standard ("nft-1.0.0")
+            event: EventLogVariant::NftMint(vec![NftMintLog {
+                //the data related with the event stored in a vector
                 owner_id: owner_id.to_string(), //owner of the token
                 token_ids: vec![token_id.to_string()], //vector of token IDs that were minted
-                memo: None, //an optional memo to include
-            }])
+                memo: None,                     //an optional memo to include
+            }]),
         };
 
         //log the serialized json
-        env::log_str(&nft_mint_log.to_string()); 
+        env::log_str(&nft_mint_log.to_string());
         // Return any extra attached deposit not used for storage
         refund_deposit(env::storage_usage() - initial_storage_usage);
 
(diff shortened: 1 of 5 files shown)


near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -272,10 +274,33 @@ impl NonFungibleToken {
         self.internal_transfer_unguarded(token_id, &owner_id, receiver_id);
 
         log!("Transfer {} from {} to {}", token_id, sender_id, receiver_id);
-        if let Some(memo) = memo {
+        if let Some(memo) = memo.clone() {
             log!("Memo: {}", memo);
```
ChaoticTempest, line 278: No need to clone:
```suggestion
        if let Some(memo) = memo.as_ref() {
            log!("Memo: {}", memo);
```

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -319,7 +344,7 @@ impl NonFungibleToken {
             env::panic_str("token_id must be unique");
         }
 
-        let owner_id: AccountId = token_owner_id;
+        let owner_id: AccountId = token_owner_id.clone();
```
ChaoticTempest, line 347: Seems like this one's an unnecessary clone. Everything compiles fine without it
```suggestion
        let owner_id: AccountId = token_owner_id;
```

near-contract-standards/src/non_fungible_token/metadata.rs
```diff
@@ -5,6 +5,8 @@ use near_sdk::serde::{Deserialize, Serialize};
 
 /// This spec can be treated like a version of the standard.
 pub const NFT_METADATA_SPEC: &str = "nft-1.0.0";
+/// This name of the NFT standard we're using
```
ChaoticTempest, line 8: nit: public comment, so better to be grammatically correct about
```suggestion
/// This is the name of the NFT standard we're using
```

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -345,7 +370,20 @@ impl NonFungibleToken {
         // Approval Management extension: return empty HashMap as part of Token
         let approved_account_ids =
             if self.approvals_by_id.is_some() { Some(HashMap::new()) } else { None };
+        
+        //construct the mint log as per the events standard
+        let nft_mint_log: EventLog = EventLog {
+            standard: NFT_STANDARD_NAME.to_string(), //standard name ("nep171")
+            version: NFT_METADATA_SPEC.to_string(), //version of the standard ("nft-1.0.0")
+            event: EventLogVariant::NftMint( vec![ NftMintLog { //the data related with the event stored in a vector
+                owner_id: token_owner_id.to_string(), //owner of the token
```
ChaoticTempest, line 379: Forgot to mention you'd also need to rename the following after getting rid of the clone:
```suggestion
                owner_id: owner_id.to_string(), //owner of the token
```

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -272,10 +274,33 @@ impl NonFungibleToken {
...
         }
 
+        //default the authorized ID to be None for the logs
+        let mut authorized_id = None; 
+        //if the approval ID was provided, set the authorized ID equal to the sender
+        if !approval_id.is_none() {
+            authorized_id = Some(sender_id.to_string()); 
+        }
+
+        //construct the mint log as per the events standard
```
ChaoticTempest, line 288: Clippy is complaining about style such as the following:
```suggestion
        // construct the mint log as per the events standard
```
Recommended you run `cargo fix` and `cargo format`, then check afterwards to see if it fixed everything with `cargo clippy`

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -272,10 +274,33 @@ impl NonFungibleToken {
         self.internal_transfer_unguarded(token_id, &owner_id, receiver_id);
 
         log!("Transfer {} from {} to {}", token_id, sender_id, receiver_id);
-        if let Some(memo) = memo {
+        if let Some(memo) = memo.as_ref() {
             log!("Memo: {}", memo);
         }
 
+        //default the authorized ID to be None for the logs
```
austinabell, line 281:
```suggestion
        // Default the authorized ID to be None for the logs.
```

Stylistic difference but would you mind changing comments to be like this to be consistent with the rest of the repo?

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -272,10 +274,33 @@ impl NonFungibleToken {
...
+        let mut authorized_id = None; 
+        //if the approval ID was provided, set the authorized ID equal to the sender
+        if !approval_id.is_none() {
+            authorized_id = Some(sender_id.to_string());
```
austinabell, line 285:
```suggestion
        if approval_id.is_some() {
            authorized_id = Some(sender_id.to_string()); 
```

austinabell: Should this not be an `AccountId` instead of a string?

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -272,10 +274,33 @@ impl NonFungibleToken {
...
+        }
+
+        //construct the mint log as per the events standard
+        let nft_transfer_log: EventLog = EventLog {
+            standard: NFT_STANDARD_NAME.to_string(), //standard name ("nep171")
+            version: NFT_METADATA_SPEC.to_string(), //version of the standard ("nft-1.0.0")
+            event: EventLogVariant::NftTransfer( vec![ NftTransferLog { //the data related with the event stored in a vector
+                authorized_id: authorized_id,
+                old_owner_id: owner_id.to_string(),
+                new_owner_id: receiver_id.to_string(),
```
austinabell, line 295: same with these, can/should they not be an `AccountId` for some reason?

near-contract-standards/src/non_fungible_token/core/core_impl.rs
BenKurrek: will do - didn't know that was a thing sorry :)

near-contract-standards/src/non_fungible_token/events.rs
```diff
@@ -0,0 +1,149 @@
...
+#[derive(Serialize, Deserialize, Debug)]
+/*
+    outputs JSON in the following order: <tag>: <field>, <content>: <field_data>
+    In this case an example would be: "event": "NftMint", "data": <NftMintLog Vector Data>
+*/
+#[serde(tag = "event", content = "data")] 
+//turns all the fields into snake case (NftMint --> nft_mint)
+#[serde(rename_all = "snake_case")] 
+#[serde(crate = "near_sdk::serde")]
+//makes it so that adding new fields isn't a breaking change
```
austinabell, line 16: I personally don't think the comments here are that helpful and make it harder to read this block. Can keep if you think they are useful, but this was confusing for me to read

near-contract-standards/src/non_fungible_token/events.rs
```diff
@@ -0,0 +1,149 @@
+use std::fmt;
+
+use near_sdk::serde::{Deserialize, Serialize};
+
+//enum that represents the data type of the EventLogJson.
+//the enum can either be an NftMintLog or an NftTransferLog
```
austinabell, line 6:
```suggestion
/// Enum that represents the data type of the EventLog.
/// The enum can either be an NftMint or an NftTransfer.
```

When documenting anything exposed, make sure to use `///` so that they show up in rustdocs

near-contract-standards/src/non_fungible_token/events.rs
```diff
@@ -0,0 +1,149 @@
...
+    NftTransfer(Vec<NftTransferLog>),
+}
+
+// Interface to capture data
+// about an event
+// Arguments
+// * `standard`: name of standard e.g. nep171
+// * `version`: e.g. 1.0.0
+// * `event`: string
+// * `data`: associate event data
```
austinabell, line 29:
```suggestion
/// Interface to capture data about an event
/// Arguments:

// * `standard`: name of standard e.g. nep171
// * `version`: e.g. 1.0.0
// * `event`: string
// * `data`: associate event data
```

near-contract-standards/src/non_fungible_token/core/core_impl.rs
BenKurrek: done

near-contract-standards/src/non_fungible_token/events.rs
```diff
@@ -0,0 +1,149 @@
...
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(crate = "near_sdk::serde")]
+pub struct EventLog {
+    pub standard: String,
+    pub version: String,
+
+    //don't have "event": {<EventLogVariant>} in the JSON, just have the contents {<EventLogVariant>}
```
austinabell, line 35:
```suggestion
    // `flatten` to not have "event": {<EventLogVariant>} in the JSON, just have the contents of {<EventLogVariant>}.
```

I don't think you need to explain the serde format in comments, but here is a suggestion if you do want to keep

near-contract-standards/src/non_fungible_token/events.rs
```diff
@@ -0,0 +1,149 @@
...
+            &serde_json::to_string(self).map_err(|_| fmt::Error)?
+        ))
+    }
+}
+
+// An event log to capture token minting
+// Arguments
+// * `owner_id`: "account.near"
+// * `token_ids`: ["1", "abc"]
+// * `memo`: optional message
```
austinabell, line 53:
```suggestion
/// An event log to capture token minting
///
/// Arguments
/// * `owner_id`: "account.near"
/// * `token_ids`: ["1", "abc"]
/// * `memo`: optional message
```

near-contract-standards/src/non_fungible_token/events.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub memo: Option<String>,
+}
+
+// An event log to capture token transfer
+// Arguments
+// * `authorized_id`: approved account to transfer
+// * `old_owner_id`: "owner.near"
+// * `new_owner_id`: "receiver.near"
+// * `token_ids`: ["1", "12345abc"]
+// * `memo`: optional message
```
austinabell, line 71: same as previous

near-contract-standards/src/non_fungible_token/events.rs
```diff
@@ -0,0 +1,149 @@
...
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(crate = "near_sdk::serde")]
+pub struct NftTransferLog {
+    //Only serialize if the option is not none
```
austinabell, line 75: I think this is intuitive based on the annotation

willemneal:
This seems to be double work of #627

BenKurrek:
> @BenKurrek This seems to be double work of #627

I didn't know this was being worked on until I was almost finished. Perhaps we can combine our PRs into one. :)

willemneal:
@BenKurrek You could use log functions from my PR to update the implementations.

austinabell:
Closing in favour of #717 #627 which have come in

PR #655, opened 2021-12-03
experimental: Cache static env variables and return references

austinabell:
Rationale, very common pattern to call into these things multiple times, and it can lead to unnecessary gas usages with going through host functions.

No real clean and safe solution I can come to wrt invalidating these caches when the context is modified in unit tests, yet.

Not worth time to continue now, but opening draft to explore later. Also, probably not worth the code size increase when a developer could just keep track of one variable themselves.


austinabell:
opinionated, closing

PR #664, opened 2021-12-09
feat(store): Re-implement LookupSet in a more efficient manner

### LINKED ISSUE #659: Optimize `LookupSet` before stabilizing
With #654, it uses `LookupMap` with `()` as the value type, which can lead to some unnecessary reads rather than just checking if a key exists. There also might be some unnecessary CPU cycles to use borsh to deserialize nothing (not sure if compiler would optimize away)

We want to make sure we do this before stabilizing, in case we need/want to change the storage format.

itegulov:
Closes #659

This PR re-implements `LookupSet` so that it tracks all elements by itself and only uses `env::storage_write` with an empty array of bytes and `env::storage_has_key` as opposed to `env::storage_read` (which is more expensive).

### DIFF
--- a/near-sdk/src/store/lookup_set/mod.rs
+++ b/near-sdk/src/store/lookup_set/mod.rs
@@ -30,10 +30,14 @@ where
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub(crate) enum EntryState {
-    Inserted, // The element is cached as freshly inserted
-    Deleted,  // The element is cached as freshly deleted
-    Present,  // The element is definitely present on the trie
-    Absent,   // The element is definitely absent from the trie
+    /// The element is cached as freshly inserted, but not necessarily absent from the trie
+    Inserted,
+    /// The element is cached as freshly deleted, but not necessarily present on the trie
+    Deleted,
+    /// The element is definitely present on the trie
+    Present,
+    /// The element is definitely absent from the trie
+    Absent,
 }
 
 impl<T, H> Drop for LookupSet<T, H>
@@ -110,7 +114,7 @@ where
                 EntryState::Absent
             }
         });
-        let entry = entry.get_mut().unwrap_or_else(|| unreachable!());
+        let entry = entry.get_mut().unwrap_or_else(|| env::abort());
         entry
     }
 
@@ -145,16 +149,17 @@ where
         Q: BorshSerialize + ToOwned<Owned = T> + Ord,
     {
         let entry_cell = self.cache.get(value.to_owned());
-        match entry_cell.get() {
-            Some(EntryState::Inserted) | Some(EntryState::Present) => true,
-            Some(EntryState::Deleted) | Some(EntryState::Absent) => false,
-            None => {
-                let lookup_key = Self::lookup_key(&self.prefix, value, &mut Vec::new());
-                let contains = env::storage_has_key(&lookup_key);
-                let _ =
-                    entry_cell.set(if contains { EntryState::Present } else { EntryState::Absent });
-                contains
+        match entry_cell.get_or_init(|| {
+            let lookup_key = Self::lookup_key(&self.prefix, value, &mut Vec::new());
+            let contains = env::storage_has_key(&lookup_key);
+            if contains {
+                EntryState::Present
+            } else {
+                EntryState::Absent
             }
+        }) {
+            EntryState::Inserted | EntryState::Present => true,
+            EntryState::Deleted | EntryState::Absent => false,
         }
     }
 
@@ -170,12 +175,8 @@ where
         let entry = self.get_mut_inner(&value);
         match entry {
             EntryState::Inserted | EntryState::Present => false,
-            EntryState::Deleted => {
-                let _ = std::mem::replace(entry, EntryState::Present);
-                true
-            }
-            EntryState::Absent => {
-                let _ = std::mem::replace(entry, EntryState::Inserted);
+            EntryState::Deleted | EntryState::Absent => {
+                *entry = EntryState::Inserted;
                 true
             }
         }
@@ -186,18 +187,15 @@ where
     /// This function will not return whether the passed value was already in the set.
     /// Use [`LookupSet::insert`] if you need that.
     pub fn put(&mut self, value: T) {
-        let entry = self.cache.get_mut(value);
-        match entry.get_mut() {
-            Some(EntryState::Inserted) | Some(EntryState::Present) => {}
-            Some(EntryState::Deleted) => {
-                let _ = entry.set(EntryState::Present);
-            }
-            Some(EntryState::Absent) | None => {
-                // It is safe to mark an entry as `Inserted` even if it is already present on trie;
-                // it just means we will invoke `env::storage_write` one more time than strictly
-                // necessary.
-                let _ = entry.set(EntryState::Inserted);
-            }
+        let entry_cell = self.cache.get_mut(value);
+        // It is safe to preemptively mark an entry as `Inserted` even if it is already present on
+        // trie; it just means we will invoke `env::storage_write` one more time than strictly
+        // necessary.
+        entry_cell.get_or_init(|| EntryState::Inserted);
+        let entry = entry_cell.get_mut().unwrap_or_else(|| env::abort());
+        match entry {
+            EntryState::Inserted | EntryState::Present => {}
+            EntryState::Deleted | EntryState::Absent => *entry = EntryState::Inserted,
         }
     }
 
@@ -211,14 +209,10 @@ where
         T: Borrow<Q>,
         Q: BorshSerialize + ToOwned<Owned = T> + Ord,
     {
-        let entry = self.get_mut_inner(&value);
+        let entry = self.get_mut_inner(value);
         match entry {
-            EntryState::Inserted => {
-                let _ = std::mem::replace(entry, EntryState::Absent);
-                true
-            }
-            EntryState::Present => {
-                let _ = std::mem::replace(entry, EntryState::Deleted);
+            EntryState::Present | EntryState::Inserted => {
+                *entry = EntryState::Deleted;
                 true
             }
             EntryState::Deleted | EntryState::Absent => false,
@@ -243,13 +237,13 @@ where
                         buf.clear();
                         let lookup_key = Self::lookup_key(&self.prefix, k, &mut buf);
                         env::storage_write(&lookup_key, &[]);
-                        let _ = std::mem::replace(entry, EntryState::Present);
+                        *entry = EntryState::Present;
                     }
                     EntryState::Deleted => {
                         buf.clear();
                         let lookup_key = Self::lookup_key(&self.prefix, k, &mut buf);
                         env::storage_remove(&lookup_key);
-                        let _ = std::mem::replace(entry, EntryState::Absent);
+                        *entry = EntryState::Absent;
                     }
                     EntryState::Present | EntryState::Absent => {}
                 }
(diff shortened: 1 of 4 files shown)


near-sdk/src/store/lookup_set/mod.rs
```diff
@@ -1,19 +1,40 @@
...
 where
     T: BorshSerialize + Ord,
     H: CryptoHasher<Digest = [u8; 32]>,
 {
-    map: LookupMap<T, (), H>,
+    prefix: Box<[u8]>,
+
+    /// Cache that keeps track the state of elements in the underlying set.
+    #[borsh_skip]
+    cache: RefCell<BTreeMap<T, Cell<EntryState>>>,
```
itegulov, line 26: I am not sure if wrapping `BTreeMap` in a `RefCell` here is a good idea, but I needed it to mutate the cache inside `contains`.

near-sdk/src/store/lookup_set/mod.rs
```diff
@@ -118,7 +210,24 @@ where
...
     /// in memory.
     pub fn flush(&mut self) {
-        self.map.flush()
+        let mut buf = Vec::new();
+        for (k, v) in self.cache.borrow_mut().iter_mut() {
+            match v.get() {
+                EntryState::Inserted => {
+                    buf.clear();
+                    let lookup_key = Self::lookup_key(&self.prefix, k, &mut buf);
+                    env::storage_write(&lookup_key, &[]);
```
austinabell, line 219: So I'm pretty sure `&[]` serializes the length (0) so should use `()` in place of these

near-sdk/src/store/lookup_set/mod.rs
austinabell: Just use `StableMap` which is in our utils and it will give you what you need for this without doing this. It is this same API just that it allows for inserting with an immutable reference and still allows mutating with a mutable reference.

See how it's used in `LookupMap`

near-sdk/src/store/lookup_set/mod.rs
itegulov: I think you are confusing de-serialization with plain `storage_write` usage here. `env::storage_write` accepts `&[u8]` as the second parameter, so you can't pass `()` there anyway.

near-sdk/src/store/lookup_set/mod.rs
itegulov: The reason why I didn't use `StableMap` is because it wraps values in `Box` which I don't think I need, but let me give it another shot

itegulov: Ok, I think it indeed looks better now. PTAL

near-sdk/src/store/lookup_set/mod.rs
```diff
@@ -54,6 +74,46 @@ where
...
+        //* ToOwned bound, which forces a clone, is required to be able to keep the value in the cache
+        let entry = self.cache.get_mut(value.to_owned());
+        entry.get_or_init(|| {
+            if Self::contains_trie_element(prefix, value) {
+                EntryState::Present
+            } else {
+                EntryState::Absent
+            }
+        });
+        let entry = entry.get_mut().unwrap_or_else(|| unreachable!());
```
austinabell, line 113: I see this is pulled from lookup map but this must have been missed when cleaning up these types of errors, this still includes file info in the code and error. Better to switch this to `env::abort`

near-sdk/src/store/lookup_set/mod.rs
```diff
@@ -1,19 +1,39 @@
...
-use crate::store::LookupMap;
-use crate::IntoStorageKey;
+use crate::{env, IntoStorageKey, StableMap};
 use borsh::{BorshDeserialize, BorshSerialize};
+use once_cell::unsync::OnceCell;
 use std::borrow::Borrow;
 use std::fmt;
+use std::marker::PhantomData;
+
+const ERR_ELEMENT_SERIALIZATION: &str = "Cannot serialize element";
```
austinabell, line 11: Noting there shouldn't be a need to duplicate this const. I've moved this to a common location and re-used in another PR, can always follow up later to remove this, but making comment so I don't forget.

near-sdk/src/store/lookup_set/mod.rs
```diff
@@ -1,19 +1,39 @@
...
+    #[borsh_skip]
+    hasher: PhantomData<H>,
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub(crate) enum EntryState {
+    Inserted, // The element is cached as freshly inserted
+    Deleted,  // The element is cached as freshly deleted
+    Present,  // The element is definitely present on the trie
+    Absent,   // The element is definitely absent from the trie
```
austinabell, line 36:
```suggestion
    /// The element is cached as freshly inserted
    Inserted,
    Deleted,  // The element is cached as freshly deleted
    Present,  // The element is definitely present on the trie
    Absent,   // The element is definitely absent from the trie
```

The standard for this is to have `///` comments above the variants so the descriptions show up with rust tooling or rust docs. Less important since private but probably still worth to change

near-sdk/src/store/lookup_set/mod.rs
```diff
@@ -80,7 +144,18 @@ where
         T: Borrow<Q>,
         Q: BorshSerialize + ToOwned<Owned = T> + Ord,
     {
-        self.map.contains_key(value)
+        let entry_cell = self.cache.get(value.to_owned());
+        match entry_cell.get() {
```
austinabell, line 148: better to use `get_or_init` here instead of setting and ignoring the error below

near-sdk/src/store/lookup_set/mod.rs
```diff
@@ -92,7 +167,38 @@ where
...
-        self.map.insert(value, ()).is_none()
+        let entry = self.get_mut_inner(&value);
+        match entry {
+            EntryState::Inserted | EntryState::Present => false,
+            EntryState::Deleted => {
+                let _ = std::mem::replace(entry, EntryState::Present);
+                true
+            }
+            EntryState::Absent => {
+                let _ = std::mem::replace(entry, EntryState::Inserted);
```
austinabell, line 178:
```suggestion
                *entry = EntryState::Present;
                true
            }
            EntryState::Absent => {
                *entry = EntryState::Inserted;
```

Don't need `mem::replace` if you don't want to use the existing value (you are ignoring here)

near-sdk/src/store/lookup_set/mod.rs
```diff
@@ -92,7 +167,38 @@ where
...
+        match entry.get_mut() {
+            Some(EntryState::Inserted) | Some(EntryState::Present) => {}
+            Some(EntryState::Deleted) => {
+                let _ = entry.set(EntryState::Present);
+            }
+            Some(EntryState::Absent) | None => {
+                // It is safe to mark an entry as `Inserted` even if it is already present on trie;
+                // it just means we will invoke `env::storage_write` one more time than strictly
+                // necessary.
+                let _ = entry.set(EntryState::Inserted);
```
austinabell, line 199: these sets will fail to change anything in every case except `None`

Here is one test you can add which will verify this:
```rust
    #[test]
    fn check_put_on_deleted() {
        let mut set = LookupSet::new(b"m");
        assert!(set.insert(8u8));
        assert!(set.remove(&8));
        set.put(8);
        assert!(set.contains(&8));
    }
```

This is why you should avoid using `let _ =` unless you are sure it cannot error. When building it out it's just safer to unwrap these to make sure you're not going to have one of these types of bugs that are hard to find

near-sdk/src/store/lookup_set/mod.rs
```diff
@@ -105,7 +211,18 @@ where
         T: Borrow<Q>,
         Q: BorshSerialize + ToOwned<Owned = T> + Ord,
     {
-        self.map.remove(value).is_some()
+        let entry = self.get_mut_inner(&value);
+        match entry {
+            EntryState::Inserted => {
+                let _ = std::mem::replace(entry, EntryState::Absent);
```
austinabell, line 217: To replace this with absent can actually lead to a bug. Consider the case where the entry is present in storage but not in memory. If you do a put to set the value to inserted in cache, then do this remove, clearing the key at lookup will not happen and the structure will be in inconsistent state. I'm surprised this wasn't caught by the arbitrary tests, but maybe they can be configured better to hit this case.

I tried making a case that will trigger this but it fails for another reason, but maybe it'll help you get on track:
```rust
    #[test]
    fn put_insert_bug() {
        let lookup_key = LookupSet::<u8>::lookup_key(b"m", &8u8, &mut Vec::new());
        {
            // Scoped to make sure set is dropped and persist changes
            let mut set = LookupSet::new(b"m");
            set.put(8u8);
        }

        assert_eq!(crate::env::storage_has_key(&lookup_key), true);
        {
            let mut set = LookupSet::new(b"m");
            set.put(8u8);
            set.remove(&8);
        }
        assert_eq!(crate::env::storage_has_key(&lookup_key), false);
        {
            let set = LookupSet::new(b"m");
            assert!(!set.contains(&8));
        }
    }
```

near-sdk/src/store/lookup_set/mod.rs
```diff
@@ -105,7 +211,18 @@ where
...
+                true
+            }
+            EntryState::Present => {
+                let _ = std::mem::replace(entry, EntryState::Deleted);
```
austinabell, line 221: again, don't need `mem::replace` when you don't need the existing value. Can just assign the value

near-sdk/src/store/lookup_set/mod.rs
```diff
@@ -118,7 +235,26 @@ where
...
+                        buf.clear();
+                        let lookup_key = Self::lookup_key(&self.prefix, k, &mut buf);
+                        env::storage_write(&lookup_key, &[]);
+                        let _ = std::mem::replace(entry, EntryState::Present);
+                    }
+                    EntryState::Deleted => {
+                        buf.clear();
+                        let lookup_key = Self::lookup_key(&self.prefix, k, &mut buf);
+                        env::storage_remove(&lookup_key);
+                        let _ = std::mem::replace(entry, EntryState::Absent);
```
austinabell, line 252: ditto on `mem::replace`

austinabell (changes requested):
When adding this custom logic, try to have tests that have path coverage, because some of these bugs could be easily caught by this

near-sdk/src/store/lookup_set/mod.rs
itegulov: Oh, I didn't realize that `OnceCell::set` returns `Result`... I thought it returns the previously set value and that the compiler was just complaining about me ignoring the output of the function (that I didn't need).

near-sdk/src/store/lookup_set/mod.rs
itegulov: Yes, I came up with this invariant in my head based on the core methods `insert`, `remove` and `contains`. But, as you have noticed, `put`, that I have added later to accommodate for `Extend`, actually breaks it. Thanks for the test and for the thorough review.

near-sdk/src/store/lookup_set/mod.rs
itegulov: Indeed, somehow I forgot that you could do that :)

near-sdk/src/store/lookup_set/mod.rs
austinabell: there wouldn't be a need to use `let _ = ` if it just returned the previous type as this should only ever be used to ignore `#[must_use]` types like `Result` when you're sure of it. 

If you could overwrite values of OnceCell it would be unsafe because when changed, the references held of the value would be invalidated, if you're curious of slightly more specifics of the API.

itegulov: Makes sense, thanks for the explanation.

austinabell (approved):

PR #665, opened 2021-12-09
feat(store): Add new `TreeMap` implementation

### LINKED ISSUE #588: v2 TreeMap
I'll take this issue since it still needs to be scoped out. The issue is that values aren't cached and the interface doesn't match std equivalents. I believe this can be done without changing the core functionality, but it may be possible to make some optimizations along the way.

This will be placed in `near_sdk::store::tree_map`

Since we are creating in parallel though, we can revisit what structure/name we would like for this. I think it makes sense to keep the same name like we are with others, but we are open to ideas :)

austinabell:
Closes #588 

So this is somewhat based on the original impl, but I changed quite a bit, so probably would have been better to just start from scratch. That was incredibly painful to make the old impl more efficient and under the new API of the underlying data structures. 

All the APIs are changed to match `std::collections::BTreeMap`. ~~I'll make a pass-through tonight or tomorrow to make sure everything we can match we do, and then I will open this up~~. Otherwise, all current functionality will remain the same if anyone wants to start understanding the changes. I tried doing changes in place, to more clearly see differences (https://github.com/near/near-sdk-rs/commit/586f0f0723548d815e0e60f4e53499e09ac460ae) but this ended up not being very helpful because a lot of things changed.


near-sdk/src/store/tree_map/mod.rs
```diff
@@ -0,0 +1,2059 @@
...
+        Tree { root: None, nodes: FreeList::new(prefix) }
+    }
+}
+
+#[derive(Clone, BorshSerialize, BorshDeserialize, Debug)]
+struct Node<K> {
+    key: K,                     // key stored in a node
+    lft: Option<FreeListIndex>, // left link of a node
+    rgt: Option<FreeListIndex>, // right link of a node
+    ht: u32,                    // height of a subtree at a node
```
itegulov, line 101: I know you have inherited the AVL implementation from the old collection and this optimization is probably not worth the trouble, but instead of storing tree height, you can actually just store the balance factor (-1, 0, 1). It is quite easy to see that it would be enough by looking at all current usages of `ht`. From what I remember upholding the balance factor invariant is a bit tricky though and the gain is minuscule (if any at all), but I thought I would share my thoughts just in case.

near-sdk/src/store/tree_map/mod.rs
```diff
@@ -0,0 +1,2059 @@
...
+}
+
+/// TreeMap based on AVL-tree
+///
+/// Runtime complexity (worst case):
+/// - `get`/`contains_key`:     O(1) - LookupMap lookup
+/// - `insert`/`remove`:        O(log(N))
+/// - `min`/`max`:              O(log(N))
+/// - `above`/`below`:          O(log(N))
+/// - `range` of K elements:    O(Klog(N))
```
itegulov, line 31: Similar thing here - if we make iterators actually hold `Node` instances, we would be able to answer `range` queries with a better complexity of `O(K + log(N))`. Every time you need `next` element, you would just follow a deterministic algorithm of finding the next/previous value in a binary tree. This might be a little more worthwhile than my last point, but I leave it to your judgement.

I am not 100% sure on my assessment of complexity as each next-element-lookup takes `O(log(N))`, so if we judge the complexity naively it would still be `O(Klog(N))`. However, if we actually do amortized analysis here, I think in a BST with inner nodes it should be `O(K + log(N))`. I can uncover some existing formal proof on this if you want or, if not, probably can come up with one myself.

itegulov (commented):
I have taken just a quick look through the code so far. Will try to dive deeper later this week. Let's finish these collections :wink:

near-sdk/src/store/tree_map/mod.rs
```diff
@@ -0,0 +1,2059 @@
...
+}
+
+impl<K, V, H> TreeMap<K, V, H>
+where
+    K: BorshSerialize + Ord,
+    V: BorshSerialize,
+    H: CryptoHasher<Digest = [u8; 32]>,
+{
+    /// Flushes the intermediate values of the map before this is called when the structure is
+    /// [`Drop`]ed. This will write all modified values to storage but keep all cached values
```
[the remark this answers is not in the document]
itegulov: On this topic, might be worth adding our usual `artibtrary`-based test suite here as well? I don't think `flush` is being tested right now.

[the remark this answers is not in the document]
austinabell: Drop wasn't added because everything in this function is implicitly called when dropped, so it is redundant. For maintainability reasons, I think I will just add the drop impl anyway to make sure if anything changes within `flush` that this is called.

As for arbitrary tests, these are basically covered by the quickcheck tests, and there would be a lot of overlap. Nonetheless will add in some

near-sdk/src/store/mod.rs
```diff
@@ -16,11 +16,18 @@ pub use self::lookup_set::LookupSet;
...
 mod index_map;
 pub(crate) use self::index_map::IndexMap;
 
 pub(crate) mod free_list;
 pub(crate) use self::free_list::FreeList;
 
-const ERR_INCONSISTENT_STATE: &str = "The collection is an inconsistent state. Did previous smart \
+pub(crate) const ERR_INCONSISTENT_STATE: &str =
+    "The collection is an inconsistent state. Did previous smart \
 										contract execution terminate unexpectedly?";
```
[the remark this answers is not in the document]
austinabell: good catch, I didn't realize tabs were added in the case of being escaped.

near-sdk/src/store/tree_map/mod.rs
```diff
@@ -0,0 +1,2059 @@
...
+        // (source: https://en.wikipedia.org/wiki/AVL_tree)
+        const B: f64 = -0.328;
+        const C: f64 = 1.440;
+        const D: f64 = 1.065;
+
+        let h = C * log2(n as f64 + D) + B;
+        h.ceil() as u32
+    }
+
+    impl<K, V, H> Debug for TreeMap<K, V, H>
```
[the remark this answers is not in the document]
austinabell: nope, just an artifact of migrating from the previous version and that it's expensive, thanks for pointing out :)

near-sdk/src/store/tree_map/mod.rs
austinabell: let's tackle this in a following PR. This is under `unstable` so let's try to refactor and optimize once we have a framework to compare the optimizations

near-sdk/src/store/tree_map/mod.rs
austinabell: It's definitely possible and should be done, but requires a decent refactor as with the current impl we don't have a way to traverse up to the parent for these operations. Perhaps like the previous point we should implement this optimization in a future PR?

near-sdk/src/store/tree_map/mod.rs
```diff
@@ -0,0 +1,2122 @@
...
+    }
+}
+
+impl<K> Tree<K>
+where
+    K: Ord + BorshSerialize + BorshDeserialize + Clone,
+{
+    fn internal_insert(&mut self, key: K)
+    where
+        K: Clone,
```
ChaoticTempest, line 537: Looks like you don't need this bound since you already declared it before
```suggestion
```

near-sdk/src/store/tree_map/mod.rs
```diff
@@ -0,0 +1,2122 @@
...
+    // - or left-most (min) node of the right subtree (containing larger keys) of node holding `key`
+    //
+    fn do_remove<Q: ?Sized>(&mut self, key: &Q) -> (Option<FreeListIndex>, Option<K>)
+    where
+        K: Borrow<Q>,
+        Q: BorshSerialize + Eq + PartialOrd,
+    {
+        // r_node - node containing key of interest
+        // p_node - immediate parent node of r_node
+        let ((r_id, mut r_node), remove_parent) = match self
```
ChaoticTempest, line 691: nit: Looks like these comments are now out of date

ChaoticTempest (commented):
Did a mini pass on this, with some minor comments. I don't understand enough about AVL-Trees so will do another pass later on when I up my understanding

near-sdk/src/store/tree_map/mod.rs
itegulov: Yeah, I don't mind. Just thought I would point this out before I forget :)

near-sdk/src/store/tree_map/mod.rs
itegulov: Sure

near-sdk/src/store/tree_map/entry.rs
```diff
@@ -0,0 +1,410 @@
...
+{
+    /// Returns a reference to this entry's key.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use near_sdk::store::TreeMap;
+    ///
+    /// let mut map: TreeMap<String, u32> = TreeMap::new(b"m");
+    /// assert_eq!(map.entry("poneyland".to_string()).key(), &"poneyland");
```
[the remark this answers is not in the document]
austinabell: This doc was just pulled from `std::collection::HashMap`. I'll remove it if it works without

ChaoticTempest (approved):
The implementation looks good to go. The interface mainly looks the same as the older one, and we should be able to refine it based on what Daniyar mentioned

near-sdk/src/store/tree_map/mod.rs
```diff
@@ -0,0 +1,2118 @@
...
+                None => {
+                    return node.map(|node| ((at, node), parent));
+                }
+            }
+        }
+    }
+
+    /// Returns (node, parent node) of right-most lower (max) node starting from given node `at`.
+    /// As min_at only traverses the tree down, if a node `at` is the minimum node in a subtree,
+    /// its parent must be explicitly provided in advance.
```
itegulov, line 400: > if a node `at` is the minimum node in a subtree, its parent must be explicitly provided in advance.

I think this bit is outdated, you no longer accept a parent node in the function signature

near-sdk/src/store/tree_map/mod.rs
```diff
@@ -0,0 +1,2118 @@
...
+        if let Some(key) = self.equal_key(key) {
+            Some(key)
+        } else {
+            self.higher(key)
+        }
+    }
+
+    /// Returns (node, parent node) of left-most lower (min) node starting from given node `at`.
+    /// As min_at only traverses the tree down, if a node `at` is the minimum node in a subtree,
+    /// its parent must be explicitly provided in advance.
```
itegulov, line 381: Same here

near-sdk/src/store/tree_map/mod.rs
```diff
@@ -0,0 +1,2118 @@
...
+                    Some(lft) => at = lft,
+                    None => break,
+                }
+            }
+        }
+        None
+    }
+
+    /// Returns (node, parent node) for a node that holds the `key`.
+    /// For root node, same node is returned for node and parent node.
```
itegulov, line 490: This whole comment seems outdated too

near-sdk/src/store/tree_map/mod.rs
```diff
@@ -0,0 +1,2118 @@
...
+    }
+
+    fn insert_at(&mut self, mut node: Node<K>, id: FreeListIndex, key: K) -> FreeListIndex {
+        if key.eq(&node.key) {
+            // This branch should not be hit, because we check for existence in insert.
+            id
+        } else {
+            if key.lt(&node.key) {
+                let idx = match node.lft {
+                    Some(lft) => self.insert_at(expect(self.node(lft)).clone(), lft, key),
```
itegulov, line 550: Have you explored whether it is possible to get rid of all these clones by writing another auxiliary method like this one:
```rust
fn node_mut(&mut self, id: FreeListIndex) -> Option<&mut Node<K>> {
    self.nodes.get_mut(id)
}
```

And then just operating with `&mut Node<K>` in this method (and maybe some others too)?

itegulov (approved):
LGTM, just some documentation nitpicks and a potential thing to look into

near-sdk/src/store/tree_map/mod.rs
austinabell: yeah, absolutely did, the issue with the current impl is you can't borrow this mutably and borrow other parts of the vec immutably. Yes we know there isn't overlap, but would need unsafe for this to happen. This is why clones currently happen in the implementation, but is still an improvement over the existing implementation because it was loading and deserializing the value where the clone was.

This can be optimized in the future without breaking changes with the current API.

near-sdk/src/store/tree_map/mod.rs
```diff
@@ -0,0 +1,2118 @@
...
+    ///
+    /// The key may be any borrowed form of the map's key type, but
+    /// [`BorshSerialize`] and [`ToOwned<Owned = K>`](ToOwned) on the borrowed form *must* match
+    /// those for the key type.
+    pub fn contains_key<Q: ?Sized>(&self, k: &Q) -> bool
+    where
+        K: Borrow<Q>,
+        Q: BorshSerialize + ToOwned<Owned = K> + Ord,
+    {
+        self.values.contains_key(k)
```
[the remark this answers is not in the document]
austinabell: I don't personally see the issue, but don't have a preference, so I'm happy to hear if you have a name that might be more clear?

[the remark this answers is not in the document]
itegulov: I just meant that it looks weird that the map has a field named `values` that also hold keys. Feel free to ignore my comment, I don't really mind either

near-sdk/src/store/tree_map/mod.rs
itegulov: I see. I thought something like that might be the case, thanks for confirming.

near-sdk/src/store/tree_map/mod.rs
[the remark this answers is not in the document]
austinabell: well, technically it doesn't hold the keys, it just is generic on the key type to do lookups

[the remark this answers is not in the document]
itegulov: That's a good point too

PR #669, opened 2021-12-13
chore: update libs and examples to 2021 edition Rust

austinabell:
Requires updating the released https://hub.docker.com/r/nearprotocol/contract-builder dockerfile to use > Rust 1.56 to avoid having to use a local dockerfile


austinabell:
Second-guessing doing this in this upcoming release. Wondering if updating this will cause issues for people using old docker images without sufficient Rust version for this. I'm thinking this should come in the release after 4.0 stable

austinabell:
> Second-guessing doing this in this upcoming release. Wondering if updating this will cause issues for people using old docker images without sufficient Rust version for this. I'm thinking this should come in the release after 4.0 stable

Given nearcore dependencies requiring Rust 2021 and less unknowns with including as a major version bump, going to include this with 4.0 stable

itegulov (approved):

PR #670, opened 2021-12-14
chore: add windows check to CI and ignore irrelevant test

austinabell:
- Removes code size check where a few contracts are compiled. We are enforcing the contracts are kept up to date with changes so no need anymore
- Adds windows compilation check to CI


near-sdk/tests/code_size.rs
```diff
@@ -21,6 +21,7 @@ fn check_example_size(example: &str) -> usize {
 }
 
 #[test]
+#[ignore = "Size of contracts currently kept in sync with changes"]
```
itegulov, line 24: Hmm, isn't the purpose of these tests to check that the contracts do not get too bloated and not that they are out of sync with changes? I am confused how the fact that we enforce that the contracts are in sync with the changes relates to their size being less than 180000 bytes.

austinabell: Because part of our CI checks actually track the exact changes of the size of the contracts, this check is not important anymore. These tests were added before we had the checks in the CI to keep the contracts in line with changes. These tests are also unnecessarily building the contracts which is not ideal to do in the CI

.github/workflows/test.yml
```diff
@@ -40,3 +37,16 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: "Setup Windows toolchain"
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+      - uses: Swatinem/rust-cache@v1
+      - run: cargo check -p near-sdk --features unstable
+      - run: cargo check -p near-contract-standards
```
ChaoticTempest, line 52: shouldn't we be adding `--target wasm32-unknown-unknown` to these checks as well?

ChaoticTempest (commented):
Unrelated, but I keep getting this warning when compiling:
```rust
warning: unnecessary `unsafe` block
   --> near-sdk\src\environment\env.rs:494:5
    |
494 |     unsafe {
    |     ^^^^^^ unnecessary `unsafe` block
```
Should we try to fix it? It's weird that clippy/fomat isn't catching it

.github/workflows/test.yml
```diff
@@ -40,3 +37,16 @@ jobs:
       - name: Test Format
         run: cargo fmt -- --check
       - run: cargo clippy --features unstable --tests -- -Dclippy::all
+  windows:
```
ChaoticTempest, line 40: forgot to mention, but `near-sdk-macros` also compiles with windows. It's just sim that doesn't

austinabell: yeah, that's fine. It all gets checked and compiled through `near-sdk` anyway

.github/workflows/test.yml
austinabell: yeah fair point, under the assumption that there can't be a case where this succeeds and wasm32 fails only for windows, but probably safe to add

austinabell:
> Unrelated, but I keep getting this warning when compiling:
> 
> ```rust
> warning: unnecessary `unsafe` block
>    --> near-sdk\src\environment\env.rs:494:5
>     |
> 494 |     unsafe {
>     |     ^^^^^^ unnecessary `unsafe` block
> ```
> 
> Should we try to fix it? It's weird that clippy/fomat isn't catching it

I was actually removing this and checking with version it was stabilized on but my computer keeps overheating and shutting off so give me a minute :D

ChaoticTempest (approved):

itegulov (approved):

PR #671, opened 2021-12-14
chore: cleanup TODOs and deprecate bad functions

austinabell:
Cleanup getting ready for stable 4.0

- Adds missing pattern for `testing_env!` macro
- Deprecate setup functions that are either not needed or a foot gun because of side effects

ref #494


ChaoticTempest (approved):
LGTM:
```
error[E0133]: call to unsafe function is unsafe and requires unsafe function or block
--
--> /host/near-sdk/src/environment/env.rs:494:5
\|
494 \|     core::arch::wasm32::unreachable();
\|     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call to unsafe function
\|
= note: consult the function's documentation for information on how to avoid undefined behavior
```
So it looks like the docker image still needs to get updated

PR #672, opened 2021-12-16
feat(store): UnorderedSet v2 caching implementation

### LINKED ISSUE #587: v2 UnorderedSet
extension of #586, to keep keys in a storage `Bucket` (#582) to be able to iterate the set. Similar tradeoffs to what the new data structures are to old, see #580.

Will exist in `near_sdk::store::unordered_set`

itegulov:
Closes #587

### DIFF
--- a/near-sdk/src/store/unordered_map/mod.rs
+++ b/near-sdk/src/store/unordered_map/mod.rs
@@ -208,7 +208,6 @@ where
     {
         for k in self.keys.drain() {
             // Set instead of remove to avoid loading the value from storage.
-            // This enforces a clone, but this is better than invoking a more costly operation.
             self.values.set(k, None);
         }
     }
--- a/near-sdk/src/store/unordered_map/mod.rs
+++ b/near-sdk/src/store/unordered_map/mod.rs
@@ -208,7 +208,7 @@ where
     {
         for k in self.keys.drain() {
             // Set instead of remove to avoid loading the value from storage.
-            // This enforces a clone, but this is better th
+            // This enforces a clone, but this is better than invoking a more costly operation.
             self.values.set(k, None);
         }
     }
--- a/near-sdk/src/store/unordered_set/iter.rs
+++ b/near-sdk/src/store/unordered_set/iter.rs
@@ -115,34 +115,16 @@ where
     type Item = &'a T;
 
     fn next(&mut self) -> Option<Self::Item> {
-        <Self as Iterator>::nth(self, 0)
+        loop {
+            let elt = self.elements.next()?;
+            if !self.other.contains(elt) {
+                return Some(elt);
+            }
+        }
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
         (0, self.elements.size_hint().1)
-    }
-
-    fn count(mut self) -> usize {
-        let mut count = 0usize;
-        for element in self.elements.by_ref() {
-            if !self.other.contains(element) {
-                count += 1;
-            }
-        }
-        count
-    }
-
-    fn nth(&mut self, n: usize) -> Option<Self::Item> {
-        let mut count = 0usize;
-        for element in self.elements.by_ref() {
-            if !self.other.contains(element) {
-                if count == n {
-                    return Some(element);
-                }
-                count += 1;
-            }
-        }
-        None
     }
 }
 
@@ -187,34 +169,16 @@ where
     type Item = &'a T;
 
     fn next(&mut self) -> Option<Self::Item> {
-        <Self as Iterator>::nth(self, 0)
+        loop {
+            let elt = self.elements.next()?;
+            if self.other.contains(elt) {
+                return Some(elt);
+            }
+        }
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
         (0, self.elements.size_hint().1)
-    }
-
-    fn count(mut self) -> usize {
-        let mut count = 0usize;
-        for element in self.elements.by_ref() {
-            if self.other.contains(element) {
-                count += 1;
-            }
-        }
-        count
-    }
-
-    fn nth(&mut self, n: usize) -> Option<Self::Item> {
-        let mut count = 0usize;
-        for element in self.elements.by_ref() {
-            if self.other.contains(element) {
-                if count == n {
-                    return Some(element);
-                }
-                count += 1;
-            }
-        }
-        None
     }
 }
 
--- a/near-sdk/src/store/unordered_set/iter.rs
+++ b/near-sdk/src/store/unordered_set/iter.rs
@@ -46,7 +46,7 @@ where
 
 impl<'a, T, H> Iterator for Iter<'a, T, H>
 where
-    T: BorshSerialize + Ord + BorshDeserialize + Clone,
+    T: BorshSerialize + Ord + BorshDeserialize,
     H: CryptoHasher<Digest = [u8; 32]>,
 {
     type Item = &'a T;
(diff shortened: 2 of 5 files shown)


near-sdk/src/store/unordered_map/mod.rs
```diff
@@ -208,7 +208,7 @@ where
     {
         for k in self.keys.drain() {
             // Set instead of remove to avoid loading the value from storage.
-            // This enforces a clone, but this is better th
+            // This enforces a clone, but this is better than invoking a more costly operation.
```
austinabell, line 211: oops I meant to delete this comment altogether, since this doesn't clone the key anymore due to `drain` impl

near-sdk/src/store/unordered_set/iter.rs
```diff
@@ -0,0 +1,418 @@
...
+pub struct Difference<'a, T, H>
+where
+    T: BorshSerialize + Ord + BorshDeserialize,
+    H: CryptoHasher<Digest = [u8; 32]>,
+{
+    elements: free_list::Iter<'a, T>,
+
+    other: &'a UnorderedSet<T, H>,
+
+    el: PhantomData<H>,
```
austinabell, line 113: You don't need this phantomdata since H is held by the set

near-sdk/src/store/unordered_set/iter.rs
```diff
@@ -0,0 +1,418 @@
...
+pub struct Drain<'a, T, H>
+where
+    T: BorshSerialize + BorshDeserialize + Ord,
+    H: CryptoHasher<Digest = [u8; 32]>,
+{
+    elements: free_list::Drain<'a, T>,
+
+    index: &'a mut LookupMap<T, FreeListIndex, H>,
+
+    el: PhantomData<H>,
```
austinabell, line 356: same here

austinabell (commented):
I know just a draft, but I had a quick scan so just a few comments

itegulov:
I wanted to get rid of more clones, but unfortunately most of the binary operations (union, intersection etc) require access to `contains` which in turn means that all of them need `Clone`. So I don't think there is an easy way out of this unless we can somehow not use `LookupMap`.

near-sdk/src/store/unordered_set/iter.rs
```diff
@@ -0,0 +1,412 @@
...
+    H: CryptoHasher<Digest = [u8; 32]>,
+{
+    fn next_back(&mut self) -> Option<Self::Item> {
+        <Self as DoubleEndedIterator>::nth_back(self, 0)
+    }
+
+    fn nth_back(&mut self, n: usize) -> Option<Self::Item> {
+        self.elements.nth_back(n)
+    }
+}
```
austinabell, line 96: I don't think a double-ended iterator impl is needed for this and doesn't make sense in this context because the set has no order. I know I added it for `UnorderedMap`, but I'm second-guessing myself here.

@matklad your opinion on this would be valuable, do you think `DoubleEndedIterator` impls should be removed from unordered collections? impls don't exist for `std` equivalents

near-sdk/src/store/unordered_set/mod.rs
```diff
@@ -0,0 +1,798 @@
...
+    T: BorshSerialize + Ord,
+    H: CryptoHasher<Digest = [u8; 32]>,
+{
+    fn deserialize(buf: &mut &[u8]) -> Result<Self, borsh::maybestd::io::Error> {
+        Ok(Self {
+            elements: BorshDeserialize::deserialize(buf)?,
+            index: BorshDeserialize::deserialize(buf)?,
+        })
+    }
+}
```
austinabell, line 52:
```suggestion
#[derive(BorshDeserialize, BorshSerialize)]
pub struct UnorderedSet<T, H = Sha256>
where
    T: BorshSerialize + Ord,
    H: CryptoHasher<Digest = [u8; 32]>,
{
    elements: FreeList<T>,
    index: LookupMap<T, FreeListIndex, H>,
}
```

There are no private types leaked here, can just derive

near-sdk/src/store/unordered_set/mod.rs
```diff
@@ -0,0 +1,798 @@
...
+        self.elements.is_empty()
+    }
+
+    /// Clears the set, removing all values.
+    pub fn clear(&mut self)
+    where
+        T: BorshDeserialize + Clone,
+    {
+        for e in self.elements.drain() {
+            self.index.remove(&e);
```
austinabell, line 130:
```suggestion
            self.index.set(e, None);
```

remove reads existing value from storage, which I don't think can be compiled away because it uses a host function to do so

near-sdk/src/store/unordered_set/mod.rs
```diff
@@ -0,0 +1,798 @@
...
+        }
+        self.elements.clear();
```
austinabell, line 132: This clear is redundant since it will be cleared when the drain iterator is dropped. I see it's probably pulled from `UnorderedMap` which did this, but this was from before `drain` was implemented so it can be removed now.

I removed here: https://github.com/near/near-sdk-rs/pull/675

near-sdk/src/store/unordered_set/mod.rs
```diff
@@ -0,0 +1,798 @@
...
+use crate::store::free_list::FreeListIndex;
+use crate::store::unordered_set::iter::{
+    Difference, Drain, Intersection, Iter, SymmetricDifference, Union,
+};
+use crate::{env, IntoStorageKey};
+use borsh::{BorshDeserialize, BorshSerialize};
+use std::borrow::Borrow;
+use std::fmt;
+
+pub struct UnorderedSet<T, H = Sha256>
```
austinabell, line 15: Can you add some docs for this? Can just pull from `std::collections::HashSet` and modify, like it seems you've done with others :)

austinabell (changes requested):
Awesome work, just made some comments to cleanup certain parts of the API :)

near-sdk/src/store/unordered_set/iter.rs
```diff
@@ -0,0 +1,412 @@
...
+    type Item = &'a T;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.iter.next()
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.iter.size_hint()
+    }
+}
```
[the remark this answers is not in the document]
itegulov: > Wondering why you used the optimized (not sure if actually optimized or it would get compiled away) version in Intersection but not here (and Union).

The honest answer is that I just could not think of a way to represent `Intersection` as a combination of other operations. Do you know one?

> Did you confirm that there is a performance difference with Intersection? (there is same amount of storage operations, so not sure if it gets compiled away)

I did not because, as per the previous paragraph, I do not know how to implement the baseline version

> If that is faster, why is it not worth doing here as well. The inner operator doesn't get the speedups of the chained iterators nth and count operations here and would default to using next for both impls

It might indeed make sense, I do not know how good Rust compiler is with this stuff.

near-sdk/src/store/unordered_set/iter.rs
itegulov: Hmm, that is a good point. It probably does not make sense to treat an unordered set as something that has a "front" and a "back".

near-sdk/src/store/unordered_set/mod.rs
itegulov: Oops, missed the most important doc of them all :)

near-sdk/src/store/unordered_set/iter.rs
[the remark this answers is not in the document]
austinabell: > The honest answer is that I just could not think of a way to represent `Intersection` as a combination of other operations. Do you know one?

To be clear I'm just talking about the optimized versions of `nth` and `count`, not the inner type. If you don't specify `nth` here it will default to using `next` exclusively for the impls

I'm not suggesting doing a different combination, as this is even what `std` does for HashSet, I'm just suggesting either adding the `nth` and `count` operations here to be consistent if it is faster, or remove from `Intersection` if the added complexity gets compiled away (which it might). These do the same amount of storage operations, so for `Intersection` I'm not sure having `nth` and `count` actually helps.

My intuition is that `nth` and `count` custom impls aren't actually helpful and are just adding unnecessary complexity for `Intersection` and `Difference`, but that's why I'm asking if you have a reason to believe otherwise.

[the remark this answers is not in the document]
itegulov: Oooh, that's what you mean. Yeah, good point. I did not take a look at the `HashSet` implementation for these iterators and just intuitively thought that this way would be more efficient. Changed.

austinabell (approved):
oops, forgot to hit approve before, LGTM :santa:

PR #673, opened 2021-12-16
feat: Payout and Royalties

willemneal:
Still need to make macro and impl


austinabell:
Seems stale, feel free to re-open if you'd like to continue :)

willemneal:
Thanks! It is a bit stale.  In tenk I went with basis points.  Will back port soon!

PR #674, opened 2021-12-16
refactor: Update LookupMap to allow and default to identity keys

austinabell:
Motivation: needed to make this backwards compatible and didn't want to commit to a breaking format change without the flexibility.

Will replace #596 as what I think this is a better alternative, since it still allows for someone to optimize by using a hash key to optimize storage with large keys, but uses the default identity hash (which is also backwards compatible with the previous version)

The reason we want LookupMap to have identity hash by default is:
- Backwards compatible and drop-in replacement of the previous version
- When clearing all keys of the map manually, you can just search keys with that prefix and remove, with a hash you cant. Fine for UnorderedMap because it keeps track of all keys, so this will default to using a crypto hash for key

I was considering the suggestion by @matklad to have a RawMap type which both `LookupMap` and `UnorderedMap` use, where lookups would be done without hashing and unordered map would. I like this approach better personally because although it adds another trait, it adds so much flexibility without adding noticeable complexity.

Optional changes:
- ~~Remove `CryptoHash` trait and allow `UnorderedMap` to use identity hash for lookups~~
  - ~~The only reason for the two is to allow it for LookupMap but not unordered~~
- Remove or re-work how the buffer works with the `StorageKeyer` trait as it in some cases forces an unnecessary clone with identity hash.
- Change default back to `Sha256` for `LookupMap`. I decided to make it Identity default because that will be the least friction switching from the old version for less experienced rust developers, also allows clearing all prefixed keys which hash keys do not allow
- ~~Rename `StorageKeyer` (obviously, but I couldn't think of a better name)~~

TODO: CHANGELOG once/if API finalized


near-sdk/src/store/lookup_map/impls.rs
```diff
@@ -25,7 +26,8 @@ impl<K, V, H, Q: ?Sized> core::ops::Index<&Q> for LookupMap<K, V, H>
 where
     K: BorshSerialize + Ord + Clone + Borrow<Q>,
```
itegulov, line 27: I don't think we actually need `Clone` here so we can relax the constraint

itegulov (commented):
> When clearing all keys of the map manually, you can just search keys with that prefix and remove, with a hash you cant

Do you then think it would make sense to implement (potentially costly) `LookupSet::is_empty` and `LookupSet::clear` when `StorageKeyer` is `Identity`?

> I like this approach better personally because although it adds another trait, it adds so much flexibility without adding noticeable complexity.

This is just my 2 cents, but I think leaving end-users an option to use `LookupMap` with hashing would be a good idea. I can run some benchmarks and see if there is significant benefit to it and what kind of key size starts affecting gas usage noticeably.

> Remove CryptoHash trait and allow UnorderedMap to use identity hash for lookups

Is there a reason why we shouldn't allow this?

> Remove or re-work how the buffer works with the StorageKeyer trait as it in **some cases** forces an unnecessary clone with identity hash.

Does not it **always** force a clone or am I missing something?

> Rename StorageKeyer (obviously, but I couldn't think of a better name)

Maybe something like `KeyIndex`/`KeyIndexer`?

austinabell:
> > When clearing all keys of the map manually, you can just search keys with that prefix and remove, with a hash you cant
> 
> Do you then think it would make sense to implement (potentially costly) `LookupSet::is_empty` and `LookupSet::clear` when `StorageKeyer` is `Identity`?

Sorry I should have given more context, this can't be done from the contract side. Externally you can iterate all keys that exist in a contract state through JSON RPC, and then technically you can filter out the ones based on the prefix then send it back into the contract to delete manually. Not something that can be a part of the API and should never be relied on.

> > I like this approach better personally because although it adds another trait, it adds so much flexibility without adding noticeable complexity.
> 
> This is just my 2 cents, but I think leaving end-users an option to use `LookupMap` with hashing would be a good idea. I can run some benchmarks and see if there is significant benefit to it and what kind of key size starts affecting gas usage noticeably.

Yeah from what I've seen it's fairly negligible, but definitely need more comprehensive benchmarks like the ones you're working towards. But yeah this is why I decided to have a trait like this, to give the flexibility

> > Remove CryptoHash trait and allow UnorderedMap to use identity hash for lookups
> 
> Is there a reason why we shouldn't allow this?

I suppose not, but at least it shouldn't be the default, since there is no benefit other than if it's cheaper for small keys (which it almost certainly is). I'll likely simplify this


> > Remove or re-work how the buffer works with the StorageKeyer trait as it in **some cases** forces an unnecessary clone with identity hash.
> 
> Does not it **always** force a clone or am I missing something?

Well, it always forces a clone, but only sometimes it's unnecessary. If the buffer is re-used, then the clone is as little cost as it would be otherwise.

> 
> > Rename StorageKeyer (obviously, but I couldn't think of a better name)
> 
> Maybe something like `KeyIndex`/`KeyIndexer`?

Yeah, this is a decent name, just might be a bit misleading because it's not indexing anything, just generating a key to use for storage.

near-sdk/src/store/lookup_map/mod.rs
```diff
@@ -78,29 +76,30 @@ type LookupKey = [u8; 32];
...
     prefix: Box<[u8]>,
     /// Cache for loads and intermediate changes to the underlying vector.
     /// The cached entries are wrapped in a [`Box`] to avoid existing pointers from being
     /// invalidated.
     #[borsh_skip]
-    cache: StableMap<K, EntryAndHash<V>>,
+    cache: StableMap<K, EntryAndHash<V, H::KeyType>>,
 
     #[borsh_skip]
     hasher: PhantomData<H>,
```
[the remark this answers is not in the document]
austinabell: ha! You're right, assumed since that's just a subset that it needed to remain

near-sdk/src/store/lookup_map/impls.rs
austinabell: yup! That was an artifact of a previous API iteration, although I don't think restrictive in any sound implementation. Removed :)

near-sdk/src/environment/hash.rs
```diff
@@ -6,27 +8,36 @@ mod private {
...
     impl Sealed for super::Keccak256 {}
+    impl Sealed for super::Identity {}
 }
 
-/// Cryptographic hashes that can be used within the SDK as a hashing function.
-pub trait CryptoHasher: self::private::Sealed {
-    /// Output type of the hashing function.
-    type Digest;
+/// Trait used to generate keys to store data based on a serializable structure.
+pub trait StorageKeyer: self::private::Sealed {
```
matklad, line 15: Usually traits are named as verbs, so perhaps something like?

```rust
trait ToKey {
  fn to_key()
}
```

near-sdk/src/store/lookup_map/mod.rs
```diff
@@ -121,14 +117,15 @@ impl<K, V, H> fmt::Debug for LookupMap<K, V, H>
...
+    H: StorageKeyer,
+    <H as StorageKeyer>::KeyType: AsRef<[u8]>,
 {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("LookupMap").field("prefix", &self.prefix).finish()
     }
 }
 
-impl<K, V> LookupMap<K, V, Sha256>
+impl<K, V> LookupMap<K, V, Identity>
```
matklad, line 128: How would the user create a lookup map with a different keyer?

austinabell: with the `with_hasher` function (naming taken from HashMap, although API slightly different since they keep an initialized hasher in the map), but perhaps this should be named differently now that it also includes identity/unhashed bytes. wdyt?

near-sdk/src/environment/hash.rs
austinabell: Yeah, this naming is probably better, I'll switch

near-sdk/src/store/unordered_map/mod.rs
```diff
@@ -28,7 +28,7 @@ const ERR_NOT_EXIST: &str = "Key does not exist in map";
 /// The default hash function for [`UnorderedMap`] is [`Sha256`] which uses a syscall
 /// (or host function) built into the NEAR runtime to hash the key. To use a custom function,
 /// use [`with_hasher`]. Alternative builtin hash functions can be found at
-/// [`near_sdk::crypto_hash`](crate::crypto_hash).
+/// [`near_sdk::store::key`](crate::store::key).
```
itegulov, line 31: By the way, don't forget to update the docs once you are set on namings and which traits are staying (e.g. `CryptoHasher` -> `ToKey`)

near-sdk/src/store/lookup_map/mod.rs
```diff
@@ -18,17 +17,16 @@ const ERR_ELEMENT_DESERIALIZATION: &str = "Cannot deserialize element";
 const ERR_ELEMENT_SERIALIZATION: &str = "Cannot serialize element";
 const ERR_NOT_EXIST: &str = "Key does not exist in map";
 
-type LookupKey = [u8; 32];
-
 /// A non-iterable, lazily loaded storage map that stores its content directly on the storage trie.
 ///
 /// This map stores the values under a hash of the map's `prefix` and [`BorshSerialize`] of the key
 /// using the map's [`CryptoHasher`] implementation.
```
itegulov, line 23: And here as well

near-sdk/src/store/lookup_set/mod.rs
```diff
@@ -1,25 +1,27 @@
...
+use crate::store::key::{Identity, ToKey};
 use crate::store::LookupMap;
 use crate::IntoStorageKey;
 use borsh::{BorshDeserialize, BorshSerialize};
 use std::borrow::Borrow;
 use std::fmt;
 
 #[derive(BorshSerialize, BorshDeserialize)]
-pub struct LookupSet<T, H = Sha256>
+pub struct LookupSet<T, H = Identity>
```
itegulov, line 11: Apparently I forgot to write docs for `LookupSet`, whoops!

near-sdk/src/store/unordered_map/mod.rs
austinabell: The docs are generated through the rust doc comments here, so they get updated on release. What here would need to change?

ChaoticTempest (approved):
LGTM with one minor suggestion.

Something to consider in a future PR:

> Change default back to Sha256 for LookupMap

Maybe a bad idea, but might be interesting if we can parameterize this default with a feature flag? Since it would be interesting if we could switch it over without requiring code changes later after someone has migrated over from the collections version

near-sdk/src/store/key.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    impl Sealed for super::Sha256 {}
+    impl Sealed for super::Keccak256 {}
+    impl Sealed for super::Identity {}
+}
+
+/// Trait used to generate keys to store data based on a serializable structure.
+pub trait ToKey: self::private::Sealed {
+    /// Output type for the generated lookup key.
+    type KeyType;
```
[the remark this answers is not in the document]
austinabell: It's a more restrictive bound so it's generally better to have it on the impls or where it is actually needed, but since this is intended to just be used internally and it simplifies the code a lot, I've made the change

near-sdk/src/store/key.rs
```diff
@@ -0,0 +1,78 @@
+use borsh::BorshSerialize;
+
+use crate::env;
+
+mod private {
+    /// Seal `ToKey` implementations to limit usage to the builtin implementations
```
vasyafromrussia, line 6: What's a reason behind limiting usage to those three?

PR #680, opened 2021-12-23
refactor: replace hashmap usages in contracts and examples to BTreeMap

austinabell:
BTreeMap should be much better in wasm contexts especially for a small number of elements. Given that we have collections to use for large data sets, there should be no reason to use a hashmap and we should stop recommending as such.

- Breaking change for NFT standard trait, which I think is reasonable breakage since it will be coming with a major version bump. This is not breaking at a storage level or by any serialization protocol we use, only breaking in the sense that if someone uses the trait manually, they will need to update the type.

Somewhat related: https://github.com/near/borsh-rs/issues/63

Not certain of this change, so I'd like to do some sort of benchmark to compare at what size of map/keys does hashmap outperform, but my intuition is that it's very unlikely `HashMap` will be better for what these are used for.


austinabell:
opinionated, closing

Issue #681, opened 2021-12-23
Change parameter serialization annotation to be on function


austinabell:
Currently, we annotate each individual parameter of a `#[near_bindgen]` function parameter with `#[serializer(borsh)]` and check through macros that the serialization protocol is the same for all arguments. This doesn't make much sense, since you can't mix protocols, there should just be one annotation on the function, like `#[args_serializer(borsh)]` or something similar.

PR #683, opened 2021-12-23
feat: improve macro spans for better errors

austinabell:
Some of these were very unhelpful, as it would just show the `#[near_bindgen]` annotation in error. 

I only didn't update the span for param serializer annotations because of #681


ChaoticTempest (approved):
The error messages look a lot better now

PR #691, opened 2022-01-07
Examples doc for random_seed

deveusss:


deveusss:
Also we could add this function to env, this is not the best version for generating random range number, in most cases this will work.

austinabell:
would you like me to help in formatting and fixing the CI, or would you like to do this?

deveusss:
All tests are ok, but Build #3871 is failing for some reason

austinabell:
> All tests are ok, but Build #3871 is failing for some reason

Yeah it's just because the wasm binaries are affected by this change and not updated. I can update them for you if you'd like? https://github.com/near/near-sdk-rs/blob/master/CONTRIBUTING.md has the info on this if you'd prefer to do yourself

deveusss:
> > All tests are ok, but Build #3871 is failing for some reason
> 
> Yeah it's just because the wasm binaries are affected by this change and not updated. I can update them for you if you'd like? https://github.com/near/near-sdk-rs/blob/master/CONTRIBUTING.md has the info on this if you'd prefer to do yourself

thank you Austin)
fails when running clippy, how you fixed this one?
https://github.com/rust-lang/rust-clippy/issues/8470

ErrorMessage:

error: could not compile `near-sdk-macros`

Caused by:
  process didn't exit successfully: 
  --extern syn=/Volumes/.../near-sdk-rs/target/debug/deps/libsyn-faaeaf152fb769db.rlib --extern proc_macro` (signal: 6, SIGABRT: process abort signal)
warning: build failed, waiting for other jobs to finish...
error: build failed

danielwpz:
this looks very useful, any chance we can have it merged?

austinabell:
> this looks very useful, any chance we can have it merged?

If the branch is updated, I can pull it in. 

Personally, I think this example is quite cryptic and for a very specific use case. I'd prefer slightly if this suggested using something that implements [SeedableRng](https://docs.rs/rand/latest/rand/trait.SeedableRng.html) and uses the std [Rng](https://docs.rs/rand/latest/rand/trait.Rng.html) which I would guess would be more intuitive and ergonomic for high-level usages.

Ideally, this example also has some comments in it, explaining what's actually happening? Seems like it would be more intuitive to use something like this with https://doc.rust-lang.org/std/ops/struct.Range.html also

agostbiro:
Thanks for the PR @deveusss ! I'm closing it for now due to lack of activity, but created a separate issue (#1051) as it'd be great to have a good example there.

PR #693, opened 2022-01-11
changed to match standard

mattlockyer:
Standard specifies no return value for this callback method.

https://nomicon.io/Standards/NonFungibleToken/ApprovalManagement.html#approved-account-contract-interface

### DIFF
--- a/near-contract-standards/src/non_fungible_token/approval/approval_receiver.rs
+++ b/near-contract-standards/src/non_fungible_token/approval/approval_receiver.rs
@@ -22,5 +22,5 @@ pub trait NonFungibleTokenApprovalReceiver {
         owner_id: AccountId,
         approval_id: u64,
         msg: String,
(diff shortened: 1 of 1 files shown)


austinabell:
hmm, strange, seems that `nft_approve` is forwarding any result from this call as its return. I don't think necessarily that this function signature violates the standard though, since adding a return doesn't break any call unless someone for some reason asserted that the return bytes were empty?

In any case, this is a bad signature because it only allows `String` return anyway. I won't weigh in on the standards part of this, but changing this from its current signature is something I've wanted to do before, but I will wait until it is agreed on the direction to go with this.

mattlockyer:
Just found it confusing.

agostbiro:
Thanks for opening this PR and raising the issue @mattlockyer ! We need to discuss more how to approach it, so I opened a new [issue](https://github.com/near/near-sdk-rs/issues/1053) for it and will close this PR in favor of that. Hope you don't mind and it would be great if you could participate in the discussion in #1053.

Issue #662, opened 2021-12-07
Deprecate or update near-sdk-sim to work with newly released nearcore crates


austinabell:
Currently, after #639, we are using the updated versions of nearcore crates for mocking in unit tests. There have been changes to the release and we no longer have crates required by near-sdk-sim to be able to update those crates. This leads to having duplicate crates and borsh dependencies when using sim tests.

Given the plans to deprecate near-sdk-sim once [workspaces-rs](https://github.com/near/workspaces-rs) is at a workable version, it is likely this will come before there is a workaround found for this. Making near-sdk-sim compatible with new crates would delay its deprecation and allow it to come out of a pre-release state for version 4.0

Issue #563, opened 2021-09-08
Port examples with simulation tests to use Workspace Testing


mikedotexe:
These examples include:

- [x] Callback results
- [x] Cross-contract high level
- [x] Cross-contract low level
- [x] Fungible token
- [x] Non-fungible token

Acceptance criteria: the simulation tests are deprecated from these examples with their tests replaced by using Sandbox Testing, preferably the Rust approach from https://github.com/near/runner-rs

PR #694, opened 2022-01-11
refactor: replace callback-results example sim tests with workspaces

austinabell:
on path to deprecate sim tests we need to replace all usages in examples

Related to https://github.com/near/near-sdk-rs/issues/563 https://github.com/near/near-sdk-rs/issues/662


examples/callback-results/src/lib.rs
```diff
@@ -65,3 +65,63 @@ impl Callback {
...
+        assert_eq!(res.json::<u8>()?, 8);
+
+        // Following tests the function call where the `call_all` function always succeeds and handles
+        // the result of the async calls made from within the function with callbacks.
+
+        // No failures
+        let res = contract
+            .call(&worker, "call_all")
+            .args_json((false, 1u8))?
+            .gas(300_000_000_000_000)
```
itegulov, line 93: Do we not have a constant for the max prepaid gas somewhere?

itegulov (approved):

ChaoticTempest (approved):

examples/callback-results/src/lib.rs
austinabell: not currently, maybe a good decision for some cases, alternative may be to have another function called `max_gas` or something to avoid having a constant that they have to find and match themselves.

@ChaoticTempest you might have an idea if any of this would be worth it for workspaces :)

matklad (approved):
A little bit more wordy, but makes so much more sense to me personally now. Excited to see workspaces shaping up, great job @ChaoticTempest!

examples/callback-results/src/lib.rs
```diff
@@ -65,3 +65,63 @@ impl Callback {
...
 }
+
+#[cfg(all(test, not(target_arch = "wasm32")))]
+mod tests {
+    use tokio::fs;
+    use workspaces::prelude::*;
+
+    #[tokio::test]
+    async fn workspaces_test() -> anyhow::Result<()> {
+        let wasm = fs::read("res/callback_results.wasm").await?;
```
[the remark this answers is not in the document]
austinabell: Related https://github.com/near/workspaces-rs/issues/54

examples/callback-results/src/lib.rs
austinabell: the only negative with defaulting max gas is that this API is agnostic of network, so this might make sense with sandbox or testnet, but could lead to issues if the developer is unaware of this default on mainnet.

Either case, this will affect when we update workspaces, so I don't think this is blocking this PR?

matklad: Def not blocking!

>the only negative with defaulting max gas is that this API is agnostic of network, so this might make sense with sandbox or testnet, but could lead to issues if the developer is unaware of this default on mainnet.

Do we actually inted workspaces to be used with `mainnet`? My gut feeling is that having a single library to be used both as a testing DSL and as the primary way to do stuff in prod is not great, as the incentives for the two use-cases are rather opposing. I'd probably expect us to:

* start with official low-level RPC-style API which is network-agnostic (this I think what `near-json-rpc-client` is)? This thing needs to be robust and complete, but can be ugly and verbose to use
* add official testing DSL which is nice to use, but not necessary maximally complete or robust
* experiment with unofficial or semi-official "nice" clints for the mainnet.

austinabell: That's a very valid point, but the issue is just that with the need for users to interact with mainnet to be able to spoon state or pull any data they can query for into the testing environment, my intuition is that we might not want to be too restrictive on what they can do.

Possible that we might want to disallow sending transactions through mainnet, but given how intentional it has to be to set up a key to do so, not sure it's really that much of a foot gun? @ChaoticTempest maybe you have some input with this

itegulov:
I have just noticed that callback-results example is not being tested by either [`test_examples`](https://github.com/near/near-sdk-rs/blob/master/.github/workflows/test_examples.yml) or [`test_examples_small`](https://github.com/near/near-sdk-rs/blob/master/.github/workflows/test_examples_small.yml). Is there some reasoning behind that or is this just an oversight?

austinabell:
> I have just noticed that callback-results example is not being tested by either [`test_examples`](https://github.com/near/near-sdk-rs/blob/master/.github/workflows/test_examples.yml?rgh-link-date=2022-01-18T00%3A19%3A17Z) or [`test_examples_small`](https://github.com/near/near-sdk-rs/blob/master/.github/workflows/test_examples_small.yml?rgh-link-date=2022-01-18T00%3A19%3A17Z). Is there some reasoning behind that or is this just an oversight?

Well, it is already tested through buildkite. Not really sure why there is this overlap tbh. Might be worth in the future removing tests from buildkite and just doing those through github actions

itegulov:
> Not really sure why there is this overlap tbh.

Yeah, that was my follow-up question :)

examples/callback-results/src/lib.rs
ChaoticTempest: The thing was that a good amount of users wanted the ability to be able to migrate their workflow from a testing environment to a live/prod environment so that they don't have to redo it all over again. That way they can easily migrate their flow having properly tested it in sandbox and having some confidence that it won't fail. I can see us disabling things or hiding things behind feature gates if things don't align well with testing like sending `mainnet` transactions though

PR #699, opened 2022-01-17
refactor: replace fungible token example sim tests with workspaces

itegulov:
I had to cut a few asserts from some of the tests since workspaces does not expose some of the stuff they needed, but at the very least all of the tests have survived the migration without losing meaningfulness. I left TODOs in appropriate places for us to revisit later.

Also, I took the liberty to delete `examples/heavy.rs` since it was originally added in #307 to showcase sim tests performance. But lmk if you would prefer for me to keep and refactor it too.

Related to #563 #662


examples/fungible-token/tests/workspaces.rs
```diff
@@ -0,0 +1,413 @@
...
+// have to be vague about the type
+async fn register_user<N: Network>(
+    worker: &Worker<N>,
+    contract: &Contract,
+    account_id: &AccountId,
+) -> anyhow::Result<()> {
+    let none: Option<bool> = None;
+    let res = contract
+        .call(&worker, "storage_deposit")
+        .args_json((account_id, none))?
```
itegulov, line 24: @austinabell is there a way to instantiate `None` with an explicit generic type in one line? I have tried `None: Option<bool>`, but apparently type ascriptions are experimental.

austinabell (commented):
oops forgot to finalize this review, sorry if any are stale comments. Was just a quick scan until the CI passes, which seemed more workspaces/sandbox utils related

examples/fungible-token/tests/workspaces.rs
```diff
@@ -0,0 +1,413 @@
...
+use near_sdk::json_types::U128;
+use near_sdk::ONE_YOCTO;
+use tokio::fs;
+use workspaces::prelude::*;
+use workspaces::{Account, Contract, DevNetwork, Network, Worker};
+
+// TODO: find/come up with a replacement for `to_yocto` from simtests
+const TEN_THOUSAND_NEAR: u128 = 100_000_000_000_000_000_000_000_000_000;
+const ONE_HUNDRED_NEAR: u128 = 100_000_000_000_000_000_000_000_000;
+const FIFTY_NEAR: u128 = 50_000_000_000_000_000_000_000_000;
```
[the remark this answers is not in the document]
itegulov: Yeah, at first I have refactored these values to look like `10000 * ONE_NEAR`, but later I found out about `parse_near!` so ended up with in-place `parse_near!("10000 N")` instead. lmk which way you prefer.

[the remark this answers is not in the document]
itegulov: Yeah, at first I have refactored these values to look like `10000 * ONE_NEAR`, but later I found out about `parse_near!` so ended up with in-place `parse_near!("10000 N")` instead. lmk which way you prefer.

examples/fungible-token/tests/workspaces.rs
```diff
@@ -0,0 +1,413 @@
...
+    // TODO: Pack into one transaction
+    let none: Option<String> = None;
+    let res = contract
+        .call(&worker, "ft_transfer_call")
+        .args_json((defi_contract.as_account().id(), transfer_amount, none, "10"))?
+        .gas(300_000_000_000_000)
+        .deposit(ONE_YOCTO)
+        .transact()
+        .await?;
+    assert!(matches!(res.status, FinalExecutionStatus::SuccessValue(_)));
```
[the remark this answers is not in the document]
itegulov: I agree, that would simplify user's life

examples/fungible-token/tests/workspaces.rs
```diff
@@ -0,0 +1,413 @@
...
+        .gas(300_000_000_000_000)
+        .deposit(ONE_YOCTO)
+        .transact()
+        .await?;
+    assert!(format!("{:?}", res.status.as_failure())
+        .contains("Can't unregister the account with the positive balance without force"));
+
+    let res = contract
+        .call(&worker, "storage_unregister")
+        .args_json(vec![Some(false)])?
```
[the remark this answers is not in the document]
itegulov: Yep, already figured this out with @ChaoticTempest's help

examples/fungible-token/tests/workspaces.rs
itegulov: The problem with `()` is that it does not match the function's signature. Even if `None` and `()` do serialize the same way, as an external user reading these examples I would be confused why we are passing `()` instead of `Option<bool>`.

`Option::<bool>::None` works though, thanks!

examples/fungible-token/tests/workspaces.rs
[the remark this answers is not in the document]
itegulov: I agree, that would simplify my life as a workspaces user

itegulov:
> oops forgot to finalize this review, sorry if any are stale comments.

No worries. Indeed, I have already refactored most of the weird stuff away.

> Was just a quick scan until the CI passes, which seemed more workspaces/sandbox utils related

Yeah, I am still trying to figure out why Github Actions don't want to run the downloaded sandbox binaries... There is some weirdness going on in Buildkite too, but I have not looked too deeply into that yet.

examples/fungible-token/tests/workspaces.rs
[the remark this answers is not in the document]
ChaoticTempest: Interesting. Normally the pattern for these was to do an `into_result()?` to forward the error, but I see that in this case we want to do assertions to be more explicit. Good idea overall, I'll add a ticket

examples/fungible-token/tests/workspaces.rs
```diff
@@ -0,0 +1,431 @@
...
+    // {
+    //     assert!(execution_error.to_string().contains("ParseIntError"));
+    // } else {
+    //     unreachable!();
+    // }
+
+    // balances remain unchanged
+    let root_balance = contract
+        .call(&worker, "ft_balance_of")
+        .args_json((contract.as_account().id(),))?
```
ChaoticTempest, line 417: It's not necessary to have the `as_account()` to retrieve the id. You can simply just do `contract.id()` for these

examples/fungible-token/tests/workspaces.rs
```diff
@@ -0,0 +1,426 @@
...
+    assert!(matches!(res.status, FinalExecutionStatus::SuccessValue(_)));
+
+    let res = ft_contract
+        .as_account()
+        .create_subaccount(&worker, "alice")
+        .initial_balance(parse_near!("10 N"))
+        .transact()
+        .await?;
+    assert!(matches!(res.details.status, FinalExecutionStatus::SuccessValue(_)));
+    let alice = res.result;
```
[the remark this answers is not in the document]
itegulov: indeed, thanks

examples/fungible-token/tests/workspaces.rs
itegulov: good point!

examples/fungible-token/tests/workspaces.rs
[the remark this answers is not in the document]
itegulov: Well, `CallExecutionDetails` does not have `into_result()` anyway. Thanks for the PR, I will take a look shortly

examples/fungible-token/tests/workspaces.rs
```diff
@@ -0,0 +1,413 @@
...
+    let initial_balance = U128::from(parse_near!("10000 N"));
+    let transfer_amount = U128::from(parse_near!("100 N"));
+    let worker = workspaces::sandbox();
+    let (contract, _, defi_contract) = init(&worker, initial_balance).await?;
+
+    // defi contract must be registered as a FT account
+    register_user(&worker, &contract, defi_contract.id()).await?;
+
+    // root invests in defi by calling `ft_transfer_call`
+    // TODO: Pack into one transaction
```
austinabell, line 197: can you explain what you mean by this? Not completely intuitive

austinabell (approved):
LGTM, I'll wait for @ChaoticTempest approval before pulling in because he might have some tricks :)

ChaoticTempest (approved):
looks good to me as well. Will be adding the promise results soon, so hopefully we can add in those commented out `promise_error` checks

examples/fungible-token/Cargo.toml
```diff
@@ -5,8 +5,13 @@ authors = ["Near Inc <hello@nearprotocol.com>"]
 edition = "2018"
 
 [dev-dependencies]
+anyhow = "1.0"
+near-primitives = "0.5.0"
```
austinabell, line 9: yeah an extension of what Phuong mentioned, this shouldn't be included
```suggestion
```

examples/fungible-token/tests/workspaces.rs
itegulov: yeah sorry for using incorrect terminology. What I meant to say is that these two actions (fn call `ft_transfer_call` and fn call `storage_unregister`) are supposed to be [batched](https://nomicon.io/RuntimeSpec/Transactions#batched-transaction) into a single transaction according to the old simtest-powered tests. Since workspaces do not support batched transactions yet, I had to split these two actions into two separate transactions. But to be fair I do not really see why you would want to batch these two specific action other than to (maybe?) save some gas... Atomicity does not matter here I think.

examples/fungible-token/Cargo.toml
itegulov: As @ChaoticTempest mentioned in the other thread I actually need this to pull `FinalExecutionStatus`

austinabell: This seems like something that shouldn't be exposed through workspaces

itegulov: Hmm, do you mean that workspaces should have its own type for expressing the call execution status?

CC @ChaoticTempest

austinabell: yeah, this is not an ideal pattern for workspace users to have to import a nearcore internal crate. Not a big deal, but something that should be changed at some point

ChaoticTempest: ooof, this was a pattern I wasn't expecting to cover with `asserts!` with `FinalExecutionStatus`. Think we can defer this later until a new minor version of workspaces is released and we can convert these `asserts!` over to using `is_success` calls

itegulov: > Think we can defer this later until a new minor version of workspaces is released and we can convert these `asserts!` over to using `is_success` calls

Yeah, we can defer for sure and `is_success` will help in this case, but `CallExecutionDetails::status` (which is a part of the public API) will still be a type from `near-primitives` so I think eventually we would want to replace it with something else.

PR #701, opened 2022-01-18
fix a typo in a comment

adsick:
not sure whether it was a "must return "0"" or "returns "0"" so correct if wrong

### DIFF
--- a/near-contract-standards/src/fungible_token/core_impl.rs
+++ b/near-contract-standards/src/fungible_token/core_impl.rs
@@ -48,7 +48,7 @@ pub trait FungibleTokenContract {
     /// Returns the total supply of the token in a decimal string representation.
     fn ft_total_supply(&self) -> U128;
 
-    /// Returns the balance of the account. If the account doesn't exist must return `"0"`.
+    /// Returns the balance of the account. If the account doesn't exist `"0"` must be returned.
     fn ft_balance_of(&self, account_id: AccountId) -> U128;
 }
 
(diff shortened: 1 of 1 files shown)


near-contract-standards/src/fungible_token/core_impl.rs
```diff
@@ -48,7 +48,7 @@ pub trait FungibleTokenContract {
     /// Returns the total supply of the token in a decimal string representation.
     fn ft_total_supply(&self) -> U128;
 
-    /// Returns the balance of the account. If the account doesn't exist must returns `"0"`.
+    /// Returns the balance of the account. If the account doesn't exist must return `"0"`.
```
willemneal, line 51:
```suggestion
    /// Returns the balance of the account. If the account doesn't exist `"0"` must be returned.
```

austinabell:
```suggestion
    /// Returns the balance of the account. If the account doesn't exist, `"0"` must be returned.
```
Comma seems important to be correct

austinabell:
Thanks for noticing this! Will merge once the comment about wording is addressed :)

near-contract-standards/src/fungible_token/core_impl.rs
```diff
@@ -48,7 +48,7 @@ pub trait FungibleTokenContract {
     /// Returns the total supply of the token in a decimal string representation.
     fn ft_total_supply(&self) -> U128;
 
-    /// Returns the balance of the account. If the account doesn't exist must returns `"0"`.
+    /// Returns the balance of the account. If the account doesn't exist `"0"` must be returned.
```
austinabell, line 51:
```suggestion
    /// Returns the balance of the account. If the account doesn't exist, `"0"` must be returned.
```
Comma is required to make sense, but I'm happy to fix in a follow up if you prefer

austinabell (approved):

PR #702, opened 2022-01-20
refactor: replace non-fungible token example sim tests with workspaces

itegulov:
Related to #563 #662


itegulov:
FYI I took the liberty to merge some of the tests introduced in https://github.com/near/near-sdk-rs/pull/627 with other already existing tests to avoid bloating the test files (and because some of them were even a strict superset of other ones).

ChaoticTempest (approved):
LGTM after all the kinks were worked out in the fungible PR

austinabell (approved):
LGTM but yeah will definitely replace the `matches!` usages for checking for success when that's available in workspaces.

PR #720, opened 2022-01-28
Add M1 support for contract builder

evgenykuzyakov:
Add support for Mac M1 hardware to use with contract builder. Requires to use `docker buildx`

### DIFF
--- a/contract-builder/README.md
+++ b/contract-builder/README.md
@@ -26,6 +26,10 @@ If you need to compile some other contracts, you can first export the path to th
 export HOST_DIR=/root/contracts/
 ```
 
+## Using on Mac M1 hardware
+
+Use `build_m1.sh` and `run_m1.sh` instead.
+
 ## Build contracts in docker
 
 Enter mounted path first:
--- a/contract-builder/run_m1.sh
+++ b/contract-builder/run_m1.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+HOST_DIR="${HOST_DIR:-$(pwd)/..}"
+
+docker run \
+     --platform linux/arm64 \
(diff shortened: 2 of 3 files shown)


contract-builder/README.md
```diff
@@ -26,6 +26,10 @@ If you need to compile some other contracts, you can first export the path to th
 export HOST_DIR=/root/contracts/
 ```
 
+## Using on Mac M1 hardware
+
+Use `build_m1.sh` and `run_m1.sh` instead.
```
itegulov, line 31: Wouldn't it make more sense to add and if check (like ``if [[ `uname -m` == 'arm64' ]]; then``) to the regular files?

contract-builder/run_m1.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+HOST_DIR="${HOST_DIR:-$(pwd)/..}"
+
+docker run \
+     --platform linux/arm64 \
```
itegulov, line 6: We should add `--platform linux/arm64` to `examples/docker_build.sh` as well. But I think we would have to publish an `arm64` docker image to our Dockerhub. Not sure who manages it though.

CC @austinabell @ChaoticTempest

itegulov (commented):
Thanks! I was meaning to do this for a little while, but never quite got my hands on it :)

contract-builder/run_m1.sh
austinabell: related @rtsai123 @anshal-savla might be worth considering if updating the docker image.

itegulov:
I believe this was handled by #751, but feel free to reopen if we missed something

PR #721, opened 2022-01-30
feat[test_utils]: new macro for unit testing with no boilerplate.

On0n0k1:
Introduce an optional macro with name "context_generator" that allows unit-testing to be smaller, easier to learn and faster to implement. Details in documentation.

### DIFF
--- a/near-sdk/src/test_utils/context_generator.rs
+++ b/near-sdk/src/test_utils/context_generator.rs
@@ -1,0 +1,406 @@
+use crate::{
+    AccountId,
+    Balance,
+    BlockHeight,
+    EpochHeight,
+    Gas,
+    PublicKey,
+    StorageUsage,
+    test_utils::{VMContextBuilder},
+};
+
+use std::convert::{
+    TryFrom,
+    TryInto,
+};
+
+fn to_bool(bytes: &[u8]) -> bool {
+    let (int_bytes, _) = bytes.split_at(std::mem::size_of::<bool>());
+    // If  the byte == 1 return true, else false
+    let byte = u8::from_le_bytes(int_bytes.try_into().unwrap());
+    assert!(byte<=1, "Invalid type for boolean, byte value is ({})", byte);
+    byte == 1
+}
+
+//
+// Most of the traits below are just copy and paste to allow the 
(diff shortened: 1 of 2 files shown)


austinabell:
I'm really not convinced this is a better direction. Using the builder pattern is better in most cases because users don't have to read the macro to understand which syntax is available. I don't believe this way is any more succinct for developers, either, so I'm having a hard time rationalizing adding this macro to the API.

@ChaoticTempest @itegulov wdyt?

On0n0k1:
When we search for testing_env! macro in the documentation, this is what we get 'https://docs.rs/near-sdk/latest/near_sdk/macro.testing_env.html'. I understand the concern, but when users have to call the builder and build for every single test, we end up with boilerplate that takes 4 to 5 lines and actual test implementation that takes 1 to 3 lines. In other words, code gets harder to read and slower to implement.

Why not have both options available? A high level option for users that only want to write code fast, and a low level option for users that want to be more specific with their tests.

On0n0k1:
This macro already creates a linter warning when there's an odd number of arguments. What if I implemented the macro in a way that it also warns what types of arguments it accepts?  Something like 'Invalid arg for method_name. Accepted args can be: "current_account_id", "signer_account_id", "signer_account_pk", "predecessor_account_id", "block_index" or "block_height", "block_timestamp", "epoch_height", "account_balance", "account_locked_balance", "storage_usage", "attached_deposit", "prepaid_gas", "random_seed" and "is_view".' .

On0n0k1:
Instead of a macro, maybe it would be better to just have a snippet then. Thanks for reviewing.

austinabell:
> I understand the concern, but when users have to call the builder and build for every single test, we end up with boilerplate that takes 4 to 5 lines and actual test implementation that takes 1 to 3 lines.

You don't actually need the boilerplate

> Why not have both options available? A high level option for users that only want to write code fast, and a low level option for users that want to be more specific with their tests.

Because more surface area on the API makes it harder to maintain and improve in the future.

> Instead of a macro, maybe it would be better to just have a snippet then. Thanks for reviewing.

Will look into better docs/examples or maybe the current API can be improved in a backwards compatible way :)

PR #727, opened 2022-02-03
fix(contract-standards): Change token_id field name on Token to id

austinabell:
Not sure which is less friction, to change this or the standard. Seems that these have been mismatched this whole time.

@mattlockyer @mikedotexe let me know if any of you have thoughts on this.

Thanks Peter and @evgenykuzyakov for pointing out the issue here https://twitter.com/salomonsen_p/status/1489264755567169539

I could have changed this to just update the serialization format, but I've changed the field name to the correct name to avoid this being a hidden change and intentional updating.


austinabell:
closed in favour of https://github.com/near/NEPs/pull/315

PR #734, opened 2022-02-16
fix(NFT Standard): add doc strings to macros so generated methods have docs

willemneal:
Currently copy pasting them, but perhaps both methods could use a macro to include an external string.


austinabell (changes requested):
Rust docs are added to impls by default unless overridden, why the need for duplicate docs? I don't see anything different here

willemneal:
If users use the macro to generate the methods then they won't get them doc
string when hovering over one in their IDE.

willemneal:
I would like to not duplicate the comments, which is what I was saying in my description.  It would be nice to be able to inject the comment.

PR #736, opened 2022-02-17
feat(sdk::env): add `block_timestamp_ms`

willemneal:
The standards define some time values in ms, which could cause confusion if users expect timestamp to return ms.  This would also prevent some users from converting to ms incorrectly.


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## [unreleased]
 
+- Added convenience function `env::block_timestamp_ms` to return ms since 1970. [PR 736](https://github.com/near/near-sdk-rs/pull/728)
```
austinabell, line 5: Can you move this to the features category?

near-sdk/src/environment/env.rs
```diff
@@ -183,6 +183,11 @@ pub fn block_timestamp() -> u64 {
     unsafe { sys::block_timestamp() }
 }
 
+/// Current block timestamp, i.e, number of non-leap-milliseconds since January 1, 1970 0:00:00 UTC.
+pub fn block_timestamp_ms() -> u64 {
+    block_timestamp() / 10u64.pow(6)
```
austinabell, line 188:
```suggestion
    block_timestamp() / 1_000_000
```
pow isn't `const` afaik

PR #738, opened 2022-02-21
fix: handle callbacks from functions without a return type in a special way

### LINKED ISSUE #729: No way to gracefully handle a callback that doesn't have return value
Currently, the `#[callback_result] a: Result<T, PromiseError>` pattern expects to deserialize `T`, by default with serde_json. In most cases, this isn't an issue but can become a weird pattern if someone tried to apply this to a callback from a function that doesn't serialize a response.

My intuition is that there should be a special case for `Result<(), PromiseError>` that ignores the deserialization (because empty bytes do not deserialize into a null value) to allow for this. The problem with this solution is that if callback results were misordered, that error would not be caught because it never reads and attempts to deserialize the result. Perhaps a workaround is to allow empty bytes or a null value for this special type. 

Definitely open to ideas here on how to improve this API

itegulov:
Fixes #729 

I fixed this in a very straightforward way by extracting the first generic type for all `#[callback_result]` function parameters and checking if it is unit or not. This also adds a nice benefit of a compile time check that all `#[callback_result]` parameters have the right type.

But I am not super happy with how it turned out in the end. Since we do not have access to semantic information in macros there is no certain way to check whether the given type is actually the kind of `Result` we want. If a user was to type a parameter as `core::result::Result` or via a `Result` type alias, the macro would panic even though it shouldn't. If the user had something like
```rust
use anyhow::Result;

fn foo(#[callback_result] a: Result<String, PromiseError>)
```

it won't panic even though it should.

The other alternative would be to change our serialization logic to make all functions that don't have an explicit return type (which is a much easier check) to return "null" as their response, but I am not sure if it is a good idea to change how all already existing contracts behave and I will be introducing maybe a miniscule, but still overhead. So what do you guys think?


near-sdk-macros/src/core_impl/code_generator/attr_sig_info.rs
```diff
@@ -191,10 +192,40 @@ impl AttrSigInfo {
...
+                            // But when someone tries to refer to their callback result with
+                            // `#[callback_result]` they specify the callback type as
+                            // `Result<(), PromiseError` which cannot be correctly deserialized from
+                            // an empty byte array.
+                            //
+                            // So instead of going through serde, we do a manual check
+                            // that the byte array is either empty or contains the string "null".
+                            syn::Type::Tuple(type_tuple) if type_tuple.elems.is_empty() => {
+                                quote! {
+                                    near_sdk::PromiseResult::Successful(data) if data.is_empty() || data == vec!(110, 117, 108, 108) =>
```
austinabell, line 212:
```suggestion
                                    near_sdk::PromiseResult::Successful(data) if data.is_empty() || &data == b"null" =>
```

`vec!` will allocate, and we don't need this to check equality.

Also, `null` will be deserialized into `()`, so this code path isn't really necessary. Also, I'm pretty sure this will fail when using another serialization protocol, try it with `borsh`

near-sdk-macros/src/core_impl/code_generator/attr_sig_info.rs
```diff
@@ -191,10 +192,40 @@ impl AttrSigInfo {
...
+                                        Ok(()),
+                                    near_sdk::PromiseResult::Successful(_) =>
+                                        panic!("Failed to deserialize callback using JSON: expected\
+                                            an empty byte array or \"null\"")
```
austinabell, line 216: This fails when the Ok type is a tuple.

near-sdk-macros/src/core_impl/code_generator/attr_sig_info.rs
```diff
@@ -191,10 +192,40 @@ impl AttrSigInfo {
...
+                            panic!("Function parameters marked with #[callback_result] should have \
+                                type Result<T, PromiseError>")
+                        );
+
+                        let deserialization_branch = match ok_type {
+                            // The unit type in this context is a bit special because functions
+                            // without an explicit return type do not serialize their response.
+                            // But when someone tries to refer to their callback result with
+                            // `#[callback_result]` they specify the callback type as
+                            // `Result<(), PromiseError` which cannot be correctly deserialized from
```
austinabell, line 205:
```suggestion
                            // `Result<(), PromiseError>` which cannot be correctly deserialized from
```

austinabell (changes requested):

near-sdk-macros/src/core_impl/code_generator/attr_sig_info.rs
itegulov: All good points. I made so that the flow goes through the regular deserialization process if the array is not empty, this should resolve all the weirdness you have pointed out.

near-sdk-macros/src/core_impl/code_generator/attr_sig_info.rs
itegulov: Not sure what you mean here?

austinabell: Just noting that if the type was `Result<(X, Y), PromiseError>` or something like this, it would have failed by the previous logic. I believe it's fixed with the changes below, but I will double-check tomorrow

itegulov: Nah, I am checking that the tuple is empty so we only enter this branch when the tuple is `()`

austinabell: ah, I misread and had a bug in a test case I made. Sorry!

near-sdk-macros/src/core_impl/code_generator/attr_sig_info.rs
```diff
@@ -191,10 +192,37 @@ impl AttrSigInfo {
                         }
                     }
                     BindgenArgType::CallbackResultArg => {
+                        let ok_type = extract_ok_type(ty).unwrap_or_else(||
+                            panic!("Function parameters marked with #[callback_result] should have \
+                                type Result<T, PromiseError>")
```
austinabell, line 197: note that this doesn't create a compiler error with span info, so the error line will be on `#[near_bindgen]` like:

```
24 | #[near_bindgen]
   | ^^^^^^^^^^^^^^^
   |
   = help: message: Function parameters marked with #[callback_result] should have type Result<T, PromiseError>
```

but you can have this error be placed on the violating line with something like https://docs.rs/syn/latest/syn/struct.Error.html and converting it into a compile error.

Can be done later, not a big deal

austinabell (approved):

near-sdk-macros/src/core_impl/code_generator/attr_sig_info.rs
itegulov: nice, didn't know there were proper diagnostics in Rust macros

PR #745, opened 2022-03-08
feat: add an optional way to handle contract errors with `Result`

### LINKED ISSUE #574: Proposal: return Result from contract rather than panicking
Hey!

I am still fairly new to writing Smart Contracts in Rust, but one thing that makes me feel really uncomfortable is the panicking.
I don’t feel like this is a very “rusty” way. In idiomatic Rust, I would rather expect to return a `Result<T, std::error::Error + Display>` and I don’t see a downside here, except that it needs to be supported by the SDK.

Returning a `Result`:
- would allow using `?` operator
- has same fail early mechanisms
- would let me use [thiserror](https://crates.io/crates/thiserror), which I absolutely love
- could use the `Display` trait for returning the error message

itegulov:
Fixes #574 

This PR provides an opt-in way to treat `Result` function return types as an error-handling mechanism. Meaning that only the underlying `Ok` values are serialized while the `Err` values result in panic. To do this, `FunctionError` must be implemented by the `Err`type (there are default implementations for types that have `Borrow<str>` and a derivation macro that generates a `ToString`-powered implementation).

The legacy behavior still works by default (`Result` is treated as a normal value and goes through full serialization), but it is supposed to generate compilation warnings (I am still trying to figure out a proper way to do this as`compile_warning!` seems to be missing rn, any suggestions are welcome). The user will have to mark the function with `#[return_result]` to opt-in for the new behavior.

### DIFF
--- a/near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
+++ b/near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
@@ -1,6 +1,7 @@
 use crate::core_impl::info_extractor::{
     AttrSigInfo, ImplItemMethodInfo, InputStructType, MethodType, SerializerType,
 };
+use crate::core_impl::utils;
 use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
 use syn::{ReturnType, Signature};
@@ -53,6 +54,7 @@ impl ImplItemMethodInfo {
             method_type,
             is_payable,
             is_private,
+            is_returns_result,
             ..
         } = attr_signature_info;
         let deposit_check = if *is_payable || matches!(method_type, &MethodType::View) {
@@ -136,6 +138,30 @@ impl ImplItemMethodInfo {
                     #method_invocation;
                     #contract_ser
                 },
+                ReturnType::Type(_, return_type)
+                    if utils::type_is_result(&return_type) && *is_returns_result =>
+                {
+                    let value_ser = match result_serializer {
+                        SerializerType::JSON => quote! {
+                            let result = near_sdk::serde_json::to_vec(&result).expect("Failed to serialize the return value using JSON.");
+                        },
+                        SerializerType::Borsh => quote! {
+                            let result = near_sdk::borsh::BorshSerialize::try_to_vec(&result).expect("Failed to serialize the return value using Borsh.");
+                        },
+                    };
+                    quote! {
+                        #contract_deser
+                        let result = #method_invocation;
+                        match result {
+                            Ok(result) => {
+                                #value_ser
+                                near_sdk::env::value_return(&result);
+                                #contract_ser
+                            }
+                            Err(err) => near_sdk::FunctionError::panic(&err)
+                        }
+                    }
+                }
                 ReturnType::Type(_, _) => {
                     let value_ser = match result_serializer {
                         SerializerType::JSON => quote! {
@@ -146,11 +172,11 @@ impl ImplItemMethodInfo {
                         },
                     };
                     quote! {
-                    #contract_deser
-                    let result = #method_invocation;
-                    #value_ser
-                    near_sdk::env::value_return(&result);
-                    #contract_ser
+                        #contract_deser
+                        let result = #method_invocation;
+                        #value_ser
+                        near_sdk::env::value_return(&result);
+                        #contract_ser
                     }
                 }
             }
--- a/near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
+++ b/near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
@@ -139,7 +139,7 @@ impl ImplItemMethodInfo {
                     #contract_ser
                 },
                 ReturnType::Type(_, return_type)
-                    if utils::type_is_result(&return_type) && *is_returns_result =>
+                    if utils::type_is_result(return_type) && *is_returns_result =>
                 {
                     let value_ser = match result_serializer {
                         SerializerType::JSON => quote! {
--- a/near-sdk-macros/src/lib.rs
+++ b/near-sdk-macros/src/lib.rs
@@ -205,8 +205,8 @@ pub fn function_error(item: TokenStream) -> TokenStream {
     };
     TokenStream::from(quote! {
         impl near_sdk::FunctionError for #name {
-            fn panic(&self) {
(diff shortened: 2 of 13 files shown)


near-sdk/compilation_tests/all.rs
```diff
@@ -19,4 +19,5 @@ fn compilation_tests() {
     t.pass("compilation_tests/cond_compilation.rs");
     t.compile_fail("compilation_tests/payable_view.rs");
     t.pass("compilation_tests/borsh_storage_key.rs");
+    t.pass("compilation_tests/function_error.rs");
```
austinabell, line 22: this test is awesome, perhaps it's worth adding a test with invalid format(s) to assert it fails consistently and with a valid error?

austinabell (commented):
Awesome work! Perhaps would be worthwhile to add this pattern to any examples where it would cleanup code, but lets do this in a follow up PR :)

near-sdk/src/types/error.rs
```diff
@@ -0,0 +1,36 @@
...
+///     }
+/// }
+/// ```
+pub trait FunctionError {
+    fn panic(&self);
+}
+
+impl<T> FunctionError for T
+where
+    T: Borrow<str>,
```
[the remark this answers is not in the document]
itegulov: Yeah, makes sense. Thanks for pointing out the implicit `Hash`, `Eq, `Ord` requirement, I was not aware of this.

near-sdk-macros/src/lib.rs
```diff
@@ -184,3 +184,30 @@ pub fn borsh_storage_key(item: TokenStream) -> TokenStream {
...
+                Span::call_site(),
+                "FunctionError can only be used as a derive on enums or structs.",
+            )
+            .to_compile_error(),
+        );
+    };
+    TokenStream::from(quote! {
+        impl near_sdk::FunctionError for #name {
+            fn panic(&self) {
+                near_sdk::env::panic_str(&std::string::ToString::to_string(&self))
```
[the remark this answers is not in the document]
itegulov: What is the actual difference between these two expressions? Is there a situation where the first `&` can be interpreted in a different way?

austinabell:
> > Perhaps would be worthwhile to add this pattern to any examples where it would cleanup code, but lets do this in a follow up PR :)
> 
> Yeah,`lockable-fungible-token` seems like a good candidate.
> 
> Also, not related to this PR in particular, but I don't think it makes sense that our only integration tests are essentially examples. I have had a few occasions where I wanted to add an e2e test but didn't want anyone to treat it as an example to follow.

Absolutely, this is a good idea to build out this scaffolding. I just suggested in examples since we do want to indicate to users that this pattern exists

near-sdk/compilation_tests/all.rs
itegulov: Ok so, I have spent half a day trying to figure out why a negative test I wrote was not failing at compile-time and finally realized that we are compiling all these files with the host target rather than wasm32-unknown-unknown, so none of the wasm-specific code we generate is tested here. I have tried making `trybuild` use a target different to the one we compile `near-sdk` crate with, but it does not seem like it is possible? I am a bit stumped here as to how to proceed. Maybe moving compilation tests into a different crate would help, but then we would still have to compile `trybuild` with wasm target and I do not know if it will play out well.

near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
```diff
@@ -136,6 +138,30 @@ impl ImplItemMethodInfo {
...
+                        let result = #method_invocation;
+                        match result {
+                            Ok(result) => {
+                                #value_ser
+                                near_sdk::env::value_return(&result);
+                                #contract_ser
+                            }
+                            Err(err) => near_sdk::FunctionError::panic(&err)
+                        }
+                    }
```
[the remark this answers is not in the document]
itegulov: Done

near-sdk/compilation_tests/all.rs
austinabell: > Ok so, I have spent half a day trying to figure out why a negative test I wrote was not failing at compile-time and finally realized that we are compiling all these files with the host target rather than wasm32-unknown-unknown, so none of the wasm-specific code we generate is tested here. I have tried making `trybuild` use a target different to the one we compile `near-sdk` crate with, but it does not seem like it is possible? I am a bit stumped here as to how to proceed. Maybe moving compilation tests into a different crate would help, but then we would still have to compile `trybuild` with wasm target and I do not know if it will play out well.

Oh, interesting. Yeah I wasn't aware of this. Yeah I’d say just leave it and we can create an issue and tackle it separately not to block this one.

austinabell (approved):

austinabell:
Oh also a changelog entry would be awesome, but nbd it can come after

near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
```diff
@@ -136,6 +138,30 @@ impl ImplItemMethodInfo {
                     #method_invocation;
                     #contract_ser
                 },
+                ReturnType::Type(_, return_type)
+                    if utils::type_is_result(return_type) && *is_returns_result =>
+                {
+                    let value_ser = match result_serializer {
+                        SerializerType::JSON => quote! {
+                            let result = near_sdk::serde_json::to_vec(&result).expect("Failed to serialize the return value using JSON.");
```
ChaoticTempest, line 146: I see that we're making similar calls to `expect` in a couple places, but @austinabell do we want to eventually get rid of these `expect` calls since they should be bringing the formatting mechanism. Or is there some reason we're keeping them?

ChaoticTempest (approved):
LGTM

near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
austinabell: iirc this expect doesn't embed file-specific data in the error message when used in a proc macro, but I could be remembering wrong why I didn't switch these.

Might be worth opening an issue and double-checking before next release, but wouldn't be a breaking change to come in after anyway

ChaoticTempest: ticket: https://github.com/near/near-sdk-rs/issues/746

PR #750, opened 2022-03-10
refactor: make lockable-ft example use `Result`-based error handling

itegulov:
Addresses https://github.com/near/near-sdk-rs/pull/745#pullrequestreview-903181784

Also covered this example with workspaces-powered tests since that is the only way to test the new error handling behavior.


examples/lockable-fungible-token/src/lib.rs
```diff
@@ -70,46 +70,54 @@ impl FunToken {
...
     /// If some amount of tokens is currently locked by the `escrow_account_id` the new allowance is
     /// decreased by the amount of locked tokens.
-    pub fn set_allowance(&mut self, escrow_account_id: AccountId, allowance: String) {
+    #[return_result]
+    pub fn set_allowance(
+        &mut self,
+        escrow_account_id: AccountId,
+        allowance: String,
+    ) -> Result<(), &'static str> {
         let allowance = u128::from_str(&allowance).expect("Failed to parse allowance");
```
ChaoticTempest, line 79: probably preferable if we switch these `.expect` calls over to early returning an error since our error type is a str anyways, but not a big deal if we leave it as is

ChaoticTempest (approved):

examples/lockable-fungible-token/src/lib.rs
```diff
@@ -119,26 +127,33 @@ impl FunToken {
...
     /// * The (`predecessor_id`) should have at least `unlock_amount` locked tokens from `owner_id`.
-    pub fn unlock(&mut self, owner_id: AccountId, unlock_amount: String) {
+    #[return_result]
+    pub fn unlock(
+        &mut self,
+        owner_id: AccountId,
+        unlock_amount: String,
+    ) -> Result<(), &'static str> {
         let unlock_amount =
             u128::from_str(&unlock_amount).expect("Failed to parse allow unlock_amount");
```
austinabell, line 146: same here, and a few other places

austinabell (commented):
oops forgot to finalize this review comment

PR #751, opened 2022-03-11
multi-architecture docker image

anshal-savla:
Ticket: https://nearinc.atlassian.net/browse/DPI-65

updated bash script to build multi-architecture Docker images (arm65/amd64) for contract-builder.


contract-builder/build.sh
```diff
@@ -1,2 +1,23 @@
...
+    docker buildx use --builder contract-builder
+else
+    docker buildx create --name contract-builder --use 
+fi
+
+if [[ ${branch} == "master" ]];then
+    docker buildx build --platform linux/amd64,linux/arm64 -t nearprotocol/"${image_name}:${branch}-${commit}" -t nearprotocol/${image_name}:latest --push .
+else 
+    docker buildx build --platform linux/amd64,linux/arm64 -t nearprotocol/"${image_name}:${branch}-${commit}" --push .
+fi
```
itegulov, line 23:
```suggestion
if docker buildx ls | grep -q contract-builder; then
    docker buildx use --builder contract-builder
else
    docker buildx create --name contract-builder --use
fi

if [[ ${branch} == "master" ]]; then
    docker buildx build --platform linux/amd64,linux/arm64 -t nearprotocol/"${image_name}:${branch}-${commit}" -t nearprotocol/${image_name}:latest --push .
else
    docker buildx build --platform linux/amd64,linux/arm64 -t nearprotocol/"${image_name}:${branch}-${commit}" --push .
fi
```

itegulov (commented):
Seems to work fine on both Linux and macOS with M1, but I am honestly not sure if contract-builder image related scripts and logic belong to this repo. If they do, shouldn't we also have a buildkite workflow that invokes this script? How can I try it out?

itegulov:
Also, while we are at it and before we publish the next version of the image, can we also update Dockerfile to use Rust 1.56 to unblock https://github.com/near/near-sdk-rs/pull/669?

mhalambek (approved):

austinabell (approved):
don't really see why this would exist within the SDK, but I suppose this could be moved out later

anshal-savla:
@itegulov @austinabell  Yes, I can update the rust version in the Docker image. I am Creating a buildkite pipeline to build and publish the docker image in the near-ops repo.

Where should this build script live if not in this repo?

austinabell:
> @itegulov @austinabell Yes, I can update the rust version in the Docker image. I am Creating a buildkite pipeline to build and publish the docker image in the near-ops repo.
> 
> Where should this build script live if not in this repo?

I suppose it's fine to be here, but maybe more ideal to move into its own repo in the future. Nothing here is tied to this SDK specifically.

austinabell:
> I suppose it's fine to be here, but maybe more ideal to move into its own repo in the future. Nothing here is tied to this SDK specifically.

Actually, second thought might be beneficial to be within SDK but only if we have building/using the image as part of the CI checks, which doesn't require building for the local script. This could be a benefit to make sure the script doesn't break examples and general use things within the SDK directly.

Otherwise, not a benefit since we just use the deployed image [here](https://github.com/near/near-sdk-rs/blob/master/examples/build_docker.sh).

PR #723, opened 2022-01-31
feat(contract-standards): add events to FT contract standard

itegulov:
As minting and burning are not a part of FT contract standard yet, these events are not being emitted

PR #752, opened 2022-03-11
feat: emit event for refund in `ft_resolve_transfer()`

think-in-universe:
In the standard `fungible_token` implementation, we have emitted `FtTransfer` in `ft_transfer()`, but no events was emitted for refund in `ft_resolve_transfer()`, which may introduce inconsistency if aggregating the data via FT events such as total deposits and withdrawals. 

To fix this issue, we can either reuse the current `FtTransfer`, `FtBurn` events or create a new `FtRefund` event if needed. In this PR, we reuse the existing events.


itegulov (commented):
Yeah, this makes sense to me. Missed this in https://github.com/near/near-sdk-rs/pull/723.

FYI you need to regenerate the examples' wasm files by running `./examples/build_all_docker.sh` in the project directory, otherwise the CI won't pass.

think-in-universe:
@itegulov thanks. updated.

austinabell:
I believe the rationale for leaving out in nft (and then what propagated to this) was that since these were optional (iirc) they should not be included in the standard. @telezhnaya @frol I don't have the context to know if this should change, could you weigh in when you get the chance?

If we change this, the NFT events should probably as well for consistency

itegulov:
> I believe the rationale for leaving out in nft (and then what propagated to this) was that since these were optional (iirc) they should not be included in the standard.

Explicit minting and burning are indeed optional, but `ft_resolve_transfer` is not. And since using `ft_resolve_transfer` can result in some tokens being burnt, I think it makes sense to emit an event.

> If we change this, the NFT events should probably as well for consistency

I have just checked NFT events, and we are already emitting a transfer event when returning an NFT to its previous owner: https://github.com/near/near-sdk-rs/blob/master/near-contract-standards/src/non_fungible_token/core/core_impl.rs#L526. And since NFT standard does not require storage deposits from users, it never has to burn NFTs by itself, so the flow there is a bit different.

telezhnaya:
If we send the token back, we should emit the event about it, otherwise, the data will be inconsistent. I agree with the proposed solution. 

I was also thinking about emitting the event only after we have the result of `ft_resolve_transfer()`. If the transfer failed, we could emit no event about it at all. But, it's better to show such unsuccessful attempts in the Wallet history, so I suggest using the current solution.

austinabell:
> If we send the token back, we should emit the event about it, otherwise, the data will be inconsistent. I agree with the proposed solution.
> 
> I was also thinking about emitting the event only after we have the result of `ft_resolve_transfer()`. If the transfer failed, we could emit no event about it at all. But, it's better to show such unsuccessful attempts in the Wallet history, so I suggest using the current solution.

Why is there this inconsistency with nft vs ft standard for the burn event? The same pattern happens with the burn in NFT https://github.com/near/near-sdk-rs/blob/e3c303f1c035f697071b0ba1bc76c0930ce498bb/near-contract-standards/src/non_fungible_token/core/core_impl.rs#L505 and it was suggested (https://github.com/near/near-sdk-rs/pull/717#issuecomment-1029458483) that only transfer be included in NFT standard.

frol:
> Why is there this inconsistency with nft vs ft standard for the burn event?

@austinabell Olga just wanted to confirm that without this fix there will, indeed, be inconsistency, and this fix is correct.

austinabell:
> > Why is there this inconsistency with nft vs ft standard for the burn event?
> 
> @austinabell Olga just wanted to confirm that without this fix there will, indeed, be inconsistency, and this fix is correct.

Reading over the logic again seems like my intuitions were incorrect. Seems the path I linked was handling the case where the token was burned already, so no need to emit the event here. I believe no other change is needed.

PR #754, opened 2022-03-15
feat: support `Result`-based error handling in init methods

itegulov:
While refactoring lockable-ft example in #750 I realized that init methods could also benefit from `Result`-based error handling. Also realized that I forgot to throw an error when a method that does not return `Result` is marked with `#[return_result]`.


near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
```diff
@@ -810,4 +810,63 @@ mod tests {
...
+        assert_eq!(expected.to_string(), actual.to_string());
+    }
+
+    #[test]
+    fn return_result_init_ignore_state() {
+        let impl_type: Type = syn::parse_str("Hello").unwrap();
+        let mut method: ImplItemMethod = parse_quote! {
+            #[init(ignore_state)]
+            #[return_result]
+            pub fn new() -> Result<Self, &'static str> { }
```
austinabell, line 851: Yeah, so this annotation is a bit misleading when used with init because nothing is actually returned. I wonder if there is a better name for this annotation.

austinabell (approved):

near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
itegulov: Yeah, not the best name for sure. How about `#[handle_result_error]` or even `#[handle_result]`?

austinabell: > Yeah, not the best name for sure. How about `#[handle_result_error]` or even `#[handle_result]`?

I personally like the latter since it's more concise (result_error is a bit redundant), but up to you!

ChaoticTempest (approved):
LGTM. And man they changed the Github `Files changed` page 🙃

itegulov:
@ChaoticTempest haha, are you talking about the file tree preview? I thought it was only available as a feature preview. Enabled it a few days ago and am very happy that they are finally adding something like this 😄

PR #757, opened 2022-03-21
feat: cause compilation to fail when trying to serialize `Result` value

### LINKED ISSUE #748: Make `Result` return value serialization print a deprecation warning
This is the logical continuation of https://github.com/near/near-sdk-rs/issues/574 - the path of gradually switching to `Result`-based error handling. As a next step, we need to make sure that using the legacy behavior of serializing a `Result` return value prints a deprecation warning.

itegulov:
Fixes #748

### DIFF
--- a/near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
+++ b/near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
@@ -4,6 +4,7 @@ use crate::core_impl::info_extractor::{
 use crate::core_impl::utils;
 use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
+use syn::spanned::Spanned;
 use syn::{ReturnType, Signature};
 
 impl ImplItemMethodInfo {
@@ -145,21 +146,21 @@ impl ImplItemMethodInfo {
                         }
                     }
                 }
-                ReturnType::Type(_, _) if *is_handles_result => {
+                ReturnType::Type(_, return_type) if *is_handles_result => {
                     return syn::Error::new(
-                        ident.span(),
+                        return_type.span(),
                         "Method marked with #[handle_result] should return Result<T, E>.",
                     )
                     .to_compile_error();
                 }
                 ReturnType::Type(_, return_type) if utils::type_is_result(return_type) => {
                     return syn::Error::new(
-                        ident.span(),
+                        return_type.span(),
                         "Serializing Result<T, E> has been deprecated. Consider marking your method \
                         with #[handle_result] if the second generic represents a panicable error or \
                         replacing Result with another two type sum enum otherwise. If you really want \
                         to keep the legacy behavior, mark the method with #[handle_result] and make \
-                        it return Result<Result<T, E>, ()>.",
+                        it return Result<Result<T, E>, near_sdk::Abort>.",
                     )
                     .to_compile_error();
                 }
@@ -296,8 +297,8 @@ fn init_method_wrapper(
                 }
             })
         }
-        ReturnType::Type(_, _) if *is_handles_result => Err(syn::Error::new(
-            ident.span(),
+        ReturnType::Type(_, return_type) if *is_handles_result => Err(syn::Error::new(
+            return_type.span(),
             "Method marked with #[handle_result] should return Result<T, E>",
         )),
         ReturnType::Type(_, _) => Ok(quote! {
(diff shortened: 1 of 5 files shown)


near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
```diff
@@ -148,7 +148,18 @@ impl ImplItemMethodInfo {
...
                     return syn::Error::new(
                         ident.span(),
-                        "Method marked with #[handle_result] should return Result<T, E>",
+                        "Method marked with #[handle_result] should return Result<T, E>.",
+                    )
+                    .to_compile_error();
+                }
+                ReturnType::Type(_, return_type) if utils::type_is_result(return_type) => {
+                    return syn::Error::new(
+                        ident.span(),
```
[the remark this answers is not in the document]
itegulov: Indeed, I have tried to create a span over `Type` values in the past unsuccessfully so I was kind of assuming it was impossible for some reason, but this time I realized I just had to use `syn::spanned::Spanned` :)

near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
```diff
@@ -148,7 +148,18 @@ impl ImplItemMethodInfo {
...
+                        "Serializing Result<T, E> has been deprecated. Consider marking your method \
+                        with #[handle_result] if the second generic represents a panicable error or \
+                        replacing Result with another two type sum enum otherwise. If you really want \
+                        to keep the legacy behavior, mark the method with #[handle_result] and make \
+                        it return Result<Result<T, E>, ()>.",
```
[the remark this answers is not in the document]
itegulov: That's fair, no idea why I thought there would be a `AsRef<str>` for `()` by default lol

I like the second way more too. I have documented it to the best of my ability and added an example so hopefully it won't confuse anyone. Lmk if you want something to be added/clarified there.

near-sdk/src/types/error.rs
```diff
@@ -32,3 +32,30 @@ where
...
+/// use near_sdk::{Abort, near_bindgen};
+///
+/// #[near_bindgen]
+/// #[derive(Default, BorshDeserialize, BorshSerialize)]
+/// pub struct Contract;
+///
+/// #[near_bindgen]
+/// impl Contract {
+///     #[handle_result]
+///     pub fn foo(&self) -> Result<Result<String, &'static str>, Abort> {
```
austinabell, line 50: Opinionated but seems more worth using the example with an actual pattern, `Abort` wouldn't have to be used for nested `Result` and this might be confusing

near-sdk/src/types/error.rs
```diff
@@ -32,3 +32,30 @@ where
...
+///         Ok(Ok("success".to_string()))
+///     }
+/// }
+/// ```
+pub struct Abort;
```
austinabell, line 55: prob best to derive Debug, Clone, PartialEq to not be prickly to use in tests and otherwise

austinabell (approved):

near-sdk/src/types/error.rs
itegulov: Hmm, I am confused. Isn't this the exact pattern we are trying to promote with `Abort`?

near-sdk/src/types/error.rs
ChaoticTempest: maybe also `impl std::error::Error` just in case someone wants to return `Box<dyn Error>` for multiple error types in the same function

ChaoticTempest (approved):

near-sdk/src/types/error.rs
austinabell: I disagree with this one. The point of this is to be optimized for minimal code size, so having this would make it silently use the formatting machinery for errors.

A bit opinionated so do either, but I think we only add this if there are valid requests from the community. I can't think of when this would be useful

near-sdk/src/types/error.rs
austinabell: Well, it makes this pattern easier, but this abort can be used more generally. If someone just wanted a minimal way to abort when errors were hit to avoid debug messages, this could be useful.

near-sdk/src/types/error.rs
itegulov: Yeah, I would rather leave out `std::error::Error` implementation. If the need arises, we can always add it.

Issue #619, opened 2021-10-29
Dynamic ft_transfer_calls in FT standard implementation


mfornet:
Current implementation of `ft_transfer_call` <span class="error">[allocates exactly 25TGas]</span>([https://github.com/near/near-sdk-rs/blob/master/near-contract-standards/src/fungible_token/core_impl.rs#L12](https://github.com/near/near-sdk-rs/blob/master/near-contract-standards/src/fungible_token/core_impl.rs#L12)) for the tx itself. In the case that less gas is required, the reminder gas can't be used for the rest of the transfer. But most importantly is the case when this gas is not enough to run the transfer, which can happen when the payload attached is too large. See [https://github.com/aurora-is-near/eth-connector/issues/27](https://github.com/aurora-is-near/eth-connector/issues/27).

I propose to determine the gas used dynamically, in such a way that the gas attached to subsequence cross contract calls is determined from the gas remaining at the end of the function.

itegulov:
@austinabell although NEP264 will help with the first half of the issue (not being able to use the leftover gas for the rest of the transfer), I don't think it will help with the second one (running out of gas before calling `ft_on_transfer` despite passing the initial gas check which is arguably more severe). We would have to make `ft_transfer_call` estimate how much gas it takes dynamically from the payload size and in general a part of a bigger problem with static gas estimation tactics.

austinabell:
> @austinabell although NEP264 will help with the first half of the issue (not being able to use the leftover gas for the rest of the transfer), I don't think it will help with the second one (running out of gas before calling `ft_on_transfer` despite passing the initial gas check which is arguably more severe). We would have to make `ft_transfer_call` estimate how much gas it takes dynamically from the payload size and in general a part of a bigger problem with static gas estimation tactics.

This is true. Perhaps the solution we need is to expose is a way to have more granular control over gas used without breaking the simple path?

Although yes it doesn't solve the issue entirely, I don't see the issue with failing before calling `ft_on_transfer`. Yes, the transaction will fail, but none of the async calls will go off so it's much more acceptable. `ft_transfer_call` estimating how much gas it takes dynamically doesn't seem that feasible since it relies on current gas prices for operations and payload size is only one factor.

Also, we can't schedule the callback using the remaining gas after each receipt execution as suggested by the issue because it requires the contract implementing `ft_on_transfer` to schedule the callback, which they can just omit if they wanted.

[...]

PR #760, opened 2022-03-27
fix(contract-standards): required minimum gas is more than needed in `*_transfer_call()`

think-in-universe:
In `ft_transfer_call()` function implemented in the FT standard,

because here `GAS_FOR_FT_TRANSFER_CALL` already includes `GAS_FOR_RESOLVE_TRANSFER`, no need to count in `GAS_FOR_FT_TRANSFER_CALL` twice in `require!()`

https://github.com/near/near-sdk-rs/blob/2d7fc59c5450dfa301fa26de1c33a6ceb69408f9/near-contract-standards/src/fungible_token/core_impl.rs#L13

The same also happens to `nft_transfer_call()` in NFT standard,

`GAS_FOR_NFT_TRANSFER_CALL` already includes `GAS_FOR_RESOLVE_TRANSFER`

https://github.com/near/near-sdk-rs/blob/8b4f0d6779906b0a7947f75218a3b277b64e5e68/near-contract-standards/src/non_fungible_token/core/core_impl.rs#L19

Though we also need to consider the required gas for `ft_on_transfer()` and `nft_on_transfer()`, we'd better add another constant for that if needed.


itegulov:
Just realized that I don't know the purpose of these explicit gas checks. Is the idea just not to waste caller's gas when we know there is not enough?

> Though we also need to consider the required gas for ft_on_transfer() and nft_on_transfer(), we'd better add another constant for that if needed.

Well, those can be implemented in an arbitrary way by a user's contract, no? So we cannot know how much gas they would require.

think-in-universe:
> Just realized that I don't know the purpose of these explicit gas checks. Is the idea just not to waste caller's gas when we know there is not enough?

yeah. that's part of the reason. also, we'd like to attach enough gas, so ensure `*_resolve_transfer()` is executed anyway. 

> Well, those can be implemented in an arbitrary way by a user's contract, no? So we cannot know how much gas they would require.

right. we cannot predict the gas cost of the token receiver. so maybe we just don't count in the  `*_on_transfer()` and only require the minimum gas. Then let developers to decide the total attached gas in their frontend code.

itegulov:
> yeah. that's part of the reason. also, we'd like to attach enough gas, so ensure *_resolve_transfer() is executed anyway.

Be aware that there is still not guarantee that the async calls will happen, especially if you have a big payload. See https://github.com/near/near-sdk-rs/issues/619. But I am also not sure why you would want to be sure that `ft_resolve_transfer()` is always executed (@austinabell gave a good explanation [here](https://github.com/near/near-sdk-rs/issues/619#issuecomment-1080733095)). Do you have a specific use case in mind?

itegulov (approved):
The changes make sense to me, but you will still need another approval from @austinabell or @ChaoticTempest

austinabell (approved):
yes, this naming is also very confusing, because `GAS_FOR_FT_TRANSFER_CALL` includes gas for `RESOLVE_TRANSFER`, which is misleading.

PR #770, opened 2022-03-28
deprecate and remove unused proc_macro_attributes

### LINKED ISSUE #766: [EXT] NEAR SDK - Init Example
https://docs.rs/near-sdk/4.0.0-pre.7/near_sdk/attr.init.html

This function is missing an example.

Examples should be simple, easy-to-follow, and useful to make onboarding and learning Rust SDK as straightforward as possible. See array primitive for examples of code snippet examples: https://doc.rust-lang.org/nightly/core/primitive.array.html), https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#784-790 for examples in code.

austinabell:
closes #766 

There was no clean way of deprecating this re-export so I've just removed it since it's coming with a major version bump. Migration should be simple enough I think. Another option is exporting modules with the same name and deprecating them, but this clutters docs a bit too much since those are shown first.

Unless someone has a better alternative to deprecate these proc_macro_attribute macros?

### DIFF
--- a/near-sdk-macros/src/lib.rs
+++ b/near-sdk-macros/src/lib.rs
@@ -81,29 +81,34 @@ pub fn ext_contract(attr: TokenStream, item: TokenStream) -> TokenStream {
 
 /// `callback` is a marker attribute it does not generate code by itself.
 #[proc_macro_attribute]
+#[deprecated(since = "4.0.0", note = "Case is handled internally by macro, no need to import")]
 pub fn callback(_attr: TokenStream, item: TokenStream) -> TokenStream {
     item
 }
 
 /// `callback_args_vec` is a marker attribute it does not generate code by itself.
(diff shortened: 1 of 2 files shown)


itegulov (approved):

ChaoticTempest (approved):

fadeevab:
@austinabell Hi! What's gonna be an alternative of `callback_vec`? E.g. I need a handler which accepts variable number of return values from cross-contract calls. Thanks.

austinabell:
> @austinabell Hi! What's gonna be an alternative of `callback_vec`? E.g. I need a handler which accepts variable number of return values from cross-contract calls. Thanks.

This annotation isn't being removed, this is just the macro that is being removed. You don't need to import the macro to use and this macro wasn't being used. You'll be able to use `#[callback_vec]` as you did before.

PR #771, opened 2022-03-28
refactor!: move macro generated types to pseudo-private module

austinabell:
Noticed these as well, should have never been exported. Assuming they aren't used anywhere.

These changes are a bit opinionated so input would be appreciated.

1. Metadata moved because the format isn't stable and will likely change
2. `BorshIntoStorageKey` moved because it's just used by macros and can be very confusing with `BorshStorageKey`


ChaoticTempest (approved):
Makes sense to me to have this change as long as its only being used in the macros

itegulov (approved):
Makes sense to me. Metadata API is most likely going to change in the next release so hiding in this release is definitely a good idea

Issue #718, opened 2022-01-27
`near_bindgen` methods with no `self` check for attached deposit


austinabell:
update: as per https://github.com/near/near-sdk-rs/pull/774#issuecomment-1561817582, what should be done to resolve this issue, 
has changed. An implementor should check if https://github.com/near/near-sdk-rs/pull/774#issuecomment-1561817582 still holds true , i.e. if calling `near_sdk::env::attached_deposit()`  from view methods still doesn't panic.

---

Methods like this will add the codegen assumed when a contract mutates state. A slightly strange pattern we have that defaults to checking this only when the parameter is `&mut self` even though in any function there can be state changes not tied to the contract state. 

Tying the check for attached deposit to whether or not contract state changes is a strange pattern IMO but the band-aid fix is to just remove this check when contract state isn't used. This needs to be done so that these methods can be called with a view operation, otherwise, the runtime will fail when trying to call `env::attached_deposit` from within a view call.

```rust
#[near_bindgen]
pub struct SomeContract;

#[near_bindgen]
impl SomeContract {
    pub fn get_val(s: &String) -> Option<u8> 

    { // ... } 

// ...
```

Test for this looks like:

```rust
#[test]
fn no_self_no_args() {
    let impl_type: Type = syn::parse_str("Hello").unwrap();
    let mut method: ImplItemMethod = syn::parse_str("pub fn method() { }").unwrap();
    let method_info = ImplItemMethodInfo::new(&mut method, impl_type).unwrap();
    let actual = method_info.method_wrapper();
    let expected = quote!(
       #[cfg(target_arch = "wasm32")]
       #[no_mangle]
       pub extern "C" fn method() 

       { near_sdk::env::setup_panic_hook(); Hello::method(); } 

     );
    assert_eq!(expected.to_string(), actual.to_string());
}
```

which is currently failing because it adds the extra check

itegulov:
Just to be clear, codegen having a check for the attached deposit is a side effect of such functions being identified as call functions.

<sub>Also, the markdown is all messed up again, curse you exalate</sub>

HumbertoTM10:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/near-sdk-rs).

### My background and how it can be leveraged
I have 4 years working in development environments, have encountered several challenges, requests and teams focusing in reusable and efficient code, so that gives me a better understanding of the code and how to solve problems efficiently.


### How I plan on tackling this issue
[...]

PR #774, opened 2022-03-31
fix: remove payable check for non-mutable functions

austinabell:
Opinionated change, probably shouldn't come in, opening for discussion. It was a foot gun for what I was just doing, but it doesn't seem super clean to just remove.

Pros:
- Removes foot gun of trying to call a method without parameters as a view call and failing

Cons:
- Silently removes non-payable check to anyone using this method and modifying state manually

Neutral:
- Moves in direction of non-payable methods by default. Unclear if developers expect this or benefit from the security of this by default (probably requires developer input cc @AustinBaggio). Is a pattern in Ethereum by default, but doesn't fit cleanly into our model because `attached_deposit` fails by default in view context. (I will follow up to ask why this is the case and not just defaulting to 0 https://near.zulipchat.com/#narrow/stream/295306-pagoda.2Fcontract-runtime/topic/attached_deposit.20view.20error/near/277355733)

### DIFF
--- a/near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
+++ b/near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
@@ -58,18 +58,19 @@ impl ImplItemMethodInfo {
             is_handles_result,
             ..
         } = attr_signature_info;
-        let deposit_check = if *is_payable || matches!(method_type, &MethodType::View) {
-            // No check if the method is payable or a view method
-            quote! {}
-        } else {
-            // If method is not payable, do a check to make sure that it doesn't consume deposit
-            let error = format!("Method {} doesn't accept deposit", ident);
-            quote! {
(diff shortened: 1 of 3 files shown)


nearmax:
Is this a backwards-compatible change from the perspective of API or behavior?

austinabell:
> Is this a backwards-compatible change from the perspective of API or behavior?

Are you referring to the change in this draft or the proposed protocol change?

The draft PR change is backwards compatible from the API surface, but just would remove the payable check by default. Not going to open this PR up, was created more for discussion.

austinabell:
Also, I have a terrible memory, because there was already an issue for this: https://github.com/near/near-sdk-rs/issues/718

austinabell:
Closing this because the logic in nearcore was changed in https://github.com/near/nearcore/pull/8433

The better solution now is to just have the non-payable check on all methods, and only remove if the attribute `payable` is applied.

PR #783, opened 2022-04-07
fix: Remove clone from Promise API

austinabell:
Having clone for this is very error-prone, especially since some constructions are cached and cloning opens up to a lot of bugs.

Not quite sure why it was added in the first place


CHANGELOG.md
```diff
@@ -13,6 +13,9 @@
 - Fixed gas assertion in `*_transfer_call` implementations of FT and NFT standards to only require what's needed. [PR 760](https://github.com/near/near-sdk-rs/pull/760)
 - Fixed events being emitted in FT standard to include refund transfers and burn events. [PR 752](https://github.com/near/near-sdk-rs/pull/752)
 
+### Fixed
+- Remove `Clone` implementation for `Promise` (error prone) https://github.com/near/near-sdk-rs/pull/783
```
ChaoticTempest, line 17: Probably better in the `Removed` section?

ChaoticTempest (approved):

CHANGELOG.md
austinabell: Yeah, I figured since it's an actual bug I would include here, guess it could go either way, though

Issue #780, opened 2022-04-05
AccountId defined twice


wackazong:
I noticed that `AccountId` is defined in `near_sdk::types` but also in `near_primitives_core::types`.

When using `near_sdk` version 4.0.0-pre.7 the `VMContext` uses the `AccountId` definition from `near_primitives_core::types`. On the other hand, e.g. `env::predecessor_account_id()` uses the definition from `near_sdk::types`.

Is that expected? To me it does not make much sense. I looked into the release histories of the packages and into the near-sdk-rs issue tracker but did not find any clues.

miraclx:
Afaik, when the `near-sdk` `AccountId` was introduced, the `near-primitives` version was still a bit bulky to be used in smart contracts. It's been stripped down and refactored a lot since then, though. Maybe @austinabell, you can take a look.

wackazong:
Ok, thanks for the explanation.

Thing is, it's being used in the definition of `VMContext`. If I do unit tests I do need to handle both versions. I use the `near-primitives` version globally now and convert from the old one using (for whom it may concern):

```
fn convert_old_account_id(old: OldAccountId) -> AccountId {
    AccountId::from_str(&old.to_string())
        .unwrap_or_else(|_| env::panic_str("INVALID_OLD_ACCOUNT_ID"))
}
```

austinabell:
Ah, yes, so it's generally recommended to use `VMContextBuilder` to generate context. This type probably shouldn't be exposed from the SDK, I'll make a change for this soon

austinabell:
Closed with #785

PR #785, opened 2022-04-08
refactor!: move VMContext type to SDK

austinabell:
Using the nearcore VMContext directly causes issues (see #780) so to improve and stabilize the API from changing in the future (as it has) this type is moved to SDK.

Some fields change from what they were in `3.x`, but toward what is used in the SDK now so the migration should be simple

The same could be done with `VMConfig` but that type has all primitive types and is rarely configured specifically so I've kept that change separate from this. Possibly worth opening an issue to migrate over, but would love to hear others' opinions if this should be done.


near-sdk/src/test_utils/context.rs
```diff
@@ -27,15 +25,70 @@ impl Default for VMContextBuilder {
...
     pub fn new() -> Self {
         Self {
             context: VMContext {
-                current_account_id: alice().try_into().unwrap(),
-                signer_account_id: bob().try_into().unwrap(),
-                signer_account_pk: vec![0u8; 32],
-                predecessor_account_id: bob().try_into().unwrap(),
+                current_account_id: alice(),
+                signer_account_id: bob(),
+                signer_account_pk: vec![0u8; 33].try_into().unwrap(),
```
ChaoticTempest, line 90: hmm, is the new `signer_account_pk` suppose to be 33 in length? Just noting it looks weird that it changed from 32, or did something change under the hood?

austinabell: nope, it was bugged/incorrect before. Guess I should have mentioned this specifically. The extra byte is the curve type.

ChaoticTempest (approved):

PR #789, opened 2022-04-13
chore: update workspaces to 0.2

austinabell:
WIP since test is broken

### DIFF
--- a/examples/callback-results/src/lib.rs
+++ b/examples/callback-results/src/lib.rs
@@ -85,9 +85,9 @@ mod tests {
     async fn workspaces_test() -> anyhow::Result<()> {
         let wasm = fs::read("res/callback_results.wasm").await?;
 
-        let worker = workspaces::sandbox();
+        let worker = workspaces::sandbox().await?;
 
-        let contract = worker.dev_deploy(wasm).await?;
+        let contract = worker.dev_deploy(&wasm).await?;
 
         // Call function a only to ensure it has correct behaviour
         let res = contract.call(&worker, "a").transact().await?;
--- a/examples/fungible-token/tests/workspaces.rs
+++ b/examples/fungible-token/tests/workspaces.rs
@@ -1,4 +1,4 @@
-use near_primitives::views::FinalExecutionStatus;
+
 use near_sdk::json_types::U128;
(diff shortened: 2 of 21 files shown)


itegulov:
I left some TODOs in nft and ft examples that should now be resolvable with workspaces 0.2. You could tackle them in this PR or I can do it in a follow up PR, up to you

austinabell:
> I left some TODOs in nft and ft examples that should now be resolvable with workspaces 0.2. You could tackle them in this PR or I can do it in a follow up PR, up to you

I'll separate that change since this is blocking #742

examples/fungible-token/tests/workspaces.rs
```diff
@@ -27,25 +27,25 @@ async fn init(
     initial_balance: U128,
 ) -> anyhow::Result<(Contract, Account, Contract)> {
     let ft_contract =
-        worker.dev_deploy(include_bytes!("../res/fungible_token.wasm").to_vec()).await?;
+        worker.dev_deploy(&include_bytes!("../res/fungible_token.wasm").to_vec()).await?;
```
ChaoticTempest, line 30: looks like you forgot to clean some of these up:
```suggestion
        worker.dev_deploy(include_bytes!("../res/fungible_token.wasm")).await?;
```

ChaoticTempest (approved):
minor nit but overall looking good

examples/callback-results/src/lib.rs
```diff
@@ -85,9 +85,9 @@ mod tests {
     async fn workspaces_test() -> anyhow::Result<()> {
         let wasm = fs::read("res/callback_results.wasm").await?;
 
-        let worker = workspaces::sandbox();
+        let worker = workspaces::sandbox().await?;
 
-        let contract = worker.dev_deploy(wasm).await?;
+        let contract = worker.dev_deploy(&wasm).await?;
```
itegulov, line 90: any reason not to use `include_bytes!` here?

itegulov (approved):

examples/callback-results/src/lib.rs
austinabell: tried to keep this change unopinionated, can follow with that change in another

PR #794, opened 2022-04-19
Support enum for near_bindgen

### LINKED ISSUE #682: Support enum for `near_bindgen`
There should be no reason why `enum` is not supported for contract state, it can be serialized/deserialized just as a struct and nothing changes wrt codegen

Definitely a less common practice, but there can definitely be usecases for representing state as an enum, and this would avoid someone having to nest this enum in a struct.

NinoLipartiia4ire:
Closes #682.


austinabell (commented):
Can you add a compilation test for this? I've written one to make sure this compiles correctly:

```rust
//! Testing that state with enum compiles correctly

use near_sdk::near_bindgen;

#[near_bindgen]
#[derive(BorshDeserialize, BorshSerialize)]
enum StateMachine {
    StateA,
    StateB,
}

impl Default for StateMachine {
    fn default() -> Self {
        Self::StateA
    }
}

#[near_bindgen]
impl StateMachine {
    pub fn swap_state(&mut self) {
        *self = match self {
            Self::StateA => Self::StateB,
            Self::StateB => Self::StateA,
        };
    }
}

fn main() {}
```

This also doesn't verify that it functions correctly, but we can do that later.

austinabell (approved):
Example seems a bit bulky for what it's showing off and would be nice to have a workspaces test to make sure this pattern works e2e if we are creating an example of this.

I think it's fine to come in and we can make the tweaks after. @itegulov do you have any thoughts on this?

examples/enum-contract/README.md
```diff
@@ -0,0 +1,9 @@
+# Status Message
+
+Records the status messages of the accounts that call this contract.
+
+## Testing
+To test run:
+```bash
+cargo test --package status-message -- --nocapture
```
ChaoticTempest, line 8: This looks like its point to wrong package
```suggestion
cargo test --package enum-contract -- --nocapture
```

near-sdk-macros/src/core_impl/code_generator/item_enum_info.rs
```diff
@@ -0,0 +1,14 @@
+use syn::ItemEnum;
+
+pub fn generate_proxy_enum(input: &ItemEnum) -> proc_macro2::TokenStream {
+    use quote::{format_ident, quote};
+    let ident = &input.ident;
+    let new_name = format_ident!("{}Contract", ident);
+    let name = quote! {#new_name};
```
itegulov, line 7: You don't need to quote the ident, it is interpolatable by itself

```suggestion
    let name = format_ident!("{}Contract", ident);
```

near-sdk-macros/src/core_impl/code_generator/item_enum_info.rs
```diff
@@ -0,0 +1,14 @@
...
+    quote! {
+         #[cfg(not(target_arch = "wasm32"))]
+         pub struct #name {
+            pub account_id: near_sdk::AccountId,
+          }
```
itegulov, line 12:
```suggestion
         pub struct #name {
             pub account_id: near_sdk::AccountId,
         }
```

examples/enum-contract/README.md
```diff
@@ -0,0 +1,9 @@
+# Status Message
```
itegulov, line 1: Maybe add something like this to distinguish it from the other one?

```suggestion
# Status Message (Enum Variant)
```

itegulov (approved):
Aside from a bunch of nitpicks looks good to me! Most of them were inherited from the struct implementation, but anyway I think we shouldn't keep copying them over.

Yeah, the example is on the heavier side for sure, but at least it is a fun one

PR #799, opened 2022-04-22
refactor: remove duplicate proc macro code and refactor enum example

austinabell:
Following #794 to remove duplicate code for generating the sim proxy struct and re-writing the added example to be more useful. I've realized creating examples of versioned contracts in docs is lacking (https://www.near-sdk.io/upgrading/production-basics#writing-upgradable-contracts) so this is one alternative of something we can point to or suggest.

Side benefit is that it's using and testing new `UnorderedMap` too

Also removes a duplicate wasm binary that was committed in that PR


near-sdk-macros/src/core_impl/code_generator/sim_proxy.rs
```diff
@@ -0,0 +1,12 @@
+use proc_macro2::Ident;
+
+pub fn generate_sim_proxy_struct(ident: &Ident) -> proc_macro2::TokenStream {
```
willemneal, line 3: Shouldn't this be deprecated or even better removed in the 4.0 release?

austinabell: It will be once sim tests are fully deprecated. Do not want to interrupt current usage patterns

ChaoticTempest (approved):

examples/versioned/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+}
+
+impl VersionedContract {
+    fn contract_mut(&mut self) -> &mut Contract {
+        if let Self::V1(contract) = self {
+            // Contract is up to date, return mutable reference to it.
+            return contract;
+        }
+        let old_contract = match self {
+            Self::V1(contract) => return contract,
```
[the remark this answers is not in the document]
austinabell: oh oops, an artifact of a refactor :D nice catch

PR #801, opened 2022-04-23
fix: Optimize nth implementation for legacy UnorderedMap

austinabell:
Made the change minimally invasive since this is somewhat of a legacy collection. The new ones are optimized and I assumed that `zip` with the optimization done to #634 would be applied to the `UnorderedMap`, but it wasn't.

exposes `unordered_map` module to allow `Iter` type to be used. Didn't add iterator types to others since no demand. There would be a benefit (in that the aux iterator traits would be available) but I don't want to return `vector::Iter` for backwards compatibility if we wanted to switch later and didn't want to just add a bunch of redundant code for no strong reason.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixes
+- Optimized `nth` operation for `UnorderedMap` iterator and implemented `IntoIterator` for it. [PR 801](https://github.com/near/near-sdk-rs/pull/801)
+  - This optimizes the `skip` operation, which is common with pagination
+
 ## [4.0.0-pre.8] - 2022-04-19
 
 ### Added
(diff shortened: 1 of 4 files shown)


ChaoticTempest (approved):
interesting, so without this optimization, skip would've loaded in the elements twice over due to calling into `nth` twice and `nth` always loaded all the elements up to `n`? That's a pretty big win if someone were still using the old collections 👏

austinabell:
> interesting, so without this optimization, skip would've loaded in the elements twice over due to calling into `nth` twice and `nth` always loaded all the elements up to `n`? That's a pretty big win if someone were still using the old collections 👏 


Yeah, zip for some reason [avoids calling nth on the inner iterator](https://doc.rust-lang.org/stable/src/core/iter/adapters/zip.rs.html#25) and just instead calls next n times

PR #803, opened 2022-04-27
refactor!: Impl Promise generics at the type level

austinabell:
Definitely not perfect since we have to follow the drop pattern with the host functions available

- This now schedules promises when `then` or `and` is called to avoid recursive generics or unsafe code to swap the generic type
- Janky pattern because Promises should only really be typed when a function call is acted on it, but not much flexibility with this single type
- Currently, if the function call returns a promise, this is expressed in the parameters (this could be useful to know it is chained, but maybe not from an interface level?) cc @itegulov your input on what you think is useful here is important
- Takes Ok type of `Result<_, PromiseError>` from new error handling logic

Generally, it's a pretty breaking change, so seems like if we are changing this API it should come in 5.0 pre-releases to be iterated on. For cases of using ACI to have typed promises for codegen, might require using this new version or creating a fallback to handle promises as some `any` type in the schema generation (this might be necessary for other language impls anyway).

@matklad if and only if you have free time, input on how promises becoming generic could be migrated to, or if we were to start from scratch if you had any considerations

### DIFF
--- a/examples/callback-results/src/lib.rs
+++ b/examples/callback-results/src/lib.rs
@@ -1,5 +1,5 @@
-use near_sdk::require;
-use near_sdk::{env, near_bindgen, Promise, PromiseError};
+use near_sdk::{env, near_bindgen, PromiseError};
+use near_sdk::{require, ScheduledFn};
 
 const A_VALUE: u8 = 8;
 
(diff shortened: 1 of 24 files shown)


austinabell:
I think this API is better, but it's opinionated and is quite a breaking change. For that reason, I'm just going to close since I won't be working on it anymore, but this could be useful if you want the promise API typed

PR #804, opened 2022-04-28
refactor: optimize read_register to read to non-zeroed buffer

austinabell:
Removed need to zero out the buffer before writing to it, seems it would be an impactful optimization.

The only opinionated point is if conversion from u64 to usize (u32 in wasm32) should be checked or not. The maximum buffer size in configs in the runtime is less than `u32::MAX` but there doesn't seem to be any blockers from this changing with a custom config -- potentially this should be unchecked for a trivial optimization in code size? The only issue is that there is UB if the register length is > `u32::MAX` because the buffer created wouldn't be large enough and memory would be written outside of the buffer. cc @matklad will there be any guarantee that the max will always be < `u32::MAX`?


matklad:
>cc @matklad will there be any guarantee that the max will always be < u32::MAX?

I think yes, this is guaranteed, by both the config and by fees. But I think the current impl with `.unwrap_or_abort()` is the best option -- better safe than unsafe, and I doubt that it'll increase the code size a lot.

near-sdk/src/environment/env.rs
```diff
@@ -114,10 +115,24 @@ pub fn setup_panic_hook() {
...
 /// Reads the content of the `register_id`. If register is not used returns `None`.
 pub fn read_register(register_id: u64) -> Option<Vec<u8>> {
-    let len = register_len(register_id)?;
-    let res = vec![0u8; len as usize];
-    unsafe { sys::read_register(register_id, res.as_ptr() as _) };
-    Some(res)
+    // Get register length and convert to a usize. The max register size in config is much less
+    // than the u32 max so the abort should never be hit, but is there for safety because there
+    // would be undefined behaviour if the case would be hit.
+    let len: usize = register_len(register_id)?.try_into().unwrap_or_else(|_| abort());
```
ChaoticTempest, line 121: Is it really undefined behavior though? From the docs, casting just truncates:
> Casting from a larger integer to a smaller integer (e.g. u32 -> u8) will truncate

but it's not ideal either for us to have a truncated `len` either, so the change still seems valid to me

ChaoticTempest (approved):

itegulov (approved):

near-sdk/src/environment/env.rs
matklad: The UB would happen not during truncation, but when the runtime writes past the truncated lenght.

Issue #800, opened 2022-04-23
nft_tokens return exceed gas limit error if starting index is high enough


ymc182:
Hi I have a collection with over 2k nft and when I try to get tokens start from around 600 , it return exceed gas limit gas, everything under that certain number is fine

austinabell:
So the issue is that this collection is a `TreeMap` which can't be optimized more because `skip(n)` requires iterating over all keys before `n` in order since the collection is ordered.

Ideally, the standards are re-written to be more efficient and low-level so this won't be a problem, but unclear what the best path for developers to optimize this. There doesn't seem to be a concise solution that _isn't_ state breaking

ymc182:
so all UnorderedMap will not be able to use skip for a relative large collection size?

austinabell:
Update on this for more context:

We don't currently have the bandwidth for a re-write to solve this and other issues in the near future. We will likely start a fork of the standards to fix and optimize, but also this can be done by anyone in the community, there requires no change in the NEP standard for NFTs.

The reasons this can't just be changed like in #810:
- `TreeMap` and `UnorderedMap` have a different storage format and will silently break existing contracts
- Some frontends might expect the ordered nature of `TreeMap`, so this could cause bugs in dapps
- `TreeMap` can't just be optimized for `nth` without a state breaking change, there would need to be additional data stored to indicate how many nodes to left/right

Medium-term plan is to re-write, optimize, and make this library more configurable, but if anyone needs changes in short term, best to fork or create new.

PR #810, opened 2022-05-09
`owner_by_id` updated from TreeMap to UnorderedMap

zavodil:
`owner_by_id` updated from TreeMap to Unordered map due to https://github.com/near/near-sdk-rs/issues/800. `simulate_mass_mint` test introduced.

Please review if my version upgrade was correct.

### DIFF
--- a/near-contract-standards/Cargo.toml
+++ b/near-contract-standards/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "near-contract-standards"
-version = "4.0.0-pre.8"
+version = "4.0.1"
 authors = ["Near Inc <hello@near.org>"]
 edition = "2018"
 license = "GPL-3.0"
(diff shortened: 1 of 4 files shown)


near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -107,7 +107,7 @@ impl NonFungibleToken {
         let mut this = Self {
             owner_id,
             extra_storage_in_bytes_per_token: 0,
-            owner_by_id: TreeMap::new(owner_by_id_prefix),
+            owner_by_id: UnorderedMap::new(owner_by_id_prefix),
```
austinabell, line 110: We can't just change this because it is state-breaking and will break for anyone trying to update their NFT version.

If you need this change, you probably just want to fork this logic

near-contract-standards/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "near-contract-standards"
-version = "4.0.0-pre.8"
+version = "4.0.1"
```
austinabell, line 3: Can't change to this

austinabell (changes requested):
As noted in the review, can't make this change because it's state breaking

PR #817, opened 2022-05-17
chore: remove near-sdk-sim from repo

### LINKED ISSUE #662: Deprecate or update near-sdk-sim to work with newly released nearcore crates

### LINKED ISSUE #759: Replace unmaintained crate `memmap` with `memmap2`
Dependency chain: `near-sdk-sim` -> `near-runtime` -> `near-vm-runner` -> `wasmer-runtime-near` -> `memmap`

We are using `near-sdk-sim` in our tests and I've noticed that an unmaintained crate is pulled indirectly via `near-sdk-sim`. The create is called `memmap` and it should be replace by its newer version `memmap2`.

Note that this issue was flagged by Halborn while doing an audit on our smart contracts.

Can we please get this fixed?

### LINKED ISSUE #802: Remove sim codegen from `near_bindgen` after 4.0 stable
This cannot happen now since we need a smooth transition from sim tests, but can be purged in the next major release/pre-release

This involves just removing the `{Struct Name}Contract` proxy structures used by sim. If we continue with this pattern, which is a bit redundant if the functionality is linked with ACI and requires the use of `rlib` on the crate, it will likely be completely re-written anyway

austinabell:
Since there is no way for us to upgrade near-sdk-sim to be compatible with new protocol version, it is necessary we not just soft deprecate and if people need these sim tests, they can stay on `4.0.0-pre.9` until they can switch

Closes #662 
closes #802 
closes #759


ChaoticTempest (approved):
Feels like people can benefit from us publishing one final patch release to near-sdk-sim to say it's deprecated and workspaces should be used in favor. WDYT?

austinabell:
> Feels like people can benefit from us publishing one final patch release to near-sdk-sim to say it's deprecated and workspaces should be used in favor. WDYT?

That's why I released `4.0.0-pre.9`, it's unlikely people would update into any temporary version we release in the ~week between now and stable release. 

Obviously not ideal, but it wasn't foreseen that we would have to hard deprecate sim tests for 4.0 given we can't update the VM logic. We can add a commit on top of the released commit as a patch version, but I'm not sure that's better since it will be the last functional version of the SDK for near-sdk-sim and dealing with deprecation warnings might not be ideal either.

itegulov:
Can we just delete all the code but leave a top-level `#[deprecated]` annotation that leads the user to workspaces-rs and migration doc and release it as a 4.0 version for the very last time?

austinabell:
> Can we just delete all the code but leave a top-level `#[deprecated]` annotation that leads the user to workspaces-rs and migration doc and release it as a 4.0 version for the very last time?

The reason this is not done is because the code will be broken using 4.0 of the SDK and will be cryptic why it doesn't work. We can always release a 4.0 with this deprecation based on this old commit, but since it's not functional anymore we should remove it from the repo.

Let's find some time to chat about the best path for this as it's a bit nuanced.

Options being:
- Add deprecation to sim tests on 4.0 release (even though broken on sim tests)
- Add compiler error on the 4.0 release with links to workspaces (and migration doc if stable link)
- Do not release `near-sdk-sim` and just communicate this deprecation to community

Any of these can be done if this PR comes in, though

PR #820, opened 2022-05-19
chore: update to 0.13.0 nearcore crates

austinabell:
WIP since the crates haven't been released, opening PR for visibility and for input as a few things had to change:
- Receipt structure removed `receipt_indices`, and added `non_exhaustive`
  - Not sure why the indices would be needed, this struct was added in 4.0 pre-releases so it's fine to remove
- Remove `outcome` function on the `MockedBlockchain` 
  - This was misleading as it only captured some gas costs, and with the new protocol change and updates, this API was not possible because getting the outcome required taking ownership of VMLogic struct
- Slightly updated interface for `created_receipts` and `logs` (nothing too breaking)
- Some additional conversions to be able to format Receipts in a usable way (like serializing Pubkey to string and back)

Most importantly, this change comes with the functionality of having the correct `promise_batch_action_function_call_weight` implementation by using the new protocol feature


near-sdk/src/environment/mock/mocked_blockchain.rs
```diff
@@ -126,6 +136,51 @@ fn sdk_context_to_vm_context(context: VMContext) -> near_vm_logic::VMContext {
...
+        PrimitivesAction::DeleteAccount(a) => {
+            VmAction::DeleteAccount { beneficiary_id: a.beneficiary_id.parse().unwrap() }
+        }
+    }
+}
+
+fn pub_key_conversion(key: &VmPublicKey) -> PublicKey {
+    // Hack by serializing and deserializing the key. This format should be consistent.
+    String::from(key).parse().unwrap()
+}
```
ChaoticTempest, line 182: feels like this conversion could also fall into the trap of semantics changing from under us if `near_crypto::PublicKey` string conversion ever changes, so a test might be worthy. This isn't too important though since this is only gonna be in unit tests, so leaving this one up to you if you wanna add

near-sdk/src/test_utils/mod.rs
```diff
@@ -89,24 +88,22 @@ macro_rules! testing_env {
...
 }
 
 /// Objects stored on the trie directly should have identifiers. If identifier is not provided
 /// explicitly than `Default` trait would use this index to generate an id.
-#[allow(dead_code)]
+#[cfg(test)]
 pub(crate) static mut NEXT_TRIE_OBJECT_INDEX: u64 = 0;
 /// Get next id of the object stored on trie.
-#[allow(dead_code)]
+#[cfg(test)]
```
ChaoticTempest, line 106: hmm, since this is the `test_utils` module don't we want to just expand further and say the whole module is under `#[cfg(test)]` instead?

ChaoticTempest (approved):
LGTM

near-sdk/src/environment/mock/mocked_blockchain.rs
```diff
@@ -126,6 +136,51 @@ fn sdk_context_to_vm_context(context: VMContext) -> near_vm_logic::VMContext {
...
+                    nonce: k.access_key.nonce,
+                }
+            }
+        },
+        PrimitivesAction::DeleteKey(k) => {
+            VmAction::DeleteKey { public_key: pub_key_conversion(&k.public_key) }
+        }
+        PrimitivesAction::DeleteAccount(a) => {
+            VmAction::DeleteAccount { beneficiary_id: a.beneficiary_id.parse().unwrap() }
+        }
```
itegulov, line 175: Very small nit: any reason why you didn't inline these like you have done with others above? Sorry for nitpicking, this is just triggering my OCD :)

itegulov (approved):
Seems very reasonable to me

near-sdk/src/environment/mock/mocked_blockchain.rs
austinabell: it's just default formatting, the line is too long and it pushes it like this

near-sdk/src/test_utils/mod.rs
austinabell: no, because these are annotated with this because they are just used within the crate, where the others are used because they are used externally (you can't use things marked with `cfg(test)` externally

PR #824, opened 2022-05-25
chore: 4.0 stable release

### LINKED ISSUE #494: Rust SDK 4.0 stabilization tracker

austinabell:
Tested new promise API against mainnet.

closes #494


ChaoticTempest (approved):
🚀 🚀 🚀 ship it

austinabell:
ignore all the force pushes, just combining commits so that this can be merged without squashing so the release hash is equivalent to the one published. Just updated the date to today and merged from master

PR #828, opened 2022-05-25
feat: enable const-generics feature by default on borsh

### LINKED ISSUE #827: Const-generics feature on a borsch dependency
Why not to enable const-generics feature on a borsch dependency ? I think it'll be very useful.

austinabell:
As noted by #827, this feature could be enabled by default. Our MSRV is already above const generics, and this default might be useful to some.

closes #827


ChaoticTempest (approved):
lgtm

PR #830, opened 2022-05-31
NEP141: safe maths fixes

sept-en:
Fixes [sdk-docs#75](https://github.com/near/sdk-docs/issues/75)


near-contract-standards/src/fungible_token/core_impl.rs
```diff
@@ -185,21 +189,35 @@ impl FungibleToken {
...
+                    let used_amount = amount
+                        .checked_sub(refund_amount)
+                        .unwrap_or_else(|| env::panic_str("Total supply overflow"));
+                    return (used_amount, 0);
                 } else {
                     // Sender's account was deleted, so we need to burn tokens.
-                    self.total_supply -= refund_amount;
+                    self.total_supply
+                        .checked_sub(refund_amount)
+                        .unwrap_or_else(|| env::panic_str("Total supply overflow"));
```
austinabell, line 220: this isn't subtracting from `self.total_supply`, so this is a bug

austinabell (changes requested):
Do you mind having all errors be some constant (or just based on a function) which indicates the overflow. I would assume the compiler can optimize away the duplicates, but also makes the code cleaner.

near-contract-standards/src/fungible_token/core_impl.rs
sept-en: Good catch, fixed in 261efd01.
Please note this actually also means that this part is not enough covered with the unit tests.

austinabell (approved):

PR #831, opened 2022-06-03
feat: `near_abi!` macro

itegulov:
Basically just the macro and ABI models without all other extra stuff from `daniyar/aci` like cargo extension, cross contract calls etc.

We have not decided in what branch this should live in, so I am targeting `master` for now.


examples/abi/README.md
```diff
@@ -0,0 +1,9 @@
+# ABI
+
+Showcases how to generate ABI from a NEAR contract.
+
+## Generating ABI
+To test run:
+```bash
+cargo near metadata
```
austinabell, line 8: Should this not be something like `cargo near abi` or something? Feel like the metadata naming might be confusing

examples/abi/res/abi.json
```diff
@@ -0,0 +1,119 @@
...
+      {
+        "id": 2,
+        "schema": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/DoublePair"
+          }
+        }
+      }
+    ],
```
austinabell, line 78: What's the point of this section if it just references the schema? Can't we just ignore the references to this?

near-sdk-macros/src/core_impl/abi/abi_visitor.rs
```diff
@@ -0,0 +1,75 @@
...
+                quote! {
+                    near_sdk::__private::AbiType { id: #id, schema: gen.subschema_for::<#t>() }
+                }
+            })
+            .collect();
+        Ok(quote! {
+            const _: () = {
+                #[no_mangle]
+                #[cfg(not(target_arch = "wasm32"))]
+                pub fn __near_abi() -> near_sdk::__private::AbiRoot {
```
austinabell, line 60: Note that this will fail to compile when there are multiple impl sections with `near_bindgen`, right? Will double check tomorrow

near-sdk-macros/src/core_impl/abi/mod.rs
```diff
@@ -0,0 +1,28 @@
+use std::collections::{hash_map::Entry, HashMap};
+use syn::Type;
+
+pub mod abi_generator;
+pub mod abi_visitor;
+
+pub struct TypeRegistry {
+    pub types: HashMap<Box<Type>, u32>,
```
austinabell, line 8: Seems like this should probably be `BTreeMap` since this should always be small. I don't believe it should matter much since this is done at compile time though

near-sdk/src/private/abi.rs
```diff
@@ -0,0 +1,101 @@
...
+    }
+}
+
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Default)]
+pub struct AbiMetainfo {
+    /// The name of the smart contract.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub name: Option<String>,
+    /// The version of the smart contract.
+    #[serde(skip_serializing_if = "Option::is_none")]
```
austinabell, line 38: Is there a reason these skip serializing if none but don't have `serde(default)` to allow deserializing without them?

near-sdk/src/private/abi.rs
```diff
@@ -0,0 +1,101 @@
...
+    /// Type identifiers of the callbacks of the function.
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    #[serde(default)]
+    pub callbacks: Vec<AbiParameter>,
+    /// Type identifier of the vararg callbacks of the function.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub callbacks_vec: Option<AbiParameter>,
+    /// Return type identifier.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub result: Option<AbiParameter>,
```
austinabell, line 81: same here, why no default for deserialize?

near-sdk/src/promise.rs
```diff
@@ -529,3 +542,13 @@ impl<T: borsh::BorshSerialize> borsh::BorshSerialize for PromiseOrValue<T> {
...
     }
 }
+
+impl<T: JsonSchema> JsonSchema for PromiseOrValue<T> {
+    fn schema_name() -> String {
+        format!("PromiseOrValue{}", T::schema_name())
+    }
+
+    fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+        T::json_schema(gen)
```
austinabell, line 552: Note that right now nothing restricts a developer from having the promise return a different schema type than `T`.

examples/abi/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+#[derive(JsonSchema, Serialize, Deserialize)]
+pub struct Pair(u32, u32);
+
+#[derive(JsonSchema, Serialize, Deserialize)]
+pub struct DoublePair {
+    first: Pair,
+    second: Pair,
+}
+
+near_abi! {
```
austinabell, line 15: Did you end up experimenting with having this as an annotation for a module or including logic within `near_bindgen` somehow?

near-sdk-macros/src/core_impl/abi/abi_visitor.rs
```diff
@@ -0,0 +1,75 @@
...
+            .iter()
+            .flat_map(|i| &i.methods)
+            .map(|m| m.abi_struct(&mut registry))
+            .collect();
+        let types: Vec<TokenStream2> = registry
+            .types
+            .iter()
+            .map(|(t, id)| {
+                quote! {
+                    near_sdk::__private::AbiType { id: #id, schema: gen.subschema_for::<#t>() }
```
austinabell, line 52: This will generate a JSON schema for borsh types, will it not?

austinabell:
> We have not decided in what branch this should live in, so I am targeting `master` for now.

Hmm, yeah either master becomes the 5.0 path and we backport non-breaking changes to some 4.x branch, or we put these changes into a branch that targets the main branch and only merge those commits in when we are ready to release the next version

near-sdk-macros/src/core_impl/abi/abi_visitor.rs
itegulov: Hmm, it should not as long as all impl sections are wrapped in the same `near_abi!` call.

examples/abi/res/abi.json
itegulov: > What's the point of this section if it just references the schema?

So what we are putting under `abi/root_schema` is called a root schema and it defines all custom types we are using in the contract. `schema` in this context refers to a subschema (which might be a better name for this json property), which **can** refer to something from the root schema. There are also cases where it can be just a primitive type (e.g. `{ "type": "integer" }`), a boolean value (e.g. `true` is a valid subschema meaning that all instances pass validation) among some other cases.

> Can't we just ignore the references to this?

I am not sure what you mean by this question, but if it helps I mentioned in Slack that type registry is actually no longer strictly necessary. We can just use subschemas directly inside `functions` since they are not that big. I am open to having a discussion on which way would be better.

near-sdk-macros/src/core_impl/abi/mod.rs
itegulov: Yeah I have tried this and the problem is that `Type` does not implement `Ord`

near-sdk/src/private/abi.rs
itegulov: Simply forgot to update these fields :)

near-sdk/src/promise.rs
itegulov: Yeah I also thought about this. Returning different things feels like a strange pattern to me, but I guess since we are not discouraging this pattern I should not presume that there is only one possible return type. I could represent this as an equivalent of `Either<Any, T>` or just simply `Any`. WDYT?

examples/abi/src/lib.rs
itegulov: Not yet, but it is in my plans

near-sdk-macros/src/core_impl/abi/abi_visitor.rs
itegulov: Nope, `gen` here is an instance of `schemars::gen::SchemaGenerator`. You are probably thinking of `T::schema_container()` from borsh.

itegulov:
> > We have not decided in what branch this should live in, so I am targeting `master` for now.
> 
> Hmm, yeah either master becomes the 5.0 path and we backport non-breaking changes to some 4.x branch, or we put these changes into a branch that targets the main branch and only merge those commits in when we are ready to release the next version. Seems like maybe the first option might be cleaner, and then we can just cherry-pick commits when we want to release any 4.x branch. Might be more work, but seems okay

Yeah sounds good to me.

Do we want to put this macro and ABI models under unstable for now?

near-sdk/src/private/abi.rs
```diff
@@ -0,0 +1,126 @@
...
+/// `TypeId`) that are then mapped to the corresponding JSON subschema.
+pub type TypeId = u32;
+
+/// Contract ABI.
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
+pub struct AbiRoot {
+    /// Semver of the ABI schema format.
+    pub abi_schema_version: String,
+    /// Meta information about the contract.
+    pub metainfo: AbiMetainfo,
```
ChaoticTempest, line 19: do we wanna keep `metainfo` instead of `metadata`?

austinabell:
> Do we want to put this macro and ABI models under unstable for now?

Yes, please.

near-sdk/src/private/abi.rs
itegulov: Do you think `metadata` is a better name here? I left `metainfo` because the previous iteration of ABI was called metadata, so no strong opinions here

near-sdk/src/promise.rs
itegulov: Well, it is not really fallible as it is non-restrictive to just `T`, so I don't think `Result` suits unless I am missing something here

near-sdk/src/lib.rs
```diff
@@ -9,6 +9,9 @@ pub use near_sdk_macros::{
     ext_contract, metadata, near_bindgen, BorshStorageKey, FunctionError, PanicOnDefault,
 };
 
+#[cfg(feature = "unstable")]
+pub use near_sdk_macros::near_abi;
```
itegulov, line 13: @austinabell am I correct in saying that this won't stop `near_abi` from being compiled without unstable feature? Is there a way I can stop that without introducing unstable feature in near-sdk-macros crate?

examples/abi/src/lib.rs
itegulov: So I have played with the idea and surprisingly everything seems to be working as is. I will run it through a few more scenarios later, in the meantime please take another look when you can.

near-sdk/src/private/abi.rs
ChaoticTempest: I feel like `metadata` would be more intuitive due to it being used more frequently to denote this kind of info, but if there's metadata being used somewhere else in SDK, then that might be of conflict. But if it's not being used anymore, then it's probably best to just take it

near-sdk/src/lib.rs
ChaoticTempest: This looks right to me. Rust won't make use of `near_abi` unless it's referenced and used somewhere. An export will just allow users to reference it, meanwhile, the feature flag will only enable the export when the unstable flag is set.

If you want to add the unstable flag to near-sdk-macros as well, you can nest feature flags, such that when near-sdk unstable is set, near-sdk-macro's unstable flag is also set like so:
```toml
# near-sdk
[features]
unstable = ["near-sdk-macros/unstable"]

# near-sdk-macros
[features]
unstable = []
```

austinabell: yeah, what Phuong mentions here with having an unstable feature flag for macros is probably required. Having the dead codegen is probably not something we want for those not using this

near-sdk/src/promise.rs
austinabell: I think that making the assumption as one type is fine, it would nudge developers in the correct pattern, but I'm just making the comment so we don't ignore the cases where maybe someone might return different types that serialize as the same thing. Can't immediately think of an implication that would be an issue, though

near-sdk-macros/src/core_impl/abi/abi_visitor.rs
austinabell: didn't realize that case is handled. It limits the ability to have impl sections in diff modules, though

examples/abi/res/abi.json
austinabell: So it seems like the tradeoff is just more duplication if there are a lot of overlapping types, but more readable and less JSON when no overlap if inlining? Am I missing other considerations?

near-sdk-macros/src/core_impl/abi/abi_visitor.rs
itegulov: True. I am unsure how to handle this though... I have tried inlining modules like this:

```rust
#[near_abi]
mod abi {
    #[path = "utility.rs"]
    mod abi_utility;

    #[near_bindgen]
    #[derive(Default, BorshDeserialize, BorshSerialize)]
    pub struct Contract {}

    ...
}
```

But apparently including non-inline modules in proc macro input is an unstable feature.

I was thinking I could maybe create a global state that would be modified by all `#[near_abi]` annotations and then add another macro that would output the accumulated state into one ABI structure. The problem is that I think there is no way to enforce macro execution order. Do you have any ideas I could explore?

examples/abi/res/abi.json
itegulov: Yeah, that's pretty much it

near-sdk/src/private/abi.rs
itegulov: Renamed

near-sdk/src/lib.rs
itegulov: Gotcha, should be better now

near-sdk-macros/src/core_impl/abi/abi_visitor.rs
itegulov: Another very vague idea I have considered is to generate a bunch of symbols (e.g. `__near_abi_lib_rs`, `__near_abi_utility_rs` etc) and then somehow get a list of them from cargo extension and call all of them. The main problem here is that I cannot find if there is a rustc/cargo API that would allow me to list all symbols from a crate.

near-sdk-macros/src/core_impl/abi/abi_generator.rs
```diff
@@ -0,0 +1,158 @@
...
+                        serialization_type: #serialization_type,
+                    }
+                }
+            })
+            .collect();
+        let callbacks: Vec<TokenStream2> = self
+            .attr_signature_info
+            .args
+            .iter()
+            .filter(|arg| matches!(arg.bindgen_ty, BindgenArgType::CallbackArg))
```
austinabell, line 72: This handles callback, but does not handle `CallbackResultArg`. Am I missing something with this?

examples/abi/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+            sum_pair(&a, &b)
+        }
+
+        pub fn add_callback(
+            &self,
+            #[callback_unwrap] a: DoublePair,
+            #[callback_unwrap] b: DoublePair,
+            #[callback_vec] others: Vec<DoublePair>,
+        ) -> DoublePair {
+            vec![b].iter().chain(others.iter()).fold(a, |acc, el| DoublePair {
```
austinabell, line 35: nit:
```suggestion
            Some(b).iter().chain(others.iter()).fold(a, |acc, el| DoublePair {
```
or just using array instead of vec which would allocate

near-sdk/Cargo.toml
```diff
@@ -25,6 +25,7 @@ near-sys = { path = "../sys", version = "0.2" }
 base64 = "0.13"
 borsh = { version = "0.9", features = ["const-generics"] }
 bs58 = "0.4"
+schemars = "0.8.8"
```
austinabell, line 28: this should be optional and only enabled by `unstable` feature

near-sdk/src/lib.rs
```diff
@@ -70,3 +73,6 @@ pub use serde;
 
 #[doc(hidden)]
 pub use serde_json;
+
+#[doc(hidden)]
+pub use schemars;
```
austinabell, line 78: let's just remove this altogether, re-exporting things like this is generally a bad pattern and the others would be purged if it wasn't very breaking of a change

near-sdk-macros/src/core_impl/abi/abi_generator.rs
itegulov: Oh, yeah, you are right. I took it from here: https://github.com/near/near-sdk-rs/blob/master/near-sdk-macros/src/core_impl/metadata/metadata_generator.rs#L69, so I guess the previous metadata implementation is not accounting for this case either. Should I fix it in a separate PR or do you think it is inconsequential at this point?

examples/abi/src/lib.rs
```diff
@@ -0,0 +1,45 @@
+use near_sdk::near_abi;
+
+#[near_abi]
+mod abi {
```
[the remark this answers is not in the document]
itegulov: Indeed, that helped!

examples/abi/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+authors = ["Near Inc <hello@nearprotocol.com>"]
+edition = "2021"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+near-sdk = { path = "../../near-sdk", features = ["unstable"] }
+serde = { version = "1", features = ["derive"] }
+schemars = "0.8"
```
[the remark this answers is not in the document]
itegulov: So maybe I am missing something but this is conflicting with your other comment about removing re-export of schemars: we need to derive `schemars::JsonSchema` for all types that are going to show up in ABI. I can create an abstraction for developers and make them use a macro to derive *something* that gives them the ability to generate ABI for these types:

```rust
#[proc_macro_attribute]
pub fn derive_near_abi(_attr: TokenStream, input: TokenStream) -> TokenStream {
    TokenStream::from(quote! {
        #[derive(schemars::JsonSchema)]
        #input
    })
}
```

The problem here is that developers will still need to have `schemars` as a dependency, right? I could use `near_sdk::schemars` here instead but you are saying we should remove it in the other thread. I also explored the option of creating my own trait `NearAbi` that would encapsulate JSON Schema logic, but unfortunately not having `schemars::JsonSchema` directly means I cannot use `schemars::Generator` which is a major problem when it comes to ABI generation.

near-sdk-macros/src/core_impl/abi/abi_generator.rs
austinabell: I guess it depends what we do with metadata, but my assumption is that it would be deprecated.

As for callback result, will this not have to be handled more specifically than [`ccfba2e` (#831)](https://github.com/near/near-sdk-rs/pull/831/commits/ccfba2e07128d73748f9e04fd881bdfc3927ea8c) to ignore the wrapping `Result` type in the schema?

examples/abi/Cargo.toml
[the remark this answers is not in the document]
austinabell: The idea I'm trying to suggest is having some wrapper type which internally uses schemars, so that if we change the internal functionality it doesn't change the API for developers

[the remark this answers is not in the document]
itegulov: Discussed with @austinabell offline and he is going to explore this separately

examples/abi/res/abi.json
itegulov: As discussed in a meeting we decided to remove type registry for now and potential revisit this decision if the size of abi file becomes a big concern.

itegulov:
@austinabell so after our last conversation I am not sure if it makes sense to leave `abi` under `unstable` feature now. My vision is that there is going to be an `abi` feature that controls whether `near_bindgen` generates ABI symbols. So I don't know if it makes sense to bundle it up with `unstable`, should I create something like `abi_unstable` for it for now?

austinabell:
> @austinabell so after our last conversation I am not sure if it makes sense to leave `abi` under `unstable` feature now. My vision is that there is going to be an `abi` feature that controls whether `near_bindgen` generates ABI symbols. So I don't know if it makes sense to bundle it up with `unstable`, should I create something like `abi_unstable` for it for now?

no, if we are adding a feature, it should be something long term. If you wanted it to be `abi_unstable`, instead you can just require someone enable the `abi` feature and `unstable`. Up to you which direction this goes into between putting it under `abi`, `unstable`, or both. My guess is `abi` is probably the best, we just need to make sure that name is stable and set before releasing the next SDK version

itegulov:
> no, if we are adding a feature, it should be something long-term because removing is unergonomic for anyone who specified it previously (breaking change). If you wanted it to be abi_unstable, instead you can just require someone enable the abi feature and unstable. Up to you which direction this goes into between putting it under abi, unstable, or both. My guess is abi is probably the best, we just need to make sure that name is stable and set before releasing the next SDK version

Makes sense, I am putting everything behind `abi` then.

near-sdk/src/private/abi.rs
```diff
@@ -0,0 +1,183 @@
...
+    pub version: Option<String>,
+    /// The authors of the smart contract.
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    #[serde(default)]
+    pub authors: Vec<String>,
+    /// Other arbitrary metadata.
+    #[serde(skip_serializing_if = "HashMap::is_empty")]
+    #[serde(default)]
+    #[serde(flatten)]
+    pub other: HashMap<String, String>,
```
austinabell, line 82:
```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub name: Option<String>,
    /// The version of the smart contract.
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub version: Option<String>,
    /// The authors of the smart contract.
    #[serde(default, skip_serializing_if = "Vec::is_empty")]
    pub authors: Vec<String>,
    /// Other arbitrary metadata.
    #[serde(default, flatten, skip_serializing_if = "HashMap::is_empty")]
    pub other: HashMap<String, String>,
```

Optional: [You can keep these within one line](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=bf440278b3f491c556cc1d506721f372) if you want.

near-sdk/src/private/abi.rs
```diff
@@ -0,0 +1,183 @@
...
+    #[serde(default)]
+    pub callbacks: Vec<AbiType>,
+    /// Type identifier of the vararg callbacks of the function.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
+    pub callbacks_vec: Option<AbiType>,
+    /// Return type identifier.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
+    pub result: Option<AbiType>,
```
austinabell, line 149: Same as previous comment, serde attributes can be combined to one line if you want

austinabell (commented):
Looks generally good! Just a few small comments and should be good

near-sdk-macros/src/core_impl/abi/abi_generator.rs
```diff
@@ -0,0 +1,168 @@
...
+                            "A function can only have one #[callback_vec] parameter.",
+                        )
+                        .to_compile_error();
+                    }
+                }
+            };
+        }
+        let callback_vec = callback_vec.unwrap_or(quote! { None });
+
+        let result = if matches!(self.attr_signature_info.method_type, MethodType::Init) {
```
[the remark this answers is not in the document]
itegulov: Oh yeah good catch

near-sdk-macros/src/core_impl/abi/abi_generator.rs
```diff
@@ -0,0 +1,168 @@
...
+    ///     result: near_sdk::__private::AbiParameter {
+    ///         type_id: 2,
+    ///         serialization_type: "json",
+    ///     }
+    /// }
+    /// ```
+    /// If args are serialized with Borsh it will not include `#[derive(borsh::BorshSchema)]`.
+    pub fn abi_struct(&self) -> TokenStream2 {
+        let function_name_str = self.attr_signature_info.ident.to_string();
+        let is_view = matches!(&self.attr_signature_info.method_type, &MethodType::View);
```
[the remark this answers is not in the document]
itegulov: Good point, created an issue: https://github.com/near/near-sdk-rs/issues/850

austinabell (approved):

near-sdk-macros/src/core_impl/abi/abi_visitor.rs
```diff
@@ -0,0 +1,67 @@
...
+        let functions: Vec<TokenStream2> =
+            public_functions.iter().map(|m| m.abi_struct()).collect();
+        let first_function_name = &public_functions[0].attr_signature_info.ident;
+        let near_abi_symbol = format_ident!("__near_abi_{}", &first_function_name);
+        Ok(quote! {
+            #[cfg(not(target_arch = "wasm32"))]
+            const _: () = {
+                #[no_mangle]
+                pub fn #near_abi_symbol() -> near_sdk::__private::AbiRoot {
+                    let mut gen = schemars::gen::SchemaGenerator::default();
```
austinabell, line 56: this should use the re-export from the SDK. Ideally behind a private module. This forces a developer to import schemars even if they don't use

This should be `near_sdk::__private::schemars::` or `near_sdk::schemars::` if that re-export is not moved from other comment

near-sdk/src/lib.rs
austinabell: Noting again this would be ideal not to export and require that we keep this API stable

austinabell (changes requested):
sorry, one thing I overlooked and then good

austinabell (approved):

near-sdk/src/lib.rs
itegulov: Gotcha, moved behind `__private` for now since I can't figure out how to remove it

PR #836, opened 2022-06-08
Add ext_contract macro to FungibleTokenMetadataProvider trait

ebramanti:
Currently, users building smart contracts can use  `ext_ft_core` to make cross-contract calls to Fungible token methods enumerated in NEP-141. However, to make a call to the `ft_metadata` method, they have to define their own trait in their smart contract because `FungibleTokenMetadataProvider` does not have a `ext_contract` macro. This PR adds the `ext_contract` macro to the `FungibleTokenMetadataProvider` trait so that users do not have to copy/paste this trait into their code.


austinabell (approved):
Very valid! Thanks for the contribution :)

PR #838, opened 2022-06-14
chore: add audit to SDK CI

austinabell:
Ignoring time audit vulnerability because it isn't exposed through chrono, which is what `nearcore` crates pull this in through

https://github.com/chronotope/chrono/issues/602

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,6 +60,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
+          toolchain: stable
       - uses: Swatinem/rust-cache@v1
       - name: Install Audit
         run: cargo install cargo-audit
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,3 +50,21 @@ jobs:
       - uses: Swatinem/rust-cache@v1
       - run: cargo check -p near-sdk --features unstable
       - run: cargo check -p near-contract-standards
+  audit:
+    name: Audit
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Sources
+        uses: actions/checkout@v2
+      - name: Install Toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+      - uses: Swatinem/rust-cache@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -50,3 +50,22 @@ jobs:
...
+    name: Audit
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Sources
+        uses: actions/checkout@v2
+      - name: Install Toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
```
itegulov, line 63: Does it matter if we only use stable here vs also MSRV (1.56.0)?

itegulov (approved):
LGTM. Do we want to add this as a required check for merging PRs?

itegulov:
> Ignoring time audit vulnerability because it isn't exposed through chrono, which is what nearcore crates pull this in through

Also noting that just because `time` is not exposed does not mean that it cannot cause segfaults but yeah makes sense that it should be solved at nearcore side

austinabell:
> > Ignoring time audit vulnerability because it isn't exposed through chrono, which is what nearcore crates pull this in through
> 
> Also noting that just because `time` is not exposed does not mean that it cannot cause segfaults but yeah makes sense that it should be solved at nearcore side

That wasn't what I was saying. I was saying that `chrono` doesn't use the section of code that causes segfaults in `time` and `chrono` is only pulled in through `nearcore` crates. This is why I linked https://github.com/chronotope/chrono/issues/602 to show that the resolution was that the vulnerable APIs are not called.

.github/workflows/test.yml
austinabell: it doesn't affect this since it's just checking the dependencies in the audit registry and not compiling things

ChaoticTempest (approved):
Also looks good to me. Might be worthwhile to add something similar in workspaces

austinabell:
> Also looks good to me. Might be worthwhile to add something similar in workspaces

It's a very low-cost sanity check. I'd recommend it. I was going to add it sooner but there were previously many vulnerable crates in previous versions of testing utilities we were using through `nearcore`

PR #843, opened 2022-06-17
change unwrap to and_then

hanakannzashi:
In case of panic when TokenMetadata Extension and Approval Extension are not implemented

PR #846, opened 2022-06-17
fix(standards): redundancy and inaccuracy in nft storage measure

### LINKED ISSUE #844: Confusing measure storage in NonFungibleToken
1.tokens_per_owner calculate twice 2.Did not remove tmp_token_id from UnorderedSet when roll back [<span class="image-wrap" style=""></span>](https://user-images.githubusercontent.com/95295094/174286829-81372e27-0930-459e-8114-555c962adff7.jpeg) [<span class="image-wrap" style=""></span>](https://user-images.githubusercontent.com/95295094/174286848-984a2246-f340-4918-a893-d095dd7dac3a.png)

austinabell:
- Fixes bug with storage measurement
- Removes redundant code and operations
- Updates key for storage keys for tokens per owner to a 32 byte array. I have no explanations or guesses why there was asymmetry between the other storage key that was also a hash

closes #844 

@BenKurrek would love your input on this

### DIFF
--- a/near-contract-standards/src/non_fungible_token/core/core_impl.rs
+++ b/near-contract-standards/src/non_fungible_token/core/core_impl.rs
@@ -5,9 +5,7 @@ use crate::non_fungible_token::core::NonFungibleTokenCore;
 use crate::non_fungible_token::events::{NftMint, NftTransfer};
 use crate::non_fungible_token::metadata::TokenMetadata;
 use crate::non_fungible_token::token::{Token, TokenId};
-use crate::non_fungible_token::utils::{
-    hash_account_id, refund_approved_account_ids, refund_deposit_to_account,
-};
+use crate::non_fungible_token::utils::{refund_approved_account_ids, refund_deposit_to_account};
 use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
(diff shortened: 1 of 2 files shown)


near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -53,7 +51,7 @@ pub struct NonFungibleToken {
 
 #[derive(BorshStorageKey, BorshSerialize)]
 pub enum StorageKey {
-    TokensPerOwner { account_hash: Vec<u8> },
+    TokensPerOwner { account_id_hash: CryptoHash },
```
willemneal, line 54: Still am advocating for this to be an AccountId.  You charge for a full length account id, but it will reduce storage on average since most real users have shorter account ids, and it will not require a host call to produce a hash.

willemneal: Furthermore a state dump will make it clear who the owner is.

austinabell: This would be a state-breaking change, though. This can come if there are breaking changes or if the standards were re-written.

austinabell:
oops, didn't see #843 

will just use/build off that

Issue #852, opened 2022-06-24
`#[handle_result]` should not rely on `Result` being referred to using specific tokens


austinabell:
Currently, the attribute expects Result<T, E> specifically, which is a subset of syntax that this functionality allows.

The main cases that should be handled are:

Pulling a result type from another module

```rust
#[handle_result]
pub fn foo() -> ::core::result::Result<(), &'static str> {...}
```

or a type alias to result:

```rust
type Result<T> = core::result::Result<T, &'static str>;

#[handle_result]
pub fn foo() -> Result<()> {...}
```

Where having both of these would allow a common pattern like:

```rust
#[handle_result]
pub fn foo() -> anyhow::Result<()> {...}
```

Motivation: [near.zulipchat.com/#narrow/stream/300659-Rust-.F0.9F.A6.80/topic/anyhow.3F/near/287320608](https://near.zulipchat.com/#narrow/stream/300659-Rust-.F0.9F.A6.80/topic/anyhow.3F/near/287320608)

uint:
At macro expansion time, the compiler operates on tokens and has no access to type information. That makes this whole deal a hacky PITA.

I'm toying with the idea of moving ["`Ok` type" extraction to the type/trait system](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ff72fc943c1952565d486af50a082f80) instead. I suspect that will work out much better.

uint:
Resolving this ticket isn't enough to support `anyhow`. Opened another issue: #1055

frol:
> I'm toying with the idea of moving ["Ok type" extraction to the type/trait system](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ff72fc943c1952565d486af50a082f80) instead. I suspect that will work out much better.

This is exactly the direction I would love to see it going. Thanks for looking into that direction.

miraclx:
Did exactly this last night for Calimero, figured I'd mention it here as well.

The idea is to inject scaffolding for any return type, and specialize for results using the trait domain. Rust treats direct association with a higher precedence than trait-enabled ones. It's the same magic that powers https://github.com/nvzqz/impls.

Once implemented, you'll no longer need the annotation, nor will you be forcing downstream users to return a `Result`.

| Method Signature | `IntoResult` Representation |
| -: | - |
| `fn()` | `Result<(), Infallible>` |
| `fn() -> T` | `Result<T, Infallible>` |
[...]

PR #854, opened 2022-06-25
feat: make #[hanlde_result] work with fully qualified paths

itegulov:
Fixes https://github.com/near/near-sdk-rs/issues/852

Opening this as a draft for now since there is one unresolved thing with `FunctionError` impl for `anyhow::Error`. There is no `AsRef<str>` implementation for `anyhow::Error` and we can't implement it ourselves (or `FunctionError` itself) due to conflicting implementations (life is hard without specialization...).

### DIFF
--- a/near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
+++ b/near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
@@ -853,4 +853,33 @@ mod tests {
         );
         assert_eq!(expected.to_string(), actual.to_string());
     }
+
+    #[test]
+    fn handle_result_anyhow() {
+        let impl_type: Type = syn::parse_str("Hello").unwrap();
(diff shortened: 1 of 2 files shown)


uint:
`AsRef<str>` can't be implemented on `anyhow::Error` anyway. Since an `anyhow::Error` might have to build a string dynamically, there's no way to guarantee a cheap pointer cast there.

@itegulov @frol What's the rationale for having `impl AsRef<str>` as the SDK error type in the first place? Is it to encourage devs to produce unhappy paths that are cheap? Or is it just convenience on our side for easily getting something that we can call [`near_sdk::env::panic_str`](https://docs.rs/near-sdk/latest/near_sdk/env/fn.panic_str.html#) on? I'd like to understand our constraints and needs here.

uint:
Stale PR. Closing.

Issue #855, opened 2022-06-27
Invalid check in `Upgradeable` implementation.


xenoliss:
Hi,

This line [https://github.com/near/near-sdk-rs/blob/master/near-contract-standards/src/upgrade/mod.rs#L69](https://github.com/near/near-sdk-rs/blob/master/near-contract-standards/src/upgrade/mod.rs#L69) seems to be wrong and should be `if self.staging_timestamp > env::block_timestamp()` if I understand correctly the goal of this contract.

itegulov:
Hmm, I think the check is right. When you call `stage_code` you are essentially saying that the upgrade will happen sometime on or after `self.staging_timestamp` (which has to be at least `self.staging_duration` from the current block timestamp at the time of calling `stage_code`). So calling `deploy_code` only works once the current block's timestamp is on or after `self.staging_timestamp`.

I was actually not aware of this file's existence. @austinabell is this actually some kind of standard? Could not find any mention of contract upgrade standardization in our docs.

xenoliss:
> will happen sometime on or **after** self.staging_timestamp

@itegulov that's actually not what is checked here. Say we are **before** `self.staging_timestamp`, then I'll be able to deploy, the condition seems reverted to me.

itegulov:
> Say we are before `self.staging_timestamp`, then I'll be able to deploy, the condition seems reverted to me.

Talking about interval condition can be confusing so let me do a little breakdown of the code:

```rust
```

itegulov:
Oh, yeah, now that I am looking at it again it seems like you are right. The check is written in a non-intuitive manner to me, so was not easy to interpret it (double negation of how I normally think about time interval checks). Needs to be fixed for sure, but also would like to hear @austinabell's opinion on how useful this is as a part of `near-contract-standard`.

austinabell:
> Oh, yeah, now that I am looking at it again it seems like you are right. The check is written in a non-intuitive manner to me, so was not easy to interpret it (double negation of how I normally think about time interval checks). Needs to be fixed for sure, but also would like to hear @austinabell's opinion on how useful this is as a part of `near-contract-standard`.

I have no idea why it was added. There is [no standards in NEPs](https://github.com/near/NEPs/tree/master/specs/Standards) about this.

I think the best call for this is just to deprecate this whole standard and then implement it only if a standard arises. Hopefully, the soft deprecation can point out why/or how this is used in the ecosystem, and if so, we can fix issues with it.

[...]

PR #856, opened 2022-06-27
chore: deprecate upgrade standards implementation

austinabell:
Not quite sure why this was added, somewhat broken as per #855 and the interface doesn't make much sense and is inefficient.

Deprecating for now and can optimize/update if there are real use cases or the standard is actually added to NEPs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Removed
+- Deprecated `near_contract_standards::upgrade`. [PR 856](https://github.com/near/near-sdk-rs/pull/856)
+  - Implementation did not match any NEAR standard and was not correct
+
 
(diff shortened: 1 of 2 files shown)


itegulov (approved):
This might make more sense as an example. WDYT?

austinabell:
@ChaoticTempest noting that the CI had a failure with the same error as before, just to note the non-deterministic failures are still happening (maybe less frequent, unclear yet)

ChaoticTempest:
@austinabell hmm, weird. Yeah keep updating me about these. Might need to up it via the env var to 100 secs, or maybe its not even a timeout issue happening, but that the sandbox node wasn't able to start up. If its the latter that's the issue, the error should have been reported earlier on... It could be that GH has a shared CPU behind its task scheduling, so you get bursty workloads that leads to this non-determinism

PR #857, opened 2022-06-27
chore: bump workspaces version to new patch

austinabell:
This patch version fixes the non-determinism in our CI failures (which is starting to get annoying)


itegulov (approved):
Nice, was definitely annoying me :)

Issue #853, opened 2022-06-24
Remove requirement of Error type being `JsonSchema` for `#[handle_result]` errors


austinabell:
<pre> #<span class="error">[handle_result]</span>
 pub fn foo() -> Result<(), anyhow::Error> 

 { Err(anyhow::anyhow!("test")) } 

</pre>

and error is:

 error<span class="error">[E0277]</span>: the trait bound `anyhow::Error: JsonSchema` is not satisfied
 --> src/lib.rs:11:1

<div class="table-wrap"> 
</div>

 11 | #<span class="error">[near_bindgen]</span>

<div class="table-wrap">
  | ^^^^^^^^^^^^^^^ the trait `JsonSchema` is not implemented for `anyhow::Error` | 
</div>

Not sure yet if the codegen isn't working, because I thought the `Ok` type was being extracted for this.

cc @itegulov

itegulov:
Oh, yeah, I have added a special case for `#[callback_result]`, but not for `#[handle_result]`. Will fix.

PR #858, opened 2022-06-28
fix: support `#[handle_result]` in ABI macro

itegulov:
Fixes https://github.com/near/near-sdk-rs/issues/853

### DIFF
--- a/near-sdk-macros/src/core_impl/abi/abi_generator.rs
+++ b/near-sdk-macros/src/core_impl/abi/abi_generator.rs
@@ -127,13 +127,13 @@ impl ImplItemMethodInfo {
                         None
                     }
                 }
-                ReturnType::Type(_, ty) if utils::type_is_result(ty) && is_handles_result => {
+                ReturnType::Type(_, ty) if is_handles_result && utils::type_is_result(ty) => {
                     let ty = if let Some(ty) = utils::extract_ok_type(ty) {
                         ty
                     } else {
                         return syn::Error::new_spanned(
                             ty,
-                            "Function marked with #[handle_result] should have return type Result<T, PromiseError>",
+                            "Function marked with #[handle_result] should have return type Result<T, E>",
                         )
                         .into_compile_error();
                     };
--- a/near-sdk-macros/src/core_impl/abi/abi_generator.rs
+++ b/near-sdk-macros/src/core_impl/abi/abi_generator.rs
@@ -133,7 +133,7 @@ impl ImplItemMethodInfo {
                     } else {
                         return syn::Error::new_spanned(
                             ty,
-                            "Function marked with #[handle_result] should have return type Result<T, E>",
+                            "Function marked with #[handle_result] should have return type Result<T, E> (where E implements FunctionError).",
                         )
                         .into_compile_error();
                     };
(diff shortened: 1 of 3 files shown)


near-sdk-macros/src/core_impl/abi/abi_generator.rs
```diff
@@ -126,6 +127,34 @@ impl ImplItemMethodInfo {
                         None
                     }
                 }
+                ReturnType::Type(_, ty) if utils::type_is_result(ty) && is_handles_result => {
```
ChaoticTempest, line 130: nit: not sure if rust reorders the conditional so that we can short circuit earlier with just `is_handles_result` instead of evaluating via `type_is_result(ty)`
```suggestion
                ReturnType::Type(_, ty) if is_handles_result && utils::type_is_result(ty) => {
```

near-sdk-macros/src/core_impl/abi/abi_generator.rs
```diff
@@ -126,6 +127,34 @@ impl ImplItemMethodInfo {
...
+                    let ty = if let Some(ty) = utils::extract_ok_type(ty) {
+                        ty
+                    } else {
+                        return syn::Error::new_spanned(
+                            ty,
+                            "Function marked with #[handle_result] should have return type Result<T, PromiseError>",
```
ChaoticTempest, line 136: Shouldn't necessarily be `PromiseError` for the `Result` error variant right? 
```suggestion
                            "Function marked with #[handle_result] should have return type Result<T, E>",
```

austinabell: Yes, that's correct, only `callback_result` uses that type. For the function error type it can be anything that implements the `FunctionError` trait. Maybe that can be expressed in this comment also

near-sdk-macros/src/core_impl/abi/abi_generator.rs
austinabell: Not certain about this, seems it isn't guaranteed from https://github.com/rust-lang/rust/issues/83623 and related things. Prob best to re-order to be safe, but this additional computation only happens at compile time to definitely a micro-optimization

near-sdk-macros/src/core_impl/abi/abi_generator.rs
itegulov: Yeah just forgot to update the text after copy-pasting this from somewhere else. Added a comment about E having to implement `FunctionError`, seems like a useful clarification indeed

near-sdk-macros/src/core_impl/abi/abi_generator.rs
```diff
@@ -126,6 +127,34 @@ impl ImplItemMethodInfo {
...
+                ReturnType::Type(_, ty) if is_handles_result && utils::type_is_result(ty) => {
+                    let ty = if let Some(ty) = utils::extract_ok_type(ty) {
+                        ty
+                    } else {
+                        return syn::Error::new_spanned(
+                            ty,
+                            "Function marked with #[handle_result] should have return type Result<T, E> (where E implements FunctionError).",
+                        )
+                        .into_compile_error();
+                    };
```
austinabell, line 139: feels like this is code duplication from the extraction from non-abi codegen, but probably fine for now

austinabell (approved):

near-sdk-macros/src/core_impl/abi/abi_generator.rs
itegulov: Yep, I kind of feel like the bindgen macro needs a pre-processing step that would do all of these "type" checks (`#[handle_result]`, `#[callback_vec]` etc) and produce a well-formed model. In this case, it would something like this:

```rust
enum BindgenReturnType {
    Regular(Ident),
    HandleResult { ident: Ident, ok_type: Ident, err_type: Ident },
}

struct BindgenFunction {
    ...
    return_type: BindgenReturnType
}
```

So that we can just safely use these models in later steps.

PR #859, opened 2022-06-28
docs: update ABI example with cargo-near install

jon-lewis:
Maybe I missed it but when I was working through the ABI example it wasn't clear where I could install this cargo-near extension.


examples/abi/README.md
```diff
@@ -2,6 +2,10 @@
 
 Showcases how to generate ABI from a NEAR contract.
 
+## Install ABI Cargo Extension
+
+The Near ABI cargo extension can be installed from source [here](https://github.com/near/cargo-near/tree/daniyar/init).
```
austinabell, line 7:
```suggestion
The Near ABI cargo extension can be installed from source [here](https://github.com/near/cargo-near).
```

PR #863, opened 2022-06-30
fix: extract #[callback_vec] vec type for abi

itegulov:
Forgot to extract the underlying type from `Vec<_>` for ABI

### DIFF
--- a/near-sdk-macros/src/core_impl/utils/mod.rs
+++ b/near-sdk-macros/src/core_impl/utils/mod.rs
@@ -9,15 +9,6 @@ pub(crate) fn path_is_result(path: &Path) -> bool {
         && path.segments.iter().next().unwrap().ident == "Result"
 }
 
-/// Checks whether the given path is literally "Vec".
-/// Note that it won't match a fully qualified name `std::vec::Vec` or a type alias like
-/// `type MyVec = Vec<String>`.
-pub(crate) fn path_is_vec(path: &Path) -> bool {
-    path.leading_colon.is_none()
-        && path.segments.len() == 1
-        && path.segments.iter().next().unwrap().ident == "Vec"
-}
-
 /// Equivalent to `path_is_result` except that it works on `Type` values.
 pub(crate) fn type_is_result(ty: &Type) -> bool {
     match ty {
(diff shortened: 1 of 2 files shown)


near-sdk-macros/src/core_impl/utils/mod.rs
```diff
@@ -39,3 +39,36 @@ pub(crate) fn extract_ok_type(ty: &Type) -> Option<&Type> {
...
     }
 }
+
+/// Checks whether the given path is literally "Vec".
+/// Note that it won't match a fully qualified name `std::vec::Vec` or a type alias like
+/// `type MyVec = Vec<String>`.
+fn path_is_vec(path: &Path) -> bool {
+    path.leading_colon.is_none()
+        && path.segments.len() == 1
+        && path.segments.iter().next().unwrap().ident == "Vec"
```
austinabell, line 49: Why is this necessary? Anything else other than a `Vec` would fail to compile by the macros, so why not allow aliases? Very niche case so fine to ignore

near-sdk-macros/src/core_impl/utils/mod.rs
```diff
@@ -39,3 +39,36 @@ pub(crate) fn extract_ok_type(ty: &Type) -> Option<&Type> {
...
+}
+
+/// Extracts the inner generic type from a `Vec<_>` type.
+///
+/// For example, given `Vec<String>` this function will return `String`.
+#[allow(dead_code)]
```
austinabell, line 55: this should just be marked as only non-wasm32 config rather than allowing dead code, to make sure if this goes unused in the future it will be warned to remove.

austinabell (commented):
Just a change to the dead_code comment is fine

near-sdk-macros/src/core_impl/utils/mod.rs
itegulov: Dead code warning is happening because it is only used in modules enabled by `abi` feature. My reasoning for not annotating this function with `#[cfg(feature = "abi")]` was that it has nothing specific to do with abi and it might be generally useful in other places (e.g. should we actually check the collection type in near_bindgen macro to report an error that makes sense? right now you would get an error saying that an iterator cannot be collected to type X which might be difficult to interpret).

If you think it still should be under abi feature I am happy to do so.

near-sdk-macros/src/core_impl/utils/mod.rs
itegulov: As I said in my other comment a verbose check just makes errors more transparent, but yeah this certainly comes with its own downsides. Happy to go either way.

near-sdk-macros/src/core_impl/utils/mod.rs
austinabell: yeah just under the abi feature is better, my bad on the assumption it was based on target. Better this way because if it's needed otherwise, can just change the config then. Harder to remember to remove the function if unused, or remove the dead_code annotation if it does become used.

near-sdk-macros/src/core_impl/utils/mod.rs
austinabell: Yeah, we can leave for now. Easier to add support for aliases later than to take it away

austinabell (approved):

PR #870, opened 2022-07-21
feat: hide nearcore crates behind testing feature

itegulov:
It seems like it is not possible to configure default features per arch, so I had to do cover everything testing-related with `#[cfg(all(not(target_arch = "wasm32"), feature = "testing"))]`. LMK if there is a better way.

Relates to https://github.com/near/cargo-near/issues/8


austinabell (approved):
All lgtm, I don't think there is a better way to do this. I wonder though, would `unit-testing` be a more clear feature name? Maybe confusing with workspaces? cc @ChaoticTempest

ChaoticTempest:
if only they allowed `cfg(test)` to be usable as a library flag as well, but I guess that's what `dev-dependencies` are for. And yeah, feel like `unit-test` or `unit-testing` would be better as `testing` is more general and likely confusing with workspaces

ChaoticTempest (approved):

itegulov:
Renamed to `unit-testing`

PR #872, opened 2022-07-27
feat: add proper support for borsh schema containers in ABI

itegulov:
The main problem with this approach is that all schema containers are independent of each other. This can make the ABI files bloated when working with many complex types. Unfortunately, it seems like `borsh-rs` does not have a generator model like `schemars` does, so I can't generate one big schema like I do with JSON Schema. Maybe this is something that would make sense to add to borsh-rs in the future though?

Related to https://github.com/near/cargo-near/issues/14


near-sdk/src/private/abi.rs
```diff
@@ -141,28 +144,142 @@ pub struct AbiFunction {
...
+        },
+        Tuple {
+            elements: Vec<Declaration>,
+        },
+        Enum {
+            variants: Vec<(VariantName, Declaration)>,
+        },
+        Struct {
+            #[serde(with = "FieldsDef")]
+            fields: Fields,
```
austinabell, line 232: The fields key feels redundant. Is there a reason this and others can't be inlined to avoid unnecessary data in abi?

examples/abi/res/abi.json
```diff
@@ -15,57 +15,116 @@
...
+            "definitions": {
+              "Pair": {
+                "Struct": {
+                  "fields": {
+                    "UnnamedFields": [
+                      "u32",
+                      "u32"
+                    ]
+                  }
+                }
```
austinabell, line 94: Intuitively, feels like this should be
```suggestion
                "Struct": [
                      "u32",
                      "u32"
                    ]
```

And I can't see why you wouldn't be able to do this because the struct is either a JSON map, an array, or empty (empty map or null, idk what it's used for), am I missing something?

near-sdk/src/private/abi.rs
itegulov: I am just following borsh-rs here: https://github.com/near/borsh-rs/blob/7325a0aab74049237b9f978e1e2c0fcf7bb799c2/borsh/src/schema.rs#L46

The structs need to be duplicated and match the other struct one for one so I can use serde's remote derivation (see https://serde.rs/remote-derive.html).

examples/abi/res/abi.json
itegulov: Hmm yeah that's completely fair. I will see if I can use remote derivation with custom serde attributes to control the format a bit more.

itegulov: Done! This introduced even more complexity to the code, but at least it seems to work. This definitely needs some tests though, will add them next week.

austinabell (approved):
The workarounds seem a bit unnecessary since we can modify borsh to include clone and update the schema format, but this seems fine for now to unlock the functionality.

Can you open an issue in borsh identifying how we want the serde serialization format to look, so we can get some input before making any changes to avoid the extra logic here? I opened https://github.com/near/borsh-rs/pull/97 because it was unopinionated

itegulov:
I finished tests and this seems to behave as I expect it to, so I will merge this.

@austinabell created the serde derivation issue in borsh repo: https://github.com/near/borsh-rs/issues/98

PR #879, opened 2022-08-05
chore: release 4.1.0-pre.1

austinabell:
Set the date in changelog to tomorrow because that's when I'll release it (probably). If I don't, I'll adjust


itegulov (approved):
> Set the date in changelog to tomorrow because that's when I'll release it

Tomorrow for many, today for some :smile:

PR #880, opened 2022-08-05
chore: rm docker build check from CI

austinabell:
Being so rigid about the wasm code changes isn't as important anymore and is slowing things down.

This removes all wasm files from the repo (I don't think there was any reason they be committed in the first place), and the check that the files be kept up to date with git. It does a smoke test for a few contracts to make sure the size isn't bloated and also prints out the code size for all examples, so worst case, we can debug where the size increased if it did.

Ideally, there is a manual trigger we can run to compare a PR against the main branch, so I'll open an issue for that


ChaoticTempest (approved):
LGTM

itegulov (approved):

PR #885, opened 2022-08-11
feat: Support newly stabilized alt_bn128 host functions

austinabell:
It's been stabilized, so devs will need this to test


austinabell:
Update: this has been confirmed to work with cargo-fuzz (by a community member)

ChaoticTempest (approved):

PR #886, opened 2022-08-12
fix: Update generic for LookupSet hasher

austinabell:
No need for the drop check on this, the structure does not have an owned value for this, just uses the generic for a pure function.

Switched the old collection generics as well. On the one hand, they do not own the values, but having the drop check might prevent some unintended behaviour.

@matklad do you have any intuitions about this? Hard to think of a reason why this shouldn't be done, but unsure if I'm missing something

### DIFF
--- a/near-sdk/src/collections/lazy_option.rs
+++ b/near-sdk/src/collections/lazy_option.rs
@@ -18,7 +18,7 @@ const ERR_VALUE_DESERIALIZATION: &str = "Cannot deserialize value with Borsh";
 pub struct LazyOption<T> {
     storage_key: Vec<u8>,
     #[borsh_skip]
-    el: PhantomData<T>,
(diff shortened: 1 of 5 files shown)


austinabell:
Okay, I failed at creating the counter case for this. Having that test while using `PhantomData<T>` doesn't fail.

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=da00ce352e1851168f5c27cdfa68f093 was what I played around with to try to get it to fail previously and not after.

The change _feels_ correct since these structures don't own the generic types in `PhantomData`, but I can't seem to write a test that shows the benefit

austinabell:
I wasn't able to find a corner case to test that this functionality breaks on the old version and the new one works. I'm not working on this anymore, so just going to close

PR #888, opened 2022-08-12
chore: Stabilize and update lazy collections

austinabell:
This one seemed like the next most unopinionated one to stabilize.

Also changed the error message for inconsistent state. Not sure why it was worded as it is. If the contract terminated unexpectedly (errored), it will roll back the state changes. I think this new wording is a bit clearer about the issue, but I'm very open to better wording.

PR is pointing at #815 so best to review that one first


near-sdk/src/collections/mod.rs
```diff
@@ -59,7 +59,7 @@ pub use lazy_option::LazyOption;
 mod tree_map;
 pub use tree_map::TreeMap;
 
-pub const ERR_INCONSISTENT_STATE: &str = "The collection is an inconsistent state. Did previous smart contract execution terminate unexpectedly?";
+pub const ERR_INCONSISTENT_STATE: &str = "The collection is an inconsistent state. Have any collections been updated without updating themselves in state?";
```
ChaoticTempest, line 62: Think this should read better:
```suggestion
pub const ERR_INCONSISTENT_STATE: &str = "The collection is in an inconsistent state. Have any collections been updated without committing changes to contract state?";
```

ChaoticTempest (approved):

PR #889, opened 2022-08-16
extract ABI primitives into near-abi-rs

miraclx:
https://github.com/near/near-abi-rs/pull/1


near-sdk/src/lib.rs
```diff
@@ -70,3 +70,7 @@ pub use serde;
 
 #[doc(hidden)]
 pub use serde_json;
+
+#[doc(hidden)]
+#[cfg(feature = "abi")]
+pub use schemars;
```
itegulov, line 76: This was actually left out on purpose, as exposing external crates is a harmful practice (@austinabell correct me if I am wrong).

Ideally SDK should have something like `#[derive(NearAbi)]` that would derive `JsonSchema` and/or `BorshSchema` automatically, but I am yet to find a technical way to do this (tried once a couple of months ago).

austinabell: yeah, the re-exports aren't a good pattern. Ideally, we have some wrapper to obfuscate the fact that we're using schemars, since that might change in the future.

I don't hold this opinion very strongly, but just trying to avoid a maintenance issue

miraclx: > Ideally SDK should have something like #[derive(NearAbi)] that would derive JsonSchema and/or BorshSchema automatically, but I am yet to find a technical way to do this

#891

near-sdk-macros/src/core_impl/abi.rs
```diff
@@ -1,51 +1,83 @@
...
+use crate::core_impl::{
+    utils, AttrSigInfo, BindgenArgType, ImplItemMethodInfo, ItemImplInfo, MethodType,
+    SerializerType,
+};
+
+pub fn generate(i: &ItemImplInfo) -> TokenStream2 {
+    let public_functions: Vec<&ImplItemMethodInfo> =
+        i.methods.iter().filter(|m| m.is_public || i.is_trait_impl).collect::<Vec<_>>();
+    if public_functions.is_empty() {
+        // Short-circuit if there are not public functions to export to ABI
```
itegulov, line 15:
```suggestion
        // Short-circuit if there are no public functions to export to ABI
```

near-sdk-macros/src/core_impl/abi.rs
```diff
@@ -1,51 +1,83 @@
...
+use syn::{Attribute, Lit::Str, Meta::NameValue, MetaNameValue, ReturnType, Type};
+
+use crate::core_impl::{
+    utils, AttrSigInfo, BindgenArgType, ImplItemMethodInfo, ItemImplInfo, MethodType,
+    SerializerType,
+};
+
+pub fn generate(i: &ItemImplInfo) -> TokenStream2 {
+    let public_functions: Vec<&ImplItemMethodInfo> =
+        i.methods.iter().filter(|m| m.is_public || i.is_trait_impl).collect::<Vec<_>>();
```
itegulov, line 13:
```suggestion
        i.methods.iter().filter(|m| m.is_public || i.is_trait_impl).collect();
```

itegulov (commented):
A couple of nitpicks, looks good otherwise. Thanks for cleaning up the visitor pattern. That was originally written when we considered applying the ABI macro to `mod` sections instead of `impl` sections.

near-sdk-macros/src/core_impl/mod.rs
```diff
@@ -1,11 +1,9 @@
 #[cfg(feature = "abi")]
-mod abi;
+pub mod abi;
```
austinabell, line 2: I don't think this needs to be exposed?

near-sdk/Cargo.toml
```diff
@@ -32,6 +32,12 @@ wee_alloc = { version = "0.4.5", default-features = false, optional = true }
 # Used for caching, might be worth porting only functionality needed.
 once_cell = { version = "1.8", optional = true, default-features = false }
 
+[dependencies.near-abi]
+git = "https://github.com/near/near-abi-rs"
+branch = "miraclx/init"
+features = ["__chunked-entries"]
+optional = true
```
austinabell, line 39: Everything seems fine, just want to make sure this doesn't come in until it points at a release (so blocked on https://github.com/near/near-abi-rs/pull/1)

examples/abi/Cargo.lock
```diff
@@ -1239,3 +1266,17 @@ dependencies = [
...
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "71e61de68ede9ffdd69c01664f65a178c5188b73f78faa21f0936016a888ff7c"
+dependencies = [
+ "borsh",
+ "byteorder",
+ "crunchy",
+ "lazy_static",
+ "rand 0.8.5",
+ "rustc-hex",
+]
```
austinabell, line 1282: why is it that these other deps are being pulled in?

near-sdk-macros/src/core_impl/abi.rs
```diff
@@ -1,51 +1,83 @@
...
     /// ```ignore
-    /// fn f3(&mut self, arg0: FancyStruct, arg1: u64) -> Result<IsOk, Error> { }
+    /// /// I am a function.
+    /// pub fn f3(&mut self, arg0: FancyStruct, arg1: u64) -> Result<IsOk, Error> { }
     /// ```
     /// will produce this struct:
     /// ```ignore
     /// near_sdk::__private::AbiFunction {
     ///     name: "f3".to_string(),
+    ///     doc: Some(" I am a function.".to_string()),
```
austinabell, line 48: Curious why the initial space isn't removed from these comments. I know unrelated, just wondering if an issue should be  opened up. cc @itegulov

austinabell (changes requested):

near-sdk-macros/src/core_impl/mod.rs
miraclx: Right. Wanted `pub(crate)` instead.

examples/abi/Cargo.lock
miraclx: `near-vm-logic`'s dependency on `zeropool-bn` used to be optional. It no longer is.

austinabell: the `near-vm-logic` dep didn't change with this, did it? Was that an accidental update? Just trying to make sure the side effects are clear

miraclx: Yeah, I don't understand why it just updated wither. Looking at the lockfile, `near-vm-logic` didn't change. Then again, it's limited to just the `abi` example. So any side effects should be sandboxed in there.

itegulov: I think we forgot to update the lockfile in this PR: https://github.com/near/near-sdk-rs/pull/885. Ever since we got rid of the examples diff check we are not enforcing lockfile to be up to date.

austinabell: ah, yes, we should probably add the git diff check back in

near-sdk-macros/src/core_impl/abi.rs
itegulov: The initial space comes from here:
```rust
/// I am a function
   ^
```

I left all original indentation intentionally as it is not clear on what exactly should be deleted. schemars actually tried doing some heuristics here, but they later realized that it does not suit all use cases and can even mess up markdown (see https://github.com/GREsau/schemars/issues/38#issuecomment-826395307). Specifically, here trimming all lines can mess with code indentation in `` ``` `` blocks. So I feel like post-processing should be left to the user depending on their use case as it is easy to remove unwanted stuff but impossible to deduce what stuff has already been removed by us.

austinabell: yeah, my assumption would be that just the first space would be trimmed, don't have opinions of any other line. Seems like that first space will always exist and might be a strange side-effect for those consuming it. Not blocking

itegulov: Hmm, I guess that is fair. I assumed that since Rust keeps this leading space, you could technically do `///this is doc`, but yeah I would not be opposed to trimming it.

itegulov (approved):

itegulov:
@austinabell anything blocking from your side? The near-abi dependency now points to a git rev which should be good enough IMO

near-sdk-macros/src/core_impl/abi.rs
austinabell: I'll just open an issue, and we can do it in a separate PR or close it if we don't want to do it. Super minor and opinionated

austinabell:
> @austinabell anything blocking from your side? The near-abi dependency now points to a git rev which should be good enough IMO

I can't do another pre-release until this points to a released version. We can do this in two-steps, where the SDK is left in an intermediate stage, or just release it and point to that (even if a pre-release). I'll leave that call to @miraclx

austinabell (approved):

miraclx:
Yeah, work on this PR is pretty much done. We can merge this right away and then wait for `abi-rs` to be published before we make a pre-release. Although, it's not my place to say how many revisions `abi-rs` will go through before it gets to that point.

PR #890, opened 2022-08-16
refactor: disentangle bindgen extractor logic

### LINKED ISSUE #926: No compiler error for using `Result` with init
Currently, any method that has `Result` as a return is given a compiler error, signposting toward `#[handle_result]`. The issue is that `#[init]` methods (as well as possibly others) don't emit this same error.

This is a footgun that someone hit from the ecosystem and the error is cryptic because it uses a serialized `Result` as the state. 

Unsure if we can safely enforce that the return type serialized is `Self`

cc @itegulov

itegulov:
Fixes #926 

I feel like trying to extract all function bindgen info in one pass has led to an error-prone code that is hard to reason about. In this PR I am refactoring it into two parts:
1. Deciding what kind of function this is (init, view, call) by doing a pass through arguments and attributes
2. Traversing all relevant parts of the function signature and calling the respective visitor methods

I am not super happy with how the polymorphic aspect of this approach turned out to be (`Box<dyn BindgenVisitor>` is unmovable, so `build()` has to clone visitor field values to the end-structure), so feel free to suggest an alternative.

The bindgen models have also been changed to only contain the values that are relevant to them (e.g., init function does not have `result_serializer`). This is technically a breaking change since it restricts some combinations that were previously compilable, but I think none of them ever made sense.

NOTE: I left the old models in place and wrote a conversion method so that the PR is reasonably sized. In a follow-up PR I will clean up the codegen side of bindgen macro, which will benefit greatly from this refactoring.

Unresolved questions:
1. Should init functions be allowed to be private?
2. Should view functions be allowed to be private? Wouldn't such a function only be callable from a transaction?

Will add some compilation tests to ensure no regression happens after we clear up the question above.


austinabell:
> Should init functions be allowed to be private?

Likely not, there is a valid pattern (a pure function calling itself in a promise to initialize the contract) but I can't think of one reason why this would ever be needed. I think it's safe to remove that ability and then just add the capability back if devs ask for it.

> Should view functions be allowed to be private? Wouldn't such a function only be callable from a transaction?

No, even calling `env::predecessor_account_id` fails in view contexts at runtime. This should be restricted at compile time. Good callout!

austinabell:
Might want to double-check https://github.com/near/near-sdk-rs/issues/718 is fixed by this (when you get to it)

itegulov:
> Might want to double-check https://github.com/near/near-sdk-rs/issues/718 is fixed by this (when you get to it)

It did not, but now it does :)

near-sdk-macros/src/core_impl/info_extractor/attr_sig_info.rs
```diff
@@ -32,8 +52,82 @@ pub struct AttrSigInfo {
...
+    fn is_view(sig: &Signature) -> bool {
+        let receiver_opt = sig.inputs.iter().find_map(|arg| match arg {
+            FnArg::Receiver(r) => Some(r),
+            _ => None,
+        });
+
+        match receiver_opt {
+            Some(receiver) => receiver.reference.is_none() || receiver.mutability.is_none(),
+            None => true,
+        }
```
miraclx, line 128: +1 treating self-less functions as view. Helps resolve function kind ambiguities when parsing ABI.

miraclx (approved):

near-sdk-macros/src/core_impl/info_extractor/visitor/view_visitor.rs
```diff
@@ -0,0 +1,75 @@
...
+            }
+            ReturnType::Type(_, typ) => Some(Returns {
+                original: return_type.clone(),
+                kind: super::parse_return_kind(typ, handle_result)?,
+            }),
+        };
+        Ok(())
+    }
+
+    fn build(&self) -> syn::Result<MethodKind> {
```
ChaoticTempest, line 64: nit: why not consume the builder via `build(self)` instead just incase you reuse the builder? That way you don't have to call clone and just claim the builder's items.

near-sdk-macros/src/core_impl/info_extractor/impl_item_method_info.rs
```diff
@@ -28,3 +28,46 @@ impl ImplItemMethodInfo {
...
+
+// Rustfmt removes comas.
+#[rustfmt::skip]
+#[cfg(test)]
+mod tests {
+    use syn::{parse_quote, Type, ImplItemMethod};
+    use crate::core_impl::ImplItemMethodInfo;
+
+    #[test]
+    fn init_no_return() {
```
ChaoticTempest, line 40: are we supporting an `init` function that returns a `Result`? if so, maybe another test to test here would also be worthwhile

near-sdk-macros/src/core_impl/info_extractor/attr_sig_info.rs
```diff
@@ -62,103 +169,70 @@ impl AttrSigInfo {
...
                 }
                 "result_serializer" => {
                     let serializer: SerializerAttr = syn::parse2(attr.tokens.clone())?;
-                    result_serializer = serializer.serializer_type;
+                    visitor.visit_result_serializer_attr(attr, &serializer)?;
                 }
                 "handle_result" => {
-                    is_handles_result = true;
+                    handles_result = true;
                 }
```
ChaoticTempest, line 205: Don't we need `vistor.visit_result` here as well?

near-sdk-macros/src/core_impl/info_extractor/visitor/view_visitor.rs
itegulov: Yeah I mentioned this in the original post:

> I am not super happy with how the polymorphic aspect of this approach turned out to be (Box<dyn BindgenVisitor> is unmovable, so build() has to clone visitor field values to the end-structure), so feel free to suggest an alternative.

Basically, there is no way to unwrap `Box` because I can't prove that there is only ever a single instance of it.

near-sdk-macros/src/core_impl/info_extractor/attr_sig_info.rs
itegulov: Nah, we call `visitor.visit_result` below (and pass `handles_result` there) so that it is called both when `handle_result` is present and when it is not. If I move it here, then it won't be called for any of the regular methods.

near-sdk-macros/src/core_impl/info_extractor/visitor/view_visitor.rs
ChaoticTempest: ahh I see, the trait object. You could also do `build(self: Box<Self>)` for the type signature for now, since this is just an internal trait

PR #891, opened 2022-08-17
impl NearSchema derive macro

miraclx:
```rust
use near_sdk::NearSchema;

#[derive(NearSchema)]
#[abi(json, borsh)]
struct Value {
    field: InnerValue
}
```

Effectively translates into

```rust
#[derive(schemars::JsonSchema, borsh::BorshSchema)]
struct Value {
    field: InnerValue
}
```

When compiled targeting WASM, this is a no-op.


miraclx:
One question though. Do we want to proxy attr macros like `#[schemars]`, `#[serde]`, `#[validate]` for schemars, and `#[borsh_skip]` for borsh?

itegulov:
Hmm, does this mean that all inner values have to implement *both* `JsonSchema` and `BorshSchema` regardless of which one is actually needed? This is one of the things I could not implement when I tried to do this: something like "if there is a derivation for `BorshSerialize` & `BorshDeserialize` then derive `BorshSchema`" (and also analogously for `JsonSchema`). Do you think this would be impossible to achieve?

itegulov:
> One question though. Do we want to proxy attr macros like #[schemars], #[serde], #[validate] for schemars, and #[borsh_skip] for borsh?

schemars actually respects many `serde` attributes when deriving a schema, so yeah that would be nice. Same goes for borsh, it checks for `borsh_skip` when deriving `BorshSchema`.

miraclx:
> "if there is a derivation for BorshSerialize & BorshDeserialize then derive BorshSchema" (and also analogously for JsonSchema). Do you think this would be impossible to achieve?

using a derive macro, unfortunately yeah.

Best we can do is leave `BorshSchema` out of `NearAbi` since `borsh` is already reexported via `near_sdk::borsh` so devs can manually implement `BorshSchema` on the types they want.

itegulov:
Maybe we could derive JsonSchema by default and add something like this for borsh? 
```rust
#[derive(NearAbi)]
#[abi(borsh)] // makes NearAbi derive `BorshSchema` instead of `JsonSchema`
struct Value {
    field: InnerValue
}
```

Does not seem like there is a great solution for this either way.

austinabell:
I agree with Daniyar, both should not be expected. Would require all inner types to have both implemented. Unclear if it would be better to have one derive with attributes or just two different derives. My intuition is that the latter seems more natural, but that is very weakly held.

The issue with your example, Daniyar, is that it limits the ability to derive _both_ types. Yes, a very niche use case, but might be a bit of a footgun

miraclx:
> limits the ability to derive both types

I actually think this might be possible, tested a small example, and it seems to work. Here's the syntax I'm working with:

```rust
#[derive(NearAbi)]
#[abi(json, borsh)]
struct Value {
    field: InnerValue
}
```

And omitting the attribute entirely defaults to `JsonSchema`.

I'll go ahead to implement this, I'll report here if I hit any blockers.

An alternative to the attribute syntax might be (if you prefer being explicit):

```rust
#[derive(NearAbi)]
#[abi(schema = "json")]
#[abi(schema = "borsh")]
struct Value {
    field: InnerValue
}
```

miraclx:
Done. So, I went with the first syntax.

```rust
#[derive(NearAbi)]
#[abi(json, borsh)]
struct Value {
    field: InnerValue
}
```

- Defaults to implement `JsonSchema` when the attribute is missing.
- `#[abi(borsh)]` implements `BorshSchema`.
- `#[abi(json, borsh)]` implements both.

This already proxies `schemars`, `serde` attributes on the item. I'll follow this with proxying field and variant attributes as well as a test suite for guaranteeing stability.

In the meantime, let me know what you think.

near-sdk-macros/src/lib.rs
```diff
@@ -213,6 +213,140 @@ pub fn metadata(item: TokenStream) -> TokenStream {
     }
 }
 
+#[proc_macro_derive(NearAbi, attributes(abi, serde, borsh_skip, schemars, validate))]
```
austinabell, line 216: The only thing about this name is that it doesn't fit in with naming of traits. Generally, they are named as verbs, so `NearAbi` might be a bit confusing, especially since it's exported from the SDK root that it might be assumed to be the Abi schema type.

@itegulov do you have any ideas for naming? I would be fine with this, just want to see if there could be something better/more clear.

near-sdk-macros/src/lib.rs
```diff
@@ -213,6 +213,140 @@ pub fn metadata(item: TokenStream) -> TokenStream {
...
+pub fn derive_near_abi(_input: TokenStream) -> TokenStream {
+    // todo! borsh & serde don't support unions
+    // todo! guess what? we don't either
+    #[cfg(not(feature = "abi"))]
+    return TokenStream::new();
+    #[cfg(feature = "abi")]
+    {
```
austinabell, line 223: This functionality seems like it shouldn't be behind a feature, the attribute should just be gated behind the feature flag. Definitely a foot gun that a developer adds this and it does a no-op if they don't realize they didn't add the `abi` feature?

Or is the reason for this that there will be a plan to add the `abi` feature to the dependency automatically in `cargo-near`?

Either way, it seems like if the feature isn't set, it should maybe still implement `BorshSchema` and `Schemafy`, just not expose the `__near_abi` functions.

I don't have a high confidence in any of these suggestions, @itegulov definitely has the most context on this

near-sdk-macros/src/lib.rs
miraclx: `NearSchema` maybe? Fits more appropriately with the traits it aliases.

near-sdk-macros/src/lib.rs
miraclx: > Or is the reason for this that there will be a plan to add the abi feature to the dependency automatically in cargo-near?

Yeah, so I mean, unless there's a reason I'm missing for contracts having to manually specify the `abi` feature. I think it's only useful in the context of `cargo-near`, so we can have that automatically activate the feature.

> Definitely a foot gun that a developer adds this and it does a no-op if they don't realize they didn't add the abi feature?

If the developer's intention for deriving adding the trait is to generate abi, they have to use `cargo-near` and with the current state of `cargo-near`, it already won't let you do anything without the `abi` feature.

But rethinking this in line with your other commend, we should just rename the trait and leave it un-feature-gated.

austinabell: > Yeah, so I mean, unless there's a reason I'm missing for contracts having to manually specify the `abi` feature. I think it's only useful in the context of `cargo-near`, so we can have that automatically activate the feature.

Note that this is not currently how `cargo-near` does it

> If the developer's intention for deriving the trait is to generate abi, they have to use `cargo-near,` and with the current state of `cargo-near`, it already won't let you do anything without the `abi` feature.
> 
> But rethinking this in line with your other comment, we should just rename the trait and leave it un-feature-gated.

The only thing here is that it would be helpful to show errors for why it can't be derived without using the `abi` feature. For example, if a sub-type that's used in the struct does not implement the relevant trait, there would be no error until someone ran a command through `cargo-near`, which would be harder to fix.

miraclx: > Note that this is not currently how cargo-near does it

Yeah, I'm saying it's how it _could_ do it.

> there would be no error until someone ran a command through cargo-near, which would be harder to fix.

Good point.

miraclx: So we're to get to the point where;

1. contracts don't need to manually specify the "abi" feature flag
2. `DeriveName` is not feature gated behind the "abi" feature flag, but its implementors are only injected when the target isn't wasm.

This would ensure `DeriveName` can be debugged without `cargo-near`.

near-sdk-macros/src/lib.rs
itegulov: Hmm, I know I was the one who suggested `NearAbi` with `#[abi(json)]`, but now I kind of feel like two separate `NearJsonSchema` & `NearBorshSchema` trait would feel more natural. Otherwise yeah `NearSchema` seems appropriate. Sorry for re-raising this point, definitely not a big deal either way.

near-sdk-macros/src/lib.rs
itegulov: My idea for "abi" feature flag is that it will be gone once we stabilize the ABI schema and devx around it in general. So yeah @miraclx your last comment describes how I see it perfectly.

> but its implementors are only injected when the target isn't wasm.

Side note, but we could even create a custom rustc target for ABI so that it does not clash with unit tests. This would also provide a uniform experience on all user platforms.

itegulov:
Also would be cool to add some use cases of this macro to `compilation_tests`.

near-sdk-macros/src/lib.rs
```diff
@@ -213,6 +213,140 @@ pub fn metadata(item: TokenStream) -> TokenStream {
     }
 }
 
+#[proc_macro_derive(NearAbi, attributes(abi, serde, borsh_skip, schemars, validate))]
+pub fn derive_near_abi(_input: TokenStream) -> TokenStream {
+    // todo! borsh & serde don't support unions
+    // todo! guess what? we don't either
```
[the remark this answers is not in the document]
miraclx: Yeah, I meant, since neither `borsh` nor `serde` (via `schemars`) support unions, we need to make sure to return a neat error saying unions are not supported for the `NearSchema` derive macro too.

near-sdk-macros/src/lib.rs
```diff
@@ -213,6 +213,140 @@ pub fn metadata(item: TokenStream) -> TokenStream {
...
+                {
+                    type_attrs.push(attr)
+                }
+                _ => continue,
+            }
+        }
+        input.attrs = type_attrs;
+
+        // todo! proxy field and variant attributes
+        // todo! serde, validate, borsh_skip
```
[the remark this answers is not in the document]
miraclx: > Haven't you handled serde already or do you mean something else?

We have for the item attribute. But not yet for the field or variant attributes.

```rust
#[derive(NearSchema)]
#[serde(...)] // <-- we're proxying these
pub struct Value {
    #[serde(...)]  // <-- but not these, not yet
    field: InnerValue
}
```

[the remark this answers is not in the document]
miraclx: > > validate
> 
> What do you mean by this?

Same as above, `schemars` uses the `#[validate]` attribute macro. We need to proxy that too.

near-sdk-macros/src/lib.rs
```diff
@@ -213,6 +213,140 @@ pub fn metadata(item: TokenStream) -> TokenStream {
...
+                    for meta in meta.nested {
+                        match meta {
+                            NestedMeta::Meta(m) if m.path().is_ident("json") => schema |= 0b01,
+                            NestedMeta::Meta(m) if m.path().is_ident("borsh") => schema |= 0b10,
+                            _ => {
+                                return TokenStream::from(
+                                    syn::Error::new_spanned(
+                                        meta.into_token_stream(),
+                                        format!(
+                                            "invalid abi argument, expected: `json` or `borsh`",
```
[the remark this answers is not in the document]
miraclx: > also error messages should be capitalized

Was just following convention in crates like `serde` – https://github.com/serde-rs/serde/blob/dc0c0dcba17dd8732cd8721a7ef556afcb04c6c0/serde_derive/src/ser.rs#L71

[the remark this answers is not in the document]
itegulov: Hmm, that is actually a good point. rustc errors are lowercase, not sure why ours are all uppercase.

near-sdk-macros/src/lib.rs
miraclx: > This would be difficult to achieve btw, as it will require either modifying the user's Cargo.toml file

Not really, `cargo build --features near-sdk/abi` works to the same effect as `near-sdk = { ..., features = ["abi"]}`

Although, I'm taking a new approach now. If you take a look at https://github.com/near/cargo-near/pull/31, what I resolved to do instead is:

- Contracts manually specify the `abi` feature flag if they want to generate ABI.
   - This gives them access to the `NearSchema` trait that they can derive on their structures. 
- `cargo-near` rejects all contracts that don't have the `abi` feature flag.
- `cargo-near` activates the `__abi-generate` feature flag when generating ABI.
- `cargo-near` activates the `__abi-embed` feature flag when embedding ABI.

(leading underscores is to ensure docs.rs hides those features, and leaving the hyphen inconsistency in there to signal contract devs not to use it)

itegulov: > Not really, cargo build --features near-sdk/abi works to the same effect as near-sdk = { ..., features = ["abi"]}

lol, how did I not think of that

> Although, I'm taking a new approach now. If you take a look at https://github.com/near/cargo-near/pull/31

Interesting, I will definitely take a look

near-sdk-macros/src/lib.rs
```diff
@@ -207,6 +209,174 @@ pub fn metadata(item: TokenStream) -> TokenStream {
...
+                if meta.nested.is_empty() {
+                    return TokenStream::from(
+                        syn::Error::new_spanned(
+                            meta.into_token_stream(),
+                            "attribute requires at least one argument",
+                        )
+                        .to_compile_error(),
+                    );
+                }
+                json_schema = false;
```
austinabell, line 243: why is this needed? Seems like a strange condition. Can you add a comment on why if it is?

near-sdk-macros/src/lib.rs
```diff
@@ -207,6 +209,174 @@ pub fn metadata(item: TokenStream) -> TokenStream {
...
+            }
+        }
+    } else {
+        quote! {}
+    };
+
+    TokenStream::from(quote! {
+        #[cfg(not(target_arch = "wasm32"))]
+        const _: () = {
+            mod __near_abi_private {
```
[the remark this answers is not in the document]
miraclx: `NearSchema`'s expansion effectively adds derives to a clone of the structure.

```rust
#[derive(NearSchema)]
struct Value {
    field: InnerValue
}
```

expands into

```rust
struct Value {
    field: InnerValue
}

const _: () = {
    mod __near_abi_private {
        #[derive(schemars::JsonSchema)]
        struct Value {
            field: InnerValue
        }
        
        impl schemars::JsonSchema for super::Value { ... }
        //                            ^^^^^
        // disambiguates the cloned structure from the original one
    }
};
```

near-sdk-macros/src/lib.rs
miraclx: Will document, but for here, it's because omitting the `#[abi]` spec defaults to deriving only `JsonSchema`. `#[abi(borsh)]` derives only `BorshSchema`. And `#[abi(json, borsh)]` derives both.

near-sdk-macros/src/lib.rs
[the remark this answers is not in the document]
austinabell: oh, I see, so you want to avoid the type name being something different from the original for the purposes of schemars, which uses the type name or something?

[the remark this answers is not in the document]
miraclx: Yeah, borsh also uses the type name.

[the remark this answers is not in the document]
miraclx: Bump on this. `super::*` requires the type to exist in a module. And would fail in this case.

```rust
mod x {
    fn a() {
        #[derive(NearSchema)]
        pub struct MyType;
    }
}
```

`super::MyType` from the injected `__near_abi_private` module would reference `x::MyType`, which is nonexistent.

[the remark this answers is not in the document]
miraclx: Resolved to expanding to this instead:

```rust
struct Value {
    field: InnerValue
}

const _: () = {
    type Proxy = Value;
    {
        #[derive(schemars::JsonSchema)]
        struct Value {
            field: InnerValue
        }
        
        impl schemars::JsonSchema for Proxy { ... }
    }
};
```

near-sdk/src/lib.rs
```diff
@@ -5,6 +5,8 @@
 #[cfg(test)]
 extern crate quickcheck;
 
+#[cfg(feature = "abi")]
```
austinabell, line 8: can we put this export under unstable as well for now? Doesn't seem clear this will be more intuitive than deriving the traits directly yet and the internal semantics aren't tested yet.

austinabell (approved):

Sdsfghy:
هلا

PR #895, opened 2022-08-24
CI: treat compiler warnings as errors

miraclx:
Context: https://github.com/near/near-sdk-rs/pull/815#discussion_r953325437


itegulov:
We are not doing `cargo check` without `unstable` feature, so this wouldn't catch the warning from the original comment anyway, right?

miraclx:
Yes, but compiling SDK without said feature always output a warning, even in downstream dependents that don't care about unstable features.

.github/workflows/test.yml
```diff
@@ -22,15 +22,19 @@ jobs:
       - uses: Swatinem/rust-cache@v1
       - run: rustup target add wasm32-unknown-unknown
       - name: Test
-        run: cargo test --all --features unstable
+        env:
+          RUSTFLAGS: -D warnings
```
matklad, line 26: I'd put this to top-level, seems right to just blanket-apply it everywhere:

https://github.com/rust-lang/rust-analyzer/blob/master/.github/workflows/ci.yaml#L12-L18

austinabell (approved):

PR #896, opened 2022-08-24
chore: update docs for `Vector` and `store` collections

austinabell:
Fixes some comments through this review (https://github.com/near/near-sdk-rs/pull/815#pullrequestreview-1083761194) and some others I think that can be improved.

~~The only functional change is the change of the field name of the storage prefix in `LazyOption`. unclear why this one was set as different. Shouldn't break any APIs, but changes the debug impl~~

### DIFF
--- a/near-sdk/src/store/unordered_set/mod.rs
+++ b/near-sdk/src/store/unordered_set/mod.rs
@@ -125,9 +125,9 @@ where
     /// # Examples
     ///
     /// ```
-    /// use near_sdk::store::UnorderedMap;
+    /// use near_sdk::store::UnorderedSet;
     ///
-    /// let mut map: UnorderedMap<String, u8> = UnorderedMap::new(b"b");
+    /// let mut map: UnorderedSet<String> = UnorderedSet::new(b"b");
     /// ```
     #[inline]
     pub fn new<S>(prefix: S) -> Self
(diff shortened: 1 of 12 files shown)


near-sdk/src/store/unordered_set/mod.rs
```diff
@@ -117,6 +117,18 @@ impl<T> UnorderedSet<T, Sha256>
...
+    ///
+    /// This prefix can be anything that implements [`IntoStorageKey`]. The prefix is used when
+    /// storing and looking up values in storage to ensure no collisions with other collections.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use near_sdk::store::UnorderedMap;
+    ///
+    /// let mut map: UnorderedMap<String, u8> = UnorderedMap::new(b"b");
```
ChaoticTempest, line 130: should be UnorderedSet instead of Map
```suggestion
    /// use near_sdk::store::UnorderedSet;
    ///
    /// let mut map: UnorderedSet<String> = UnorderedSet::new(b"b");
```

ChaoticTempest (approved):
Just a couple documentation errors I noticed

austinabell:
> Just a couple documentation errors I noticed

oops! those were careless :D

PR #897, opened 2022-08-24
fix: Make Lazy and LazyOption debug impls match

austinabell:
Noticed when working on related things that these were not matching. This should be resolved before #888 comes in, but is a separate issue

What this proposes is including the cached value in debug output but keeping the type as an `Option` so that we can change the implementation without breaking this format. Can alternatively just keep the default debug derive or ignore cache, but the cache seems important for debug purposes. (Maybe not obfuscating the cache entry type is valuable to see if a value has been flushed to storage)?

### DIFF
--- a/near-sdk/src/store/lazy/impls.rs
+++ b/near-sdk/src/store/lazy/impls.rs
@@ -100,7 +100,7 @@ where
         } else {
             f.debug_struct("Lazy")
                 .field("storage_key", &self.storage_key)
-                .field("cached_value", &self.cache.get().and_then(|v| v.value().as_ref()))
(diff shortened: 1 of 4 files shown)


ChaoticTempest:
hmm, maybe adding an extra field where we show whether it has been flushed or not based on the `EntryState` would also be nice? Like:
```
cached { flushed: true, value: Some(8) }
```

austinabell:
Just debug output the `CacheEntry` type. Not worth adding code size for a new type with debug

ChaoticTempest (approved):

itegulov (approved):

PR #904, opened 2022-08-29
feat: add JsonSchema derives to NFT exposed types

austinabell:
needed for building an ABI with the NFT example


itegulov (approved):
Meant to do this + BorshSchema after #891 comes in, but not sure if we have time to wait much longer. Thanks for handling this

PR #906, opened 2022-08-30
update test case to include decompression step

miraclx:
Since the abi file output is now prefixed with the crate name, it's a little weird to have `abi_abi.json`, lol.

But let me know if you think we should leave it as-is.


itegulov (approved):
> Since the abi file output is now prefixed with the crate name, it's a little weird to have abi_abi.json, lol.

Yeah honestly I am not sure of the usefulness of this example long-term. Once we enable `abi` for `nft`, `ft` and other examples this is probably going to be redundant. And then any advanced examples of `abi` should live in `near-sdki-abi`.

austinabell (approved):

austinabell:
The CI is failing due to the folder name being `abi` and the size check is assuming the same name for the wasm file. This doesn't change the SDK so I'll just release before this comes in, unless you want to update it soon :)

miraclx:
Ah, right. Fixing.

PR #909, opened 2022-08-30
fix: cargo check with all features

miraclx:
Before

```console
$ cargo check --all-features
error: environment variable `CARGO_NEAR_ABI_PATH` not defined
 --> near-sdk-macros/src/core_impl/abi/abi_embed.rs:5:20
  |
5 |     let abi_path = env!("CARGO_NEAR_ABI_PATH");
  |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: this error originates in the macro `env` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `near-sdk-macros` due to previous error
```

Now, the only time this error comes up is when all three of these conditions are true:

1. You're building something that invokes the `#[near_bindgen]` attribute.
2. You have the `near-sdk/__abi-embed` feature flag activated.
3. The `CARGO_NEAR_ABI_PATH` env is unset.

And as far as the flow with `cargo near` is well-defined, this should never happen.

So, you really have to go looking to find this error. Or, here:

```console
$ cd near-sdk/examples/abi
$ cargo build --features near-sdk/__abi-embed
error: the `__abi-embed` feature flag is private and should not be activated manually

       help: consider using https://github.com/near/cargo-near
  --> src/lib.rs:15:1
   |
15 | #[near_bindgen]
   | ^^^^^^^^^^^^^^^
   |
   = note: this error originates in the attribute macro `near_bindgen` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `abi` due to previous error
```

### DIFF
--- a/near-sdk-macros/src/core_impl/abi/abi_embed.rs
+++ b/near-sdk-macros/src/core_impl/abi/abi_embed.rs
@@ -1,15 +1,17 @@
-use proc_macro2::{Span, TokenStream as TokenStream2};
+use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
 
 pub fn embed() -> TokenStream2 {
     let abi_path = match option_env!("CARGO_NEAR_ABI_PATH") {
         Some(path) => path,
         None => {
-            return syn::Error::new(
-                Span::call_site(),
-                "CARGO_NEAR_ABI_PATH environment variable is not set",
-            )
-            .to_compile_error()
+            return quote! {
+                compile_error!(
+                    "the `__abi-embed` feature flag is private and should not be activated manually\n\
+                    \n\
+                    help\x1b[0m: consider installing https://github.com/near/cargo-near"
+                );
+            };
         }
     };
     quote! {
--- a/near-sdk-macros/src/core_impl/abi/abi_embed.rs
+++ b/near-sdk-macros/src/core_impl/abi/abi_embed.rs
@@ -1,8 +1,17 @@
-use proc_macro2::TokenStream as TokenStream2;
+use proc_macro2::{Span, TokenStream as TokenStream2};
 use quote::quote;
 
 pub fn embed() -> TokenStream2 {
-    let abi_path = env!("CARGO_NEAR_ABI_PATH");
+    let abi_path = match option_env!("CARGO_NEAR_ABI_PATH") {
+        Some(path) => path,
+        None => {
+            return syn::Error::new(
+                Span::call_site(),
+                "CARGO_NEAR_ABI_PATH environment variable is not set",
+            )
+            .to_compile_error()
+        }
+    };
     quote! {
         const _: () = {
             const __CONTRACT_ABI: &'static [u8] = include_bytes!(#abi_path);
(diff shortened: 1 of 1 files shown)


near-sdk-macros/src/core_impl/abi/abi_embed.rs
```diff
@@ -1,8 +1,17 @@
...
 use quote::quote;
 
 pub fn embed() -> TokenStream2 {
-    let abi_path = env!("CARGO_NEAR_ABI_PATH");
+    let abi_path = match option_env!("CARGO_NEAR_ABI_PATH") {
+        Some(path) => path,
+        None => {
+            return syn::Error::new(
+                Span::call_site(),
+                "CARGO_NEAR_ABI_PATH environment variable is not set",
```
austinabell, line 10: Can you improve this error so it indicates that the `__abi-embed` feature should not be set, and is only used through cargo-near? This error isn't any more descriptive than the previous one

miraclx: Was just about asking for error message suggestions. What do you think?

- <code>\`__abi-embed\` feature flag should not be set manually, use \`cargo near\`</code>
- <code>invalid use of private feature flag \`__abi-embed\`, please remove this feature</code>

And if we're mentioning the use of `cargo near`, should we add a link to it?

austinabell: up to you on that, either sounds great! And yeah, I think a link to it would be useful

miraclx: Went through a couple of error message iterations and settled on this:

CleanShot 2022-08-30 at 20 59 56@2x

Would've loved to add the hint into the notes to more seamlessly blend in with cargo / rustc style error reports but the interface for that is yet to be stabilized. https://github.com/rust-lang/rust/issues/54140

near-sdk-macros/src/core_impl/abi/abi_embed.rs
```diff
@@ -2,7 +2,18 @@ use proc_macro2::TokenStream as TokenStream2;
...
 pub fn embed() -> TokenStream2 {
-    let abi_path = env!("CARGO_NEAR_ABI_PATH");
+    let abi_path = match option_env!("CARGO_NEAR_ABI_PATH") {
+        Some(path) => path,
+        None => {
+            return quote! {
+                compile_error!(
+                    "the `__abi-embed` feature flag is private and should not be activated manually\n\
+                    \n\
+                    help\x1b[0m: consider installing https://github.com/near/cargo-near"
```
austinabell, line 12:
```suggestion
                    help\x1b[0m: consider using https://github.com/near/cargo-near"
```

this wording may indicate that all one needs to do is install cargo-near and it will compile with `cargo check`, which isn't true. wdyt?

austinabell (approved):

near-sdk-macros/src/core_impl/abi/abi_embed.rs
miraclx: lgtm.

PR #913, opened 2022-09-01
chore: remove rlib from examples

itegulov:
Seems like we have missed a couple of them


austinabell (approved):
Ah, yes, they were here when the sim tests hadn't been migrated over. Good catch

PR #916, opened 2022-09-06
feat: Default optional input fields of (N)FT metadata

austinabell:
Not a breaking change, just doesn't require all fields with `null` added in the JSON. Simplifies the devX and reduces input size necessary.

Unsure why this would be intended, JSON is already not canonical so shouldn't matter the ambiguity here afaik

@mattlockyer maybe you might have some context or reason why this shouldn't be done, though

### DIFF
--- a/near-contract-standards/src/fungible_token/metadata.rs
+++ b/near-contract-standards/src/fungible_token/metadata.rs
@@ -12,8 +12,11 @@ pub struct FungibleTokenMetadata {
     pub spec: String,
     pub name: String,
     pub symbol: String,
+    #[serde(default)]
     pub icon: Option<String>,
+    #[serde(default)]
     pub reference: Option<String>,
+    #[serde(default)]
     pub reference_hash: Option<Base64VecU8>,
     pub decimals: u8,
 }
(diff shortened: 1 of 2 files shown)


ChaoticTempest:
hmm, this could still be a breaking change if people are directly comparing those fields to `null`. Since now when referencing something like `nft_meta.icon`, it would yield `undefined` instead. I'm no expert on how these (N)FT metadata gets used though, so this is just my perspective from using it via `near-api-js` or from `workspaces-rs`

austinabell:
> hmm, this could still be a breaking change if people are directly comparing those fields to `null`. Since now when referencing something like `nft_meta.icon`, it would yield `undefined` instead. I'm no expert on how these (N)FT metadata gets used though, so this is just my perspective from using it via `near-api-js` or from `workspaces-rs`

no, because the data still gets serialized as null with default, it just allows it to be omitted for input. For serialization/output, https://serde.rs/field-attrs.html#skip_serializing_if is used

ChaoticTempest:
ahh, I see. Then looks good to me then

ChaoticTempest (approved):

austinabell:
I was wrong! This is [done automatically](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0a820dcea4b90f494b16fc4f835c63e5)

no need for this

PR #917, opened 2022-09-16
fix: optimize new set implementation and update docs

austinabell:
There is no benefit to the cache in the set's case, as no APIs depend on it, and since no data loaded, the cache adds more overhead than benefit. This change also removes the `Ord` and `Clone` requirements on the set's keys.

Benchmarking these changes are ~15% decrease in gas for a small number of elements, and ~25% for larger amounts of modifications


near-sdk/src/store/lookup_set/mod.rs
```diff
@@ -1,54 +1,64 @@
...
 use borsh::{BorshDeserialize, BorshSerialize};
-use once_cell::unsync::OnceCell;
 use std::borrow::Borrow;
 use std::fmt;
 use std::marker::PhantomData;
 
+/// A non-iterable implementation of a set that stores its content directly on the storage trie.
+///
+/// This set stores the values under a hash of the set's `prefix` and [`BorshSerialize`] of the
+/// value and transformed using the map's [`ToKey`] implementation.
```
itegulov, line 13: I think you meant to say "key" here

```suggestion
/// This set stores the values under a hash of the set's `prefix` and [`BorshSerialize`] of the
/// key and transformed using the map's [`ToKey`] implementation.
```

itegulov (approved):
> Benchmarking these changes are ~15% decrease in gas for a small number of elements, and ~25% for larger amounts of modifications

Interesting, would you mind sharing the benchmark source code?

In any case, LGTM and sounds likely that the caching overhead outweighs the benefits from what I remember from the benchmarks I did back in January.

near-sdk/src/store/lookup_set/mod.rs
itegulov: Oh, this is a set, ignore me :see_no_evil:

austinabell:
> Interesting, would you mind sharing the benchmark source code?

Yeah, was hacked together quickly, and didn't update all the values or point to a git dep, but here it is: https://github.com/austinabell/storage_collections_bench/commit/b3864e370f256541eb1010a858748e17dd16d242

PR #923, opened 2022-09-20
chore: put legacy collections under feature flag

### LINKED ISSUE #869: Put old collections under `legacy` feature
This needs to be done after #818 but the old collections should be put under a feature, "legacy", and be default for now, with the intentions of the feature not being set by default in the next major version.

Perhaps other APIs should go in this bucket of things we need to keep support for and don't want to deprecate, but should be avoided from base API

austinabell:
closes #869

This doesn't functionally change anything because it's enabled by default, but allows devs to remove this section, and this will be deprecated in the next major version

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,7 @@ jobs:
       - uses: Swatinem/rust-cache@v1
       - run: rustup target add wasm32-unknown-unknown
       - name: Test
-        run: cargo test --all --features unstable --features legacy
+        run: cargo test --all --features unstable,legacy
   lint:
     name: Clippy and fmt
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - uses: Swatinem/rust-cache@v1
       - run: rustup target add wasm32-unknown-unknown
       - name: Test
-        run: cargo test --all --features unstable
+        run: cargo test --all --features unstable --features legacy
```
ChaoticTempest, line 27: optional: it's kinda weird that `--features` can be supplied multiple times like this. Normally I just put `--features unstable,legacy`, but up to you

ChaoticTempest (approved):
hmm, that one test failure for `simulate_simple_approve` seems weird, but maybe bringing up at least 16 sandbox nodes at the same time could make it very slow. Did you ever try upping the timeout seconds?

PR #925, opened 2022-09-20
chore: update readme to include versioning and fixes

### LINKED ISSUE #246: Promote documented features to heading level 3 so they can be deep linked
when trying to reference specific features of `near-sdk-rs` it's impossible to deep link directly to the section of the `README.md` that describes the feature since they are not formatted as headings. 

one solution would be to make each feature it's own heading level 3 (or similar) to retain current document format but allow linking directly to specific features like the `#<span class="error">[payable]</span>`, `#<span class="error">[init]</span>` and `#<span class="error">[private]</span>` macros from other locations.

### LINKED ISSUE #384: Document API evolution strategy
Uplifting [https://github.com/near/near-sdk-rs/pull/366#issuecomment-823984897](https://github.com/near/near-sdk-rs/pull/366#issuecomment-823984897) to an issue:

>This is going to break legacy code. log! serves the purpose.

Do we have our backwards compatibility policy for SDK documented anywhere? Taking a bird's eye view, the current public API of SDK feels like it has a lot of rough edges from Rust API design point of view, and it seems like long term smoothing those out would be beneficial. However, it's unclear to me at least what is the best path towards that.

I can see several options for this particular change, and for this class of changes:

*   do nothing and consider the API frozen / append only
*   keep the current major version, but use deprecation to slowly clean-up the API (so, apply #<span class="error">[deprecated]</span> & #<span class="error">[doc(hidden)]</span> attributes to the log function, introduce `log_str`, re-implement `log` in terms of `log_str`)
*   use release trains: release a point release which deprecates `log` function now, and release a new semver-incompatible version not-to-far in the future, which adds log with the correct signature.
*   prepare for "big-bang" style next major version (don't implement this right now, but add a BreakingChange label and release a new version when we accumulate a bunch of those).
*   re-design the API from the first principles, using the lessons learned, but not the code, from the existing sdk (aka "don't tell Joel Spolsky")

Neither of them is inherently better than the other (even the last one can be beneficial in the exceptional circumstances), but it's unclear which one we use.

### LINKED ISSUE #405: Document MSRV policy
We should document which rust version is supported by the nearsdk. It seems like we shouldn't overthink this and just say "latest stable", but it still is important to actually state that for users of the sdk to have correct expectations.

austinabell:
There were a few issues to be closed relating to the README, so I figured I'd just do them.

Rendered https://github.com/near/near-sdk-rs/blob/austin/readme-update/README.md
closes #405 
closes #384
closes #246 

- Update badge to gh actions (was bugging me that buildkite badge was borked)
- Fix links
- Include build information, linking to cargo-near

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
     <a href="https://crates.io/crates/near-sdk"><img src="https://img.shields.io/crates/d/near-sdk.svg?style=flat-square" alt="Download" /></a>
     <a href="https://docs.rs/near-sdk"><img src="https://docs.rs/near-sdk/badge.svg" alt="Reference Documentation" /></a>
     <a href="https://discord.gg/gBtUFKR"><img src="https://img.shields.io/discord/490367152054992913.svg" alt="Join the community on Discord" /></a>
-    <a href="https://buildkite.com/nearprotocol/near-sdk-rs"><img src="https://badge.buildkite.com/3bdfe06edbbfe67700833f865fe573b9ac6db517392bfc97dc.svg" alt="Buildkite Build" /></a>
+    <a href="https://github.com/near/near-sdk-rs/actions"><img src="https://github.com/near/near-sdk-rs/actions/workflows/test.yml/badge.svg" alt="GitHub Actions Build" /></a>
   </p>
 
    <h3>
@@ -64,13 +64,12 @@ impl StatusMessage {
 
 ## Features
 
-* **Unit-testable.** Writing unit tests is easy with `near-sdk`:
+### **Unit-testable.** 
+Writing unit tests is easy with `near-sdk`:
 
     ```rust
     #[test]
     fn set_get_message() {
-        let context = get_context(vec![]);
-        testing_env!(context);
         let mut contract = StatusMessage::default();
         contract.set_status("hello".to_string());
         assert_eq!("hello".to_string(), contract.get_status("bob_near".to_string()).unwrap());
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -64,13 +64,12 @@ impl StatusMessage {
...
+Writing unit tests is easy with `near-sdk`:
 
     ```rust
     #[test]
     fn set_get_message() {
-        let context = get_context(vec![]);
-        testing_env!(context);
         let mut contract = StatusMessage::default();
         contract.set_status("hello".to_string());
         assert_eq!("hello".to_string(), contract.get_status("bob_near".to_string()).unwrap());
```
itegulov, line 75: This is no longer rendered correctly, you need to remove the tabulation here

README.md
```diff
@@ -82,18 +81,20 @@ impl StatusMessage {
...
-    to see various usages of cross contract calls, including **system-level actions** done from inside the contract like balance transfer (examples of other system-level actions are: account creation, access key creation/deletion, contract deployment, etc).
+Follow [examples/cross-contract-high-level](examples/cross-contract-calls/high-level)
+to see various usages of cross contract calls, including **system-level actions** done from inside the contract like balance transfer (examples of other system-level actions are: account creation, access key creation/deletion, contract deployment, etc).
 
-* **Initialization methods.** We can define an initialization method that can be used to initialize the state of the contract. `#[init]` verifies that the contract has not been initialized yet (the contract state doesn't exist) and will panic otherwise.
+### **Initialization methods.** 
+We can define an initialization method that can be used to initialize the state of the contract. `#[init]` verifies that the contract has not been initialized yet (the contract state doesn't exist) and will panic otherwise.
 
     ```rust
     #[near_bindgen]
```
itegulov, line 100: Also here

README.md
```diff
@@ -64,13 +64,12 @@ impl StatusMessage {
 
 ## Features
 
-* **Unit-testable.** Writing unit tests is easy with `near-sdk`:
+### **Unit-testable.**
```
itegulov, line 67: I don't think `**<text>**` make a difference for headers

PR #931, opened 2022-09-27
Added a nix flake based developer environment

DavidM-D:
First this allows contributors to easily spin up a development environment which is consistent across machines. This may make things simpler when debugging build issues.

Second it allows people developing smart contracts externally to have a correctly configured compiler, IDE and linting tools ect. This removes a couple of steps for newer rust developers who use nix (admittedly a vanishingly small group of people).

One thing to note is that the flake instructions will not work until this PR is merged.


DavidM-D:
Closed until we decide whether this is a good idea

PR #932, opened 2022-09-27
explicit return added

mahmudsudo:
explicit is always better than implicit,for more readability for new developers


itegulov:
Rust is an expression-based language, so there is nothing wrong with "implicit" returns. They are widely used in our (and others) codebases.

PR #934, opened 2022-10-07
Enforce Standard Logging Compliance by Abstraction

coleFD:
The goal of this commit is to force compliance of standard logging practices so more efficient indexing features can be developed in the future such as a logs bloom filter by event name. See the original discussion here: https://github.com/near/NEPs/issues/297

! NOTE:
Please see discussion in the linked issue above to determine if it would be better to make the `standard` and `version` field optional

! BREAKING CHANGE: 
For this change to be useful, we will need to privatize/remove the `env::log()` and `env::log_str()` functions, which has not been implemented yet in this commit. May not be a breaking change, but adding this to this section, so it is recognized


near-sdk/src/environment/env.rs
```diff
@@ -1021,4 +1046,41 @@ mod tests {
...
+        // Event with json data
+        let mut data = std::collections::HashMap::new();
+        data.insert("vector", vec![1, 2, 3, 4]);
+        event(
+            "nepXXX",
+            "1.0.0",
+            "json_data_event",
+            Some(&(serde_json::to_string(&data).unwrap())),
+            EventSerializationFormat::JSON,
+        );
```
frol, line 1070: APIs like this are error-prone, and I would not teach our users bad practices.

Can you take inspiration from [FT](https://github.com/near/near-sdk-rs/blob/770cbce018a1b6c49d58276a075ace3da96d6dc1/near-contract-standards/src/fungible_token/events.rs) / [NFT](https://github.com/near/near-sdk-rs/blob/master/near-contract-standards/src/non_fungible_token/events.rs) events and design the API that would help people to design their events reliably? E.g. encourage people to define their events as enum, and bake it into the flow somehow; here is just an illustration of how that could look like:

```rust
enum Mintbase001Event {
    Mint { owner_id: AccountId, token_id: String },
    Burn { ... },
    Transfer { ... },
}

impl JsonEventEmit for Mintbase001Event { /* there will be a default `.emit()` implementation which just serializes `&self` and logs it */ }

impl Tokens {
    fn mint(&mut self, token: Token) {
         self.insert(token);
         Mintbase001Event::Mint { ... }.emit();
    }
}

self.tokens.mint(token);
```

coleFD: Sorry, I am new to open source dev. Is it best practice to `--amend` the commit since this was more of a proposal commit or to make a new commit?

Also, is a logging structure like this `Event::emit(MyEvent::Mint { ... })` acceptable, or is the structure above the only acceptable structure?

frol: @coleFD we will squash PR commits here, so feel free to just push new commits

coleFD:
Changed to WIP because I I would like to convert to a single derive attribute with kwargs, and I need to resolve issues with lifetimes. Any feedback on this pattern (see `./src/types/events` test for pattern)

near-sdk/src/environment/env.rs
frol: > Also, is a logging structure like this Event::emit(MyEvent::Mint { ... }) acceptable, or is the structure above the only acceptable structure?

That makes sense to me. My proposed structure was just an illustration of how to promote the events to be reliable instead of ad-hoc.

frol:
@coleFD Exposing events via ABI would be awesome! I feel it is a much heavier lift than the initial idea, but I see it as a cool challenge :-D

coleFD:
> near_bindgen(event)

For sure, I dont think adding events to the abi should be added to this issue. I dont want to add to many changes into 1 issue, but If I utilize `near_bindgen` instead of creating a new macro, it will put us in a better spot to add the abi functionality in the future.

near-sdk-macros/src/core_impl/event/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+/// function execution based on what type is returned by the function. By default, this will be
+/// done through `serde` serialized as JSON, but this can be overwritten using
+/// `#[result_serializer(borsh)]`.
+///
+/// # Examples
+///
+/// ```ignore
+/// use near_sdk::{near_events, Event};
+///
+/// #[near_events]
```
frol, line 27: I feel I might be missing something, why do we need two ways of achieving the same thing? I mean there is `#[near_bindgen(events)]` and this `#[near_events]`.

near-sdk-macros/src/core_impl/event/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+///
+///    #[event_standard("empty_standard")]
+///    #[event_version("3.0.0")]
+///    EmptyEvent
+/// }
+///
+/// #[near_bindgen]
+/// impl Contract {
+///     pub fn some_function(&self) {
+///         Event::emit {
```
frol, line 45: How does this `Event::emit {` work? Shouldn't it be a function call with `()` instead of `{}`?

frol:
> I still need to remove the previous logging functions.

Please, clarify which logging functions you refer to.

frol:
@austinabell Could you recommend someone to review this PR?

near-sdk-macros/src/core_impl/event/mod.rs
coleFD: I will change this. I made the conversion from `#[near_events]` to `#[near_bindgen(events)]` yesterday.

near-sdk-macros/src/core_impl/event/mod.rs
coleFD: yep, I will change this

coleFD:
> > I still need to remove the previous logging functions.
> 
> Please, clarify which logging functions you refer to.

The goal of this is to prevent logging without having the standard metadata fields `standard`, `version`, `event`, and `data`, so any logging functionality that does not do that, I think should be removed from the users' toolbox. I think the only functions that do this are `env::log()`, `env::log_str()`, and the macro `log!()`. 

Are there any concerns with doing this?

frol:
log_str and its derivatives should be left as is. There is no problem with using logs in contracts, though, we should definitely promote usage of events. Also, I would imagine we reimplement NFT and FT events using this facility. I see that @austinabell is out of office until next week, so we will need to wait until he gets back and gets time to review it.

frol:
By the way, I have just learned about this prior art: https://github.com/NEARFoundation/near-contract-tools#events

coleFD:
> By the way, I have just learned about this prior art: https://github.com/NEARFoundation/near-contract-tools#events

interesting, is there still incentive to add this to `near-sdk`?

frol:
I would like to hear what @austinabell thinks

near-sdk/src/types/event.rs
```diff
@@ -0,0 +1,95 @@
...
+    use super::Event;
+    use super::StandardEvent;
+    use crate as near_sdk;
+
+    #[near_bindgen(events)]
+    pub enum TestEvents<'a, 'b, T>
+    where
+        T: crate::serde::Serialize + Clone,
+    {
+        #[event_meta(standard = "swap_standard", version = "1.0.0")]
```
austinabell, line 28: Curious why the decision to group all events in one enum? Would this not just be simpler if it was an attribute on a struct? I suppose this is a bit opinionated, but wondering if there is a reason this _must_ be an enum

near-sdk-macros/src/lib.rs
```diff
@@ -316,3 +371,81 @@ pub fn function_error(item: TokenStream) -> TokenStream {
...
+                fn format(&self) -> String {
+                    let (standard, version): (String, String) = match self {
+                        #(#event_meta),*
+                    };
+                    let event = EventBuilder {standard, version, event_data: self.clone() };
+                    near_sdk::serde_json::to_string(&event).unwrap_or_else(|_| near_sdk::env::abort())
+                }
+
+                fn emit(&self) {
+                    near_sdk::Event::emit(self.clone());
```
austinabell, line 441: Clone shouldn't be a required bound. You don't need an owned copy of the data to serialize. I haven't checked if this compiles, but it would look something like this:
```suggestion
            struct EventBuilder<'a> #impl_generics #where_clause {
                standard: String,
                version: String,
                #[serde(flatten)]
                event_data: &'a #name #type_generics
            }
            impl #impl_generics near_sdk::StandardEvent for #name #type_generics #where_clause {
                fn format(&self) -> String {
                    let (standard, version): (String, String) = match self {
                        #(#event_meta),*
                    };
                    let event = EventBuilder {standard, version, event_data: self };
                    near_sdk::serde_json::to_string(&event).unwrap_or_else(|_| near_sdk::env::abort())
                }

                fn emit(&self) {
                    near_sdk::Event::emit(self);
```

near-sdk/src/types/event.rs
```diff
@@ -0,0 +1,95 @@
+pub trait StandardEvent {
+    fn format(&self) -> String;
+    fn emit(&self);
```
austinabell, line 3: This could have a default implementation to prevent devs from having to implement it manually if they choose.
```suggestion
    fn emit(&self) {
        crate::env::log_str(&format!("EVENT_JSON:{}", &standard_event.format()));
    }
```

near-sdk/src/types/event.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+pub struct Event {}
+impl Event {
+    pub fn emit(standard_event: impl StandardEvent) {
+        crate::env::log_str(&format!("EVENT_JSON:{}", &standard_event.format()));
+    }
+}
```
austinabell, line 11: This should just be removed altogether. It's basically a duplicate of the other trait and would be confusing to have this type. 

Also, side note is you don't need an owned value of `standard_event` here

austinabell (commented):
Thanks for working on this! I think this can be

near-sdk/src/types/event.rs
coleFD: This was added to provide the pattern `Event::emit(event)`. Personally, I like this more than using the `.emit()` method since it is easier to locate when skimming a contract, but we can go the `.emit()` method route if yall like that more. I do agree there should only be one option to emit events.

near-sdk/src/types/event.rs
```diff
@@ -0,0 +1,95 @@
+pub trait StandardEvent {
+    fn format(&self) -> String;
```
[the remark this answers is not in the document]
coleFD: if we decide not to implement the `Event` type, I could just remove this and add it to the `.emit()` method

near-sdk/src/types/event.rs
austinabell: You could still do that pattern using the other trait, though. What's the need to have it as a separate trait?

near-sdk/src/types/event.rs
austinabell: nope, that's fine. It's opinionated, was just sharing my intuition. I suppose we could also just support _both_ if it doesn't increase complexity meaningfully.

cc @itegulov wdyt?

near-sdk/src/types/event.rs
coleFD: okay, I see what you are saying

near-sdk/src/types/event.rs
frol: I treat events as some sort of standard and an enum in my head represents the fact that there is a finite number of types of events that a contract can emit. That said, I would consider extracting `standard = "..."` to the enum declaration level instead of the enum variants level. This way we could have FtEvents, NftEvents, MtEvents, FayyrEvents, etc, so one could import and use them in a composable way.

near-sdk/src/types/event.rs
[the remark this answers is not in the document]
austinabell: > thoughts on this interface?
> 
> ```
> pub trait StandardEvent {
>     fn format(&self) -> std::borrow::Cow<str>;
>     fn emit(&self) {
>         crate::env::log_str(&format!("EVENT_JSON:{}", self.format()));
>     }
>     fn emit_event(standard_event: &Self) {
>         crate::env::log_str(&format!("EVENT_JSON:{}", standard_event.format()));
>     }
> }
> ```

`emit_event` is the same interface as `emit`, just slightly more limiting. I don't see the point of it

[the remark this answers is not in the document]
coleFD: What about returning `Cow<str>` for `format()`?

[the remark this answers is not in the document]
austinabell: Oh, didn't notice. Why not just have it as an associated type bound on `std::fmt::Display` or `AsRef<str>`? Adding `Cow` code to everyone's contract, even if they don't want to use it, seems unnecessary. We currently avoid `Cow` in the SDK for that reason.

near-sdk-macros/src/core_impl/utils/mod.rs
```diff
@@ -73,3 +73,24 @@ pub(crate) fn extract_vec_type(ty: &Type) -> Option<&Type> {
...
+                    if path.is_ident("standard") {
+                        v.0 = Some(value.clone());
+                    } else if path.is_ident("version") {
+                        v.1 = Some(value.clone());
+                    }
+                };
+            })
+        }
+    });
+    v
```
itegulov, line 95: nit: any reason to pre-wrap `v` as a pair? I personally find that this looks more readable:
```suggestion
pub(crate) fn get_event_args(var: &syn::Variant) -> (Option<LitStr>, Option<LitStr>) {
    let mut standard = None;
    let mut version = None;
    var.attrs.iter().for_each(|attr| {
        let test = attr.parse_args_with(
            syn::punctuated::Punctuated::<syn::MetaNameValue, syn::Token![,]>::parse_terminated,
        );
        if let Ok(item) = test {
            item.iter().for_each(|kwargs| {
                if let syn::MetaNameValue { path, lit: syn::Lit::Str(value), .. } = kwargs {
                    if path.is_ident("standard") {
                        standard = Some(value.clone());
                    } else if path.is_ident("version") {
                        version = Some(value.clone());
                    }
                };
            })
        }
    });
    (standard, version)
```

near-sdk-macros/src/lib.rs
```diff
@@ -316,3 +371,81 @@ pub fn function_error(item: TokenStream) -> TokenStream {
         }
     })
 }
+
+/// The attribute macro `near_events` injects this macro and should be the used over this
```
itegulov, line 375:
```suggestion
/// The attribute macro `near_bindgen(events)` injects this macro and should be the used over this
```

near-sdk-macros/src/core_impl/utils/mod.rs
```diff
@@ -73,3 +73,24 @@ pub(crate) fn extract_vec_type(ty: &Type) -> Option<&Type> {
...
+        let test = attr.parse_args_with(
+            syn::punctuated::Punctuated::<syn::MetaNameValue, syn::Token![,]>::parse_terminated,
+        );
+        if let Ok(item) = test {
+            item.iter().for_each(|kwargs| {
+                if let syn::MetaNameValue { path, lit: syn::Lit::Str(value), .. } = kwargs {
+                    if path.is_ident("standard") {
+                        v.0 = Some(value.clone());
+                    } else if path.is_ident("version") {
+                        v.1 = Some(value.clone());
```
itegulov, line 89: Should we prohibit multiple conflicting `#[standard]`/`#[version]` attributes?

near-sdk-macros/src/lib.rs
```diff
@@ -43,8 +43,63 @@ use syn::{parse_quote, File, ItemEnum, ItemImpl, ItemStruct, ItemTrait, WhereCla
...
+///    Swap { token_in: AccountId, token_out: AccountId, amount_in: u128, amount_out: u128 },
+///
+///    #[event_standard("string_standard")]
+///    #[event_version("2.0.0")]
+///    StringEvent(String),
+///
+///    #[event_standard("empty_standard")]
+///    #[event_version("3.0.0")]
+///    EmptyEvent
+/// }
```
itegulov, line 82:
```suggestion
/// #[near_bindgen(events)]
/// pub enum MyEvents {
///    #[event_meta(standard = "swap_standard", version = "1.0.0")]
///    Swap { token_in: AccountId, token_out: AccountId, amount_in: u128, amount_out: u128 },
///
///    #[event_meta(standard = "string_standard", version = "2.0.0")]
///    StringEvent(String),
///
///    #[event_meta(standard = "empty_standard", version = "3.0.0")]
///    EmptyEvent
/// }
```

near-sdk-macros/src/core_impl/event/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+///    Swap { token_in: AccountId, token_out: AccountId, amount_in: u128, amount_out: u128 },
+///
+///    #[event_standard("string_standard")]
+///    #[event_version("2.0.0")]
+///    StringEvent(String),
+///
+///    #[event_standard("empty_standard")]
+///    #[event_version("3.0.0")]
+///    EmptyEvent
+/// }
```
itegulov, line 40:
```suggestion
/// pub enum MyEvents {
///    #[event_meta(standard = "swap_standard", version = "1.0.0")]
///    Swap { token_in: AccountId, token_out: AccountId, amount_in: u128, amount_out: u128 },
///
///    #[event_meta(standard = "string_standard", version = "2.0.0")]
///    StringEvent(String),
///
///    #[event_meta(standard = "empty_standard", version = "3.0.0")]
///    EmptyEvent
/// }
```

near-sdk-macros/src/lib.rs
```diff
@@ -316,3 +371,81 @@ pub fn function_error(item: TokenStream) -> TokenStream {
...
+                }
+
+                fn emit(&self) {
+                    near_sdk::Event::emit(self.clone());
+                }
+            }
+        })
+    } else {
+        TokenStream::from(
+            syn::Error::new(Span::call_site(), "NearEvent can only be used as a derive on enums.")
```
itegulov, line 447:
```suggestion
            syn::Error::new(Span::call_site(), "EventMetadata can only be used as a derive on enums.")
```

near-sdk-macros/src/core_impl/utils/mod.rs
```diff
@@ -73,3 +73,24 @@ pub(crate) fn extract_vec_type(ty: &Type) -> Option<&Type> {
...
     }
 }
+
+pub(crate) fn get_event_args(var: &syn::Variant) -> (Option<LitStr>, Option<LitStr>) {
+    let mut v: (Option<LitStr>, Option<LitStr>) = (None, None);
+    var.attrs.iter().for_each(|attr| {
+        let test = attr.parse_args_with(
+            syn::punctuated::Punctuated::<syn::MetaNameValue, syn::Token![,]>::parse_terminated,
+        );
+        if let Ok(item) = test {
```
itegulov, line 83: Shouldn't we be checking that `attr` is `event_meta` just in case something else has `standard`/`version` args?

near-sdk/src/types/event.rs
itegulov: > @itegulov wdyt about this? Do you think the implications of enum vs struct annotation would have any implication on integrating into the ABI?

I don't think it matters, we should be able to generate ABI for both options.

Personally, I find that having a mega-enum covering all standards used in this contract is a bit of an antipattern, but not a strong opinion. What @frol suggested makes more sense to me - having an enum per standard looks cleaner.

near-sdk-macros/src/lib.rs
```diff
@@ -316,3 +359,93 @@ pub fn function_error(item: TokenStream) -> TokenStream {
         }
     })
 }
+
+/// Adding the `events` argument for the `#[near_bindgen]` macro injects this derive macro and should be the used instead of this
```
frol, line 363: It was hard to understand what you mean here. I hope I managed to unpack it correctly:

```suggestion
/// NOTE: This is an internal implementation for `#[near_bindgen(events(standard = ...))]` attribute.
```

frol (approved):
Looks good to me.

@austinabell @itegulov What do you think?

frol:
@coleFD CI is not happy about fmt.

@austinabell Buildkite CI failed with the following error:

image

I don't see this PR changing Cargo.* or rust toolchain in any way.

itegulov:
> Buildkite CI failed with the following error:

This was fixed in https://github.com/near/near-sdk-rs/pull/942. Just needs to be rebased I believe.

near-sdk/src/types/event.rs
```diff
@@ -0,0 +1,84 @@
+pub trait StandardEvent {
+    type EventString: std::fmt::Display;
+    fn format(&self) -> Self::EventString;
+    fn emit(&self) {
+        crate::env::log_str(&format!("EVENT_JSON:{}", self.format()));
+    }
```
austinabell, line 6: Sorry, I initially mentioned `std::fmt::Display` as an alternative to `AsRef<str>` but thinking about it more, it doesn't make much sense. This is because it forces `fmt` machinery in, even if someone wanted to have a static event be emitted.

```suggestion
    type EventString: std::convert::AsRef<str>;
    fn format(&self) -> Self::EventString;
    fn emit(&self) {
        format!("EVENT_JSON:{}", self.format().as_ref());
    }
```

near-sdk/src/types/event.rs
```diff
@@ -0,0 +1,84 @@
+pub trait StandardEvent {
+    type EventString: std::fmt::Display;
```
austinabell, line 2: Wondering if anyone else has opinions, but having this named as `EventJson` might be more clear?

Honestly not thrilled that our events and API is being forced as JSON, and we don't have much flexibility to use a more efficient format in the future, but this would make it more clear what this format is?

wdyt @frol ?

frol: Yeah, naming is hard. The standard uses `EVENT_JSON:` prefix specifically to keep the opportunity open for future extension. Thus, `EventJson` makes total sense to me as it matches the prefix. Maybe even consider renaming `#[near_bindgen(event)]` to `#[near_bindgen(event_json)]`?

frol: I don't want to drag ourselves into the conversation of how other formats could look like - borsh does not have a schema, and binary data cannot be dumped into UTF-8 string, so before getting dumped into logs, we either would need to make a protocol change in NEAR Protocol that would turn the logs to be an array of bytes instead of a valid UTF-8, or encode borsh-serialized event into, just for the sake of example, base64.

Thus, I gravitate towards actually renaming all the mentions of just `event` into `event_json` and get back to this discussion later, when we will have several events standards.

coleFD: should I move forward with @frol recommendation?

austinabell: I don't have an opinion if you're asking me. What frol suggested SGTM :)

mooori:
Using [`darling`](https://github.com/TedDriggs/darling) to parse attributes passed to `#[near_bindgen]` could help to remove custom parsing logic.

Another benefit of using `darling` is related to supporting other attributes besides `events`. That could be done by adding fields to the structs into which attributes are parsed, instead of adding more custom parsing logic.

For example, [this PR](https://github.com/near/near-sdk-rs/pull/952) shows how `darling` could be used for `near_bindgen` and handling multiple attributes might [look like this](https://github.com/near/near-sdk-rs/pull/952#issuecomment-1296883879).

near-sdk-macros/src/lib.rs
```diff
@@ -43,8 +43,51 @@ use syn::{parse_quote, File, ItemEnum, ItemImpl, ItemStruct, ItemTrait, WhereCla
...
+///     pub fn some_function(&self) {
+///            MyEvents::StringEvent(
+///                 String::from("some_string")
+///             ).emit();
+///     }
+///
+/// }
+/// ```
 #[proc_macro_attribute]
 pub fn near_bindgen(_attr: TokenStream, item: TokenStream) -> TokenStream {
```
austinabell, line 86: since the param is used now, can remove the `_` prefix

near-sdk/src/types/event.rs
```diff
@@ -0,0 +1,84 @@
...
+
+        TestEvents::LifetimeTestA::<String>("lifetime").emit();
+
+        TestEvents::LifetimeTestB::<String>("lifetime_b").emit();
+
+        let logs = get_logs();
+
+        assert!(
+            logs[0]
+                == r#"EVENT_JSON:{"standard":"test_standard","version":"1.0.0","event":"swap","data":{"token_in":"wrap.near","token_out":"test.near","amount_in":100,"amount_out":200,"test":"tst"}}"#
```
austinabell, line 65: nit: this should be `assert_eq` to have a better error message if this is ever broken

near-sdk/src/types/event.rs
```diff
@@ -0,0 +1,84 @@
+pub trait EventJson {
+    type EventString: std::convert::AsRef<str>;
```
austinabell, line 2: My previous comment was referring to changing the name of the associated type, not the trait name, but I suppose this is fine. Only thing I'll mention is that it is ambiguous what the `EventString` is referring to -- whether it be the whole event message or just the JSON data

near-sdk-macros/src/lib.rs
```diff
@@ -316,3 +359,93 @@ pub fn function_error(item: TokenStream) -> TokenStream {
...
+        let (custom_impl_generics, ..) = generics.split_for_impl();
+
+        if attr_error > 0 {
+            return TokenStream::from(
+                syn::Error::new(
+                    Span::call_site(),
+                    "Near events must have `event_version`. Must have a single string literal value.",
+                )
+                .to_compile_error(),
+            );
```
austinabell, line 415: confused, if you're just erroring this static message if at least one error, why not just do it where you increment `attr_error`?

near-sdk-macros/src/lib.rs
```diff
@@ -316,3 +359,93 @@ pub fn function_error(item: TokenStream) -> TokenStream {
...
+
+            #[derive(near_sdk::serde::Serialize)]
+            #[serde(crate="near_sdk::serde")]
+            #[serde(rename_all="snake_case")]
+            struct EventBuilder #custom_impl_generics #where_clause {
+                standard: String,
+                version: String,
+                #[serde(flatten)]
+                event_data: &#event_lifetime #name #type_generics
+            }
```
austinabell, line 428: This will conflict if you have two event derives in the same module. Since this is a hidden type, ideally it's scoped only for the implementation of `EventJson`. Am I missing something? Why else would it be used otherwise?

near-sdk-macros/src/lib.rs
```diff
@@ -316,3 +359,93 @@ pub fn function_error(item: TokenStream) -> TokenStream {
...
+
+        // handle lifetimes, generics, and where clauses
+        let (impl_generics, type_generics, where_clause) = &input.generics.split_for_impl();
+        // add 'near_event lifetime for user defined events
+        let mut generics = input.generics.clone();
+        let event_lifetime = syn::Lifetime::new("'near_event", Span::call_site());
+        generics.params.insert(
+            generics.params.len(),
+            syn::GenericParam::Lifetime(syn::LifetimeDef::new(event_lifetime.clone())),
+        );
```
austinabell, line 405: why insert the lifetime param at the end? Lifetimes must come before type generics. (honestly confused as why this isn't broken in test, so might be missing something here)

near-sdk-macros/src/lib.rs
coleFD: Yes, it's only used for the implementation of `EventJson` to wrap the user defined events enum and add the `standard` and `version` field with correct formatting.

near-sdk-macros/src/lib.rs
coleFD: compiler errors in the `.for_each()`. I can convert to a for loop and do what you proposed

near-sdk-macros/src/lib.rs
coleFD: `syn` handles ordering in the `ToTokens` implementation for generics. I will change this to insert it at the 0 index so it's not confusing

coleFD:
any feedback on the latest commit?

austinabell:
> any feedback on the latest commit?

There's actually a much simpler way to make sure there is no overlap here. I made the commit https://github.com/near/near-sdk-rs/commit/1e6c09edf1778fb6e2ee1604965df09bcc11545c.

Feel free to cherry-pick, make a similar change yourself, or let me know if I can push it to this branch. It's a small nit but with this, you can still have a collision when you don't need that to be the case.

austinabell:
So when making that other change, it wasn't clear to me why the trait is necessary at all. Seems unergonomic to require devs to import that trait to use this. Why not just remove it like in https://github.com/near/near-sdk-rs/commit/1cc5c19d134c5e72ef3ec5cfe7e1f5701b82ed7f ?

coleFD:
> So when making that other change, it wasn't clear to me why the trait is necessary at all. Seems unergonomic to require devs to import that trait to use this. Why not just remove it like in [1cc5c19](https://github.com/near/near-sdk-rs/commit/1cc5c19d134c5e72ef3ec5cfe7e1f5701b82ed7f) ?

For my first iteration, the plan was to have a `StandardEvent` trait to standardize the interface for all current and future serialization formats. I agree it doesnt make much sense if we just have json

austinabell:
> For my first iteration, the plan was to have a `StandardEvent` trait to standardize the interface for all current and future serialization formats. I agree it doesn't make much sense if we just have JSON

I just don't see the benefit in having a trait for this if all it's going to do is add the convenience `env::log_str("EVENT_JSON:...")`. Can always add the trait back later, but I'd rather keep the API as slim as possible.

I'll push that change to this branch then since it seems like you're on board. Would be nice to have a macro test for the codegen, since I was able to make both changes without any tests failing, which is not ideal. This is not required though, improved tests can come after this PR comes in. Sorry for so much back and forth, this is just an important interface and it would be painful to migrate later.

frol:
@austinabell @coleFD Thanks for pushing it forward! I was out of the office, and I just had a chance to sync back. I see there was a huge progress! 🙏

austinabell (approved):
LGTM but would be good to have another approval since I was the last to make a commit on this branch

near-sdk/src/types/mod.rs
```diff
@@ -13,6 +13,9 @@ pub use self::account_id::{AccountId, ParseAccountIdError};
 mod gas;
 pub use self::gas::Gas;
 
+#[cfg(test)]
+mod event_tests;
```
frol, line 17: Is it actually the right place for the tests? Should we move it to `near-sdk/tests`?

near-sdk-macros/src/lib.rs
```diff
@@ -43,8 +43,51 @@ use syn::{parse_quote, File, ItemEnum, ItemImpl, ItemStruct, ItemTrait, WhereCla
...
+///    #[event_version("3.0.0")]
+///    EmptyEvent
+/// }
+///
+/// #[near_bindgen]
+/// impl Contract {
+///     pub fn some_function(&self) {
+///            MyEvents::StringEvent(
+///                 String::from("some_string")
+///             ).emit();
```
frol, line 80:
```suggestion
///         MyEvents::StringEvent(
///             String::from("some_string")
///         ).emit();
```

near-sdk-macros/src/core_impl/event/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+/// In addition, this function extracts the event's `standard` value and injects it as a constant to be used by
+/// the `near_sdk::EventMetadata` derive macro
+pub(crate) fn near_events(attr: TokenStream, item: TokenStream) -> TokenStream {
+    // get standard from attr args
+    let standard = get_standard_arg(&syn::parse_macro_input!(attr as syn::AttributeArgs));
+    if standard.is_none() {
+        return TokenStream::from(
+            syn::Error::new(
+                Span::call_site(),
+                "Near events must have a `standard` value as an argument for `event_json` in the `near_bindgen` arguments. The value must be a string literal.",
```
frol, line 16:
```suggestion
                "Near events must have a `standard` value as an argument for `event_json` in the `near_bindgen` arguments. The value must be a string literal, e.g. \"nep999\", \"mintbase-marketplace\".",
```

frol (approved):
Overall, looks good to me. I only left a few nits in the comments

austinabell:
thank you! Sorry for all the back and forth

coleFD:
No worries! thank you guys for being responsive and taking the time to dig into it!

PR #938, opened 2022-10-17
chore: make examples use the root target folder

itegulov:
Not sure if there is a reason not to share the target directory across all examples. This saves ~35% of disk space on a clean build (1.7GB vs 2.6GB).


austinabell (approved):
noting that this would be adding a bit of friction if devs try to copy these examples, but seems worth it

ChaoticTempest (approved):
LGTM

PR #939, opened 2022-10-19
feat: return raw bytes from ABI chunk generators

miraclx:
`cargo-near` + `near-sdk`, at any point in time, could use versions of `near-abi` with conflicting in-memory representations of `ChunkedAbiEntry`, making it impossible to simply cast one as the other.

Since `near-abi` already hosts a deserialization implementation for `ChunkedAbiEntry`, with `schema_version` validation, it makes sense to just use that.

This means the ABI chunk generator has to return a JSON string directly, which `cargo-near` can then parse, returning a neat error if an error occurred.

This is linked to https://github.com/near/cargo-near/pull/74.

### DIFF
--- a/near-sdk-macros/src/core_impl/abi/abi_generator.rs
+++ b/near-sdk-macros/src/core_impl/abi/abi_generator.rs
@@ -23,13 +23,15 @@ pub fn generate(i: &ItemImplInfo) -> TokenStream2 {
         #[cfg(not(target_arch = "wasm32"))]
         const _: () = {
             #[no_mangle]
-            pub fn #near_abi_symbol() -> near_sdk::__private::ChunkedAbiEntry {
+            pub fn #near_abi_symbol() -> String {
                 let mut gen = near_sdk::__private::schemars::gen::SchemaGenerator::default();
                 let functions = vec![#(#functions),*];
-                near_sdk::__private::ChunkedAbiEntry::new(
-                    functions,
-                    gen.into_root_schema_for::<String>()
-                )
+                near_sdk::serde_json::to_string(
+                    &near_sdk::__private::ChunkedAbiEntry::new(
+                        functions,
+                        gen.into_root_schema_for::<String>()
+                    )
+                ).unwrap()
             }
         };
     }
--- a/near-sdk-macros/src/core_impl/abi/abi_generator.rs
+++ b/near-sdk-macros/src/core_impl/abi/abi_generator.rs
@@ -23,18 +23,19 @@ pub fn generate(i: &ItemImplInfo) -> TokenStream2 {
         #[cfg(not(target_arch = "wasm32"))]
         const _: () = {
             #[no_mangle]
-            pub extern "C" fn #near_abi_symbol() -> *const std::ffi::c_char {
+            pub extern "C" fn #near_abi_symbol() -> (*const u8, usize) {
                 let mut gen = near_sdk::__private::schemars::gen::SchemaGenerator::default();
                 let functions = vec![#(#functions),*];
-                std::ffi::CString::new(
+                let mut data = std::mem::ManuallyDrop::new(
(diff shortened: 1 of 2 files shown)


austinabell:
Confused why you need this to happen in the SDK. If you wanted to serialize and deserialize, couldn't you just do that on the `cargo-near` side? This feels much more error prone and unwieldy for anyone else to use

miraclx:
> If you wanted to serialize and deserialize, couldn't you just do that on the cargo-near side? This feels much more error prone and unwieldy for anyone else to use

1. Currently, `near-sdk`'s codegen doesn't `export` anything but raw bytes anyway, these ABI generators are the only exemption.

2. The problem we're trying to mitigate is dynamically loading the code generated methods, and unexpectedly casting between incompatible `ChunkedAbiEntry`-ies. Because `near-sdk`'s `near-abi` version at any time is different from whatever the consumers' is.

3. This crosses an FFI boundary, we can't make any guarantees about the in-memory representation of the `ChunkedAbiEntry`, unless we always force `near-sdk` to recompile the contract every single time. Ideally, if we're crossing an FFI boundary, we should have “stable” guarantees, in this case, deferring to the C ABI.

4. And finally, this is hidden behind the `__abi-generate` feature flag.

austinabell (approved):
This all seems reasonable. I am under the assumption we would need to co-ordinate the cargo-near and near-sdk-rs release to happen together and basically indicate that you cannot mix the newer released version with a previous version of either, correct?

miraclx:
This is correct, but to be clear, a new release of `near-sdk` only forces a new release of `cargo-near` if the new version of the SDK also exports an upgraded ABI schema version.

So if a new release of the SDK is cut but still exports a compatible ABI schema version, there'd be no need to upgrade your version of `cargo-near`.

Would be neat if every successive release of `cargo-near` could build contracts with any `schema_version` less than the maximum supported schema version.

Especially since it's a utility, that we expect devs to install globally, and not on a per-project basis. So, jumping between contracts that use different versions of `near-sdk`, potentially exporting different `schema_versions`.

`cargo-near` should be versatile enough to support all of them.

austinabell:
> This is correct, but to be clear, a new release of `near-sdk` only forces a new release of `cargo-near` if the new version of the SDK also exports an upgraded ABI schema version.
> 
> So if a new release of the SDK is cut but still exports a compatible ABI schema version, there'd be no need to upgrade your version of `cargo-near`.
> 
> Would be neat if every successive release of `cargo-near` could build contracts with any `schema_version` less than the maximum supported schema version.
> 
> Especially since it's a utility, that we expect devs to install globally, and not on a per-project basis. So, jumping between contracts that use different versions of `near-sdk`, potentially exporting different `schema_versions`.
> 
> `cargo-near` should be versatile enough to support all of them.

I'm more referring to the handling of the change of this method signature. For example, using the current version of `cargo-near` with this branch of the SDK or the https://github.com/near/cargo-near/pull/74 branch with a currently released SDK version (pre-release).

Not a big deal, since ABI has only been available for pre-releases, just want to make sure I understand

miraclx:
Oh, yes. They're interlinked.

austinabell:
> Oh, yes. They're interlinked. Looking forward, after we drop the pre-releases, this method interface should stay consistent. It's the data that changes.. (should help mitigate any issues of incompatibility)

Sounds good to me, feel free to merge this whenever you're ready

austinabell:
One last thing I'll suggest. Have you considered just having the interface bytes rather than string? This would allow in the future if we want to change the format going across the boundary, it doesn't have to be utf8?

miraclx:
It's already this, unless I'm missing something. `CString` is not UTF-8.

miraclx:
Took the approach in #941, lmk what you think.

austinabell (approved):

PR #941, opened 2022-10-21
abi generator: return raw byte array instead of CString

miraclx:
Context: https://github.com/near/near-sdk-rs/pull/939#issuecomment-1285987782

Linked with: https://github.com/near/cargo-near/pull/80

### DIFF
--- a/near-sdk-macros/src/core_impl/abi/abi_generator.rs
+++ b/near-sdk-macros/src/core_impl/abi/abi_generator.rs
@@ -23,18 +23,17 @@ pub fn generate(i: &ItemImplInfo) -> TokenStream2 {
         #[cfg(not(target_arch = "wasm32"))]
         const _: () = {
             #[no_mangle]
-            pub extern "C" fn #near_abi_symbol() -> *const std::ffi::c_char {
+            pub extern "C" fn #near_abi_symbol() -> (*const std::ffi::c_uchar, usize, usize) {
                 let mut gen = near_sdk::__private::schemars::gen::SchemaGenerator::default();
                 let functions = vec![#(#functions),*];
-                std::ffi::CString::new(
+                let data = std::mem::ManuallyDrop::new(
                     near_sdk::serde_json::to_vec(&near_sdk::__private::ChunkedAbiEntry::new(
                         functions,
                         gen.into_root_schema_for::<String>(),
                     ))
                     .unwrap(),
-                )
-                .unwrap()
-                .into_raw()
+                );
+                (data.as_ptr(), data.len(), data.capacity())
             }
         };
     }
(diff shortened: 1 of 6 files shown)


near-sdk-macros/src/core_impl/abi/abi_generator.rs
```diff
@@ -23,18 +23,19 @@ pub fn generate(i: &ItemImplInfo) -> TokenStream2 {
...
+            pub extern "C" fn #near_abi_symbol() -> (*const std::ffi::c_uchar, usize) {
                 let mut gen = near_sdk::__private::schemars::gen::SchemaGenerator::default();
                 let functions = vec![#(#functions),*];
-                std::ffi::CString::new(
+                let mut data = std::mem::ManuallyDrop::new(
                     near_sdk::serde_json::to_vec(&near_sdk::__private::ChunkedAbiEntry::new(
                         functions,
                         gen.into_root_schema_for::<String>(),
                     ))
                     .unwrap(),
```
austinabell, line 34: curious why not use `into_vec` here so instead of `(*const std::ffi:c_uchar, _)` it would be `(*const u8, _)`?

miraclx: Just saw that `std::ffi::c_uchar` was only just stabilized in `rustc 1.64.0`, I'll revert to `u8`.

austinabell: > PS: what's the MSRV of the SDK?

eh, kinda 1.56 https://github.com/near/near-sdk-rs/blob/master/.github/workflows/test.yml#L16, but I'm not sure that's the MSRV of building a contract with the SDK.

> Well, a [std::ffi:c_uchar](https://doc.rust-lang.org/std/ffi/type.c_uchar.html) already is a u8, just thought to use the alias for clarity.

but having it as bytes means `c_uchar` would be confusing, especially if changed to a binary format in the future. Very much a nit btw, but yeah I think the MSRV requirement is a good reason to have it as a `*const u8` instead anyway

austinabell (approved):
There shows some extra changes since the target branch isn't updated from `master`, but the changes LGTM

near-sdk-macros/src/core_impl/abi/abi_generator.rs
miraclx: > eh, kinda 1.56 [master/.github/workflows/test.yml#L16](https://github.com/near/near-sdk-rs/blob/master/.github/workflows/test.yml?rgh-link-date=2022-10-21T14%3A19%3A17Z#L16), but I'm not sure that's the MSRV of building a contract with the SDK.

Looks like it's `1.60` since that's `near-vm-logic`'s MSRV.

miraclx: Actually, I take that back, just building a contract is indeed `1.56`. But `cargo-near` fails at ABI generation for toolchains under `1.60`.

PR #943, opened 2022-10-21
ci: update actions/checkout to v3

armujahid:


austinabell (approved):
Skimming through https://github.com/actions/checkout/blob/main/CHANGELOG.md I don't see anything restricting us from updating this, but I'd love to get another approval from someone else on this

ChaoticTempest:
Looking at https://github.com/actions/checkout/issues shows a couple issues that's happening with docker containers, but our docker containers are a completely separate from tests, so we don't have to worry about it. We don't plan on adding anything docker related to our tests in the foreseeable future right? @austinabell. Otherwise, this looks good to be added

itegulov (approved):
Looked through the changelog too and it seems like a reasonable upgrade. We can always rollback if we start experiencing issues.

austinabell:
thanks @armujahid !

PR #948, opened 2022-10-25
refactor: make PromiseIndex a newtype

### LINKED ISSUE #945: Make `PromiseIndex` a newtype to avoid misuse
There shouldn't be any reasonable use to define a PromiseIndex manually, so it would make the API much harder to misuse if this wasn't an alias for `u64`.

This was originally set as an alias to avoid breaking changes in-between major versions, but this really should be updated before the next major version increase (5.0).

austinabell:
I thought this was done with a previous major version bump.

This shouldn't be breaking in practical cases because promise indices should never be manually constructed, but is technically breaking and should come in with a major version bump. (I'll probably keep this as a draft until we are ready to start pulling in breaking changes)

closes #945

### DIFF
--- a/near-sdk/src/environment/env.rs
+++ b/near-sdk/src/environment/env.rs
@@ -404,7 +404,7 @@ pub fn promise_create(
     gas: Gas,
 ) -> PromiseIndex {
     let account_id = account_id.as_bytes();
-    unsafe {
+    PromiseIndex(unsafe {
         sys::promise_create(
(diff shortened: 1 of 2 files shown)


austinabell:
Noting that this is a safer API to use, but is a breaking change and would have to come with a major version update. It is a bit opinionated, though, so I'll just close and if someone wants to pull in with a major release -- feel free. Nothing needed to be updated about this

PR #951, opened 2022-10-28
Remove unused tokens in compilation test

mooori:
The `#[near_bindgen]` macro currently ignores tokens after the attribute name, for example `#[near_bindgen(this_is_ignored)]`. This is due to `_attr` (which [contains these tokens](https://doc.rust-lang.org/reference/procedural-macros.html#attribute-macros)) being ignored:

https://github.com/near/near-sdk-rs/blob/039cc110c9c01907b95250a50d9ff8ccbad99cbf/near-sdk-macros/src/lib.rs#L46-L47

There's a contract used in compilation test which has tokens after the attribute name: `#[near_bindgen(init => new)]`. They are removed in this PR.

## Context

For [`near-plugins`](https://github.com/aurora-is-near/near-plugins) we're experimenting with passing extra tokens like `#[near_bindgen(some_config(value_a, value_b))]`. When adding logic to parse these tokens via [`darling`](https://docs.rs/darling/latest/darling/index.html) the compilation of the contract with `#[near_bindgen(init => new)]` fails.

### DIFF
--- a/near-sdk/compilation_tests/cond_compilation.rs
+++ b/near-sdk/compilation_tests/cond_compilation.rs
@@ -9,7 +9,7 @@ struct Incrementer {
     value: u32,
 }
 
-#[near_bindgen(init => new)]
+#[near_bindgen]
 impl Incrementer {
     #[cfg(feature = "myfeature")]
     pub fn new() -> Self {


austinabell:
Sorry if I'm missing the motivation here, but what is the purpose of changing this within our compilation tests? This shouldn't affect what you're building?

Are you suggesting that something within our macros could be changed to allow it to be extended in external libs?

mooori:
The feature proposed in #952 uses tokens passed after the name of `#[near_bindgen]`. [This comment](https://github.com/near/near-sdk-rs/pull/952#discussion_r1008259741) tries to extend on the motivation and highlights the code that breaks the compilation of `cond_compilation.rs`.

austinabell:
> Assuming `init => new` is ignored, it might make sense to remove it even if #952 will be rejected. To avoid that someone stumbles upon this in the future.

Yeah, that makes sense, I'm just making it clear we may use the attributes in the future. #934 is one that proposes this kind of change

austinabell (approved):

PR #952, opened 2022-10-28
feat: Blacklist attributes not to forward to ContractExt methods

mooori:
# Background

Smart contract plugins like [`near-plugins`](https://github.com/aurora-is-near/near-plugins) use attribute macros to modify contract methods.

For example, `access_control_any` in the following snippet injects code that panics if the caller is not a grantee of `Role::A` or `Role::B`:

```rust
#[near_bindgen]
impl Contract {
    #[access_control_any(roles(Role::LevelA, Role::LevelC))]
    pub fn restricted_method(&mut self) { /* */ }
}
```

# Problem
`#[near_bindgen]` will create a struct `ContractExt` with:

```rust
#[near_bindgen]
impl ContractExt {
    // _All_ attributes of Contract::restricted_method are forwarded here.
    //
    // We want to avoid forwarding plugin attributes (i.e. access_control_any),
    // as it may not make sense and lead to compilation errors.
    #[access_control_any(roles(Role::LevelA, Role::LevelC))]
    pub fn restricted_method(self) -> near_sdk::Promise { /* */ }
}
```

# Proposed solution
Make `#[near_bindgen]` accept a blacklist of attributes which it should _not_ forward to methods of the automatically generated `*Ext` type. This PR adds `blacklist_ext_fn_attrs` for this purpose:

```rust
#[near_bindgen(blacklist_ext_fn_attrs(access_control_any))]
impl Contract {
    #[access_control_any(roles(Role::LevelA, Role::LevelC))]
    pub fn restricted_method(&mut self) { /* */ }
}
```

# Implementation overview
- Add `MacroConfig` which allows configuring the behavior of macros.
- Pass it down to where the corresponding `ContractExt` methods are created.
- Forward only attributes which are _not_ blacklisted.

# Backwards compatibility
This should be backwards compatible since:

- Currently any tokens following the name of `near_bindgen` are ignored (ref. #951).
- After this change, usage of `#[near_bindgen]` without attributes will result in using the default value of `MacroConfig` which does not alter behavior.

# Thoughts / Question
Maybe function attributes shouldn't be forwarded to `Ext` methods by default? In that case, it should be possible to turn the blacklist feature proposed here into a whitelist feature. Though that would be a breaking change.

### DIFF
--- a/near-sdk-macros/src/lib.rs
+++ b/near-sdk-macros/src/lib.rs
@@ -4,13 +4,18 @@ extern crate proc_macro;
 mod core_impl;
 
 use core_impl::ext::generate_ext_structs;
+use core_impl::MacroConfig;
 use proc_macro::TokenStream;
 
 use self::core_impl::*;
+use darling::FromMeta;
 use proc_macro2::Span;
 use quote::{quote, ToTokens};
 use syn::visit::Visit;
-use syn::{parse_quote, File, ItemEnum, ItemImpl, ItemStruct, ItemTrait, WhereClause};
+use syn::{
+    parse_macro_input, parse_quote, AttributeArgs, File, ItemEnum, ItemImpl, ItemStruct, ItemTrait,
+    WhereClause,
+};
 
 /// This attribute macro is used on a struct and its implementations
 /// to generate the necessary code to expose `pub` methods from the contract as well
@@ -44,7 +49,15 @@ use syn::{parse_quote, File, ItemEnum, ItemImpl, ItemStruct, ItemTrait, WhereCla
 /// }
 /// ```
 #[proc_macro_attribute]
-pub fn near_bindgen(_attr: TokenStream, item: TokenStream) -> TokenStream {
+pub fn near_bindgen(attrs: TokenStream, item: TokenStream) -> TokenStream {
+    let attr_args = parse_macro_input!(attrs as AttributeArgs);
+    let config = match MacroConfig::from_list(&attr_args) {
+        Ok(args) => args,
+        Err(e) => {
+            return TokenStream::from(e.write_errors());
+        }
+    };
+
     if let Ok(input) = syn::parse::<ItemStruct>(item.clone()) {
         let ext_gen = generate_ext_structs(&input.ident, Some(&input.generics));
         #[cfg(feature = "__abi-embed")]
@@ -95,7 +108,7 @@ pub fn near_bindgen(_attr: TokenStream, item: TokenStream) -> TokenStream {
         let generated_code = item_impl_info.wrapper_code();
 
         // Add wrapper methods for ext call API
-        let ext_generated_code = item_impl_info.generate_ext_wrapper_code();
+        let ext_generated_code = item_impl_info.generate_ext_wrapper_code(&config);
         TokenStream::from(quote! {
             #ext_generated_code
             #input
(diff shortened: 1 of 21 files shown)


near-sdk-macros/src/lib.rs
```diff
@@ -44,7 +49,15 @@ use syn::{parse_quote, File, ItemEnum, ItemImpl, ItemStruct, ItemTrait, WhereCla
...
 #[proc_macro_attribute]
-pub fn near_bindgen(_attr: TokenStream, item: TokenStream) -> TokenStream {
+pub fn near_bindgen(attrs: TokenStream, item: TokenStream) -> TokenStream {
+    let attr_args = parse_macro_input!(attrs as AttributeArgs);
+    let config = match MacroConfig::from_list(&attr_args) {
+        Ok(args) => args,
+        Err(e) => {
+            return TokenStream::from(e.write_errors());
+        }
+    };
```
mooori, line 59: To extend on the motivation for #951: Previously attributes passed to `near_bindgen` were ignored. Now `darling` tries to parse them into `MacroConfig`. The tokens `init => new` in the compilation test were previously ignored. Now this test will fail since `init => new` cannot be parsed into `MacroConfig`.

near-sdk/compilation_tests/cond_compilation.rs
```diff
@@ -9,7 +9,7 @@ struct Incrementer {
     value: u32,
 }
 
-#[near_bindgen(init => new)]
```
mooori, line 12: See #951. Sneaked in here to make CI happy.

mooori:
Another open PR that's about to use attributes with `#[near_bindgen]` is #934. A possible way to handle multiple attributes would be:

```rust
use darling::util::PathList;
use darling::FromMeta;

#[derive(Default, FromMeta, Clone, Debug)]
pub struct MacroConfig {
    pub events: Option<EventsConfig>,
    pub blacklist_ext_fn_attrs: Option<PathList>,
}

pub struct EventsConfig {
    standard: String,
    another_config_value: T, // darling supports many types
}
```

Using [`darling`](https://github.com/TedDriggs/darling) the attributes passed in the following use cases can be parsed into `MacroConfig`:

```rust
#[near_bindgen(events(standard = "nepXXX", another_config_value = "..."))]
pub enum MyEvents { /* ... */ }


#[near_bindgen(blacklist_ext_fn_attrs(some_plugin_attr))]
impl Contract { /* ... */ }
```

mfornet (approved):

austinabell (commented):
Seems fine. A little unfortunate to add a dependency like this for something that doesn't affect the vast majority, but this seems reasonable if it enables more tooling.

None of the `_Ext` methods should require the attributes in general, possibly an oversight there. I'm not sure what developers would prefer in general.

I can't think of any attribute that a developer would want to keep, seems like there might be another option of just removing the attributes and if developers mention breakages we could do something like this? Just seems like a bit of an unintuitive pattern to have the `#[near_bindgen(blacklist_ext_fn_attrs(attr1, attr2))]` pattern. @itegulov wdyt?

Also, one other thought is that `blocklist` naming would be more clear, just as an alternative to consider.

mooori:
> I can't think of any attribute that a developer would want to keep, seems like there might be another option of just removing the attributes and if developers mention breakages we could do something like this?

Yes, also blacklisting is an extra step that has to be taken by contract developers that use plugin attributes. So not having to do this because attributes aren't forwarded to `_Ext` methods would be preferred, from that point of view.

mooori:
Closing this since it shouldn't be required anymore now that #959 is merged.

If there'll ever be a need to allow contracts dev to whitelist some attributes to forward to `_Ext` methods, the blacklist feature proposed here could be turned into a whitelist feature.

PR #954, opened 2022-11-02
chore: use near-abi 0.3.0

itegulov:
Related: https://github.com/near/cargo-near/pull/83

### DIFF
--- a/examples/adder/Cargo.toml
+++ b/examples/adder/Cargo.toml
@@ -17,7 +17,7 @@ workspaces = "0.5.0"
 serde_json = "1.0"
 tokio = { version = "1.14", features = ["full"] }
 anyhow = "1.0"
-near-abi = "0.1.0-pre.0"
+near-abi = { version = "0.3.0", features = ["__chunked-entries"] }
 zstd = "0.11"
 
 [profile.release]
--- a/examples/adder/Cargo.toml
+++ b/examples/adder/Cargo.toml
@@ -17,7 +17,7 @@ workspaces = "0.5.0"
 serde_json = "1.0"
 tokio = { version = "1.14", features = ["full"] }
 anyhow = "1.0"
-near-abi = { version = "0.3.0", features = ["__chunked-entries"] }
+near-abi = "0.3.0"
 zstd = "0.11"
 
 [profile.release]
--- a/examples/adder/src/lib.rs
+++ b/examples/adder/src/lib.rs
@@ -62,7 +62,7 @@ mod tests {
         let abi_root =
             serde_json::from_slice::<AbiRoot>(&zstd::decode_all(&res.result[..])?).unwrap();
 
-        assert_eq!(abi_root.schema_version, "0.1.0");
+        assert_eq!(abi_root.schema_version, "0.3.0");
         assert_eq!(abi_root.metadata.name, Some("adder".to_string()));
         assert_eq!(abi_root.metadata.version, Some("0.1.0".to_string()));
         assert_eq!(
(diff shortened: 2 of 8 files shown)


near-sdk-macros/src/core_impl/abi/abi_generator.rs
```diff
@@ -87,13 +87,27 @@ impl ImplItemMethodInfo {
...
-            &self.attr_signature_info.method_type,
-            &MethodType::Init | &MethodType::InitIgnoreState
-        );
-        let AttrSigInfo { is_payable, is_private, is_handles_result, .. } =
-            self.attr_signature_info;
+        let mut modifiers = vec![];
+        let kind = match &self.attr_signature_info.method_type {
+            &MethodType::View => quote! { near_sdk::__private::AbiFunctionKind::View },
+            &MethodType::Regular => {
+                quote! { near_sdk::__private::AbiFunctionKind::Call }
```
miraclx, line 94: I think it's worth noting for clarity that until https://github.com/near/near-sdk-rs/pull/890 is merged, the ABI will represent `self`-less functions as call functions. Meaning clients will think they have to send signed transactions and not simply making view calls. Then again, I don't have any stats on how often these are used or if at all, so it might be a trivial nit.

examples/adder/Cargo.toml
```diff
@@ -17,7 +17,7 @@ workspaces = "0.5.0"
 serde_json = "1.0"
 tokio = { version = "1.14", features = ["full"] }
 anyhow = "1.0"
-near-abi = "0.1.0-pre.0"
+near-abi = { version = "0.3.0", features = ["__chunked-entries"] }
```
miraclx, line 20: This feature isn't part of the public API, no example should use it directly. Doesn't look like it's needed here anyway, compiles fine without it.

miraclx (changes requested):

examples/adder/Cargo.toml
itegulov: oh, yeah, added by accident

near-sdk-macros/src/core_impl/abi/abi_generator.rs
miraclx: Yep!

examples/adder/src/lib.rs
```diff
@@ -75,13 +75,18 @@ mod tests {
...
+        assert_eq!(add_function.kind, AbiFunctionKind::View);
+        assert_eq!(add_function.modifiers, vec![]);
+        match &add_function.params {
+            AbiParameters::Json { args } => {
+                assert_eq!(args.len(), 2);
+                assert_eq!(args[0].name, "a".to_string());
+                assert_eq!(args[1].name, "b".to_string());
+            }
+            AbiParameters::Borsh { .. } => {
+                assert!(false);
```
miraclx, line 87: _nit_: [`unreachable!()`](https://doc.rust-lang.org/std/macro.unreachable.html)?

miraclx (approved):

examples/adder/src/lib.rs
itegulov: Well, it is not actually unreachable if the thing we are trying to test happens to have borsh serialization though

miraclx: Hm, but isn't that the point of the test? Asserting that it is JSON serialized and not borsh?

itegulov: Yeah, but `unreacheable!` to me is something you know (statically) will never be reached, but can't convince compiler to be the case. This test, on the other hand, depends on some external input (a wasm file) and tests its properties.

This is arguing very subtle semantics and honestly it feels like `AbiParameters` should just have `json()?` and `borsh()?` methods instead of this match/fail pattern.

miraclx: That's fair, wasn't a strong opinion. I'm not sure if we're at the point where we need those helper methods, so we can proceed as-is. 👍🏽

austinabell: In that case, still seems like panic!() might be more readable?

austinabell (approved):

ChaoticTempest (approved):

PR #959, opened 2022-11-08
Stop forwarding all non-bindgen attributes to _Ext methods

mooori:
As mentioned in [this comment](https://github.com/near/near-sdk-rs/pull/952#pullrequestreview-1165458157), non-bindgen attributes on `Contract` methods shouldn't be forwarded to `ContractExt` methods.

In some cases it might lead to compilation errors or unintended behavior, see #952 for example.

**Note**: I assume technically that's a breaking change - but then, it's hard to think of a case where contract developers want these attributes to be forwarded (besides `cfg`).

### DIFF
--- a/near-sdk-macros/src/core_impl/code_generator/ext.rs
+++ b/near-sdk-macros/src/core_impl/code_generator/ext.rs
@@ -1,7 +1,7 @@
 use crate::core_impl::{serializer, AttrSigInfo};
-use proc_macro2::{Ident, TokenStream as TokenStream2};
+use proc_macro2::{Ident, Span, TokenStream as TokenStream2};
 use quote::{format_ident, quote, ToTokens};
-use syn::{Generics, Signature};
+use syn::{parse_quote, Attribute, Generics, Path, Signature};
 
 /// Generates inner ext code for structs and modules. If intended for a struct, generic details
 /// for the struct should be passed in through `generic_details` and the `ext` method will be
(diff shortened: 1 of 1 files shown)


austinabell (approved):
I think it's more correct to not forward on the attributes. I'm inclined to include this in our release going out today or tomorrow, anything else you wanted to change about this?

Edit: and we can consider #952 in a future change, but I'd like to avoid adding extra semantics if possible

mooori:
Thanks for the quick review, would be great if that could make it into the upcoming release. There's nothing left to do from my side (just kept it in draft mode while CI was running).

_Not_ forwarding any attributes also seems preferable to me. However that breaks this compilation test:

https://github.com/near/near-sdk-rs/blob/557c22ca9f19981178c159f40f11cd2ceea68c39/near-sdk/compilation_tests/cond_compilation.rs#L12-L22

Without forwarding `#[cfg(...)]` the methods `new` and `inc` will be defined twice for `Incrementer`.

How about merging as is so `compilation_tests/cond_compilation` needs no special treatment? Then we could:

- Implement a whitelist feature (by "inverting" #952).
- Once that's done:
  - Remove the code which forwards `cfg` by default.
  - Use the whitelist feature to make `cond_compilation.rs` pass, kind of like this:

```rust
#[near_bindgen(ext_fn_attr_whitelist(cfg))]
impl Incrementer { /* */ }
```

austinabell:
> _Not_ forwarding any attributes also seems preferable to me

Possibly, I don't see an issue with forwarding cfg options. This seems more intuitive to me. Why does this seem preferable not to forward anything?

mooori:
> Why does this seem preferable not to forward anything?

Now there's a whitelist that's kind of hidden from contract devs and I'm not sure if they _always_ want to forward cfg attributes. Not forwarding cfg by default would make forwarding more explicit and allow to _not_ forward cfg if it's not desired.

Though if the only expected use case for cfg attributes on contract methods is something like in `compilation_tests/cond_compilation.rs`, then forwarding cfg by default is probably the most convenient option for contract devs.

PR #960, opened 2022-11-10
fix: Error const reference without default features

austinabell:
Bit sneaky this slipped in, but added some checks to the CI to make sure this doesn't fail in the future


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+## [4.1.1] - 2022-11-10
```
miraclx, line 5: missing link here

miraclx (approved):

CHANGELOG.md
austinabell: of course, forgot to update that part! Thanks :)

PR #962, opened 2022-11-11
Fix empty owner tokens `start_index` error

YellingOilbird:
When the user's `token_set` is empty this method panics now in case of use
```
{
        ...
        from_index: None,
        limit: None,
}
```
But i propose for an empty list returned in this situation


austinabell (approved):
that's valid for sure. I wonder why there is the require below is there at all. Going to pull this in since it's just for a view call and this is minimally changing from previous

austinabell:
thank you!

Issue #868, opened 2022-07-15
Deprecate `LegacyTreeMap`


austinabell:
Probably best to mark this as deprecated so that it doesn't accidentally get used. If anyone is using that type and can't update, they can just suppress the warnings

```rust
    #[deprecated(since = "4.1.0", note = "...")]
```

PR #963, opened 2022-11-15
Depreciated legacy tree map  #868

DavidM-D:
I have the alternative of marking each individual use of legacy tree map in `legacy_tree_map.rs` as depreciated rather than the whole module, but since the whole source file is depreciated I figured it wasn't worth the effort

### DIFF
--- a/near-sdk/src/collections/legacy_tree_map.rs
+++ b/near-sdk/src/collections/legacy_tree_map.rs
@@ -1,6 +1,8 @@
 //! Legacy `TreeMap` implementation that is using `UnorderedMap`.
 //! DEPRECATED. This implementation is deprecated and may be removed in the future.
 #![allow(clippy::all)]
+// This suppresses the depreciation warnings for uses of LegacyTreeMap in this module
+#![allow(deprecated)]
 
 use borsh::{BorshDeserialize, BorshSerialize};
 use std::ops::Bound;
@@ -18,6 +20,7 @@ use crate::IntoStorageKey;
 /// - `above`/`below`:          O(log(N))
 /// - `range` of K elements:    O(Klog(N))
 ///
+#[deprecated(since = "4.1.0", note = "Use tree_map::TreeMap instead")]
 #[derive(BorshSerialize, BorshDeserialize)]
 pub struct LegacyTreeMap<K, V> {
     root: u64,
--- a/near-sdk/src/collections/mod.rs
+++ b/near-sdk/src/collections/mod.rs
@@ -39,6 +39,7 @@
 //! that seemlessly integrated with the rest of the Rust standard library.
 
 mod legacy_tree_map;
+#[allow(deprecated)]
 pub use legacy_tree_map::LegacyTreeMap;
 
 mod lookup_map;


near-sdk/src/collections/legacy_tree_map.rs
```diff
@@ -18,6 +20,7 @@ use crate::IntoStorageKey;
 /// - `above`/`below`:          O(log(N))
 /// - `range` of K elements:    O(Klog(N))
 ///
+#[deprecated(since = "4.1.0", note = "Use tree_map::TreeMap instead")]
```
austinabell, line 23: Unfortunately, it's not that simple. Just swapping to use this type is a state-breaking change. At least the note should indicate this.

Also, the `tree_map` module is not exposed, so you might want to change this to say `near_sdk::collections::TreeMap` instead.

Optionally, this PR could introduce some migration of the old collection to the new one as a method, for example something like `LegacyTreeMap::to_tree_map(&self)` or something like this and then this comment can point to this to make the migration easy. This is less important, though, as I presume no one is really using this and if they absolutely want it, they can just ignore the deprecation warnings

DavidM-D: So many mistakes in one line!

DavidM-D:
An issue I'm having with this migration code is that it appears to be hitting the gas limit after successive runs. How should I go about resetting the gas limit after each run?

The gas limit also seems very low, it seems to fall over with 100 elements, is that to be expected or am I doing something stupid?

near-sdk/src/collections/legacy_tree_map.rs
```diff
@@ -585,6 +585,16 @@ where
         self.save(&n);
         self.tree.pop();
     }
+
+    pub fn to_tree_map<S: IntoStorageKey>(self, prefix: S) -> super::TreeMap<K, V> {
+        // We read this into the heap first so one can use the same prefix without getting into trouble
+        let elements: Vec<(K, V)> = self.into_iter().collect();
+        let mut tm = super::TreeMap::new(prefix);
+        for (k,v) in elements.iter() {
```
austinabell, line 593: This all happens synchronously, what would be a case that this would solve a race condition that using the iterator directly in the loop wouldn't?

near-sdk/src/collections/legacy_tree_map.rs
```diff
@@ -585,6 +585,16 @@ where
         self.save(&n);
         self.tree.pop();
     }
+
+    pub fn to_tree_map<S: IntoStorageKey>(self, prefix: S) -> super::TreeMap<K, V> {
```
austinabell, line 589:
```suggestion
    pub fn into_tree_map<S: IntoStorageKey>(self, prefix: S) -> super::TreeMap<K, V> {
```

Just to follow https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv naming conventions.

Also, would be nice to have a small comment describing this fn

near-sdk/src/collections/legacy_tree_map.rs
```diff
@@ -585,6 +585,16 @@ where
...
+        // We read this into the heap first so one can use the same prefix without getting into trouble
+        let elements: Vec<(K, V)> = self.into_iter().collect();
```
austinabell, line 591: So I'm curious about if this should be clearing the state when doing this conversion.

Seems like if the API is `(self) -> Self` it would be intuitive that the storage would be cleared and the new collection would be filled with this data.

Otherwise, we could consider this pattern of not clearing and have the method be something like:

```rust
fn copy_to_tree_map(&self, ..) -> super::TreeMap...
```

with a better name and whatnot.

Also, the other part that might seem opinionated is if the prefix is added as a parameter here. This makes sense if we go the copy route, but if the API is for an owned `self`, it might be more intuitive to just remove the prefix and re-use the one from which you're clearing.

To summarize, my intuitions are that this should either:
- Remove the prefix from parameters, re-use and clear the state, put state back in at the same prefix
- Change this API to `&self` and make it clear that the elements are not being removed from storage
  - Clearly document this fact and make it clear to devs that they should not use the same prefix as before (seems like a footgun and also requires devs to manually `clear` the state of the old collection)
- Another option for fun: Just create a `drain` iterator that clears and gives all key-value pairs back. Let the user decide if they want to clear the elements and put it into another collection or to just use the current iterator and use that to just fill another collection
  - This gives more freedom to migrate this state into other data structures, rather than just the `collections::ltm` -> `collections::tm` path
  
Seems like it might be good to get another opinion on this. @itegulov, you might be most familiar with these APIs, wdyt?

austinabell:
> An issue I'm having with this migration code is that it appears to be hitting the gas limit after successive runs. How should I go about resetting the gas limit after each run?
> 
> The gas limit also seems very low, it seems to fall over with 100 elements, is that to be expected or am I doing something stupid?

That sounds expected. This tree-map is very inefficient. I wonder if that's even more reason to do a `drain(..)` iterator for this so that the migration could be done in chunks.

near-sdk/src/collections/legacy_tree_map.rs
```diff
@@ -18,6 +20,10 @@ use crate::IntoStorageKey;
 /// - `above`/`below`:          O(log(N))
 /// - `range` of K elements:    O(Klog(N))
 ///
+#[deprecated(
+    since = "4.1.0",
+    note = "Use near_sdk::collections::TreeMap, if you have an existing LegacyTreeMap use LegacyTreeMap::to_tree_map to migrate"
```
austinabell, line 25: So this comment is out of date now, just remove the comment about migration and this is good

near-sdk/src/collections/legacy_tree_map.rs
```diff
@@ -18,6 +20,7 @@ use crate::IntoStorageKey;
 /// - `above`/`below`:          O(log(N))
 /// - `range` of K elements:    O(Klog(N))
 ///
+#[deprecated(since = "4.1.0", note = "Use near_sdk::collections::TreeMap")]
```
austinabell, line 23: noting that the since "4.1.0" is a bit unideal, but given we don't really have a plan what will be the next minor/major version, and no one uses this, it's probably fine

austinabell (approved):

near-sdk/src/collections/legacy_tree_map.rs
DavidM-D: So yeah, how do we feel about depreciations? Are the major release only, or do we just do it whenever?

PR #964, opened 2022-11-17
perf (`TreeMap.range`): Update the TreeMap->Range logic

### LINKED ISSUE #928: Optimize `TreeMap` range query implementation
`TreeMap` is awkward to use as its range query complexity is `O(Klog(N))`, which means iterating it is very costly. This happens because `tree_map::iter::Keys` holds a bounded range defining the iterator elements, but not their position in the tree:

```rust
pub struct Keys<'a, K: 'a>
where
    K: BorshSerialize + BorshDeserialize + Ord,
{
    tree: &'a Tree<K>,
    length: u32,
    min: Bound<&'a K>,
    max: Bound<&'a K>,
}
```

So every time we search for the next (or previous) element in the tree we have to search it from the top of the tree:
```rust
fn next(&mut self) -> Option<&'a K> {
    ...
    let next = next_asc(self.tree, self.min);
    ...
}

fn next_asc<'a, K>(tree: &'a Tree<K>, bound: Bound<&'a K>) -> Option<&'a K>
where
    K: BorshSerialize + Ord + BorshDeserialize,
{
    match bound {
        Bound::Unbounded => tree.min(),
        Bound::Included(bound) => tree.ceil_key(bound),
        Bound::Excluded(bound) => tree.higher(bound),
    }
}
```

This does not have to be the case, the range query can be answered in `O(K + log(N))` if we were to save the actual bounding tree nodes and find the next (or previous) element in `O(1)` (amortized). This cannot be easily done with the current implementation however as there is no way to traverse up to the parent for a given node. So a decent refactoring is required, but we could actually avoid storing the parent index in the `Node` struct if we hold a `cur_path: Vec<FreeListIndex>` in the iterator instead (i.e. a list of node indices we have gone through to reach the current node).

Part of #818

askoa:
Closes #928 

Update the `TreeMap->Range` logic to keep track of current path. See the Issue description for detailed explanation.

### DIFF
--- a/near-sdk/src/store/tree_map/iter.rs
+++ b/near-sdk/src/store/tree_map/iter.rs
@@ -600,8 +600,7 @@ where
         //If the last returned key does not have right node then return the
         //last value in the stack.
         if seen.is_none() && !self.stack_asc.is_empty() {
-            let len = self.stack_asc.len();
-            if let Some(result_idx) = self.stack_asc.get(len - 1) {
+            if let Some(result_idx) = self.stack_asc.last() {
                 seen = self.tree.node(*result_idx).map(|f| &f.key);
             }
         }
@@ -643,8 +642,7 @@ where
         //If the last returned key does not have left node then return the
         //last value in the stack.
         if seen.is_none() && !self.stack_desc.is_empty() {
-            let len = self.stack_desc.len();
-            if let Some(result_idx) = self.stack_desc.get(len - 1) {
+            if let Some(result_idx) = self.stack_desc.last() {
                 seen = self.tree.node(*result_idx).map(|f| &f.key);
             }
         }
--- a/near-sdk/src/store/tree_map/iter.rs
+++ b/near-sdk/src/store/tree_map/iter.rs
@@ -303,8 +303,8 @@ where
 {
     tree: &'a Tree<K>,
     length: u32,
-    min: Find<&'a K>,
-    max: Find<&'a K>,
+    min: FindUnbounded,
+    max: FindUnbounded,
     //The last element in the stack is the latest value returned by the iterator
     stack_asc: Vec<FreeListIndex>,
     stack_desc: Vec<FreeListIndex>,
@@ -318,8 +318,8 @@ where
         Self {
             tree,
             length: tree.nodes.len(),
-            min: Find::First { bound: Bound::Unbounded },
-            max: Find::First { bound: Bound::Unbounded },
+            min: FindUnbounded::First,
+            max: FindUnbounded::First,
             stack_asc: Vec::new(),
             stack_desc: Vec::new(),
         }
@@ -339,13 +339,13 @@ where
         }
 
         let next = match self.min {
-            Find::First { bound: min } => self.find_min(self.tree.root.as_ref(), min),
-            Find::Next { bound: _ } => self.find_next_asc(),
+            FindUnbounded::First => self.find_min(self.tree.root.as_ref()),
+            FindUnbounded::Next => self.find_next_asc(),
         };
 
-        if let Some(next) = next {
+        if next.is_some() {
             // Update minimum bound.
-            self.min = Find::Next { bound: Bound::Excluded(next) };
+            self.min = FindUnbounded::Next;
 
             // Decrease count of potential elements
             self.length -= 1;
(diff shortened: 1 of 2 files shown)


near-sdk/src/store/tree_map/iter.rs
```diff
@@ -507,6 +558,100 @@ where
...
+                    Some(rgt) => self.find_min(Some(&rgt), Bound::Unbounded),
+                    None => None,
+                }
+            }
+        }
+        //If the last returned key does not have right node then return the
+        //last value in the stack.
+        if seen.is_none() && !self.stack_asc.is_empty() {
+            let len = self.stack_asc.len();
+            if let Some(result_idx) = self.stack_asc.get(len - 1) {
```
itegulov, line 604:
```suggestion
            if let Some(result_idx) = self.stack_asc.last() {
```

near-sdk/src/store/tree_map/iter.rs
```diff
@@ -507,6 +558,100 @@ where
...
+                    Some(lft) => self.find_max(Some(&lft), Bound::Unbounded),
+                    None => None,
+                }
+            }
+        }
+        //If the last returned key does not have left node then return the
+        //last value in the stack.
+        if seen.is_none() && !self.stack_desc.is_empty() {
+            let len = self.stack_desc.len();
+            if let Some(result_idx) = self.stack_desc.get(len - 1) {
```
itegulov, line 647:
```suggestion
            if let Some(result_idx) = self.stack_desc.last() {
```

itegulov (commented):
The implementation looks solid, great job! The same should be done for `Keys` (in the same file) and we should be good.

askoa:
>  The same should be done for `Keys` (in the same file) and we should be good.

Done.

askoa:
The build failure seems unrelated to this PR.

```
thread 'actix-rt|system:0|arbiter:1' panicked at 'failed to start listening on server_addr=0.0.0.0:19718 e=Os { code: 98, kind: AddrInUse, message: "Address already in use" }', chain/network/src/peer_manager/peer_manager_actor.rs:302:29
```

itegulov (approved):
Ran some benchmarks and it does seem like this implementation's asymptotic is better than the old one. Source: https://github.com/itegulov/storage_collections_bench/tree/tree-map-iter. The farthest I could test without hitting the max gas limit was ~32 insert operations + ~32 iterations, where this implementation is 20% more efficient.

@austinabell you have the most context on this, so could you take a look before we merge this just in case?

itegulov:
> The build failure seems unrelated to this PR.

Yeah, don't worry about it. Just some flaky workspaces behaviour.

near-sdk/src/store/tree_map/iter.rs
```diff
@@ -507,6 +638,98 @@ where
...
+        //If the last returned key does not have left node then return the
+        //last value in the stack.
+        if seen.is_none() && !self.stack_desc.is_empty() {
+            if let Some(result_idx) = self.stack_desc.last() {
+                seen = self.tree.node(*result_idx).map(|f| &f.key);
+            }
+        }
+        seen
+    }
+}
```
austinabell, line 731: At a glance, this logic seems identical to that for `Keys`. What is preventing it from being abstracted to be shared between the two?

austinabell (approved):
Everything should be good to go. I think existing tests are exhaustive enough and if the benchmarks are showing improvement, that's probably good enough!

The only comment is just on the duplication of the logic, but that refactor could happen in a follow-up PR if needed

austinabell:
> where this implementation is 20% more efficient.

Honestly would have expected this to be more of an optimization. I suppose because now since the values are cached anyway, it's only extra computing that is the overhead whereas if this was done for `sdk::collections::TreeMap` the performance gain would be much more? (Asking out of curiosity)

near-sdk/src/store/tree_map/iter.rs
askoa: `Keys` seems to be special case of `KeysRange` and can be easily replaced. May be a followup Issue to remove `Keys` and replace with `KeysRange` might be a better solution. What do you think?

austinabell: yeah, that's fine, can do in a follow up!

@itegulov feel free to merge this if you're happy with it, unsure if you had any other thoughts since you seem to be the stakeholder for that issue

askoa: Created #987 to track the issue.

itegulov:
@askoa thanks again for handling this!

PR #968, opened 2022-11-22
refactor: switch all arithmetic in standards to explicitly checked

austinabell:
I made the statement today that we do checked arithmetic for everything, so this PR intends to be extra careful.

Ideally and in most cases, overflow checks should be done through Rust (and recommended https://docs.near.org/sdk/rust/best-practices#enable-overflow-checks and https://docs.near.org/sdk/rust/introduction#create-a-new-project), but this is done as an extra-safety measure

Would be great to get another set of eyes to see if I switched one unnecessarily or something like this

### DIFF
--- a/near-contract-standards/src/non_fungible_token/approval/approval_impl.rs
+++ b/near-contract-standards/src/non_fungible_token/approval/approval_impl.rs
@@ -8,6 +8,7 @@ use crate::non_fungible_token::utils::{
     refund_approved_account_ids_iter, refund_deposit,
 };
 use crate::non_fungible_token::NonFungibleToken;
+use crate::ERR_ARITHMETIC_OVERFLOW;
 use near_sdk::{assert_one_yocto, env, require, AccountId, Gas, Promise};
 
 const GAS_FOR_NFT_APPROVE: Gas = Gas(10_000_000_000_000);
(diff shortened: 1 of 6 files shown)


near-contract-standards/src/non_fungible_token/approval/approval_impl.rs
```diff
@@ -59,7 +61,7 @@ impl NonFungibleTokenApproval for NonFungibleToken {
         // if given `msg`, schedule call to `nft_on_approve` and return it. Else, return None.
         msg.map(|msg| {
             ext_nft_approval_receiver::ext(account_id)
-                .with_static_gas(env::prepaid_gas() - GAS_FOR_NFT_APPROVE)
+                .with_static_gas(Gas(env::prepaid_gas().0.saturating_sub(GAS_FOR_NFT_APPROVE.0)))
```
itegulov, line 64: I think you missed a similar place here: https://github.com/near/near-sdk-rs/blob/master/near-contract-standards/src/non_fungible_token/core/core_impl.rs#L399

itegulov (commented):
LGTM, but weren't we trying to move away from using `expect`?

near-contract-standards/src/non_fungible_token/approval/approval_impl.rs
austinabell: No, that one is checked because of the require above it

austinabell:
Closing because opinionated and I'm not working on this anymore

Issue #970, opened 2022-11-22
cargo test with all features fails


austinabell:
Just opening for tracking, I don't think there is a clean way around this for what's needed in cargo-near, but the tests fail if `__abi-embed` feature is set.

This can be a bit of friction for developers contributing using `cargo t --all-features` or some config that enables these features by default (which some configs do).

miraclx:
One option is to add a feature flag that negates it, `__abi-embed-unset` which will also get activated by `--all-features`😄

Or, and I'm leaning toward this, a build script that activates another feature flag that cargo is unaware of when both the `__abi-embed` feature flag is set and the env is defined.

miraclx:
Also, looks like `--all-features` has been failing since 9ab6df8ed35e67c247299059d8277c2170a0a91b.

PR #971, opened 2022-11-23
fix: `__abi-embed` compilation error

miraclx:
`--all-features` causes cargo to activate all features in the `Cargo.toml` file.

This patch introduces an intermediary flag. Unknown to cargo, but injected with a build script if;

- The `__abi-embed` feature flag is activated.
- The `CARGO_NEAR_ABI_PATH` env is defined.

This way, it's never implicitly activated. Resolving the compilation error in #970.

This also means, we gravitate from this compilation error:

CleanShot 2022-11-24 at 03 34 06@2x

.. to this compilation warning ..

CleanShot 2022-11-24 at 03 31 04@2x

### DIFF
--- a/near-sdk-macros/build.rs
+++ b/near-sdk-macros/build.rs
@@ -1,0 +1,10 @@
+fn main() {
+    if cfg!(feature = "__abi-embed") {
+        if option_env!("CARGO_NEAR_ABI_PATH").is_some() {
+            println!("cargo:rustc-cfg=feature=\"__abi-embed-checked\"");
+        } else {
+            println!("cargo:warning=the `__abi-embed` feature flag is private and should not be activated manually, ignoring");
+            println!("cargo:warning=\x1b[1mhelp\x1b[0m: consider using https://github.com/near/cargo-near");
+        }
+    }
+}
--- a/near-sdk-macros/Cargo.toml
+++ b/near-sdk-macros/Cargo.toml
@@ -16,11 +16,11 @@ proc-macro = true
 
 [dependencies]
 proc-macro2 = "1.0"
-syn = {version = "1", features = ["full", "fold", "extra-traits", "visit"] }
+syn = { version = "1", features = ["full", "fold", "extra-traits", "visit"] }
 quote = "1.0"
 Inflector = { version = "0.11.4", default-features = false, features = [] }
 
 [features]
 abi = []
-__abi-embed = []
-__abi-generate = []
+__abi-embed = ["abi"]
+__abi-generate = ["abi"]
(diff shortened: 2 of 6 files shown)


itegulov (commented):
Wait, so if we can just depend on features that are not declared in `Cargo.toml`, couldn't we just remove `__abi-embed` and `__abi-generate` from `Cargo.toml`?

miraclx:
`cargo` only lets us enable features (with `--features`) that exist in the manifest. Any other, results in this error:

```console
$ cargo build --features near-sdk/__abi-generate
error: failed to select a version for `near-sdk`.
    ... required by package `adder v0.1.0 (/Users/miraclx/git/near/near-sdk-rs/examples/adder)`
versions that meet the requirements `*` (locked to 4.1.1) are: 4.1.1

the package `adder` depends on `near-sdk`, with features: `__abi-generate` but `near-sdk` does not have these features.

failed to select a version for `near-sdk` which could resolve this conflict
```

itegulov (approved):
A little hacky, but looks reasonable to me! 👍

near-sdk-macros/build.rs
```diff
@@ -0,0 +1,10 @@
+fn main() {
+    if cfg!(feature = "__abi-embed") {
+        if option_env!("CARGO_NEAR_ABI_PATH").is_some() {
+            println!("cargo:rustc-cfg=feature=\"__abi-embed-checked\"");
```
austinabell, line 4: curious -- I wonder if it is possible to manually pass the feature flag to the compiler from outside of the build script? This way we would be able to remove the pseudo-private flags altogether since the overhead doesn't matter for cargo-near?

miraclx: You mean through `RUSTFLAGS`? That applies to all `rustc` instances cargo spawns throughout its execution. Can't guarantee anything else (deps) doesn't change its behavior with that config. But through `cargo`, we constrain the activation to the crate in particular. And then only conditionally activate `__abi-embed-checked`, also specifically for the crate.

An alternative I experimented with early on, was to use [`cargo-near` as `RUST_WRAPPER`](https://github.com/near/cargo-near/blob/bcf9ae57ec3302a979f450b200113f74c61a97c7/cargo-near/src/contract.rs#L81), so when generating ABI, it embeds a unique identifier for the sdk in it's chunks. And when embedding the ABI, it selectively [injects the embed flag](https://github.com/near/cargo-near/blob/bcf9ae57ec3302a979f450b200113f74c61a97c7/cargo-near/src/rustc_wrapper.rs#L45) specifically for the SDK.

But that was cumbersome, so I stripped it.

austinabell: > Can't guarantee anything else (deps) maintains its behavior with that flag activated

What would be the concern there, that there would be a dependency used in a contract being built that for some reason uses a `__abi-embed` feature flag?

I think this solution is good enough for now, but maybe we could have at least a tracking issue to see if we can get rid of these pseudo-private feature flags

miraclx: > What would be the concern there, that there would be a dependency used in a contract being built that for some reason uses a __abi-embed feature flag?

Yup, that's the primary concern. Arguably, a good approach would be to localize effects instead of globalizing.

PR #975, opened 2022-11-24
fix: make event `emit` public

danielwpz:


austinabell (approved):
ah, yes, this is very valid. Do you mind updating the test case in https://github.com/near/near-sdk-rs/blob/master/near-sdk/tests/event_tests.rs to make sure we don't remove this in the future?

sth like:
```rust
mod private {
    use super::*;
    
    #[near_bindgen(event_json(standard = "another_standard"))]
    pub enum AnotherEvent {
        #[event_version("1.0.0")]
        Test,
    }
}

...

private::AnotherEvent::Test.emit()
```

which fails on current master

danielwpz:
@austinabell sure, will do!

austinabell:
thanks for the contribution!

Issue #871, opened 2022-07-26
Document 0 allowance interaction with `add_access_key` methods


austinabell:
This pattern is a bit of a foot gun because someone could accidentally specify 0 allowances (which is handled as unlimited by the runtime), which can lead to the access key recipient draining the funds by spamming the function.

The short solution is documenting the interaction with `Balance` being zero for `add_access_key` and `add_access_key_with_nonce` of the `Promise` API and `env::promise_batch_action_add_key_with_function_call`.

Long-term considerations is the API can be changed to `Option<NonZeroU128>` to be the least error prone or `Option<Balance>` to be slightly more clear about what it is without being unergonomic. Another option would be to introduce an enum for allowance that looks something like this:

```rust
enum Allowance {
    Unbounded,
    Bounded(Balance),
}
```

PR #976, opened 2022-11-25
Took out a footgun with allowances

DavidM-D:
Address's #871 .

I expect this is a reasonable final state for the library, but how should I manage the migration?

Create new function names for the new behavior and depreciate the old endpoints?

### DIFF
--- a/near-sdk/src/environment/env.rs
+++ b/near-sdk/src/environment/env.rs
@@ -557,10 +557,7 @@ pub fn promise_batch_action_add_key_with_full_access(
 
 /// This is a short lived function while we migrate between the Balance and the allowance type
 pub(crate) fn migrate_to_allowance(allowance: Balance) -> Allowance {
-    match allowance {
-        0 => Allowance::Unlimited,
-        x => Allowance::limited(x).expect("This must be non zero"),
-    }
+    Allowance::limited(allowance).unwrap_or(Allowance::Unlimited)
 }
 
 #[deprecated(since = "4.1.1", note = "Use add_access_key_allowance instead")]
--- a/near-sdk/src/environment/env.rs
+++ b/near-sdk/src/environment/env.rs
@@ -560,7 +560,7 @@ pub(crate) fn migrate_to_allowance(allowance: Balance) -> Allowance {
     Allowance::limited(allowance).unwrap_or(Allowance::Unlimited)
 }
 
-#[deprecated(since = "5.0", note = "Use add_access_key_allowance instead")]
+#[deprecated(since = "5.0.0", note = "Use add_access_key_allowance instead")]
 pub fn promise_batch_action_add_key_with_function_call(
     promise_index: PromiseIndex,
     public_key: &PublicKey,
--- a/near-sdk/src/environment/env.rs
+++ b/near-sdk/src/environment/env.rs
@@ -560,7 +560,7 @@ pub(crate) fn migrate_to_allowance(allowance: Balance) -> Allowance {
     Allowance::limited(allowance).unwrap_or(Allowance::Unlimited)
 }
 
-#[deprecated(since = "4.1.1", note = "Use add_access_key_allowance instead")]
+#[deprecated(since = "5.0", note = "Use add_access_key_allowance instead")]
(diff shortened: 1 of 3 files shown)


austinabell:
> Create new function names for the new behavior and depreciate the old endpoints?

That would be the safest way, for sure. A little unfortunate that this small change touches three different endpoints currently. Unclear what the alternative would be, other than just making the change with a major release and documenting the migration/accepting breakages.

near-sdk/src/environment/env.rs
```diff
@@ -553,15 +554,48 @@ pub fn promise_batch_action_add_key_with_full_access(
         )
     }
 }
+
+/// This is a short lived function while we migrate between the Balance and the allowance type
+pub(crate) fn migrate_to_allowance(allowance: Balance) -> Allowance {
+    match allowance {
+        0 => Allowance::Unlimited,
+        x => Allowance::limited(x).expect("This must be non zero"),
+    }
```
austinabell, line 563: feels like this should be:

```suggestion
    Allowance::limited(allowance).unwrap_or(Allowance::Unlimited)
```

Sorry if that doesn't compile, I didn't check it, but want to avoid the expect which comes with unnecessary bloat to compiled code size

near-sdk/src/promise.rs
```diff
@@ -2,10 +2,31 @@ use borsh::BorshSchema;
...
+use crate::env::migrate_to_allowance;
 use crate::{AccountId, Balance, Gas, GasWeight, PromiseIndex, PublicKey};
 
+/// Allow an access key to spend either an unlimited or limited amount of gas
+// This wrapper prevents incorrect construction
+#[derive(Clone, Copy)]
+pub enum Allowance {
+    Unlimited,
+    Limited(NonZeroU128),
+}
```
ChaoticTempest, line 17: I wonder if we should add a note that a zero allowance is allowed outside of promises. A zero allowance is allowed with add keys calls through RPC into the chain. This is just a special case of the host functions defining allowance as a ptr to a u64

austinabell: wait, what? You're saying you can create a function call access key without an allowance through rpc?

ChaoticTempest: yup. For example the following workspaces code:
```rust
use serde_json::json;
use workspaces::types::{KeyType, SecretKey};
use workspaces::{AccessKey, Account};

const STATUS_MSG_WASM_FILEPATH: &str = "./examples/res/status_message.wasm";

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    let worker = workspaces::sandbox().await?;
    let wasm = std::fs::read(STATUS_MSG_WASM_FILEPATH)?;
    let contract = worker.dev_deploy(&wasm).await?;

    let sk = SecretKey::from_seed(KeyType::ED25519, "test");
    let pk = sk.public_key();
    let ak = AccessKey::function_call_access(contract.id(), &["set_status"], Some(0));
    let account = Account::from_secret_key(contract.id().clone(), sk, &worker);

    contract
        .batch()
        .add_key(pk, ak)
        .transact()
        .await?
        .into_result()?;

    let outcome = account
        .call(contract.id(), "set_status")
        .args_json(json!({
            "message": "hello_world",
        }))
        .transact()
        .await?;
    println!("set_status: {:?}", outcome);

    Ok(())
}
```
would yield the error when calling into `set_status`:
```
Error: unable to broadcast the transaction to the network

Caused by:
    handler error: [An error happened during transaction execution: InvalidAccessKeyError(NotEnoughAllowance { account_id: AccountId("dev-20221130043518-92793473638831"), public_key: ed25519:DcA2MzgpJbrUATQLLceocVckhhAqrkingax4oJ9kZ847, allowance: 0, cost: 1561286367017332100000 })]
```

austinabell: honestly had a vague intuition that this path would use the same logic as the runtime host function. Yeah, could document the difference. Honestly, there could be a reason (albeit bad) to add a key with no allowance just to record it as on-chain data. It's how some protocols verify accounts given there is no arbitrary signing right now

austinabell (approved):
naming might be a bit confusing but I don't know of a better solution here

PR #977, opened 2022-11-25
Removed the not ready enum type

### LINKED ISSUE #796: Remove `NotReady` variant for `PromiseError`
This variant cannot be constructed by the runtime, so no need to have it within our types. We can always add it back in the future if the functionality gets added.

[https://github.com/near/near-sdk-rs/blob/c0219b80de590945d8a203be046150c23f296cf9/near-sdk/src/types/vm_types.rs#L39](https://github.com/near/near-sdk-rs/blob/c0219b80de590945d8a203be046150c23f296cf9/near-sdk/src/types/vm_types.rs#L39)

DavidM-D:
It can't be generated by the runtime fixes #796

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,94 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1667395993,
+        "narHash": "sha256-nuEHfE/LcWyuSWnS8t12N1wc105Qtau+/OdUAjtQ0rA=",
+        "owner": "numtide",
+        "repo": "flake-utils",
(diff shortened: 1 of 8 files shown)


austinabell (approved):
I think this is the correct way to go as it's unclear what the semantics would even look like if this was to be handled from the application level.

Would love another opinion on this. Also, we probably want to hold this PR until we are ready to make a major version release as this is a breaking change

itegulov:
Shouldn't we also remove it from https://github.com/near/near-sdk-rs/blob/master/near-sdk/src/types/vm_types.rs#L18?

DavidM-D:
🤦‍♂️

Issue #974, opened 2022-11-24
ABI generation fails with lifetimes


miraclx:
Similar to #972.

```rust
#[near_bindgen]
impl MyType {
    pub fn slice<'a>(data: &'a String) -> &'a str { ... }
}
```

```console
$ cargo build --features near-sdk/__abi-generate
   Compiling adder v0.1.0 (/Users/miraclx/git/near/near-sdk-rs/examples/adder)
error[E0261]: use of undeclared lifetime name `'a`
  --> src/lib.rs:3:17
   |
 3 |     pub fn slice<'a>(data: &'a String) -> &'a str {
   |                 -                          ^^ undeclared lifetime
   |                 |
   |                 help: consider introducing lifetime `'a` here: `<'a>`

For more information about this error, try `rustc --explain E0261`.```

**Comment by itegulov** [comment]
Just a note that the existing lifetime support is limited to top-level lifetimes. For example this does not compile:
```rust
pub fn slice<'a>(data: Cow<'a, [u8; 32]>) -> Cow<'a, [u8; 32]> {
    data
}
```

miraclx:
I think lifetimes in argument position is pretty much invalid form. As part of the return type, which is more reasonable, it compiles just fine.

itegulov:
@miraclx 

> I think lifetimes in argument position is pretty much invalid form.

Can you elaborate on why you think it is invalid form? The `foo` example above seems reasonable to me.

miraclx:
Well, really, my thought was: in the context of the code-generated contract API, reference to data that was deserialized just for you doesn't seem very useful. But yeah, the scope extends larger than this. Enough to be generalized, so it could technically be called by another method in the contract that only has a reference to pass to it, so that makes sense. You can ignore the original comment.

But, in that case, we should make an effort to resolve the issue you highlighted in https://github.com/near/near-sdk-rs/issues/974#issuecomment-1330048298.

Two options I've experimented with locally use `TokenStream` sanitization as an alternative to #982, like in #1001;

 - The first approach strips all lifetime patterns from type signatures.
   - `&'a Cow<'b, 'c, [u8; 32]>` -> `&Cow<[u8; 32]>`
 - While the second approach replaces all lifetime patterms with the implicit lifetime.
   - `&'a Cow<'b, 'c, [u8; 32]>` -> `&'_ Cow<'_, '_, [u8; 32]>`

[...]

PR #980, opened 2022-11-28
fix: prohibit NEAR function generics

### LINKED ISSUE #979: Use of generics returns a confusing error
We should consider returning a neater error message stating that generics aren't supported, instead of the confusing error it currently returns.

Especially since, it's perfectly valid code without the `#[near_bindgen]` attr.

```rust
#[near_bindgen]
impl MyType {
    pub fn method<T>(arg: T) -> T { ... }
}
```

```console
$ cargo build
   Compiling contract v0.1.0 (/path/to/contract)
error[E0401]: can't use generic parameters from outer function
  --> src/lib.rs:10:28
   |
10 |     pub fn method<T>(data: T) -> T {
   |            ---------       ^ use of generic parameter from outer function
   |            |      |
   |            |      type parameter from outer function
   |            help: try using a local generic parameter instead: `method<T, T>`

For more information about this error, try `rustc --explain E0401`.
error: could not compile `contract` due to previous error
```

But technically, it's clear what the hurdle will be here. It's understandably tricky disambiguating between generic identifiers and explicitly typed ones. Would be interesting to investigate.

itegulov:
Closes #979

### DIFF
--- a/near-sdk-macros/src/core_impl/code_generator/ext.rs
+++ b/near-sdk-macros/src/core_impl/code_generator/ext.rs
@@ -267,7 +267,7 @@ mod tests {
         let mut method: ImplItemMethod = parse_quote! {
             pub fn method(&self, k: &String) { }
         };
-        let method_info = ImplItemMethodInfo::new(&mut method, false,impl_type).unwrap().unwrap();
+        let method_info = ImplItemMethodInfo::new(&mut method, false, impl_type).unwrap().unwrap();
         let actual = generate_ext_function(&method_info.attr_signature_info);
         let expected = quote!(
             pub fn method(self, k: &String,) -> near_sdk::Promise {
@@ -298,7 +298,7 @@ mod tests {
         let mut method: ImplItemMethod = parse_quote! {
           pub fn borsh_test(&mut self, #[serializer(borsh)] a: String) {}
         };
-        let method_info = ImplItemMethodInfo::new(&mut method, false,impl_type).unwrap().unwrap();
+        let method_info = ImplItemMethodInfo::new(&mut method, false, impl_type).unwrap().unwrap();
         let actual = generate_ext_function(&method_info.attr_signature_info);
         let expected = quote!(
           pub fn borsh_test(self, a: String,) -> near_sdk::Promise {
--- a/near-sdk-macros/src/core_impl/code_generator/ext.rs
+++ b/near-sdk-macros/src/core_impl/code_generator/ext.rs
@@ -241,7 +241,7 @@ mod tests {
             #[warn(unused)]
             pub fn method(&self) { }
         };
-        let method_info = ImplItemMethodInfo::new(&mut method, impl_type).unwrap();
+        let method_info = ImplItemMethodInfo::new(&mut method, false,impl_type).unwrap().unwrap();
         let actual = generate_ext_function(&method_info.attr_signature_info);
 
         // Note: only whitelisted non-bindgen attributes are forwarded.
@@ -267,7 +267,7 @@ mod tests {
         let mut method: ImplItemMethod = parse_quote! {
             pub fn method(&self, k: &String) { }
         };
-        let method_info = ImplItemMethodInfo::new(&mut method, impl_type).unwrap();
+        let method_info = ImplItemMethodInfo::new(&mut method, false,impl_type).unwrap().unwrap();
         let actual = generate_ext_function(&method_info.attr_signature_info);
         let expected = quote!(
             pub fn method(self, k: &String,) -> near_sdk::Promise {
@@ -298,7 +298,7 @@ mod tests {
         let mut method: ImplItemMethod = parse_quote! {
           pub fn borsh_test(&mut self, #[serializer(borsh)] a: String) {}
         };
-        let method_info = ImplItemMethodInfo::new(&mut method, impl_type).unwrap();
+        let method_info = ImplItemMethodInfo::new(&mut method, false,impl_type).unwrap().unwrap();
         let actual = generate_ext_function(&method_info.attr_signature_info);
         let expected = quote!(
           pub fn borsh_test(self, a: String,) -> near_sdk::Promise {
--- a/near-sdk-macros/src/core_impl/code_generator/ext.rs
+++ b/near-sdk-macros/src/core_impl/code_generator/ext.rs
@@ -241,7 +241,7 @@ mod tests {
             #[warn(unused)]
             pub fn method(&self) { }
         };
-        let method_info = ImplItemMethodInfo::new(&mut method, false,impl_type).unwrap().unwrap();
+        let method_info = ImplItemMethodInfo::new(&mut method, false, impl_type).unwrap().unwrap();
         let actual = generate_ext_function(&method_info.attr_signature_info);
 
         // Note: only whitelisted non-bindgen attributes are forwarded.
--- a/near-sdk-macros/src/core_impl/info_extractor/attr_sig_info.rs
+++ b/near-sdk-macros/src/core_impl/info_extractor/attr_sig_info.rs
@@ -38,6 +38,12 @@ impl AttrSigInfo {
         original_attrs: &mut Vec<Attribute>,
         original_sig: &mut Signature,
     ) -> syn::Result<Self> {
+        if original_sig.generics.params.len() > 0 {
+            return Err(Error::new(
+                original_sig.generics.span(),
+                "Contract API is not allowed to have generics.",
+            ));
+        }
         if original_sig.asyncness.is_some() {
             return Err(Error::new(
                 original_sig.span(),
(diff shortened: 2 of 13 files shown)


itegulov:
Hmm, this also prohibits lifetimes (see #974). I can make it so that this only blocks certain types of generic parameters, but honestly not sure if there is a legitimate pattern for lifetime params at all.

near-sdk-macros/src/core_impl/info_extractor/attr_sig_info.rs
```diff
@@ -38,6 +38,12 @@ impl AttrSigInfo {
         original_attrs: &mut Vec<Attribute>,
         original_sig: &mut Signature,
     ) -> syn::Result<Self> {
+        if !original_sig.generics.params.is_empty() {
```
miraclx, line 41: Let's maybe only do this for `pub`-lic functions?

itegulov:
Decided to leave lifetimes out of this PR for now, we can discuss this separately under https://github.com/near/near-sdk-rs/issues/974.

This became more of a refactoring PR, but I feel like the current logic is backward from what it should be. I left async/variadic/extern checks for all impl methods (even private), but honestly not sure that is necessary. For example, an async function is compilable for wasm on its own, you just wouldn't be able to use it in any meaningful way since public functions can't be async. @austinabell you would probably have the most context on what the original intention of these checks was.

near-sdk-macros/src/core_impl/info_extractor/impl_item_method_info.rs
```diff
@@ -1,22 +1,48 @@
...
+        }
+        if sig.variadic.is_some() {
+            return Err(Error::new(
+                sig.span(),
+                "Contract API is not allowed to have variadic arguments.",
+            ));
+        }
+
+        Ok(())
+    }
```
miraclx, line 31: Aren't these checks useful for `TraitItemMethodInfo`?

near-sdk-macros/src/core_impl/code_generator/ext.rs
```diff
@@ -241,7 +241,7 @@ mod tests {
             #[warn(unused)]
             pub fn method(&self) { }
         };
-        let method_info = ImplItemMethodInfo::new(&mut method, impl_type).unwrap();
+        let method_info = ImplItemMethodInfo::new(&mut method, false,impl_type).unwrap().unwrap();
```
miraclx, line 244: curious, doesn't look like cargo-fmt acted on these..

```suggestion
        let method_info = ImplItemMethodInfo::new(&mut method, false, impl_type).unwrap().unwrap();
```

near-sdk-macros/src/core_impl/info_extractor/attr_sig_info.rs
```diff
@@ -38,25 +38,12 @@ impl AttrSigInfo {
         original_attrs: &mut Vec<Attribute>,
         original_sig: &mut Signature,
     ) -> syn::Result<Self> {
-        if original_sig.asyncness.is_some() {
+        if original_sig.generics.params.iter().any(|g| matches!(g, GenericParam::Type(_))) {
```
miraclx, line 41: I think `GenericParam::Const(_)` also qualifies for restriction here.

near-sdk/compilation_tests/generic_function.rs
```diff
@@ -0,0 +1,19 @@
...
+#[derive(Default, BorshDeserialize, BorshSerialize)]
+struct Ident {
+    value: u32,
+}
+
+#[near_bindgen]
+impl Ident {
+    pub fn is_ident<T>(&self, val: T) -> T {
+        val
+    }
```
miraclx, line 16: Once the logic is extended to disallow const generics, lets add a case here.

miraclx (commented):
> I left async/variadic/extern checks for all impl methods (even private), but honestly not sure that is necessary. For example, an async function is compilable for wasm on its own, you just wouldn't be able to use it in any meaningful way since public functions can't be async.

Eh, should be fine to retain those checks.

A small side effect of the current logic (or perhaps intentional?) is we're no longer checking them for `#[ext_contract]` traits. And while technically, neither async traits nor C-variadic methods are stable, explicit definition of support on our end would be nice.

near-sdk-macros/src/core_impl/code_generator/ext.rs
itegulov: Hmm, that's interesting. Looks like a bug, wondering if I can create a minimal reproduction.

near-sdk-macros/src/core_impl/info_extractor/impl_item_method_info.rs
itegulov: Yes, good catch. This also removes the side effect of not checking `#[ext_contract]` which you mentioned (was not intended). I think it makes more sense to prohibit that stuff until we find an actual use for them.

near-sdk/compilation_tests/generic_function.rs
itegulov: Created a separate test case so that we can check the error message (it only reports one at a time, probably because the macro errors work differently from regular compilation errors).

near-sdk-macros/src/core_impl/info_extractor/attr_sig_info.rs
```diff
@@ -38,25 +38,17 @@ impl AttrSigInfo {
...
+            .generics
+            .params
+            .iter()
+            .any(|g| matches!(g, GenericParam::Type(_) | GenericParam::Const(_)))
+        {
             return Err(Error::new(
-                original_sig.span(),
-                "Contract API is not allowed to be async.",
+                original_sig.generics.span(),
+                "Contract API is not allowed to have generics.",
```
miraclx, line 49: What do you think about reporting errors on a case-by-case basis, so it highlights the generic itself and not the `<` span?

We can use [`syn::Error::combine`](https://docs.rs/syn/latest/syn/parse/struct.Error.html#method.combine) to combine multiple errors, so it highlights all of them. ([example](https://github.com/near/near-sdk-rs/blob/0075253c6d4b862817a8a7fe5105ac87dc223614/near-sdk-macros/src/lib.rs#L368-L370))

And since we've decided to leave generics out, it only highlights the generics but not the lifetimes.

near-sdk/compilation_tests/generic_function.rs
miraclx: .. should resolve this issue

itegulov: Unfortunately, resolving this issue would require doing the same refactoring in `ItemImplInfo`, `ItemTraitInfo` and potentially some other places. I think this is better done in a separate PR.

near-sdk-macros/src/core_impl/info_extractor/attr_sig_info.rs
```diff
@@ -38,16 +38,27 @@ impl AttrSigInfo {
...
+        for generic in &original_sig.generics.params {
+            match generic {
+                GenericParam::Type(type_generic) => {
+                    errors.push(Error::new(
+                        type_generic.span(),
+                        "Contract API is not allowed to have generics.",
+                    ));
+                }
+                GenericParam::Const(const_generic) => {
+                    // `generic.span()` points to the `const` part of const generics, so we use `ident` explicitly.
```
miraclx, line 51: +1

miraclx (approved):

near-sdk-macros/src/core_impl/code_generator/ext.rs
```diff
@@ -267,7 +267,7 @@ mod tests {
         let mut method: ImplItemMethod = parse_quote! {
             pub fn method(&self, k: &String) { }
         };
-        let method_info = ImplItemMethodInfo::new(&mut method, impl_type).unwrap();
+        let method_info = ImplItemMethodInfo::new(&mut method, false,impl_type).unwrap().unwrap();
```
miraclx, line 270:
```suggestion
        let method_info = ImplItemMethodInfo::new(&mut method, false, impl_type).unwrap().unwrap();
```

near-sdk-macros/src/core_impl/code_generator/ext.rs
```diff
@@ -298,7 +298,7 @@ mod tests {
         let mut method: ImplItemMethod = parse_quote! {
           pub fn borsh_test(&mut self, #[serializer(borsh)] a: String) {}
         };
-        let method_info = ImplItemMethodInfo::new(&mut method, impl_type).unwrap();
+        let method_info = ImplItemMethodInfo::new(&mut method, false,impl_type).unwrap().unwrap();
```
miraclx, line 301:
```suggestion
        let method_info = ImplItemMethodInfo::new(&mut method, false, impl_type).unwrap().unwrap();
```

near-sdk/compilation_tests/generic_function.rs
miraclx: Ahh, right! That's unfortunate.. combining errors doesn't apply then, since `AttrSigInfo` is called per-method signature. But hey, at least we get good spans 😅.

But refactoring the macro system is in order tbh.

PR #981, opened 2022-11-28
chore:  cleanup impl_item_method_info

willemneal:
This breaks up `method_wrapper` that was over 100 lines long into smaller methods.

It also moves one function that should have been a method into the impl.

I have another PR in the works, but this was needed so that `self` could be accessible within `init_method_wrapper`.


frol:
@willemneal Hey, we are picking up near-sdk-rs development, and I am happy to merge this PR once it is rebased. Would you have the capacity to do that?

agostbiro:
Thanks for implementing this @willemneal ! This PR is now superseded by #1033 , so I'm closing it.

PR #982, opened 2022-11-29
fix: strip return types of lifetimes

### LINKED ISSUE #974: ABI generation fails with lifetimes

itegulov:
Closes #974 

We are already stripping function params of their lifetimes, seems reasonable to also do this for the return type.

### DIFF
--- a/near-sdk/compilation_tests/bad_argument.stderr
+++ b/near-sdk/compilation_tests/bad_argument.stderr
@@ -1,4 +1,4 @@
-error: Unsupported argument type.
+error: Unsupported contract API type.
   --> compilation_tests/bad_argument.rs:30:56
    |
 30 |     pub fn insert(&mut self, key: TypeA, value: TypeB, t: impl MyTrait) -> Option<TypeB> {
--- a/near-sdk-macros/src/core_impl/info_extractor/arg_info.rs
+++ b/near-sdk-macros/src/core_impl/info_extractor/arg_info.rs
@@ -58,7 +58,8 @@ impl ArgInfo {
                 ));
             }
         };
-        let (reference, mutability, ty) = utils::extract_ref_mut(original.ty.as_ref())?;
+        let (reference, mutability, ty) =
+            utils::extract_ref_mut(original.ty.as_ref(), original.span())?;
         // In the absence of callback attributes this is a regular argument.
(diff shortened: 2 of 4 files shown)


near-sdk/compilation_tests/bad_argument.stderr
```diff
@@ -1,4 +1,4 @@
-error: Unsupported argument type.
+error: Unsupported contract API type.
   --> compilation_tests/bad_argument.rs:30:56
    |
 30 |     pub fn insert(&mut self, key: TypeA, value: TypeB, t: impl MyTrait) -> Option<TypeB> {
```
miraclx, line 4: Looks like the error message now highlights the `impl` as opposed to the ident `t`. Should we override the span?

miraclx (approved):

near-sdk/compilation_tests/bad_argument.stderr
miraclx: Looks like highlighting the ident is `rustc`'s approach.

```console
error[E0277]: the trait bound `MyType: Serialize` is not satisfied
    --> src/lib.rs:23:19
     |
23   |     pub fn method(a: &Self) -> Self {
     |                   ^ the trait `Serialize` is not implemented for `MyType`
     |
```

itegulov: I have parametrized `extract_ref_mut` by span to distinguish between places where we have access to ident and where we don't. Having to always calculate span is not super clean, but probably not a big deal.

miraclx (approved):

ChaoticTempest (approved):

PR #986, opened 2022-11-30
feat: Lazy trait and `riff` attribute that allows loading and storing contract state through Lazy API

willemneal:
This PR depends on #981

This adds an attribute `riff` to the `near_bindgen` macro. This allows the state of an impl to be loaded using a new `Lazy` trait.  This allows users to define their own storage key. This allows breaking the contract into "riff"'s which can then lazily load other riffs when needed.

This is very similar to the LazyOption fields used currently but allows them to define their own interface not have to load the core STATE struct every time.

I wanted to achieve this with a separate crate, but in the end needed to intercept the current `env::load_state` and `env::store_state`.  For the past several months I've been relying on a fork of the sdk to enable [`contract riffs`](https://github.com/AhaLabs/contract-riffs), but it would be better to be able to just depend on this for the macro.


frol:
@willemneal Do you have an example code-base where we can see how to use it? If I understand it correctly, I would expect some umbrella contract would import riff contract struct and somehow host it, and the challenge you bring seems to be that `STATE` key is colliding, but I also see a risk of contract function names colliding as well, so I wonder if simple nesting/composition would do better:

```rust
struct RiffContract {
    some_storage_data: String,
}

#[near_bindgen]
impl RiffContract {
    fn get_data(&self) -> String { ... }
}

// -----
// Umbrella contract:

struct Contract {
    riff_contract: near_sdk::store::Lazy<RiffContract>,
}

#[near_bindgen]
impl Contract {
    fn get_data(&self) -> String {
         // Just delegate the call to the inner struct:
         self.riff_contract.get().get_data()
    }
}
```

PR #989, opened 2022-12-03
refactor: abstract common functions in `Keys` and `KeysRange`

### LINKED ISSUE #987: refactor: abstract common functions in `Keys` and `KeysRange`
New description:
Instead of replacing `Keys` with `KeysRange` the common functions in them will be abstracted out.

Old description: 
[`Keys`](https://github.com/near/near-sdk-rs/blob/2400a34272a9ecfe6f020c16fda6cdf2029543cf/near-sdk/src/store/tree_map/iter.rs#L296) is a sub case of [`KeysRange`](https://github.com/near/near-sdk-rs/blob/2400a34272a9ecfe6f020c16fda6cdf2029543cf/near-sdk/src/store/tree_map/iter.rs#L386) and can be replaced with it. 

Replacing `Keys` with `KeysRange` will remove duplicate code. Please see https://github.com/near/near-sdk-rs/pull/964#discussion_r1032590540 for more details.

askoa:
Closes #987

### DIFF
--- a/near-sdk/src/store/tree_map/iter.rs
+++ b/near-sdk/src/store/tree_map/iter.rs
@@ -45,7 +45,7 @@ where
     V: BorshSerialize,
     H: ToKey,
 {
-    keys: KeysRange<'a, K>,
+    keys: Keys<'a, K>,
     values: &'a LookupMap<K, V, H>,
 }
 
@@ -56,10 +56,7 @@ where
     H: ToKey,
 {
     pub(super) fn new(map: &'a TreeMap<K, V, H>) -> Self {
-        Self {
-            keys: KeysRange::new(&map.tree, (Bound::Unbounded, Bound::Unbounded)),
-            values: &map.values,
-        }
+        Self { keys: Keys::new(&map.tree), values: &map.values }
     }
 }
 
@@ -151,7 +148,7 @@ where
     H: ToKey,
 {
     /// Values iterator which contains empty and filled cells.
-    keys: KeysRange<'a, K>,
+    keys: Keys<'a, K>,
     /// Exclusive reference to underlying map to lookup values with `keys`.
     values: &'a mut LookupMap<K, V, H>,
 }
@@ -163,10 +160,7 @@ where
     H: ToKey,
 {
     pub(super) fn new(map: &'a mut TreeMap<K, V, H>) -> Self {
-        Self {
-            keys: KeysRange::new(&map.tree, (Bound::Unbounded, Bound::Unbounded)),
-            values: &mut map.values,
-        }
+        Self { keys: Keys::new(&map.tree), values: &mut map.values }
     }
 }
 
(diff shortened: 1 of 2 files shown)


near-sdk/src/store/tree_map/iter.rs
```diff
@@ -584,7 +398,11 @@ where
 
     fn size_hint(&self) -> (usize, Option<usize>) {
         let len = self.length as usize;
-        (0, Some(len))
+        (len, Some(len))
```
austinabell, line 401: So the main concern with this is that this is stating that there is an exact amount of elements in this iterator, which is not the case for `KeysRange`. I do think that this difference can be useful to users, if they want to use `Keys` as an exact size iterator.

I would recommend trying to keep `Keys` and `KeysRange` as separate types, but just share the internal logic with shared types or abstracting the methods to just take the raw components.

near-sdk/src/store/tree_map/mod.rs
```diff
@@ -9,10 +9,10 @@ use crate::store::LookupMap;
 use crate::{env, IntoStorageKey};
 use borsh::{BorshDeserialize, BorshSerialize};
 pub use entry::Entry;
-pub use iter::{Iter, IterMut, Keys, Range, RangeMut, Values, ValuesMut};
+pub use iter::{Iter, IterMut, KeysRange, Range, RangeMut, Values, ValuesMut};
```
austinabell, line 12: Oh... interesting... seems like the `KeysRange` type was not exported previously. That was a mistake :D

askoa: You want both `Keys` and `KeysRange` in here?

near-sdk/src/store/tree_map/iter.rs
askoa: Cool! I'll add `Keys` back and have shared methods.

near-sdk/src/store/tree_map/mod.rs
austinabell: oh, actually. Sorry for the false positive here. No, KeysRange is only used internally; it was correct not to be exposed.

near-sdk/src/store/tree_map/iter.rs
```diff
@@ -294,6 +294,116 @@ where
     }
 }
 
+fn find_min<'a, K>(
+    tree: &'a Tree<K>,
+    root: Option<&FreeListIndex>,
```
austinabell, line 299: small nit: the root is coming from the tree, so there should be no need to pass in the root in this and `find_max`, right?

askoa: I think its needed. Its the root of the `find_min` and not root of the tree. See below.

https://github.com/near/near-sdk-rs/blob/1351cad09653a8a333be7509fb9941fcbf79bc61/near-sdk/src/store/tree_map/iter.rs#L366

austinabell: oops! I missed that when I skimmed this

austinabell (approved):
LGTM! Thanks :) 

@itegulov do you want to have a skim through to make sure I didn't miss something nuanced here?

itegulov (approved):

PR #992, opened 2022-12-07
<WIP> feat: `defrag` function for `FreeList`

### LINKED ISSUE #990: New collections using `FreeList` internally might consume much more gas than using `Vector` depending on use case
Hello,

ok so first I'm sorry for the confusingly long title, but I didn't know how else to tell this in a short sentence.
Let me explain my use case:

In my contract I keep track of work that needs to be done, but might not fit into current execution of contract due to gas limitations. Instead I postpone this work into future executions of the contract.
I keep track of this work using the new `store::UnorderedMap`.

I noticed a weird behavior using this new API where gas usage increased when I removed elements from the start while iterating through them. The implementation reveals that internally for iteration a `FreeList` is used which does not re-arrange elements on removal. This is quite bad for my use case, because I need to optimize gas usage for worst case.

In a worst case scenario my `UnorderedMap` will be filled quite heavily in a single contract call. Then in later contract calls this work will be executed. Unfortunately iterating and removing the elements at the start results in the iterator visiting empty elements in successive contract calls. The gas usage increases with every call!

When I switch back to the legacy `UnorderedMap` which internally uses `Vector` (I assume?) for iteration, I no longer have this problem. The gas usage on first iteration is a little more, but it does not increase with every iteration.

Please let me know if I need to explain further and eventually give a code example.

My suggested solution would be to have another `store::UnorderedMap` that also uses `Vector` for storing keys. The user should be able to pick one of both and the difference would need to be explained. Since other collection types also use `FreeList` under the hood these collections would also need a second implementation for consistency reasons.

askoa:
Closes #990 

`defrag` function for `FreeList` will replace `Solt::Empty` entries in front of the list with ``Slot::Occupied` entries from back of the list. Please see the issue for detailed description.

### DIFF
--- a/near-sdk/src/store/free_list/mod.rs
+++ b/near-sdk/src/store/free_list/mod.rs
@@ -221,21 +221,26 @@ where
         Drain::new(self)
     }
 
-    pub fn defrag(&mut self) {
+    /// Empty slots in the front of the list is swapped with occupied slots in back of the list.
+    /// Defrag helps reduce gas cost in certain scenarios where lot of elements in front of the list are
+    /// removed without getting replaced. Please see https://github.com/near/near-sdk-rs/issues/990
+    pub fn defrag(&mut self) -> HashMap<u32, u32> {
         let elements_size = self.elements.len();
         let mut forward_ix = 0;
         let mut backward_ix = elements_size - 1;
         let mut swaps: HashMap<u32, u32> = HashMap::default();
         while forward_ix < backward_ix {
+            //increment forward index until empty slot is found
             let mut forward_element = self.elements.get(forward_ix);
             while let (true, Some(Slot::Occupied(_))) = (forward_ix < backward_ix, forward_element)
             {
                 forward_ix += 1;
                 forward_element = self.elements.get(forward_ix);
             }
-            if forward_element.is_none() || forward_ix >= backward_ix {
+            if forward_ix >= backward_ix {
                 continue;
             }
+            //decrement backward index until occupied slot is found
             let mut backward_element = self.elements.get(backward_ix);
             while let (true, Some(Slot::Empty { next_free: _ })) =
                 (backward_ix > forward_ix, backward_element)
@@ -246,19 +251,52 @@ where
             if forward_ix >= backward_ix {
                 continue;
             }
+
+            //If matching indexes found, add to swapping indexes
             if let (Some(Slot::Occupied(_)), Some(Slot::Empty { next_free: _ })) =
-                (forward_element, backward_element)
+                (backward_element, forward_element)
             {
+                //forward_ix is the old empty slot, backward_ix is the new empty slot
                 swaps.insert(forward_ix, backward_ix);
-                self.elements.swap(forward_ix, backward_ix);
+                self.elements.swap(forward_ix, backward_ix)
             }
         }
 
+        //update `first_free` if its swapped during defrag
         if let Some(free_list_index) = self.first_free {
             if let Some(to_index) = swaps.get(&free_list_index.0) {
                 self.first_free = Some(FreeListIndex(*to_index))
             }
         }
+
+        //Follow the free indexes chain and update for swapped indexes
+        let mut curr_free = self.first_free;
+        while let Some(free_index) = curr_free {
+            let next_free_slot = self.elements.get(free_index.0);
+            if let Some(Slot::Empty { next_free: next_index }) = next_free_slot {
+                if let Some(free_next_index) = next_index {
+                    //Check if the next index is swapped during defrag.
+                    //If the next index is swapped then update the empty slot to point to new next index
+                    if let Some(new_next_index) = swaps.get(&free_next_index.0) {
+                        //The next index is swapped during defrag and is in index new_next_index
+                        let new_next_slot =
+                            Slot::Empty { next_free: Some(FreeListIndex(*new_next_index)) };
+                        self.elements.replace(free_index.0, new_next_slot);
+                        curr_free = Some(FreeListIndex(*new_next_index));
+                    } else {
+                        //The next index is not swapped during defrag
+                        curr_free = *next_index;
+                    }
+                } else {
+                    //End of free indexes chain
+                    curr_free = None;
+                }
+            } else {
+                //Should not happen. The free indexes chain should all be `Slot::Empty`
+                curr_free = None;
+            }
+        }
+        swaps
     }
 }
 
--- a/near-sdk/src/store/free_list/mod.rs
+++ b/near-sdk/src/store/free_list/mod.rs
@@ -6,7 +6,7 @@ use crate::{env, IntoStorageKey};
 
 use borsh::{BorshDeserialize, BorshSerialize};
 
-use std::{fmt, mem};
+use std::{collections::HashMap, fmt, mem};
 
 /// Index for value within a bucket.
 #[derive(BorshSerialize, BorshDeserialize, Debug, Hash, PartialEq, Eq, Clone, Copy)]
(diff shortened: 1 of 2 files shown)


near-sdk/src/store/free_list/mod.rs
```diff
@@ -220,6 +220,84 @@ where
...
+    /// Empty slots in the front of the list is swapped with occupied slots in back of the list.
+    /// Defrag helps reduce gas cost in certain scenarios where lot of elements in front of the list are
+    /// removed without getting replaced. Please see https://github.com/near/near-sdk-rs/issues/990
+    pub fn defrag(&mut self) -> HashMap<u32, u32> {
+        let elements_size = self.elements.len();
+        let mut forward_ix = 0;
+        let mut backward_ix = elements_size - 1;
+        let mut swaps: HashMap<u32, u32> = HashMap::default();
+        while forward_ix < backward_ix {
+            //increment forward index until empty slot is found
```
austinabell, line 233: You can traverse this chain using `self.first_free` and then pulling the next indices from each element loaded. This _should_ contain all of the freed elements (given no bugs anywhere). This avoids the need to traverse everything manually

near-sdk/src/store/free_list/mod.rs
```diff
@@ -220,6 +220,84 @@ where
     pub fn drain(&mut self) -> Drain<T> {
         Drain::new(self)
     }
+
+    /// Empty slots in the front of the list is swapped with occupied slots in back of the list.
+    /// Defrag helps reduce gas cost in certain scenarios where lot of elements in front of the list are
+    /// removed without getting replaced. Please see https://github.com/near/near-sdk-rs/issues/990
+    pub fn defrag(&mut self) -> HashMap<u32, u32> {
```
austinabell, line 227: since this is just used internally, I'd suggest an API that just does one defrag at a time, which would make it easier to use in other collections, or have this function accept a callback, that gives back the old and new indices.

If this method is kept, can just keep this in a `Vec<(u32, u32)>` to avoid the HashMap overhead here

near-sdk/src/store/free_list/mod.rs
askoa: I don't understand what you are explaining. 

1. `self.first_free` would be last removed. If items are being removed at the start of iterator then the `n`th element removed (at index `n-1`) would be self.first_free. It might not be efficient to start replacing there. I thought we want to populate from the front.
2. Should we design this for a generic scenario of fragmentation or only for the scenario described in the issue?

near-sdk/src/store/free_list/mod.rs
askoa: Replacing one element at a time might not be a good approach. If we defragmented 5 elements then while defragmenting 6th element we'll have to traverse past the first five. Also, we'll not know if the list is already defragmented or not until we traverse all free indexes.

near-sdk/src/store/free_list/mod.rs
austinabell: > I thought we want to populate from the front.

Nope, the freelist is unordered; it doesn't matter where the values are shifted from and to in the underlying `Vector`

> It might not be efficient to start replacing there

The assumption here, I assume, is that you won't have to move values twice if the array shrinks, but this all would happen in-memory, so it would be cheap. What you could do is traverse all freed nodes to get the indices and check for this case so that you only move the values at most once. This avoids loading values that will not be moved with this operation.

> Should we design this for a generic scenario of fragmentation or only for the scenario described in the issue?

I don't know what you mean by this; can you clarify?

askoa: For a scenario like below, we will end up swapping Occupied1 and Occupied2 with Empty1 and Empty2 with first two entries (Empty4 and Empty3) still empty. For efficient defragmentation we might have to start from the beginning. Did I understand your approach correctly?

Empty4, Empty3, Empty2, Empty1=(self.first_free), Occupied1, Occupied2

near-sdk/src/store/free_list/mod.rs
austinabell: yeah, you're right. That was an inaccurate suggestion

askoa: I like the idea of callback function. Would it be okay to call the callback function every swap rather than at the end?

near-sdk/src/store/free_list/mod.rs
askoa: Let me update the commit with checking `< self.occupied_count`.

near-sdk/src/store/free_list/mod.rs
austinabell: yes, I think that would be most sound, given that there aren't multiple swaps of the same element as discussed on the other thread.

near-sdk/src/store/free_list/mod.rs
```diff
@@ -301,6 +306,44 @@ mod tests {
...
+        bucket.remove(i2);
+        bucket.remove(i4);
+        bucket.remove(i1);
+        bucket.remove(i7);
+        bucket.remove(i3);
+        bucket.remove(i8);
+        assert_eq!(bucket.occupied_count, 2);
+
+        let free_items = navigate_empty_slots(&bucket);
+        assert_eq!(free_items, 6);
```
austinabell, line 339: so one note about this implementation, this should actually just clear the last elements from storage altogether (and shrink the underlying Vector, set the `first_free` back to None). This way storage costs can be refunded on this operation

askoa: I see! This should make the logic simpler, I guess.

austinabell: yeah, I would suggest using `drain(<num of elements>..)` and then iterating through to pull filled cells to move into the vacant entries, but up to you and what you had planned to implement!

askoa: If the free slots chain is in drained vector then it becomes difficult to traverse through them. I just pushed a PR. But, now, I think the better option is to 
1. Traverse through free slots chain and store the free slots that needs replacement (`self.occupied_count`) in a `Vec`
2. Pop items from the `self.elements` and if its occupied, then place it in one of the free slots in the `Vec`

Let me know if you are okay with above approach?

askoa:
@austinabell Did you had time to review the PR? I can make incremental changes to the PR. If you want a conceptually different solution then this is where I stop. I'll close this PR and keep an eye for eventually acceptable solution.

near-sdk/src/store/free_list/mod.rs
```diff
@@ -220,6 +220,86 @@ where
...
+                //Could not find an occupied slot to fill the free slot
+                env::panic_str(ERR_INCONSISTENT_STATE)
+            }
+        }
+        self.elements.len = self.occupied_count;
+    }
+
+    fn next_free_slot(&mut self) -> Option<FreeListIndex> {
+        while let Some(curr_free_index) = self.curr_free_slot {
+            let curr_slot = self.elements.values.remove(curr_free_index.0);
```
austinabell, line 282: This feels like it would lead to a bug if you are modifying the vector's inner elements manually. The index is cheap to copy, it feels safer to not leave the inner vector in an inconsistent state. wdyt @itegulov ?

near-sdk/src/store/free_list/mod.rs
```diff
@@ -220,6 +220,86 @@ where
...
+                callback(value, curr_free_index.0);
+                //The entry at curr_free_index.0 should have `None` by now.
+                //Moving it to `occupied_index` will make that entry empty.
+                self.elements.swap(curr_free_index.0, occupied_index);
+            } else {
+                //Could not find an occupied slot to fill the free slot
+                env::panic_str(ERR_INCONSISTENT_STATE)
+            }
+        }
+        self.elements.len = self.occupied_count;
```
austinabell, line 277: Inline with previous comment, modifying a type that is meant to be internal to `Vector` seems unideal since it seems like the same functionality could be done with the safe public APIs of `Vector`.

near-sdk/src/store/free_list/mod.rs
```diff
@@ -220,6 +220,86 @@ where
...
+    }
+
+    fn next_free_slot(&mut self) -> Option<FreeListIndex> {
+        while let Some(curr_free_index) = self.curr_free_slot {
+            let curr_slot = self.elements.values.remove(curr_free_index.0);
+            self.curr_free_slot = match curr_slot {
+                Some(Slot::Empty { next_free }) => next_free,
+                _ => None,
```
austinabell, line 285: This branch seems like it should be unreachable, perhaps we want to panic with inconsistent state if hit, no? Or am I missing a case where this could happen?

austinabell (commented):
Thanks so much for exploring this! The implementation looks great with just a few points I added. Totally understandable if you don't have the bandwidth to make more adjustments, appreciate you being open to prior feedback and suggestions ❤️

@itegulov will likely be the one to approve this

near-sdk/src/store/unordered_map/mod.rs
```diff
@@ -630,6 +630,23 @@ where
...
+    K: BorshSerialize + BorshDeserialize + Ord + Clone,
+    V: BorshSerialize + BorshDeserialize,
+    H: ToKey,
+{
+    pub fn defrag(&mut self) {
+        self.keys.defrag(|key, new_index| {
+            // Check if value is in map to replace first
+            let entry = self.values.get_mut_inner(key);
+            if let Some(existing) = entry.value_mut() {
+                _ = mem::replace(&mut existing.key_index, FreeListIndex(new_index));
```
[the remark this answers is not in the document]
askoa: I used the format from `insert` function. I apply you suggestion and run the tests.

https://github.com/near/near-sdk-rs/blob/18621c73c16266e3f2551fc2001314e514c84216/near-sdk/src/store/unordered_map/mod.rs#L491-L499

near-sdk/src/store/free_list/mod.rs
askoa: I think its unavoidable to have the vector in inconsistent state until the defrag is complete. This is similar to `Drain`. If the drain is not complete then the vector will be in inconsistent state. I was wondering if I can include a `remove` function visible only to the crate.

near-sdk/src/store/free_list/mod.rs
askoa: It can have `None`  but not `Self::Occupied`. I'll fix it.

near-sdk/src/store/free_list/mod.rs
askoa: I did not see a suitable API to meet the needs of `defrag`. If you have any specific suggestion I can look into it.

PR #997, opened 2022-12-16
Add documentation to collection cache types

austinabell:
It also removes some unused types. I can't remember why those were kept in if it was a prediction of what would be needed, but they shouldn't be needed.


ChaoticTempest (approved):
📚

Issue #972, opened 2022-11-23
ABI generation fails with `Self` return types


miraclx:
```rust
#[near_bindgen]
impl MyType {
    pub fn new() -> Self { ... }
}
```

```console
$ cargo build --features near-sdk/__abi-generate
   Compiling adder v0.1.0 (/Users/miraclx/git/near/near-sdk-rs/examples/adder)
error[E0411]: cannot find type `Self` in this scope
 --> src/lib.rs:3:12
  |
3 |     pub fn new() -> Self {
  |            ---      ^^^^ `Self` is only available in impls, traits, and type definitions
  |            |
  |            `Self` not allowed in a function

For more information about this error, try `rustc --explain E0411`.
error: could not compile `adder` due to 2 previous errors
```

itegulov:
Is there a common use for `Self` outside of `init` functions though? Your particular example works if you mark the function with `#[init]`.

austinabell:
> Is there a common use for `Self` outside of `init` functions though? Your particular example works if you mark the function with `#[init]`.

Not a common one, but I think it is a valid one. I've seen devs try this, and it only doesn't really work when there are storage collections in the state. But if everything is in-memory and you just want to return all of the state, there could be a reason to do so.

`new` might be a confusing naming for the use case, though.

To confirm, this fails also, right?

```rust
#[near_bindgen]
// derive default, serde, ....
struct MyType {
  data: u8,
}

#[near_bindgen]
impl MyType {
  pub fn get_state(self) -> Self { self }
  // or
  pub fn get_state(&self) -> &Self { self }
}
```

itegulov:
> Not a common one, but I think it is a valid one. I've seen devs try this, and it only doesn't really work when there are storage collections in the state. But if everything is in-memory and you just want to return all of the state, there could be a reason to do so.

Makes sense.

I have taken a look at what `ink!` does with `Self` and they just ban its usage in non-constructor functions altogether:
![latest-screenshot](https://user-images.githubusercontent.com/7879134/203930372-d3c76d11-7ad0-4580-a413-90651e7b3da5.png)

So maybe banning all occurrences of `Self` makes more sense from the maintenance point of view.

austinabell:
> So maybe banning all occurrences of Self makes more sense from the maintenance point of view.

My intuition would agree with this. If they _really_ wanted to serialize the state as a return value, they could use the type name explicitly or a number of other ways to work around.

[...]

Issue #1000, opened 2023-01-21
ABI generation fails with `Self` arg types


miraclx:
```rust
#[near_bindgen]
#[derive(Serialize, Deserialize)]
struct MyType {
    data: u8,
}

#[near_bindgen]
impl MyType {
    fn set_state(&mut self, new_state: Self) {
        std::mem::replace(self, new_state);
    }
}
```

```
$ cargo build --features near-sdk/__abi-generate
   Compiling adder v0.1.0 (/Users/miraclx/git/near/near-sdk-rs/examples/adder)
error[E0308]: mismatched types
  --> src/lib.rs:12:1
   |
12 | #[near_bindgen]
   | ^^^^^^^^^^^^^^^ expected struct `Input`, found struct `MyTypeExt`
   |
   = note: expected reference `&Input<'_>`
              found reference `&MyTypeExt`
   = note: this error originates in the attribute macro `near_bindgen` (in Nightly builds, run with -Z macro-backtrace for more info)

Some errors have detailed explanations: E0308, E0411.
For more information about an error, try `rustc --explain E0308`.
error: could not compile `adder` due to 4 previous errors
```

PR #1001, opened 2023-01-21
fix: concretize `Self` references in method signatures

miraclx:
Resolves: #972, #1000.

This patch replaces any `Self` references in method signatures with the concrete type of the `impl` block itself.

Example:

```rust
#[near_bindgen]
#[derive(Serialize, Deserialize)]
struct MyType {
    data: u8,
}

#[near_bindgen]
impl MyType {
    fn set_state(&mut self, new_state: Self) -> Self {
        std::mem::replace(self, new_state)
    }
}
```

the impl block is transformed into:

```rust
impl MyType {
    fn set_state(&mut self, new_state: MyType) -> MyType {
        std::mem::replace(self, new_state)
    }
}
```

this transformation is maintained regardless of type complexity or nesting level.

Worth noting, this does not transform `Self` references in generic bounds. But they're going to be prohibited following https://github.com/near/near-sdk-rs/pull/980 anyway. So I left it out.

### DIFF
--- a/near-sdk-macros/src/core_impl/abi/abi_generator.rs
+++ b/near-sdk-macros/src/core_impl/abi/abi_generator.rs
@@ -1,5 +1,5 @@
 use proc_macro2::{Span, TokenStream as TokenStream2};
-use quote::{format_ident, quote, ToTokens};
+use quote::{format_ident, quote};
 use syn::spanned::Spanned;
 use syn::{Attribute, Lit::Str, Meta::NameValue, MetaNameValue, ReturnType, Type};
 
@@ -16,7 +16,7 @@ pub fn generate(i: &ItemImplInfo) -> TokenStream2 {
         return TokenStream2::new();
     }
 
-    let functions: Vec<TokenStream2> = public_functions.iter().map(|m| m.abi_struct(i)).collect();
+    let functions: Vec<TokenStream2> = public_functions.iter().map(|m| m.abi_struct()).collect();
     let first_function_name = &public_functions[0].attr_signature_info.ident;
     let near_abi_symbol = format_ident!("__near_abi_{}", first_function_name);
     quote! {
@@ -81,7 +81,7 @@ impl ImplItemMethodInfo {
     /// }
     /// ```
     /// If args are serialized with Borsh it will not include `#[derive(borsh::BorshSchema)]`.
-    pub fn abi_struct(&self, i: &ItemImplInfo) -> TokenStream2 {
+    pub fn abi_struct(&self) -> TokenStream2 {
         let function_name_str = self.attr_signature_info.ident.to_string();
         let function_doc = match parse_rustdoc(&self.attr_signature_info.non_bindgen_attrs) {
             Some(doc) => quote! { Some(#doc.to_string()) },
@@ -117,8 +117,7 @@ impl ImplItemMethodInfo {
             let arg_name = arg.ident.to_string();
             match arg.bindgen_ty {
                 BindgenArgType::Regular => {
-                    let typ = if typ.to_token_stream().to_string() == "Self" { &i.ty } else { typ };
-                    let schema = generate_schema(typ, &arg.serializer_ty);
+                    let schema = generate_schema(&typ, &arg.serializer_ty);
                     match arg.serializer_ty {
                         SerializerType::JSON => params.push(quote! {
                             near_sdk::__private::AbiJsonParameter {
@@ -135,7 +134,7 @@ impl ImplItemMethodInfo {
                     };
                 }
                 BindgenArgType::CallbackArg => {
-                    callbacks.push(generate_abi_type(typ, &arg.serializer_ty, &i.ty));
+                    callbacks.push(generate_abi_type(typ, &arg.serializer_ty));
                 }
                 BindgenArgType::CallbackResultArg => {
                     let typ = if let Some(ok_type) = utils::extract_ok_type(typ) {
@@ -148,7 +147,7 @@ impl ImplItemMethodInfo {
                         )
                         .into_compile_error();
                     };
-                    callbacks.push(generate_abi_type(typ, &arg.serializer_ty, &i.ty));
+                    callbacks.push(generate_abi_type(typ, &arg.serializer_ty));
                 }
                 BindgenArgType::CallbackArgVec => {
                     if callback_vec.is_none() {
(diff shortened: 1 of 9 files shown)


willemneal:
Shouldn't this be fixed in the abi? Seems like it should be handled there and not banned. Witgen handles it.

miraclx:
> Shouldn't this be fixed in the abi? Seems like it should be handled there and not banned. Witgen handles it.

Well, the issue here is mostly that the intermediate code generated that eventually generates the ABI derives the schema for a type from a method call that is generic over the type. Specifically a line that does:

```rust
gen.subschema_for::<T>();
```

and in the case where `T` is `Self`, `Self` will be undefined because the generated code lives inside an `extern "C"` block which is unbound to an impl block.

At the moment, the example code just fails on master (See #972). This patch fixes that to make sure it both compiles and has ABI generated for the appropriate type.

An alternative as mentioned in #972 would just be to warn / fail with a useful error nudging users to use concrete types, but I wanted to submit this first so we know it's possible and not just jump head-first into disallowing it.

near-sdk-macros/src/core_impl/utils/mod.rs
```diff
@@ -73,3 +75,62 @@ pub(crate) fn extract_vec_type(ty: &Type) -> Option<&Type> {
...
+                .collect();
+            TokenTree::Group(Group::new(Delimiter::None, replace_with))
+        }
+        TokenTree::Group(group) => {
+            let stream = _sanitize_self(group.stream(), replace_with);
+            TokenTree::Group(Group::new(group.delimiter(), stream))
+        }
+        rest => rest,
+    });
+    trees.collect()
```
itegulov, line 98: Not a big deal at all, but I think you can avoid creating some `TokenTree`s here if you operate on `impl Iterator<Item = TokenStream2>` instead.

near-sdk-macros/src/core_impl/utils/mod.rs
```diff
@@ -73,3 +75,62 @@ pub(crate) fn extract_vec_type(ty: &Type) -> Option<&Type> {
...
+}
+
+pub fn sanitize_self(typ: &Type, replace_with: &TokenStream2) -> syn::Result<Type> {
+    syn::parse2(_sanitize_self(quote! { #typ }, replace_with)).map_err(|original| {
```
itegulov, line 102: Sanitizing `TokenStream` instead of `Type` is a nice trick. I remember trying to sanitize `Self` in the past, and it required so much boilerplate to traverse all `Type` variants.

itegulov (approved):
LGTM, just a couple of nitpicks. This is a great fix for the current major version of SDK, but I wonder if we should still explore whether banning `Self` makes sense in 5.0. Not because this PR does an unsatisfactory job but rather because it seems like a potentially harmful practice as was pointed out [here](https://github.com/near/near-sdk-rs/issues/972#issuecomment-1327720519). I can create an issue if you agree that it is worth exploring.

near-sdk-macros/src/core_impl/abi/abi_generator.rs
```diff
@@ -117,7 +117,7 @@ impl ImplItemMethodInfo {
             let arg_name = arg.ident.to_string();
             match arg.bindgen_ty {
                 BindgenArgType::Regular => {
-                    let schema = generate_schema(typ, &arg.serializer_ty);
+                    let schema = generate_schema(&typ, &arg.serializer_ty);
```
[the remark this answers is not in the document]
miraclx: Nah. Just that `typ` is already a `&Type`, so there's no need for an extra reference.

But the compiler already coerces references between themselves, so there's no real difference.

near-sdk-macros/src/core_impl/utils/mod.rs
miraclx: `TokenTree` really is what helps us with this level of granularity, though.

near-sdk-macros/src/core_impl/abi/abi_generator.rs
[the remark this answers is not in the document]
itegulov: > so there's no need for an extra reference.

But you have added an extra reference though, no?

near-sdk-macros/src/core_impl/utils/mod.rs
itegulov: WDYM? Couldn't you do something like (untested):

```rust
    let trees = typ.into_iter().map(|t| match t {
        TokenTree::Ident(ident) if ident == "Self" => replace_with
            .clone()
            .into_iter()
            .map(|mut t| {
                t.set_span(ident.span());
                t
            })
            .collect::<TokenStream2>(),
        TokenTree::Group(group) => {
            let stream = _sanitize_self(group.stream(), replace_with);
            TokenTree::Group(Group::new(group.delimiter(), stream)).into()
        }
        rest => rest.into(),
    });
    trees.collect()
```

miraclx: Yeah, tried that approach in [`6df05aa` (#1001)](https://github.com/near/near-sdk-rs/pull/1001/commits/6df05aad394007e506a1664c7a4b561e801d975b)? Hard to get over the `TokenTree` from the group part. Another issue might also just be the unnecessary allocation that `TokenStream` does behind the scenes, whereas operating on `TokenTree`'s esp in an iterator context can be more efficient.

near-sdk-macros/src/core_impl/abi/abi_generator.rs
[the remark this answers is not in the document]
miraclx: Ah, shoot.. relic of earlier iteration 😅 Nice catch!

miraclx:
> it seems like a potentially harmful practice as was pointed out https://github.com/near/near-sdk-rs/issues/972#issuecomment-1327720519. I can create an issue if you agree that it is worth exploring.

Yup, Sounds good to me!

near-sdk-macros/src/core_impl/utils/mod.rs
itegulov: > Another issue might also just be the unnecessary allocation that TokenStream does behind the scenes

Hmm, yeah I peeked the `FromIterator<TokenTree>` implementation and it seems like they do a bunch of optimizations there. Probably more efficient indeed, but your call in the end.

miraclx: Well, the change is small, and I don't foresee any sufficiently complex type signatures that cause OOMs 😆 
So we'll leave it as is.

near-sdk-macros/src/core_impl/utils/mod.rs
miraclx: Yeah, nightmarish 😨

itegulov:
Done: https://github.com/near/near-sdk-rs/issues/1005

mitinarseny:
Hey Team,
This fix makes it impossible to define a trait with `#[init]` method that returns `Self`:
```rust
#[ext_contract(ext_my_contract)]
pub trait MyContract {
    #[init]
    fn new() -> Self;
}
```

as it expands into
```rust
pub trait MyContract {
    fn new() -> MyContract;
}

pub mod ext_my_contract {
    // ...
}
```

```
error[E0782]: trait objects must include the `dyn` keyword
 --> src/lib.rs:2:16:
  |
6 |     fn new() -> Self;
  |                 ^^^^
  |
help: add `dyn` keyword before this trait
  |
6 |     fn new() -> dyn Self;
  |                 +++
```

Issue #936, opened 2022-10-12
Standardized errors


itegulov:
Right now it is unclear what kind of errors a contract (or even a function) can fail with. It would be nice to have a standardized way of declaring contract errors, something like this:

```
#[near_error]
pub enum MyError {
  #[message("Caller is not the owner of this contract")]
  NotTheOwner,
  #[message("Caller is not allowed to call this function")]
  NotAllowed,
}
```

We can then include all contract errors as a part of the ABI, this proposal also goes really well with the `Result`-based error handling as we can statically infer what errors a function can fail with.

frol:
> I assume tuple/struct variants would be forbidden here?

I don't think this is the intent.

This issue is more about making error handling explicit as currently the norm is that contract developers just panic in the middle of the execution, and that is treated as an error.

Let me bring some context here:

In NEAR, transaction execution often happens across several blocks (please, [watch/read this nice article](https://docs.near.org/concepts/data-flow/near-data-flow)). Contract call action is processed when receipt is processed and if during the execution the function wants to make a cross-contract call or make a transfer it will initiate more receipts. Each receipt execution is atomic (either completely successful and the changes are committed to the blockchain or completely rejected (e.g. on panic) and changes are not saved). Note that transactions in NEAR (in contrast to individual receipts) are not atomic.

That said, `panic` is often used to indicate that no changes should be saved, but the result string and logs are still returned, see this [example](https://explorer.near.org/transactions/6qywTYaousCwdLiZFBhCCraVCrwUi39XVvg39w7N7qs6):

<img width="596" alt="image" src="https://github.com/near/near-sdk-rs/assets/304265/bf5334f7-58fa-46cb-95c7-af7f886e0992">

This issue suggests we standardize this error handling and instead of returning a text message, we actually return some structured error.

On a similar note, it is sometimes useful to communicate an error without failing the execution (let it save the changes it did so far) [[here is one recent discussion about it](https://github.com/near/NEPs/pull/452#discussion_r1176434582)], so it is worth addressing this scenario together while we are discussing it. cc @robert-zaremba @fadeevab

uint:
I looked into how things work a bit. If I understand correctly, the VM currently conflates these two ideas:
- a guest error happened, and
- a receipt rollback is needed.

It sounds a lot like you'd want to disentangle those in nearcore. Is that right? @frol

frol:
[...]

PR #1008, opened 2023-02-20
feat: adds Default implementation for PromiseOrValue

encody:
- Only implements for `PromiseOrValue<T>` where `T: Default`.
- Returns the `Value(..)` variant.

### DIFF
--- a/near-sdk/src/promise.rs
+++ b/near-sdk/src/promise.rs
@@ -569,6 +569,15 @@ pub enum PromiseOrValue<T> {
     Value(T),
 }
 
+impl<T> Default for PromiseOrValue<T>
+where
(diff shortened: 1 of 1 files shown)


frol:
I am a bit concerned about the semantics of `PromiseOrValue::default()`. I will need to think about it more. Adding it is not a breaking change, but removing would be a breaking change, so I would prefer to be on the safer side of things.

agostbiro:
Thanks for submitting this PR @encody ! I'm inclined to agree with @frol : adding a default implementation for sum types can lead to surprising results, and changes are to be expected to `Promise` in #936, so I'm leaning towards respectfully declining this for now.

Issue #918, opened 2022-09-16
contract standards rewrite meta issue


austinabell:
There are numerous things that could/should be fixed when making state-breaking changes, so worth to batch and trying to do everything at once before asking anyone to migrate. This is not a final list, just going to start accumulating and indexing other issues:

- `no_std` and optimize size #443 (somewhat optional, this isn't a state-breaking change, but better to do before)
- Switch mapping from using `TreeMap` -- it's incredibly unscalable for iterating through pagination #800 #810
  - Alternatively, could change/implement a new ordered map that stores extra data to be able to paginate more efficiently, but ordering isn't part of the NFT spec
- Tweak gas usage using gas weighted function calls. This has been updated in `4.0` to be splitting unused gas evenly, but the static gas cannot be removed without being breaking and causes issues. https://github.com/near/near-sdk-rs/issues/619 https://github.com/near/near-sdk-rs/issues/559
- Explore https://github.com/near/near-sdk-rs/issues/568
- Update macros to be derive macros https://github.com/near/near-sdk-rs/issues/422
- Explore https://github.com/near/near-sdk-rs/issues/775
- Fix the generic return type for approval receiver. This should be no return, or ideally if possible to have an associated type on the trait (this might not play nicely with bindgen macros) https://github.com/near/near-sdk-rs/pull/693

Issue #568, opened 2021-09-15
Revisit nft_resolve_transfer args


mikedotexe:
According to the standard:
 [https://nomicon.io/Standards/NonFungibleToken/Core.html#nft-interface](https://nomicon.io/Standards/NonFungibleToken/Core.html#nft-interface)

`nft_resolve_transfer` has an argument `approved_account_ids` that is supposed to be `string[]` but I see there's a HashMap here. There's also a place where that field calls `approvals`, which should be fixed.

It seems like we don't need to pass the entire HashMap and can stick to the standard and just send a Vec<String> instead.

This is one of those tickets where I noticed it while I'm in the middle of something, so don't have more info other than this at the moment.

austinabell:
What we probably want here is a Set (probably `BTreeSet`) for this. A set will serialize/deserialize as an array but comes with the property that we don't have to check for duplicates manually and can still do the lookup more efficiently than iterating through.

I had a quick look and still don't understand the use of storing an integer value for each one. I'll dig into this more tomorrow.

austinabell:
dug slightly more into this, but it's clear there needs to be discussion around this because I need more context around what the flow is intended to be, and the fact that this is quite a breaking change so if the API is changed we need to be careful.

frol:
Resolved in #1019

PR #1019, opened 2023-05-13
fix: rename param `approvals` to `approved_account_ids`

hanakannzashi:
According to [NEP171](https://github.com/near/NEPs/blob/master/neps/nep-0171.md), the last param of `nft_resolve_transfer ` is `approved_account_ids`, and the [macro implementation](https://github.com/near/near-sdk-rs/blob/master/near-contract-standards/src/non_fungible_token/macros.rs#L47) is also `approved_account_ids`, but in trait definition, it named `approvals`, it will cause unexpected JSON serialization result when cross call `nft_resolve_transfer` because the trait use `#[ext_contract]` macro. Currently, approval ids can not be revoked when `nft_transfer_call` failed.

### DIFF
--- a/near-contract-standards/src/non_fungible_token/core/resolver.rs
+++ b/near-contract-standards/src/non_fungible_token/core/resolver.rs
@@ -26,7 +26,7 @@ pub trait NonFungibleTokenResolver {
     /// * `previous_owner_id`: the owner prior to the call to `nft_transfer_call`
     /// * `receiver_id`: the `receiver_id` argument given to `nft_transfer_call`
     /// * `token_id`: the `token_id` argument given to `ft_transfer_call`
-    /// * `approvals`: if using Approval Management, contract MUST provide
+    /// * `approved_account_ids`: if using Approval Management, contract MUST provide
     ///   set of original approved accounts in this argument, and restore these
     ///   approved accounts in case of revert.
     ///
@@ -36,6 +36,6 @@ pub trait NonFungibleTokenResolver {
         previous_owner_id: AccountId,
         receiver_id: AccountId,
         token_id: TokenId,
-        approvals: Option<HashMap<AccountId, u64>>,
+        approved_account_ids: Option<HashMap<AccountId, u64>>,
     ) -> bool;
 }


frol (approved):

ChaoticTempest (approved):
currently, this won't fail since the generated JSON serialization underneath for `ext` interfaces just serializes arguments without the related parameter name. So it's doing something like serializaing `("arg1", 2, 3)` into a function that takes `fn(&str, u32, usize)` and passing each item positionally.

But I think this is good to push in anyways since there's a naming disconnect that this would fix if someone were to look at this code. It shouldn't break anything since that `ext` isn't exposed. What do you think @itegulov?

itegulov (approved):
Agreed with @ChaoticTempest. I actually never understood if the positional way of providing arguments is standardized/documented anywhere. I tried to look into it last year, but no one I talked to knew where this came from. From what I know, this is a near-sdk-rs specific behaviour.

frol:
> currently, this won't fail since the generated JSON serialization underneath for `ext` interfaces just serializes arguments without the related parameter name. So it's doing something like serializaing `("arg1", 2, 3)` into a function that takes `fn(&str, u32, usize)` and passing each item positionally.

WAT? This cannot be true. [Here](https://github.com/near/neardevhub-contract/blob/6af1cc49d9e8f436f5dfa8b4f36d6aaca21bdb01/src/social_db.rs#L8) is DevHub contract, and [here](https://explorer.near.org/transactions/CGWbGGuHZ5X6yCpcUak5pYuyX8Kb1sC3iDY7JK5HD4Yz#37dxaa6o3KR5fX7TYss7TPECdGjCNQVmKYxTfRu64rNc) is the cross-contract call on chain. The arguments are serialized as JSON object with the `data` key there.

Do I miss something?

frol:
> I actually never understood if the positional way of providing arguments is standardized/documented anywhere

It seems to be a side-effect of `serde` deserialization that supports array-parsing as tuples of unnamed arguments: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1f8a035d839bf4b14fe800bdb238450e

itegulov:
> It seems to be a side-effect of serde deserialization that supports array-parsing as tuples of unnamed arguments

Yep, I understand the technical part of it, but just not sure if it was ever an intended side effect or not

frol:
> not sure if it was ever an intended side effect or not

It was not. I learned about this side effect just a couple of weeks ago.

hanakannzashi:
> 

`nft_transfer_call` does not indeed cause panic, what I mean is if its final return result is `false`, we believe that this NFT Transfer Operation should be revoked, but `approved_account_ids` cannot be revoked correctly because `nft_resolve_transfer` always receive `None` as `approved_account_ids`.

austinabell:
> It was not. I learned about this side effect just a couple of weeks ago.

Yeah, it was something I brought up and asked about when I first joined. It was an unknown that this was a side effect of serde, but I couldn't think of a reason to remove it since the intended format is JSON, which is not canonical anyway and allowing users to supply args as an array saves gas by avoiding serializing keys (like what is done with the Ext api).

Be careful with releasing this change. The original issue is here https://github.com/near/near-sdk-rs/issues/568 and was planned to be fixed with the standards rewrite to group in with other breaking changes https://github.com/near/near-sdk-rs/issues/918.

If you release this fix, it will break some existing usages in the ecosystem for such a trivial detail. It might be a cleaner alternative to just update the spec for something like this (note that JS has the spec naming, though https://github.com/near/near-sdk-js/blob/develop/packages/near-contract-standards/src/non_fungible_token/core/resolver.ts#L36)

PR #1020, opened 2023-05-13
fix: remove receiver approval

hanakannzashi:
We should remove any approvals receiver already set.


frol (approved):
@hanakannzashi Thanks for catching it!

ChaoticTempest (approved):
LGTM: the NFT tests don't seem to be a part of the GH actions pipeline but ran it locally and everything seems to be working fine. I'm surprised we haven't ran into this issue more but probably because `nft_resolve_transfer` always gets passed a `approved_account_ids` which overrides that `token_id` key anyways.

Thanks for noticing this edge case!

PR #1021, opened 2023-05-13
fix: approved account ids

hanakannzashi:
According to the [comments](https://github.com/near/near-sdk-rs/blob/master/near-contract-standards/src/non_fungible_token/core/resolver.rs#L29) and [`nft_token` implementation](https://github.com/near/near-sdk-rs/blob/master/near-contract-standards/src/non_fungible_token/core/core_impl.rs#L412:L415), I think 'Extension Not Exists' vs 'Extension Exist But Value Not Exists' is different. i.e. If a contract doesn't have Approval Extension,  when we get `approved_account_ids`, it should be `None`, but if a contract does have Approval Extension, then we get `approved_account_ids`, it should be at least empty map, NEVER be `None`.

### DIFF
--- a/near-contract-standards/src/non_fungible_token/core/core_impl.rs
+++ b/near-contract-standards/src/non_fungible_token/core/core_impl.rs
@@ -206,7 +206,7 @@ impl NonFungibleToken {
 
         // clear approvals, if using Approval Management extension
         // this will be rolled back by a panic if sending fails
-        let approved_account_ids = 
+        let approved_account_ids =
             self.approvals_by_id.as_mut().map(|by_id| by_id.remove(token_id).unwrap_or_default());
 
         // check if authorized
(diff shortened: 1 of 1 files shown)


frol (approved):
@hanakannzashi This is definitely a good usability improvement!

miraclx (commented):
Seems the benefit here (in the case that approval extensions are supported, but the map is empty) is that the error moves from being "Unauthorized" to "Sender not approved" and if approval extensions are not supported, the panic message is "Approval extension is disabled"

Just clearing that out of the way because even though account approvals are supported, but an empty map is provided, WE WILL STILL panic, just with an arguably, better error message.

The only case that only ever successfully executes, is if account approvals are supported (AND the non-empty associative map contains the `sender_id`)

frol:
@miraclx Yes, that is it. It is a minor yet welcome UX improvement. I am not a codeowner here, can you approve it if it looks fine to you?

miraclx (approved):
Lgtm.

PR #1023, opened 2023-06-16
Freelist defrag

### LINKED ISSUE #990: New collections using `FreeList` internally might consume much more gas than using `Vector` depending on use case

uint:
Closes #990 

This PR picks up where #992 left off (thanks @askoa!). The implementation is mostly the same except:

- `FreeList::defrag` does not interact with `Vector` internals (see comments in #992)
- `Vector` gets the new `remove_tail` method to support the above in a slightly more efficient fashion
- the `defrag` test does not assert stuff irrelevant to the feature - those assertions are instead pulled into new, more atomic tests (`new_bucket_is_empty`, `occupied_count_gets_updated`)

### DIFF
--- a/near-sdk/src/store/free_list/mod.rs
+++ b/near-sdk/src/store/free_list/mod.rs
@@ -228,56 +228,37 @@ where
     where
         F: FnMut(&T, u32),
     {
-        let elements_size = self.elements.len();
-        let mut backward_index = elements_size.checked_sub(1);
+        let mut occupied_index = self.occupied_count;
         let mut curr_free: Option<FreeListIndex> = self.first_free;
 
-        while let (Some(_), Some(mut curr_free_index)) = (backward_index, curr_free) {
+        //Navigate free slots chain
+        while let Some(curr_free_index) = curr_free {
+            //Replace the free slot with `None`
+            let curr_free_slot = self.elements.values.remove(curr_free_index.0);
+
             if curr_free_index.0 < self.occupied_count {
-                //Get first `Slot::Occupied` index from back
-                let value: Option<&T>;
-                (backward_index, value) = self.get_prev_occupied(backward_index);
-                if let (Some(backward_ix), Some(value)) = (backward_index, value) {
-                    //call callback function.
-                    callback(value, curr_free_index.0);
-
-                    self.elements.swap(curr_free_index.0, backward_ix);
-
-                    curr_free_index = FreeListIndex(backward_ix);
+                //Find occupied entry to populate the free slot
+                while occupied_index < self.elements.len {
+                    if let Some(Slot::Occupied(value)) = self.elements.get(occupied_index) {
+                        callback(value, curr_free_index.0);
+                    } else {
+                        occupied_index += 1;
+                        continue;
+                    }
+                    //`curr_free_index.0` should have `None` by now. Swapping will move to `occupied_index`.
+                    self.elements.swap(curr_free_index.0, occupied_index);
+                    break;
                 }
-            }
-
-            let curr_slot = self.elements.get(curr_free_index.0);
-            curr_free = match curr_slot {
-                Some(Slot::Occupied(_)) => None,
-                Some(Slot::Empty { next_free }) => *next_free,
-                None => None,
-            }
-        }
-        while self.elements.len > self.occupied_count {
-            match self.elements.pop() {
-                Some(Slot::Occupied(_)) => {
-                    //Something is wrong, all `Slot::Occupied` should be within self.occupied_count
-                    env::panic_str(ERR_INCONSISTENT_STATE)
-                }
-                Some(Slot::Empty { next_free: _ }) => {}
-                None => {
-                    //Something is wrong, the value removed should be `Slot::Empty`
+                if occupied_index == self.elements.len {
+                    //Could not find an occupied slot to fill the free slot
                     env::panic_str(ERR_INCONSISTENT_STATE)
                 }
             }
+            curr_free =
+                if let Some(Slot::Empty { next_free }) = curr_free_slot { next_free } else { None };
         }
+        self.elements.len = self.occupied_count;
         self.first_free = None;
-    }
-
-    fn get_prev_occupied(&self, mut index: Option<u32>) -> (Option<u32>, Option<&T>) {
-        while let Some(ix) = index {
-            if let Some(Slot::Occupied(value)) = self.elements.get(ix) {
-                return (Some(ix), Some(value));
-            }
-            index = ix.checked_sub(1)
-        }
-        (None, None)
     }
 }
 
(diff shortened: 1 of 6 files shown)


near-sdk/src/store/unordered_map/mod.rs
```diff
@@ -630,6 +630,23 @@ where
     }
 }
 
+impl<K, V, H> UnorderedMap<K, V, H>
+where
+    K: BorshSerialize + BorshDeserialize + Ord + Clone,
+    V: BorshSerialize + BorshDeserialize,
+    H: ToKey,
+{
+    pub fn defrag(&mut self) {
```
frol, line 639: It definitely needs documentation with an example of where it is useful and why (best practices).

near-sdk/src/store/unordered_map/mod.rs
```diff
@@ -630,6 +630,23 @@ where
...
+        self.keys.defrag(|key, new_index| {
+            // Check if value is in map to replace first
+            let entry = self.values.get_mut_inner(key);
+            if let Some(existing) = entry.value_mut() {
```
frol, line 643: This works just fine unless I miss something:

```suggestion
            if let Some(existing) = self.values.get_mut(key) {
```

frol (approved):
@uint Thanks for the contribution! This seems good to go once those two comments I left are addressed

frol:
@austinabell If you would have the capacity, it would be great to hear your input here.

near-sdk/src/store/unordered_map/mod.rs
uint: I did my best, but would appreciate a look!

near-sdk/src/store/free_list/mod.rs
```diff
@@ -223,6 +223,92 @@ where
...
+        }
+    }
+
+    fn defrag<F>(&mut self, mut callback: F)
+    where
+        F: FnMut(&T, u32),
+    {
+        while let Some(curr_free_index) = self.next_free_slot() {
+            if let Some((value, occupied_index)) = self.next_occupied() {
+                callback(value, curr_free_index.0);
```
austinabell, line 271: wouldn't it be necessary for the callback to include the new and old indices? For example, when defragging, indices might need to be updated from old to new? I can't remember context here of why that might only include the new index.

near-sdk/src/store/unordered_map/mod.rs
uint: Yeah, looks like it does the exact same thing under the hood :+1:

near-sdk/src/store/free_list/mod.rs
```diff
@@ -223,6 +223,92 @@ where
     pub fn drain(&mut self) -> Drain<T> {
         Drain::new(self)
     }
+
+    /// Empty slots in the front of the list is swapped with occupied slots in back of the list.
+    /// Defrag helps reduce gas cost in certain scenarios where lot of elements in front of the list are
+    /// removed without getting replaced. Please see https://github.com/near/near-sdk-rs/issues/990
+    pub fn defrag<F>(&mut self, callback: F)
```
austinabell, line 230: Might make sense to have the callback be optional? In most cases I would assume devs won't care about how the indices change. That or make this method `pub(crate)` to ensure that if this type (`FreeList`) becomes public, this API doesn't get exposed as such.

near-sdk/src/store/free_list/mod.rs
```diff
@@ -223,6 +223,92 @@ where
...
+            curr_free_slot: list.first_free,
+        }
+    }
+
+    fn defrag<F>(&mut self, mut callback: F)
+    where
+        F: FnMut(&T, u32),
+    {
+        while let Some(curr_free_index) = self.next_free_slot() {
+            if let Some((value, occupied_index)) = self.next_occupied() {
```
austinabell, line 270: Intuitively seems like it would be a cleaner option to just call `.drain()` on the underlying `self.elements` for the range of `self.occupied_count..` and then with that iterator you place the filled values in each next free slot. Just a suggestion, I think it might be fine to have as is where you are swapping all values then removing the tail, it just might reduce the amount of operations done.

austinabell (commented):
I don't have bandwidth to review in depth right now, but shared some thoughts from skimming the PR. Hope it's helpful! :)

near-sdk/src/store/free_list/mod.rs
frol: I think for the matter of this PR, I would make FreeList::defrag to be `pub(crate)` and call it a day as it is out of scope to define public API for FreeList.

@austinabell Has FreeList been ever planned to become part of public API?

near-sdk/src/store/unordered_map/mod.rs
frol: The documentation looks great.

P.S. It would be great if we also extend the documentation on `.remove` and `.remove_entry` with the reference to `.defrag`

near-sdk/src/store/free_list/mod.rs
austinabell: > @austinabell Has FreeList been ever planned to become part of public API?

I mean, just depends if community asks for it. There was no one pushing for this to be public when I was working on it. No strict reason it can't be. +1 on just making the method non-public to make sure that it isn't accidentally exposed if so

near-sdk/src/store/free_list/mod.rs
uint: Ah, fair point! I think I saw that `FreeList::drain` doesn't take a range (which makes sense) and assumed it would be the same for `Vector`, but now I see that's not the case. I'll use that and get rid of `remove_tail`.

uint:
I know this is horribly out of scope, but `cargo doc` kept complaining about broken links in doc comments and it was an easy fix, so I added that.

near-sdk/src/store/unordered_map/mod.rs
uint: Done!

frol (approved):
@austinabell Thanks for the review! I feel @uint addressed all of them, and I am ready to merge this PR once all tests pass.

PR #1024, opened 2023-06-19
fix: add compiler error for using Result with init

agostbiro:
As reported in #926, currently an `#[init]` method that returns a `Result`, but misses a `#[handle_result]` annotation compiles successfully, but causes problems at runtime.

This PR fixes that, so writing

```rust
#[init]
pub fn new() -> Result<Self, &'static str> { }
```

Now fails to compile with the 

```
Serializing Result<T, E> has been deprecated. Consider marking your method with #[handle_result]...
```

message, but writing

```rust
#[init]
#[handle_result]
pub fn new() -> Result<Self, &'static str> { }
```

compiles successfully as before.


frol (approved):
@agostbiro Thanks for the contribution! This looks good and has the necessary test checking for it, so I am ready to merge it once CI is green.

PR #188, opened 2020-06-12
Add precompiled contracts for alt_bn128 curve

snjax:
This feature adds to near following precompiled contracts:

- alt_bn128_g1_multiexp
- alt_bn128_pairing_check

Example of usage is at [zeropoolnetwork/near-groth16-verifier](https://github.com/zeropoolnetwork/near-groth16-verifier)

PR #1028, opened 2023-06-21
feat: alt_bn128 curve

uint:
Picks up where #188 left off to expose alt_bn128 functions to smart contract devs.

The implementation is already there in nearcore: near/nearcore#3971

I tested briefly that these are callable from test smart contracts, but if something's missing, let me know!


frol (changes requested):
Let's add tests to these. Also, if there are good examples of how to use these functions and references to the relevant specifications / descriptions about what these "alt" are, please, add links in the doc comment as well

uint:
@frol I wasn't sure if we should test the implementation here when it lives in nearcore, but I see it's done for other cyphers. Fair enough.

I'll try to figure out/look up test vectors and explanations. Happy for hints on where to look since I'm not very knowledgeable about cryptography.

uint:
The closest I got to finding test vectors is the test cases section here: https://eips.ethereum.org/EIPS/eip-196 Sadly, most of this is way over my head.

frol:
@uint Fair point. Well, I am not familiar with this topic either, so I looked up nearcore tests: https://github.com/near/nearcore/blob/f5e8746ed85202bea244d3a2d4c698f7739d05bb/runtime/near-test-contracts/estimator-contract/src/lib.rs#L557-L720. Let's try porting those tests, as I am not comfortable shipping new functions without ever running them.

uint:
@frol I think (?) what you linked to are gas benchmarks. They don't include expected outputs, but maybe I could get them by performing the same operations using an alternative implementation of this cypher. Worth a shot. At least we'll have the happy path covered.

I still think we should include these correctness tests in nearcore only, since that's where the implementation lives. Since we only have bindings here, I could maybe add some very simple tests asserting that the bindings are pointing to existing host fns. What do you think?

frol:
> They don't include expected outputs, but maybe I could get them by performing the same operations using an alternative implementation of this cypher. Worth a shot. At least we'll have the happy path covered.

Yes, that is exactly what I had in mind. Just test the happy path.

> I still think we should include these correctness tests in nearcore only, since that's where the implementation lives.

Totally agree, but I want to catch the situation if nearcore breaks the behavior early.

> Since we only have bindings here, I could maybe add some very simple tests asserting that the bindings are pointing to existing host fns. What do you think?

I feel that calling the functions would be exactly the test (I don't suggest we copy all the test vectors from nearcore - call it once to test the interface is enough).

frol (approved):
I found more tests in nearcore: https://github.com/near/nearcore/blob/8cd095ffc98a6507ed2d2a8982a6a3e42ebc1b62/runtime/near-vm-runner/src/logic/tests/alt_bn128.rs#L87-L271

Added return value validation.

Issue #1005, opened 2023-01-23
Prohibit the use of `Self` in non-init functions


itegulov:
The use of `Self` in non-init context might be harmful as sometimes it indicates that the function was meant to be marked with `#[init]`. It might make sense to gradually prohibit this pattern, e.g. give a deprecation warning in 4.2 and give a compilation error in 5.0.

Originally pointed out in https://github.com/near/near-sdk-rs/issues/972#issuecomment-1327720519.

uint:
Once the deprecation warning is there, should we create a 5.0 milestone and attach this issue to it so that we don't forget to switch to the compilation error when it's time?

uint:
Also, should passing a `Self` argument still be fine? The original motivation was about using it as the return type from what I see.

itegulov:
We already have a tracking issue for 5.0: https://github.com/near/near-sdk-rs/issues/946

> Also, should passing a Self argument still be fine? The original motivation was about using it as the return type from what I see.

I personally think that all `Self` usage is potentially harmful in this context, but this is very opinionated

PR #1030, opened 2023-06-22
Deprecate `Self` in non-init function's return type

uint:
This is part of #1005, but doesn't close the issue.

I really wanted to produce nice compiler warnings with spans using the Proper Way, but [the API is unstable](https://doc.rust-lang.org/proc_macro/struct.Diagnostic.html). I've turned to the dark side and am using `println` instead. Happy for ideas on how to improve this!

The code already collects spans of all forbidden `Self` tokens. Once we switch to prohibiting this completely, it should be easy to emit compile errors using these spans with a much better DX.

### DIFF
--- a/near-sdk-macros/src/core_impl/info_extractor/arg_info.rs
+++ b/near-sdk-macros/src/core_impl/info_extractor/arg_info.rs
@@ -58,7 +58,7 @@ impl ArgInfo {
                 ));
             }
         };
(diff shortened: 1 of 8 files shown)


uint:
Updated to also handle `Self` in args, as per https://github.com/near/near-sdk-rs/issues/1005#issuecomment-1603457254.

frol (approved):
@uint Thanks! Looks good to me

Issue #1032, opened 2023-06-23
Refactor ABI generator to use `AttrSigInfoV2`


agostbiro:
Refactor ABI generator to use `AttrSigInfoV2`. Blocked by #1031

Issue #1031, opened 2023-06-23
Add test coverage for ABI generator


agostbiro:
The [ABI generator implementation](https://github.com/near/near-sdk-rs/blob/e4cb471631c55fb8831403ffea12afc71030abc1/near-sdk-macros/src/core_impl/abi/abi_generator.rs) currently lacks test coverage which makes it difficult to refactor.

It'd be good to add test coverage similar to the [code generator.](https://github.com/near/near-sdk-rs/blob/e4cb471631c55fb8831403ffea12afc71030abc1/near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs#L30)

itegulov:
I know it's not the same, but you might find cargo-near [integration tests](https://github.com/near/cargo-near/tree/main/integration-tests) useful. They basically test the whole ABI generation flow e2e, including SDK's ABI generator.

agostbiro:
Great, thanks @itegulov !

agostbiro:
I'm working on this.

PR #1033, opened 2023-06-23
refactor: disentangle bindgen code generation

### LINKED ISSUE #1027: Refactor bindgen code generation to take advantage of `MethodKind`
PR #1025 (continued from #890) introduces a new `AttrSigInfo` struct that collects the attributes data that varies based on the method kind (call, view or init) into a `MethodKind` enum.

This allows us to refactor the bindgen code generation to make it less error prone.

agostbiro:
This PR picks up where #1025 left off to bring the benefits of disentangling bindgen code extraction to code generation. It resolves #1027.

### Code generator

I made the changes in two steps to make it easier to review:

1. I first switched over to using `AttrSigInfoV2` where the structure of the code is unaffected by this in e3c1245b07609e2d57f31cea2c5e32c0a52d1667.
2. Then in  e404e74b0fda4b49febf55acf6afba34d34089d5 I thoroughly reworked the `method_wrapper` implementation to disentangle logic based on the return type of the method and logic based on the method kind (call, init or view). It was also possible to remove some duplicated error handling since the [code extractor](https://github.com/near/near-sdk-rs/blob/e4cb471631c55fb8831403ffea12afc71030abc1/near-sdk-macros/src/core_impl/info_extractor/visitor.rs) takes care of that now.

The purpose of the first few commits is just to make this two step approach possible. 

#### `method_wrapper`

The semantics of the code emitted from `method_wrapper` didn't change, and the only change in the form of the generated code is for fallible `init` methods. It changed from

```rust
let result = Hello::new();
match result {
    Ok(contract) => near_sdk::env::state_write(&contract),
    Err(err) => near_sdk::FunctionError::panic(&err)
}
```

to

```rust
let contract = Hello::new();
match contract {
    Ok(contract) => {
        near_sdk::env::state_write(&contract);
    }
    Err(err) => near_sdk::FunctionError::panic(&err)
}
```

This change is a consequence of removing the duplicated logic for `init` methods.

The `method_wrapper` is thoroughly tested, so while the refactoring changed a lot, I'm reasonably confident in its correctness. (I also added a new test case in 4f0918c954f653d3aa0afe408095612fa00f7001 for an uncovered edge case.)

Note that some methods added in the refactoring could be moved from `code_generator/impl_item_method_info.rs` to `code_generator/attr_sig_info.rs` for consistency. I left them there, because the code is not changed, it only got extracted to separate methods and hopefully it's easier to review this way although I see now that the GitHub PR UI can't connect them. In IntelliJ the method bodies are marked as unchanged for me. Let me know if I should move them over.

### ABI generator

The ABI generator is not covered by this refactoring, so `AttrSigInfoV1` cannot be removed yet. The reason for this is that it lacks test coverage, and I didn't want to bloat this PR by including ABI generator tests in it. Instead I made follow up issues #1031 and #1032. The temporarily derived `Clone` implementations in db86a6cdb43fefe60fa6d8dcb62d1da18eec347a are still needed for the glue code between `AttrSigInfoV1` and `AttrSigInfoV2` in the ABI generator.


frol (approved):
@agostbiro Great job! This PR looks good to me

@itegulov Could you also review it? 🙏

itegulov (approved):
LGTM!

agostbiro:
thanks @frol , it got merged now 🙏

Issue #1040, opened 2023-06-29
Disallow private init methods


agostbiro:
In https://github.com/near/near-sdk-rs/pull/890 we made private init functions emit a compilation error. I just found this example in the wild that uses private init methods with `near-sdk-rs` v4 that'd be broken by that change:

https://github.com/NearSocial/social-db/blob/915689009fef74b2f0cf189e7951bbb8d9f5c886/contract/src/upgrade.rs#L6-L8

I'm not sure  why the method is marked as 

```rust
#[private]
#[init(ignore_state)]
```

but it tells me that instead of a compilation error we should emit a deprecation warning. What do you think?

Edit: changed to emitting a warning in #1043. The warning should be turned into an error in the 5.0 release.

agostbiro:
Added a PR to just emit a warning in 4.x for private init methods: https://github.com/near/near-sdk-rs/pull/1043

And added this issue to the [5.0 tracking issue](https://github.com/near/near-sdk-rs/issues/946)

frol:
> I'm not sure why the method is marked as

Well, it is not intuitive, but it makes sense and it does work. It is used to restrict the permission to call the method only to the contract own account itself (private), and don’t try to deserialize the current STATE, while actually saving the returned value to STATE. It is abusing the attributes a bit, but it actually makes sense for migrations when your new contract structure has changed and you need to deserialize it with the old structure and save it as a new structure. Why exactly did we deprecate the combination of private + init? I still feel it is a fair usecase to limit who is allowed to initialize the contract (and even feel that by default it might be good to only allow init methods to be called by account where the code is deployed [private])

willemneal:
Do you mean you need to submit the init with a full access key for the contract? Ideally their shouldn't be a full access key. In practice this seems a suggestion.

frol:
In order to call such method (`migrate_state`) one would either need to either of these:

1. have a full-access key to sign a transaction calling `migrate_state` function
2. have a limited access key that is allowed to call `migrate_state` to sign the transaction
3. have a cross-contract call to itself to `migrate_state`

uint:
> I still feel it is a fair usecase to limit who is allowed to initialize the contract (and even feel that by default it might be good to only allow init methods to be called by account where the code is deployed [private])

@frol That sounds beyond reasonable TBH.

I see the original motivation for disallowing private + init was that there didn't seem to be a sane use case, but you've presented one and it's out there in the wild. Shall we just allow it again?

[...]

PR #1043, opened 2023-06-30
refactor: change private init method from error to warning

agostbiro:
As discussed in #1040, we jumped the gun on disallowing private init methods. In 4.x they should be deprecated and only disallowed in 5.0.

This PR rolls back emitting a compilation error for private init methods and instead just logs a warning. 

We cannot use the `#[deprecated]` attribute, since the `#[private]` attribute itself is not deprecated. 

Using the [`Diagnostic`](https://doc.rust-lang.org/proc_macro/struct.Diagnostic.html#method.warning) would be more elegant than a `println`, but unfortunately it's unstable at this time.

### DIFF
--- a/near-sdk-macros/src/core_impl/info_extractor/visitor.rs
+++ b/near-sdk-macros/src/core_impl/info_extractor/visitor.rs
@@ -86,7 +86,7 @@ impl Visitor {
         }
     }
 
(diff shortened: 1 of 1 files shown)


uint (approved):
LGTM

PR #1048, opened 2023-07-07
A manual command for comparing contract sizes

### LINKED ISSUE #881: Add manual trigger to compare compiled wasm builds against main branch
We need a way to easily compare the changes in wasm code size without relying on tracking specific wasm binary changes (#880).

Ideally, this is a manual trigger (maybe a GitHub actions trigger that runs on a specific comment on a PR) and this will build and compare the branch against the main branch and possibly outputs a comment with the summary in the PR.

How this is done doesn't really matter, but would be nice to have this data available for when we optimize the code sizes in the future.

uint:
Closes #881

# Changes

I made the following backwards-compatible changes to existing bash build scripts:
- ~~`contract-builder` docker containers spawned by build scripts now include a hash of their workdir in their names - this means you can work on two workdirs with different paths without unexpected behavior~~
- ~~docker build scripts get the `--cargo-cache-dir` parameter - if provided, they bind `.cargo` cache directories to host fs so they can be persisted as needed~~
- ~~docker builds are internally performed by the same user UID as the one running the script - this means all resulting artifacts are owned by that user and don't require root permissions to clean up.~~

I chose Python for this new script because I know it, it's convenient and popular, and has better abstractions than bash.

# Usage
In a PR, post a comment with `/compare` as the body and wait a few minutes. After a while a comparison should be posted in a new comment.

You can also run the script locally, as long as you're not on the `master` branch and the `master` branch is available in `git`. `master` will be checked out to a tempdir and later cleaned up, so no worries about polluting your local fs.

```
ci/compare_sizes/compare_sizes.py
```

See it in action [here](https://github.com/uint/near-sdk-rs/pull/2#issuecomment-1625035924).

# Notes
The workflow file needs to be on the default branch to be usable.


uint:
I suspect the buildkite failure is related to it persisting old root-owned artifacts. I'll try to clean its cache once I get the permissions!

frol:
@uint Here is the error message:

image

Nothing jumps to my head immediately. Do you know what I can do there?

frol:
@uint I don't see any cache setup on BuiltKite. Here is the whole workflow config there:

```yaml
steps:
  - command: " source ~/.cargo/env && rustc --version && cargo --version && rustup update 1.69 && rustup default 1.69 && rustup target add wasm32-unknown-unknown && ./ci-test.sh"
    label: "run tests"
    artifacts:
    - "examples/*/res/*/*.png"
    agents:
    - "distro=amazonlinux"
    timeout: "180"
```

Any clues?

uint:
> Any clues?

@frol No idea. Instead of trying to figure that out, I reverted my changes to the bash scripts and made it so that the new Python one doesn't need them. We're back to green CI.

How's this?

frol (approved):
Looks great to me. I almost missed this PR, sorry for the delay. Let's merge it now

Issue #1049, opened 2023-07-09
Deprecate declarative macros for fungible tokens 


agostbiro:
This is a follow up task from #1042 which deprecated declarative macros for non-fungible tokens. The same should be done for fungible tokens as well.

PR #1054, opened 2023-07-15
[Enhancement] Deprecate Fungible Token declarative macros.

ruseinov:
Closes https://github.com/near/near-sdk-rs/issues/1049


near-contract-standards/src/fungible_token/mod.rs
```diff
@@ -7,5 +7,8 @@ pub mod receiver;
 pub mod resolver;
 pub mod storage_impl;
 
+pub use self::core::FungibleTokenCore;
+pub use crate::storage_management::StorageManagement;
 pub use core_impl::FungibleToken;
 pub use macros::*;
+pub use resolver::FungibleTokenResolver;
```
frol, line 14: Let's align on one way of importing things. I like `self::`, but it is not the style used before, so let's not use it now.

examples/fungible-token/ft/src/lib.rs
```diff
@@ -18,7 +18,10 @@ NOTES:
 use near_contract_standards::fungible_token::metadata::{
     FungibleTokenMetadata, FungibleTokenMetadataProvider, FT_METADATA_SPEC,
 };
-use near_contract_standards::fungible_token::FungibleToken;
+use near_contract_standards::fungible_token::{
+    FungibleToken, FungibleTokenCore, FungibleTokenResolver, StorageManagement,
+};
+use near_contract_standards::storage_management::{StorageBalance, StorageBalanceBounds};
```
frol, line 24: Given that `StorageBalance` and `StorageBalanceBounds` are imported from storage_management, let's import `StorageManagement` from there as well, instead of re-exporting it through `fungible_token`.

examples/fungible-token/ft/src/lib.rs
```diff
@@ -94,8 +98,86 @@ impl Contract {
...
+    fn ft_resolve_transfer(
+        &mut self,
+        sender_id: AccountId,
+        receiver_id: AccountId,
+        amount: U128,
+    ) -> U128 {
+        let (used_amount, burned_amount) =
+            self.token.internal_ft_resolve_transfer(&sender_id, receiver_id, amount);
+        if burned_amount > 0 {
+            self.on_tokens_burned(sender_id, burned_amount);
```
frol, line 140: I feel `on_tokens_burned` and `on_account_closed` are not necessary now as that was an artifact of the "macro API". The logic can be inlined into the appropriate trait methods now.

near-contract-standards/src/fungible_token/mod.rs
```diff
@@ -7,5 +7,8 @@ pub mod receiver;
 pub mod resolver;
 pub mod storage_impl;
 
+pub use self::core::FungibleTokenCore;
+pub use crate::storage_management::StorageManagement;
```
frol, line 11: Why do we need to re-export StorageManagement here?

frol (changes requested):

near-contract-standards/src/fungible_token/mod.rs
ruseinov: It is the same way we re-export in `non_fungible_token`. The reason is that we expose `pub mod core` several lines above, so we can't use `crate` anymore.

near-contract-standards/src/fungible_token/mod.rs
ruseinov: We don't, I kept this consistent with how macros were organised.

near-contract-standards/src/fungible_token/mod.rs
ruseinov: But in fact since we only have one entity there - we can just re-export that directly.

ruseinov: Agreed.

ruseinov: There is a reason I used `self` there and it has to do with the fact that otherwise we get an ambiguity error. I have changed that to use the fully qualified path `crate::fungible_token`..

ruseinov: Note, that we do use `self` for `non_fungible_token` imports. For that same reason - there is a name ambiguity problem with `core`.

examples/fungible-token/ft/src/lib.rs
ruseinov: Yes, it makes sense. Felt redundant.

ruseinov:
This seems to be an a good shape for another round of reviews.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 
 ### Removed
 - Deprecated declarative macros for NFT impl code generation. [PR 1042](https://github.com/near/near-sdk-rs/pull/1042)
+- Deprecated declarative macros for FT impl code generation. TODO: FIX LINK [PR 1042](https://github.com/near/near-sdk-rs/pull/1042)
```
frol, line 15: Should the TODO be removed now?

near-contract-standards/src/fungible_token/mod.rs
```diff
@@ -7,5 +7,7 @@ pub mod receiver;
 pub mod resolver;
 pub mod storage_impl;
 
+pub use crate::fungible_token::core::FungibleTokenCore;
```
frol, line 10: Let's add module-level documentation with a link to the example, so when developers visit [the doc](https://docs.rs/near-contract-standards/latest/near_contract_standards/fungible_token/index.html), they can quickly find the relevant information.

frol (approved):
Overall, looks good to me. See the last couple of comments from me.

near-contract-standards/src/fungible_token/mod.rs
frol: @ruseinov We should consider migrating all the imports to use `self::` or `crate::` for local imports, but it is a separate discussion.

ruseinov: `self` is interesting for disambiguation scenarios, because the path ends up being shorter.

frol:
@ruseinov The only remaining item left is the CI. Add the offending packages to the ignore list and create a separate issue to address it in the future.

Issue #749, opened 2022-03-10
Drop the legacy `Result` serialization behavior in 5.x


itegulov:
Something to consider for 5.x (or another major release): 
 1\. Making [https://github.com/near/near-sdk-rs/pull/745](https://github.com/near/near-sdk-rs/pull/745) the default behavior for all functions that return `Result`
 2\. Deleting `#[handle_result]` annotation
 3\. Removing compilation error introduced in #748

uint:
Following the discussion in https://github.com/near/near-sdk-rs/pull/1057#discussion_r1272593025, I don't think we're going to do this. Should we close this, @frol @agostbiro?

uint:
David Tolnay invented [a way to emulate specialization](https://github.com/dtolnay/case-studies/blob/master/autoref-specialization/README.md) that basically abuses [deref coercion](https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/book/second-edition/ch15-02-deref.html#implicit-deref-coercions-with-functions-and-methods).

This weekend, I figured out a way to use this to [emulate associated functions with trait specialization](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=92a13d8edc91fef7e2b8e7f780e46aa1).

This looks like a promising way to reliably detect `Result<_, _>` types in macro expansions, including for the purposes of generating schemas. I might experiment with that sometime soon.

frol:
@uint Oh, I missed the previous comment, sorry. Well, if we can do that and the UX won't be confusing, I am all ears!

PR #1057, opened 2023-07-20
Have `#[handle_result]` support `Result` types regardless of how they're referred to

### LINKED ISSUE #852: `#[handle_result]` should not rely on `Result` being referred to using specific tokens

uint:
Closes #852


near-sdk-macros/src/core_impl/abi/abi_generator.rs
```diff
@@ -329,7 +333,7 @@ mod tests {
                 callbacks: vec![],
                 callbacks_vec: None,
                 result: Some(near_sdk::__private::AbiType::Json {
-                    type_schema: gen.subschema_for::<IsOk>(),
+                    type_schema: gen.subschema_for::< <Result<IsOk, Error> as near_sdk::__private::ResultTypeExt>::Okay>(),
```
agostbiro, line 336: Just an informational comment: I had trouble eyeballing whether this is syntactically correct so I ran the compilation tests with feature `__abi-generate` and they passed 👍

near-sdk-macros/src/core_impl/info_extractor/visitor.rs
```diff
@@ -209,14 +209,7 @@ fn is_view(sig: &Signature) -> bool {
 
 fn parse_return_kind(typ: &Type, handles_result: bool) -> syn::Result<ReturnKind> {
     if handles_result {
-        if let Some(ok_type) = utils::extract_ok_type(typ) {
-            Ok(ReturnKind::HandlesResult { ok_type: ok_type.clone() })
-        } else {
-            Err(Error::new(
```
agostbiro, line 215: I think we should return an error if here if the return type is not `Result` as the generated code for the `handles_result` attribute assumes a `Result` and will try to match on it: 

```rust
#[no_mangle]
pub extern "C" fn method() {
    near_sdk::env::setup_panic_hook();
    let contract: Hello = near_sdk::env::state_read().unwrap_or_default();
    let result = contract.method();
    match result {
        Ok(result) => {
            let result = near_sdk::serde_json::to_vec(&result)
                .expect("Failed to serialize the return value using JSON.");
            near_sdk::env::value_return(&result);
        }
        Err(err) => near_sdk::FunctionError::panic(&err),
    }
}

```

near-sdk-macros/src/core_impl/info_extractor/impl_item_method_info.rs
```diff
@@ -62,18 +62,6 @@ mod tests {
         assert!(matches!(actual, ReturnType::Type(_, ty) if ty.as_ref() == &expected));
     }
 
-    #[test]
-    fn handle_result_incorrect_return_type() {
```
agostbiro, line 66: I think we should put this back, see https://github.com/near/near-sdk-rs/pull/1057/files#r1272593025

near-sdk-macros/src/core_impl/info_extractor/mod.rs
```diff
@@ -84,5 +84,5 @@ pub struct Returns {
 pub enum ReturnKind {
     Default,
     General(Type),
-    HandlesResult { ok_type: Type },
+    HandlesResult { ty: Type },
```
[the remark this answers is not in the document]
uint: Ah, good catch. Will change.

near-sdk-macros/src/core_impl/info_extractor/visitor.rs
uint: Well, the generated code will fail to compile if given something else than a `Result`. (Come to think of it, should we use fully qualified paths for `Ok`/`Err` here? Users could shadow these names.)

Do you mean the macro should return an error if the type is not `Result`? The problem is we have no way of asserting whether the return type is `Result` at macro expansion time.

near-sdk-macros/src/core_impl/info_extractor/impl_item_method_info.rs
uint: The whole point of this PR is to remove this check at the macro level because it's too restrictive. It prevents users from using type aliases and such (like `anyhow::Result<T>`).

near-sdk-macros/src/core_impl/info_extractor/visitor.rs
uint: > then an option could be to move this feature behind a flag (e.g. #[handle_result(aliased)]) that turns off the error check.

One problem I see there is I don't believe our macros should care at all if something's aliased/renamed or not. We're really choosing between two un-ideal approaches. I lean toward mine slightly since I prefer a cleaner API and a simpler implementation here, and the error message seems like an acceptable (albeit sad) cost.

That said, it's not a hill I'd die on. If you feel strongly about your approach, I can do it.

uint: Ticket for the global symbol paths: #1058

agostbiro: > I don't believe our macros should care at all if something's aliased/renamed or not

That's a good point!

The way I'm thinking about this tradeoff is making the SDK beginner friendly vs adaptable for advanced use cases. For me, it seems more important to keep the SDK beginner friendly with the explicit error message as NEAR is trying to attract developers and it seems that the advanced feature was only requested by one user, but I might be wrong. Adding the `#[handle_result(aliased)]` flag could be a way to reconcile the two interests at the cost of clean design.

I feel we should get more input here from others cc @frol @austinabell 🙏

frol: I wonder if we can create a dummy function that will fail to compile with a more meaningful message.

To be honest, I wish we wouldn't need to have `#[handle_result]` attribute in the first place, and just use `Result`-like types, but keeping in mind reverting and non-revering error signaling (https://github.com/near/near-sdk-rs/issues/936#issuecomment-1657110020), I believe it is not realistic to get rid of `#[handle_result]` anyway.

uint: > I wonder if we can create a dummy function that will fail to compile with a more meaningful message.

That's an idea I haven't considered! I'll look into that.

> To be honest, I wish we wouldn't need to have `#[handle_result]` attribute in the first place (#749), and just use `Result`-like types, but keeping in mind reverting and non-revering error signaling ([#936 (comment)](https://github.com/near/near-sdk-rs/issues/936#issuecomment-1657110020)), I believe it is not realistic to get rid of `#[handle_result]` anyway.

Ha! So I've been toying with that idea on my own time. I feel like it could be possible, and would allow us to move more logic out of the macro system and into the type system. [Here's a very rough "sketch".](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=57801643ec5b8bb1a23c937531c6dd9a)

I figure customization of behavior should be possible with e.g. composition. There could exist a `Literal<Result<_, _>>` wrapper that just serializes a `Result` instead of handling it. Similarly, you could have a `Result<_, NoRollback<ErrorType>>` thing.

It's probably out of the scope of this PR, but I'd be happy to look into that more.

uint: @frol @agostbiro I don't think there is a way to fully customize these compilation error messages.

Here's another error message I can produce by doing something akin to a static type assertion.

```rs
error[E0271]: type mismatch resolving `<u32 as ResultTest>::Type == Result<_, _>`
  --> compilation_tests/handle_result_alias.rs:12:1
   |
12 | #[near_bindgen]
   | ^^^^^^^^^^^^^^^
   | |
   | expected this to be `Result<_, _>`
   | type mismatch resolving `<u32 as ResultTest>::Type == Result<_, _>`
   |
   = note: expected enum `Result<_, _>`
              found type `u32`
note: required by a bound in `ResultTest::assert_result`
  --> compilation_tests/handle_result_alias.rs:12:1
   |
12 | #[near_bindgen]
   | ^^^^^^^^^^^^^^^ required by this bound in `ResultTest::assert_result`
   = note: this error originates in the attribute macro `near_bindgen` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Here's the type-fu code that produces this:
```rs
            trait ResultTest {
                type Type;

                fn assert_result<T, E>()
                where
                    Self: ResultTest<Type = Result<T, E>>,
                {
                }
            }

            impl<T> ResultTest for T {
                type Type = T;
            }

            <#ty>::assert_result(); // #[handle_result] expects a `Result<_, _>`
```

frol: > Ha! So I've been toying with that idea on my own time. I feel like it could be possible, and would allow us to move more logic out of the macro system and into the type system. [Here's a very rough "sketch".](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=57801643ec5b8bb1a23c937531c6dd9a)

This is really nice. I wonder what would be the downsides (I fear it might affect error messages readability).

> Similarly, you could have a Result<_, NoRollback<MyError>> thing.

I actually love the `NoRollback` wrapper idea!

uint: > This is really nice. I wonder what would be the downsides (I fear it might affect error messages readability).

Any specific error messages you have in mind? The one in this PR becomes irrelevant if there's no `handle_result` attribute.

What I don't love is I'd probably end up implementing this for most built-in/near-sdk types, and expect the contract dev to implement the trait for their types. From the POV of the contract dev, if they need to use a third-party type, they'll probably have to create a wrapper that implements `MethodReturn` even if the type already implements something like `serde::{Deserialize, Serialize}`.

Trait specialization might be handy here, but oh well.

uint: Oh, I suppose if the return type does not implement `MethodReturn`, we have little control over the error message there.

frol: > Oh, I suppose if the return type does not implement MethodReturn, we have little control over the error message there.

@uint yes, exactly this. It will become cryptic and every single custom return type would need this trait impl for no apparent reason (from a newbie perspective)

austinabell: So an option you could do to avoid requiring specialization here is to create wrappers for the different serialization types (JSON, Borsh, etc) and change the API to explicitly wrap the return types (and even params) like [axum does](https://docs.rs/axum/latest/axum/index.html). 

So for something that is currently:
```rust
    #[result_serializer(borsh)]
    pub fn add_borsh(#[serializer(borsh)] a: u8, #[serializer(borsh)] b: u8) -> u8 {
        a + b
    }
```

Would be:

```rust
    pub fn add_borsh(#[serializer(borsh)] a: u8, #[serializer(borsh)] b: u8) -> Borsh<Pair> {
       Borsh(a + b)
    }
```

Where the trait would be something like:
```rust
pub trait NearResponse {
  fn handle_response(self);
}

impl NearResponse for Result<T, E> { ... }
impl NearResponse for Borsh<T> { ... } 
```

So you can handle Result specifically in the trait. Semantics would be a bit strange when trying to apply to params based on how the semantics are already established (and might be hard to enforce at compile time). Probably not a great idea for this lib, but maybe worth considering. The fact that it's a bit magical how the result and params are serialized I've never liked.

austinabell: Out of curiosity, I did this pattern for a different framework, but the concepts apply the same: https://github.com/austinabell/derisc0 if it's at all helpful to see the comparison

uint: @austinabell So you're basically suggesting we strictly require one of our own wrappers to be used or `Result<Wrapper<_>, _>`? I guess that does remove some vague magic from the contract dev's perspective.

austinabell: Nope, not suggesting anything. There are reasons why doing this isn't the best idea in this case, I'm just providing a slight alternative to consider.

uint: > I'm just providing a slight alternative to consider

That's the very definition of suggesting something!

What are the reasons doing this isn't the best idea? We haven't decided to go through with it and it'd be great to collect opinions and weigh benefits/drawbacks.

austinabell: I just meant that I'm not suggesting that this is the best path forward.

The main drawbacks in this case that intuitively come to mind are:
- Devs are currently very set with the semantics of `near_bindgen` and this would be breaking for almost every contract which requires a major version bump
- input parameters were chosen that each represents a field in JSON, which makes it hard to incorporate these wrapper types for that (either strange semantics or more semantic breaking changes)
- Might require a decent amount of work to create useful errors for code, or do something like `axum` with https://docs.rs/axum/latest/axum/attr.debug_handler.html for suggestions, which is a weird pattern.
  - This API requires more Rust knowledge to understand how to use it and debug incompatibilities, and most developers I've seen build in the ecosystem don't come from that background
- More verbose and explicit (which could also be a positive, depending on perspective)

uint: Okay, so the main thing I'm getting from this discussion is it's a priority to guide contract devs and we don't want to compromise on helpful error messages.

In that case, I give up on the `MethodReturn` trait idea - I'm pretty sure I wouldn't be able to get it to produce the error messages we want, though I'm also fairly sure it would make the API less verbose in common cases. I'll do something like @agostbiro's suggestion to make the new behavior an opt-in thing.

frol:
This PR was superseded by #1099

PR #1060, opened 2023-07-25
Use global paths in macro expansions

### LINKED ISSUE #1058: Use global paths in all macro expansions
This concerns the `near-sdk-macros` crate.

Review all symbols used in macro expansions and convert them to global paths - e.g. `Ok` would become `::std::result::Result::Ok`.

This helps guard against the intended symbols being shadowed by users. For non-standard-library stuff, it does not guard against crate renames and isn't perfect, but it's probably an improvement.

uint:
Closes #1058 

Switching to global paths (the ones that start with `::`) means the intended symbols can't be shadowed by user code except maybe by more advanced things like aliasing crates via `extern`.


agostbiro (approved):
Thanks, LGTM!

agostbiro:
I'm wondering if we can somehow lint for this?

PR #1062, opened 2023-07-28
Introduce `AccountIdRef`

### LINKED ISSUE #944: Enable API around borrowed AccountId
The current API forces allocations for `AccountId` and can be annoying to deal with in certain cases with lifetimes. We could build an API, similar to https://doc.rust-lang.org/nightly/std/path/struct.Path.html vs https://doc.rust-lang.org/nightly/std/path/struct.PathBuf.html semantics to allow referencing and validating account ID strings without allocating and taking ownership.

Roughly, the structs would look like:

```rust
pub struct AccountId(String);

pub struct AccountIdRef<'a>(&'a str);
```

> Alternatively could be `AccountId` and `AccountIdBuf`, but this would be more breaking from current APIs

Which can be much more flexible in cases where you never need an owned value because you don't need to create a binding to that value, you can just reference some static string or something with an adequate lifetime.

An extension of this is that we could build a macro to alleviate the construction and have the account id validation happen at compile time rather than being forced to do at runtime.

uint:
Closes #944

## Description
This PR introduces the `AccountIdRef` type for dealing with borrowed account IDs without unnecessarily allocating new strings.

The new type acts a lot like `Path` in the standard library. It wraps `str` and is unsized, meaning you can't create a value of type `AccountIdRef`, but you can create useful references: `&AccountIdRef` and `&mut AccountIdRef`.

The two types can also now be used with `Cow<'_, AccountIdRef>`.

`&AccountIdRef` supports zero-copy deserialization with `serde`.

## Extra changes
There are some extra improvements to the account ID API:
- idiomatic fallible `new` constructors,
- constructors accept a range of "stringy" types via generics,
- `PartialEq` and `PartialOrd` implementations across account id and string types.

### DIFF
--- a/near-sdk/src/environment/env.rs
+++ b/near-sdk/src/environment/env.rs
@@ -505,7 +505,7 @@ pub fn promise_and(promise_indices: &[PromiseIndex]) -> PromiseIndex {
 }
 
 pub fn promise_batch_create(account_id: &AccountId) -> PromiseIndex {
-    let account_id = account_id.as_ref();
+    let account_id = account_id.as_str();
     unsafe { sys::promise_batch_create(account_id.len() as _, account_id.as_ptr() as _) }
 }
 
--- a/near-sdk/src/types/account_id.rs
+++ b/near-sdk/src/types/account_id.rs
@@ -1,7 +1,9 @@
 use borsh::{maybestd::io, BorshDeserialize, BorshSchema, BorshSerialize};
 use serde::{de, Deserialize, Serialize};
+use std::borrow::{Borrow, Cow};
 use std::convert::TryFrom;
 use std::fmt;
+use std::ops::{Deref, DerefMut};
 
 use crate::env::is_valid_account_id;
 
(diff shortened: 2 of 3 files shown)


uint:
TODO: review existing contracts and try to switch to `&AccountIdRef` where applicable.
- [ ] examples
- [ ] standards

frol:
Oh, before it is too late, let's extract [near-account-id from nearcore](https://github.com/near/nearcore/tree/master/core/account-id) into a standalone repo, and port all the changes there, and use it here instead of maintaining 3 copies of `AccountId` (the third one was in workspaces-rs, and fortunately it actually moved to use the nearcore's near-account-id, but it suffers from lack of proper semver versioning)

uint:
@frol Sounds good to me. #944 also suggested enabling static address validation, which would require us to extract some of this logic to a standalone crate anyway.

frol:
@uint I have created the repo: https://github.com/near/near-account-id, and invited you to be the maintainer. Please, set it up there (I will take care of publishing pipeline, don't worry about that part)

uint:
> @uint I have created the repo: https://github.com/near/near-account-id, and invited you to be the maintainer. Please, set it up there (I will take care of publishing pipeline, don't worry about that part)

@frol Are there other types like this we'd want to pull out and make this repo a `near-common` one?

uint:
Closing this in favor of https://github.com/near/near-account-id/pull/2

Issue #1006, opened 2023-01-23
Refactor `info_extractor` to accumulate compilation errors


itegulov:
Currently `ItemImplInfo`, `ItemTraitInfo` (and potentially some others) return a compilation error as soon as they encounter a single method with a compilation error. Instead, they should accumulate such errors and combine them with [`syn::Error::combine`](https://docs.rs/syn/latest/syn/parse/struct.Error.html#method.combine). This would give a saner developer experience to the users whose contracts have multiple problems at once.

Originally pointed out in https://github.com/near/near-sdk-rs/pull/980#discussion_r1083681681.

jaswinder6991:
I want to work on this.

uint:
> I want to work on this.

@jaswinder6991 Sorry for the wait. Feel free to start a PR for this and reach out if you have any questions!

jaswinder6991:
Hi @uint, opened this PR https://github.com/near/near-sdk-rs/pull/1065 to address this issue.

PR #1065, opened 2023-08-08
Implemented syn::Error::combine for ItemImplInfo and ItemTraitInfo

jaswinder6991:
This is wip PR which addresses https://github.com/near/near-sdk-rs/issues/1006 
This initial commit it to get feedback on how to proceed further with this.

1. What is the best way to test this other than running run_tests.sh
2. The issue mentioned there could be potentially more implementations in info_extractor. I could not find more, can someone more experienced confirm this?
3. What else is needed to conclude the PR?


uint:
What you have is a good start and exactly what we want here!

> * What is the best way to test this other than running run_tests.sh

Good question. Take a look at [`compilation_tests`](https://github.com/near/near-sdk-rs/tree/a11c6f36eb37f2ab9b511ffca123a71afcf3d37e/near-sdk/compilation_tests) in `near-sdk`. [`all.rs`](https://github.com/near/near-sdk-rs/blob/a11c6f36eb37f2ab9b511ffca123a71afcf3d37e/near-sdk/compilation_tests/all.rs) collects all tests. Each of the remaining files is one test.

[`trybuild`](https://github.com/dtolnay/trybuild) is used here to compile source files and test the output of the compiler (pass vs fail with a specific message). See the [`trybuild` repo for details](https://github.com/dtolnay/trybuild).

It should be possible to add a test that's supposed to fail compilation and produce multiple errors, and have `trybuild` verify the output matches what's expected.

To run just the compilation tests, try this inside the `near-sdk` directory:

```
cargo test --features=unstable compilation_tests
```

> * The issue mentioned there could be potentially more implementations in info_extractor. I could not find more, can someone more experienced confirm this?

Yeah. I'd review all functions in the `info_extractor` directory that return `Result<_, syn::Error>`. It's a bit of a chore, I know.

If such a function returns errors at multiple points, I'd attempt to collect and combine them like you've done for `ItemImplInfo::new`. It might not always be possible and that's fine. Sometimes a later error check might rely on a previous check passing for some data to be available.

For example, [here's one more fn I found](https://github.com/jaswinder6991/near-sdk-rs/blob/info_extractor-error_combine/near-sdk-macros/src/core_impl/info_extractor/arg_info.rs#L46). At a glance, it has 4 points where it might return an error and they look like they can all be combined.

> * What else is needed to conclude the PR?

I think you've figured it all out. Implementation + tests is what we need. Then if all checks pass and there's an approving review, we're good to merge.

jaswinder6991:
Hi @uint, I started working on the arg_info.rs function that you mentioned. I could use some advice here:

1. I am defining dummy values like below in the function. Is this an expected practice for the sdk? Do you have some other suggestion.
https://github.com/near/near-sdk-rs/blob/9b410783791781ee6becf71e6561af53e47b0b56/near-sdk-macros/src/core_impl/info_extractor/arg_info.rs#L58-L67

2. For catching sanitisation errors, if I use a dummy value for sanitize_self of there is an error, wouldn't that cause problems later on? 
Also in extracting reference, mutability and type. I could not find suitable dummy variables to be used if there is an error. Mostly because of syn::Type.
Should I pursue this more or is it okay to leave as is.
https://github.com/near/near-sdk-rs/blob/9b410783791781ee6becf71e6561af53e47b0b56/near-sdk-macros/src/core_impl/info_extractor/arg_info.rs#L70-L73

3. I am writing compilation test which would fail and check if all combined errors are returned.

uint:
> 1. I am defining dummy values like below in the function. Is this an expected practice for the sdk? Do you have some other suggestion.

The code is correct, but at this point the whole `new` constructor is going to grow into a small ball of spaghetti. It's probably a good time to refactor this a bit.

> 2. For catching sanitisation errors, if I use a dummy value for sanitize_self of there is an error, wouldn't that cause problems later on?

I looked at why `sanitize_self` and `extract_ref_mut` might fail and it seems they'd fail for the exact same reasons (upon encountering [types other than expected](https://github.com/near/near-sdk-rs/blob/a11c6f36eb37f2ab9b511ffca123a71afcf3d37e/near-sdk-macros/src/core_impl/utils/mod.rs#L86-L92)) with the same error messages. I don't think we need to combine these two.

My approach here would be to first [collect all the data behind `Result` variables](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=70dedfc659bb653514fafe49a1bad20d), and then [match against all these result/error types at once](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e1799ea6a426b56e216451b20943314f).

The linked snippets are broad strokes, not a complete solution. Hope this helps!

jaswinder6991:
Hi @uint, thanks for you suggestions and the code snippets, They were really helpful. In my latest commit https://github.com/near/near-sdk-rs/pull/1065/commits/5eca06cd691f686e1fd142018ef2f182ee3cc4fc, I did a refactor accordingly. Collecting all results and matching them against possible Ok or Error outcomes and returning the errors if any. 
I also wrote a compilation test to check whether combined errors are returned. 
Let me know if you have any further suggestions.

near-sdk-macros/src/core_impl/info_extractor/arg_info.rs
```diff
@@ -99,18 +102,37 @@ impl ArgInfo {
...
+                reference: *reference,
+                mutability: *mutability,
+                ty: ty.clone(),
+                bindgen_ty,
+                serializer_ty,
+                self_occurrences: sanitize_self.self_occurrences.clone(),
+                original: original.clone(),
+            }),
+            _ => {
+                more_errors.extend(pat_info.err());
```
uint, line 124: Clever!

jaswinder6991:
@uint Thanks for pointing out the error span problems. I actually ended up reading more about it and used syn::new_spanned which gives better error spans and now the errors are more accurate. Maybe we should use it across the whole project. Let me know if you have more suggestions.

frol:
This PR was superseded by #1097

PR #1066, opened 2023-08-10
Make `PromiseIndex` a newtype

### LINKED ISSUE #945: Make `PromiseIndex` a newtype to avoid misuse

uint:
Closes #945

This is **breaking** and should only be released with a major version bump (5.0.0)

### DIFF
--- a/near-sdk/src/environment/env.rs
+++ b/near-sdk/src/environment/env.rs
@@ -456,7 +456,7 @@ pub fn promise_create(
 ) -> PromiseIndex {
     let account_id = account_id.as_bytes();
     unsafe {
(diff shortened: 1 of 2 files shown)


frol (approved):
Looks good to me. Given that we already have quite a couple of breaking changes, I am going ahead and merging it into master. I will cherry-pick some of the PRs to make 4.x release if necessary.

PR #1072, opened 2023-08-18
Delete the `metadata` module from `near-sdk-macros`

### LINKED ISSUE #1034: Deprecate metadata module
As per @itegulov's [comment](https://github.com/near/near-sdk-rs/pull/1033#discussion_r1243042318) the `metadata` module should be deleted as it was replaced by `abi`. 

It should be deprecated first and then removed in 5.0.

uint:
Closes #1034

This is **breaking** and should only be released with a major version bump (5.0.0)!


uint:
I also deleted the `__private::metadata` module from `near-sdk` - it also doesn't seem to be used by `cargo-near`, `near-abi` or anything of the sort, but please confirm.

frol (approved):
Let's resolve the conflict and merge it. There are several breaking changes, and I would suggest we release 5.0 sooner rather later, and iterate faster

frol:
This PR was superseded by #1098

PR #1074, opened 2023-08-18
Un-deprecate private init functions

### LINKED ISSUE #1040: Disallow private init methods

uint:
Closes #1040 (note the [discussion further down](https://github.com/near/near-sdk-rs/issues/1040#issuecomment-1687126452))

~~This is **breaking** and should only be released with a major version bump (5.0.0)!~~


frol:
Before we proceed, let’s resolve this: https://github.com/near/near-sdk-rs/issues/1040#issuecomment-1685012081

frol (approved):

Issue #51, opened 2020-01-05
Postponed execution through events


nearmax:
As part of adding higher-level building blocks on the top of near-bindgen we can consider the following form of postponed function execution implemented through events.

Consider the contact that allows tracking some external chain by accepting transactions. This contract also allows other Near contracts to verify whether some transaction in the given block.

```rust
struct Rollups {
  transactions: HashMap<BlockHeight, Vec<Transaction>>,
}

impl Rollups {
  pub fn contains_transaction(&self, block_height: BlockHeight, hash: Vec<u8>) -> bool {
    self.transactions[block_height].iter().any(|t| t.hash == hash)
  }
  pub fn add_block(&mut self, block_height: BlockHeight, block: Vec<Transaction>) {
    self.transactins.insert(block_height, block);
  }
}
```

Unfortunately, if `Alice` calls `Rollups::contains_transaction` with `block_height` that was not submitted yet by a relayer we cannot verify whether the transaction made it to the block.

To improve DevX we can add the following `Event` helper.
```rust
struct Rollups {
  transactions: HashMap<BlockHeight, Vec<Transaction>>,
  block_added_event: Event<BlockHeight>,
}

impl Rollups {
  pub fn contains_transaction(&mut self, block_height: BlockHeight, hash: Vec<u8>, callback_method: Option<Vec<u8>>) -> Some<bool> {
  let prepaid_gas = env::prepaid_gas();
   if let Some(block) = self.transactions.get(block_height) {
       Some(block.iter().any(|t| t.hash == hash))
    } else if Some(callback_method) = callback_method {
       let predecessor = env::predecessor_account_id();
       self.block_added_event
         .when(move |height| height == block_height)
         .execute(move |this: &Self| {
           let result = this.contains_transaction(block_height, hash, callback_method);
           env::promise_create(predecessor, &callback_method,
            json!({ "result": result }).to_string().as_bytes(),
            0,
            1_000_000,
        );
         });
       None
   }
  }

  pub fn add_block(&mut self, block_height: BlockHeight, block: Vec<Transaction>) {
    self.block_added_event.register(block_height);
    self.transactins.insert(block_height, block);
  }
}
```

A more black-box version of the API can be the following:
```rust
struct Rollups {
  transactions: HashMap<BlockHeight, Vec<Transaction>>,
  block_added_event: Event<BlockHeight>,
}

impl Rollups {
[...]

PR #1075, opened 2023-08-18
chore: borsh version update

dj8yfo:
This pr is targeted to resolve the following sub-issue of [borsh-rs #51](https://github.com/near/borsh-rs/issues/51):
-  test alpha release on near-sdk-rs and capture migration guide steps, create a draft PR (only merge after 1.0.0 release)


near-sdk/Cargo.toml
```diff
@@ -24,7 +24,7 @@ serde_json = "1"
 near-sdk-macros = { path = "../near-sdk-macros", version = "~4.1.1" }
 near-sys = { path = "../near-sys", version = "0.2" }
 base64 = "0.13"
-borsh = { version = "0.9", features = ["const-generics"] }
+borsh = { version = "1.0.0-alpha.3", features = ["derive", "schema"] }
```
frol, line 27: Is `schema` only needed by `near-abi`? Sounds like this feature only needs to be enabled when `abi` is enabled.

near-sdk/src/store/index_map.rs
```diff
@@ -21,7 +21,7 @@ where
     ///
     /// Note: u32 indices are used over usize to have consistent functionality across architectures.
     /// Some functionality would be different from tests to Wasm if exceeding 32-bit length.
-    #[borsh_skip]
+    #[borsh(skip, bound(deserialize = ""))]
```
frol, line 24: @dj8yfo Should we fix it on borsh side? I feel `#[borsh(skip)]` should be enough and adding `bound()` here is an indicator of a deficiency in borsh.

near-sdk/src/store/tree_map/iter.rs
```diff
@@ -981,12 +981,26 @@ where
     }
 }
 
+// FIXME: borsh 0.9 didn't feature ser/de implementation for `std::ops::Bound`
+// nor does 1.0.0, so the `BorshDeserialize`, `BorshSerialize` were and are pointless
```
frol, line 985: How does it even get compiled then? I also don't see a reason for this type fo implement BorshSerialize / BorshDeserialize, do I miss something (would it break if we completely drop them here)?

examples/adder/Cargo.toml
```diff
@@ -10,14 +10,15 @@ crate-type = ["cdylib"]
 [dependencies]
 near-sdk = { path = "../../near-sdk", features = ["unstable"] }
 serde = { version = "1", features = ["derive"] }
+borsh = { version = "1.0.0-alpha.3", features = ["derive"] }
```
frol, line 13: Is there a way to avoid direct dependency on borsh (and serde), and keep using the re-exported version from near-sdk-rs?

dj8yfo: yes, i'll address it in a new pr into `borsh` as discussed out of band, and post link to it in this thread.

near-sdk/src/store/tree_map/iter.rs
dj8yfo: It shouldn't break if we completely drop them here.

the bound is `std::ops::Bound<K>: borsh::ser::BorshSerialize`, so the code compiles.
If an implementation of trait is added for this type in `borsh`, the code will even become runnable.

near-sdk/src/collections/lazy_option.rs
```diff
@@ -17,7 +17,7 @@ const ERR_VALUE_DESERIALIZATION: &str = "Cannot deserialize value with Borsh";
 #[derive(BorshSerialize, BorshDeserialize)]
 pub struct LazyOption<T> {
     storage_key: Vec<u8>,
-    #[borsh_skip]
+    #[borsh(skip)]
     el: PhantomData<T>,
```
dj8yfo, line 21: `PhantomData` is hardcoded [exception](https://github.com/near/borsh-rs/blob/master/borsh-derive/src/internals/generics.rs#L238), when type parameter is skipped from being added to any kinds of bounds.

near-sdk/src/store/lazy/mod.rs
```diff
@@ -66,7 +66,7 @@ where
 {
     /// Key bytes to index the contract's storage.
     storage_key: Box<[u8]>,
-    #[borsh_skip]
+    #[borsh(skip, bound(deserialize = ""))]
     /// Cached value which is lazily loaded and deserialized from storage.
     cache: OnceCell<CacheEntry<T>>,
```
dj8yfo, line 71: `OnceCell` has a `Default` [implementation](https://docs.rs/once_cell/1.18.0/src/once_cell/lib.rs.html#425) of an empty cell, not requiring `Default` on the contained type.

near-sdk/src/store/index_map.rs
dj8yfo: Here's a link to [this case in migration guide](https://github.com/near/borsh-rs/pull/201/files#diff-d94d69a1dd0393d69f6594f304958b8893aa8749dfbc32cb15cd4addddbdd046R348).

1. We've inherited/copied this behaviour from `serde`. 
2. `core::default::Default` for skipped field derivation
It's a logical thing to do in majority of cases: straightforward implementation of `Default` for `struct SomeStruct<T>`  will require `T: Default`, e.g., when one of fields contains `T` directly. 
In this case, `StableMap` is a collection, and it has a `Default` implementation, returning empty collection irrelevant of the values and bounds on type parameters.

near-sdk/src/store/tree_map/mod.rs
```diff
@@ -101,6 +98,7 @@ where
     K: BorshSerialize,
 {
     root: Option<FreeListIndex>,
+    #[borsh(bound(deserialize = ""))]
     nodes: FreeList<Node<K>>,
```
dj8yfo, line 102: This crate contains a number of unsual types.

E.g. `IndexMap<T>` doesn't technically require any bounds on `T` to perform `BorshSerialize` or `BorshDeserialize`. 
It only requires that there's an implementation of both for [Box<[u8]>](https://github.com/near/near-sdk-rs/blob/master/near-sdk/src/store/index_map.rs#L17), which is provided by `borsh` library. 
Instead, the bounds `T: BorshSerialize`, `T: BorshDeserialize` are baked deep into the type's implementation and basic operations with it, 
[flush, BorshSerialize](https://github.com/near/near-sdk-rs/blob/master/near-sdk/src/store/index_map.rs#L61), [get, BorshDeserialize](https://github.com/near/near-sdk-rs/blob/master/near-sdk/src/store/index_map.rs#L107). 

Due to this pattern being pervasive across the repo, the automatic bound `K: BorshDeserialize` on type parameter in non-skipped field is dropped on commented line with `#[borsh(bound(deserialize = ""))]`, as it wasn't technically required for the trait's implementation in the code previous to this pr. In all similar occasions the bounds were made as close to the version in `master` as possible with these `#[...(bound(...))]` attributes.

near-sdk/src/store/index_map.rs
dj8yfo: A similar example has aslo been published to [1.0.0-alpha.3/borsh doc](https://docs.rs/borsh/1.0.0-alpha.3/borsh/derive.BorshDeserialize.html#interaction-with-borshskip)

dj8yfo: By the way, the derived implementations `#[derive(BorshSerialize, BorshDeserialize)]` won't actually serialize/deserialize the collection, which was my first impression, the heavy lifting is done in `flush` and `get` methods, 
they will only serialize/deserialize the `prefix` of the collection.

near-sdk/Cargo.toml
dj8yfo: Nope, there're direct imports outside of `abi` feature's scope:

```bash
cargo check -p near-sdk --no-default-features
```

```bash
 1  error[E0432]: unresolved import `borsh::BorshSchema`
  --> near-sdk/src/promise.rs:1:5
 2  error[E0432]: unresolved imports `borsh::schema`, `borsh::BorshSchema`
  --> near-sdk/src/private/metadata.rs:1:13
 3  error[E0432]: unresolved import `borsh::BorshSchema`
  --> near-sdk/src/types/account_id.rs:1:31
 4  error[E0432]: unresolved import `borsh::BorshSchema`
  --> near-sdk/src/types/gas.rs:1:31
 5  error: cannot determine resolution for the derive macro `BorshSchema`
  --> near-sdk/src/private/metadata.rs:6:44
 6  error: cannot determine resolution for the derive macro `BorshSchema`
   --> near-sdk/src/private/metadata.rs:21:44
 7  error: cannot determine resolution for the derive macro `BorshSchema`
   --> near-sdk/src/types/account_id.rs:37:84
 8  error: cannot determine resolution for the derive macro `BorshSchema`
   --> near-sdk/src/types/gas.rs:18:5
 9  error[E0433]: failed to resolve: could not find `schema` in `borsh`
    --> near-sdk/src/promise.rs:232:43
10  error[E0433]: failed to resolve: could not find `schema` in `borsh`
    --> near-sdk/src/promise.rs:232:71
11  error[E0433]: failed to resolve: could not find `schema` in `borsh`
    --> near-sdk/src/promise.rs:237:32
12  error[E0433]: failed to resolve: could not find `schema` in `borsh`
    --> near-sdk/src/promise.rs:577:43
13  error[E0433]: failed to resolve: could not find `schema` in `borsh`
    --> near-sdk/src/promise.rs:577:71
14  error[E0433]: failed to resolve: could not find `schema` in `borsh`
    --> near-sdk/src/promise.rs:582:32

```
There is, e.g., 
```rust
impl BorshSchema for Promise {

```

but 
```rust
#[cfg(feature = "abi")]
impl schemars::JsonSchema for Promise {

```

dj8yfo: Some extra-redundant cross-referencing: [pr](https://github.com/near/borsh-rs/pull/210)

examples/adder/Cargo.toml
dj8yfo: Some extra-redundant cross-referencing: [pr](https://github.com/near/borsh-rs/pull/210)

near-sdk/src/store/index_map.rs
dj8yfo: Actually i've discovered a [test](https://github.com/near/near-sdk-rs/blob/c2352df2491dcc979febfc4650c65bdba4bceb6c/near-sdk/src/store/vec/mod.rs#L885) of `Vector<T>`, based on `IndexMap<T>`, which tests exactly this behaviour of restoring collection just by restoring its `prefix`, so `#[derive(BorshSerialize, BorshDeserialize)]` isn't accidental, they were added by design.

examples/mission-control/src/agent.rs
```diff
@@ -1,12 +1,13 @@
...
 use crate::asset::*;
 use crate::rate::*;
-use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
+use near_sdk::borsh::{BorshDeserialize, BorshSerialize};
 use near_sdk::serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
 #[derive(Clone, Serialize, Deserialize, BorshDeserialize, BorshSerialize)]
 #[serde(crate = "near_sdk::serde")]
+#[borsh(crate = "::near_sdk::borsh")]
```
frol, line 10: I would align these and I think `::` prefix is too verbose and could be confusing to users, so I suggest we drop it from the borsh attribute.

```suggestion
#[serde(crate = "near_sdk::serde")]
#[borsh(crate = "near_sdk::borsh")]
```

near-sdk/Cargo.toml
frol: @dj8yfo Please, feature-gate uses of BorshSchema to be part of `abi` feature 🙏

frol (commented):
Looks great to me!

frol (approved):
@dj8yfo This change looks great to me!

examples/adder/Cargo.toml
```diff
@@ -17,7 +17,7 @@ workspaces = "0.7.0"
 serde_json = "1.0"
 tokio = { version = "1.14", features = ["full"] }
 anyhow = "1.0"
-near-abi = "0.3.0"
+near-abi = { version = "0.4.0", git = "https://github.com/dj8yfo/near-abi-rs.git", branch = "borsh_update_alpha_5" }
```
frol, line 20: near-abi 0.4.0 is now released, please, update and let's merge this PR

frol (approved):
🎉

Issue #539, opened 2021-08-13
Add commit hook for styling


willemneal:
I usually have fmt on save set in my IDE but when it fails and I only learn of this when my tests fail. In `npm` there are several easy to use commit hooks and I'm wondering if that would be a good solution here so that we don't waste CI time on failed formatting.

austinabell:
> I usually have fmt on save set in my IDE but when it fails and I only learn of this when my tests fail. In `npm` there are several easy to use commit hooks and I'm wondering if that would be a good solution here so that we don't waste CI time on failed formatting.

Yeah, this is an option, but the hook can't run by default, it needs to be initialized manually by any contributor. Grin has one [here](https://github.com/mimblewimble/grin/blob/master/.hooks/pre-commit) as an example, but this is the only project I know off-hand that does this for a rust project.

willemneal:
Aright sounds good to me.  It can be documented and we can keep the CI hook.  I mainly want it for myself as I've had this happen too often when the IDE silently fails to fmt.  

Also this from their contributing docs:

> When you make a commit, rustfmt will be run and we also automatically reformat the .rs files that your commit is touching.

Makes it sound like that hook will be used automatically, just not stage or commit the file.

curryrasul:
> Dfkl

uint:
Closed by #1078

Issue #809, opened 2022-05-09
Remove status-message-collections and merge with status-message


austinabell:
Actually never realized this example exists. Should prob migrate to one contract and remove the bloat that currently exists (like why there is a `LookupSet` in collections that is completely unused)

The API should be the same as what exists within `status-message`, so the factory-contract example that uses this won't break

shariffdev:
Hey, I made a PR on this here, https://github.com/near/near-sdk-rs/pull/1076
I also could not find any places where the `status-message-collections` was being used in the factory-contract.

uint:
> Hey, I made a PR on this here, #1076 I also could not find any places where the `status-message-collections` was being used in the factory-contract.

I'm confident the intention here is to have one contract called `status-message` that uses a `LookupMap` under the hood.

I believe @austinabell assumed the current `status-message` contract would be deleted and then `status-message-collections` would be renamed to `status-message` and adapted, but it shouldn't matter much which angle you tackle this from.

uint:
Closed by #1076

PR #1076, opened 2023-08-19
merge status-message-collections

shariffdev:
Addresses: https://github.com/near/near-sdk-rs/issues/809

This PR eliminates the `status-message-collection` example, leaving only one `status-message`-esque example.


uint (changes requested):
Hi! Unfortunately, this is not quite what we want here. We'd like there to be one `status-message` contract using the `LookupMap`.

`LookupMap` is more efficient (in the NEAR blockchain context) and the appropriate storage option to choose in situations like these. Since examples need to show contract devs best practices, we really need to advertise it over `HashMap` here.

shariffdev:
> Hi! Unfortunately, this is not quite what we want here. We'd like there to be one `status-message` contract using the `LookupMap`.
> 
> `LookupMap` is more efficient (in the NEAR blockchain context) and the appropriate storage option to choose in situations like these. Since examples need to show contract devs best practices, we really need to advertise it over `HashMap` here.

Understood, making chnages.

uint (approved):
LGTM!

uint:
@shariffdev One note: please include the phrase `Closes #539` when opening a PR for a specific issue. That way the issue gets automatically closed when we merge.

PR #778, opened 2022-04-04
Feat/payout

ymc182:
Takes #673 closer to the final stage

PR #1077, opened 2023-08-20
Feat/payout

ruseinov:
A continuation of https://github.com/near/near-sdk-rs/pull/778


ruseinov:
@frol I'm not sure if the `sim` change should belong to this PR at all to be honest. Thoughts?

frol (changes requested):

Cargo.toml
```diff
@@ -2,6 +2,7 @@
 members = [
     "near-sdk",
     "near-sdk-macros",
+#    "near-sdk-sim",
```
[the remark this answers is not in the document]
ruseinov: So I thought, removed!

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -47,6 +48,7 @@ pub struct NonFungibleToken {
     // required by approval extension
     pub approvals_by_id: Option<LookupMap<TokenId, HashMap<AccountId, u64>>>,
     pub next_approval_id_by_id: Option<LookupMap<TokenId, u64>>,
+    pub royalties: Option<Royalties>,
```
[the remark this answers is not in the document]
ruseinov: Is a CHANGELOG entry an option?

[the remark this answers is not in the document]
frol: @ruseinov Yes, it is, just need to make it really visible.

frol:
@ruseinov May I ask you to submit a new PR to fix the CI failure? (I believe it is not related to this change)

ruseinov:
> @ruseinov May I ask you to submit a new PR to fix the CI failure? (I believe it is not related to this change)

definitely!

ruseinov:
> @ruseinov May I ask you to submit a new PR to fix the CI failure? (I believe it is not related to this change)

This seems to only be reproducible on this branch.
I have tried reducing the `gas` amount to `10_000_000_000_000`. Then `assert!(res.is_failure());` works, but `assert_eq!(res.logs().len(), 0);` still does not.

examples/non-fungible-token/tests/workspaces/test_core.rs
```diff
@@ -185,7 +185,7 @@ async fn simulate_transfer_call_receiver_panics_and_nft_resolve_transfer_produce
             Some("transfer & call"),
             "incorrect message",
         ))
-        .gas(30_000_000_000_000)
+        .gas(1)
```
ruseinov, line 188: @frol FYI: this fixed the issue that was so confusing.

Somehow the gas amount for this one has changed, I guess it'd be nice for me to investigate how those are calculated. 
There's also an option to supply more gas and still let it fail, but then it manages to emit a log message and fails 
when the assertion happens. 
It seems to me like a somewhat flaky test, perhaps it'd be nice to revisit those in the near future and have some
good docs describing the behaviour.

ruseinov:
This is ready for another round of reviews. All the checks pass.

examples/non-fungible-token/tests/workspaces/test_core.rs
frol: @ruseinov Well, this is totally unexpected that it succeeds as the function call definitely cannot complete within 1 gas limit, especially if before that it could not succeed within 30 Tgas limit. Please, dig more into it.

ruseinov: > @ruseinov Well, this is totally unexpected that it succeeds as the function call definitely cannot complete within 1 gas limit, especially if before that it could not succeed within 30 Tgas limit. Please, dig more into it.

The way that it currently works is:
1. If we give it 30 Tgas - it succeeds, because it needs way less
2. If we give it something around 20 - then it fails, but emits a log message, which fails the test.
3. With 1 gas as an extreme it does fail and emits no log messages.

I'll look into why.

ruseinov: I'm not sure though what has changed the gas requirements in the first place. And that's what bothers me the most.

ruseinov:
> If 30 Tgas works file, keep it that way, and if you feel it is too much, let's revisit it in a separate PR.

It's too much, the request succeeds then. This test fails with 30 Tgas, it needs way less to succeed.
The idea of this test is to simulate a failed request that emits no logs. 
We can't revisit this in a separate PR if we want the CI to be green. But I will adjust the value to make sure it's
the largest possible still satisfying the requirements.

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -391,7 +396,6 @@ impl NonFungibleTokenCore for NonFungibleToken {
         msg: String,
     ) -> PromiseOrValue<bool> {
         assert_one_yocto();
-        require!(env::prepaid_gas() > GAS_FOR_NFT_TRANSFER_CALL, "More gas is required");
```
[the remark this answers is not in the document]
ruseinov: Right, I see what I've missed.

frol:
@ruseinov Please, fix CI, it seems that we need to downgrade more crates here: https://github.com/near/near-sdk-rs/actions/runs/6353767202/workflow?pr=1077#L26-L31

ruseinov:
> @ruseinov Please, fix CI, it seems that we need to downgrade more crates here: https://github.com/near/near-sdk-rs/actions/runs/6353767202/workflow?pr=1077#L26-L31

It's done!

ruseinov:
> > @ruseinov Please, fix CI, it seems that we need to downgrade more crates here: https://github.com/near/near-sdk-rs/actions/runs/6353767202/workflow?pr=1077#L26-L31
> 
> It's done!

Also done [here](https://github.com/near/near-sdk-rs/pull/1089) as a separate PR.

near-contract-standards/src/non_fungible_token/payout/payout_impl.rs
```diff
@@ -0,0 +1,75 @@
...
+        };
+        let rest = balance - royalty_payment;
+        let owner_payout: u128 = payout.payout.get(owner_id).map_or(0, |x| x.0) + rest;
+        payout.payout.insert(owner_id.clone(), owner_payout.into());
+        payout
+    }
+}
+
+fn apply_percent(percent: BasisPoint, int: u128) -> u128 {
+    int * percent as u128 / 100u128
```
frol, line 51: 1. Let's explicitly use `.checked_mul()` to avoid potential vulnerabilities due to integer overflows.
2. Use `.into()` to upcast, so if we ever change BasisPoint to signed int or whatnot we actually catch it during compilation time.

frol (changes requested):

near-contract-standards/src/non_fungible_token/payout/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+pub struct Payout {
+    pub payout: HashMap<AccountId, U128>,
+}
+
+#[derive(Deserialize, Serialize, BorshDeserialize, BorshSerialize, Default, Debug)]
+#[serde(crate = "near_sdk::serde")]
+pub struct Royalties {
+    key_prefix: Vec<u8>,
+    pub accounts: HashMap<AccountId, BasisPoint>,
+    pub percent: BasisPoint,
```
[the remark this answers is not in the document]
ruseinov: I think I have overlooked some of this, I've removed the occurrences, but not the definition. I will browse through this code again to make sure everything makes sense.

near-contract-standards/src/non_fungible_token/payout/payout_impl.rs
ruseinov: Yep, I'm about to review this again. I'm sure something else will pop up.

near-contract-standards/src/non_fungible_token/payout/mod.rs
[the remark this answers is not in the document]
ruseinov: Alright, that's used [here](https://github.com/ruseinov/near-sdk-rs/blob/ca930d78c3cafd4e68c9ef5aa92cf9850fb8b6b9/near-contract-standards/src/non_fungible_token/payout/payout_impl.rs#L50).

near-contract-standards/src/non_fungible_token/payout/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+#[serde(crate = "near_sdk::serde")]
+pub struct Payout {
+    pub payout: HashMap<AccountId, U128>,
+}
+
+#[derive(Deserialize, Serialize, BorshDeserialize, BorshSerialize, Default, Debug)]
+#[serde(crate = "near_sdk::serde")]
+pub struct Royalties {
+    key_prefix: Vec<u8>,
+    pub accounts: HashMap<AccountId, BasisPoint>,
```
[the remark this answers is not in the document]
ruseinov: I can see that `std::collections::HashMap` is used in a variety of places. We should probably create an issue to change that.

near-contract-standards/src/non_fungible_token/payout/payout_impl.rs
```diff
@@ -0,0 +1,75 @@
+use super::NonFungibleTokenPayout;
+use crate::non_fungible_token::core::NonFungibleTokenCore;
+use crate::non_fungible_token::payout::*;
+use crate::non_fungible_token::NonFungibleToken;
+use near_sdk::{assert_one_yocto, env};
+use near_sdk::{require, AccountId, Balance, IntoStorageKey};
+use std::collections::HashMap;
+impl Royalties {
```
agostbiro, line 8: nit: please add a space between the `impl` and the imports and between the methods in the `impl`.

agostbiro, line 8: nit: please add a space between the impl and the imports and between the methods in the impl.

near-contract-standards/src/non_fungible_token/payout/payout_impl.rs
```diff
@@ -0,0 +1,75 @@
...
+        require!(
+            self.accounts.len() <= 10,
+            "can only have a maximum of 10 accounts spliting royalties"
+        );
+        let mut total: BasisPoint = 0;
+        self.accounts.iter().for_each(|(_, percent)| {
+            require!(*percent <= 100, "each royalty should be less than 100");
+            total += percent;
+        });
+        require!(total <= 100, "total percent of each royalty split  must be less than 100")
```
agostbiro, line 30: It's not so clear for me what's the invariant here. Before reading this line I thought that `self.percent` should be equal to the `total`. Could you explain some more please to make sure I understand?

near-contract-standards/src/non_fungible_token/payout/payout_impl.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+fn apply_percent(percent: BasisPoint, int: u128) -> u128 {
+    int.checked_mul(percent as u128).unwrap_or_else(|| env::panic_str("royalty overflow")) / 100u128
+}
+
+impl NonFungibleTokenPayout for NonFungibleToken {
+    #[allow(unused_variables)]
+    fn nft_payout(&self, token_id: String, balance: U128, max_len_payout: Option<u32>) -> Payout {
```
agostbiro, line 56: Could you explain why `max_len_payout` is not used here? It's a MUST condition [in the standard](https://github.com/willemneal/NEPs/blob/18828873648eff1a2e8464db234aefd70918b3e0/neps/nep-0199.md#specification) to support it, unless I'm missing something.

near-contract-standards/src/non_fungible_token/payout/payout_impl.rs
```diff
@@ -0,0 +1,75 @@
...
+            "can only have a maximum of 10 accounts spliting royalties"
+        );
+        let mut total: BasisPoint = 0;
+        self.accounts.iter().for_each(|(_, percent)| {
+            require!(*percent <= 100, "each royalty should be less than 100");
+            total += percent;
+        });
+        require!(total <= 100, "total percent of each royalty split  must be less than 100")
+    }
+    pub fn create_payout(&self, balance: Balance, owner_id: &AccountId) -> Payout {
```
agostbiro, line 32: It'd be good to add some unit tests for the payout calculation especially around edge cases (0 inputs, inputs that can produce results that overflow, returned payout sum is not higher than the balance)

agostbiro (changes requested):
Thanks for this difficult PR @ruseinov ! I went through it and added some comments, but I'll probably need a second pass to understand all the changes.

near-contract-standards/src/non_fungible_token/payout/payout_impl.rs
ruseinov: Ah, now I see what the field is for. I think this just got lost in translation, will fix.

near-contract-standards/src/non_fungible_token/payout/payout_impl.rs
ruseinov: agreed.

ruseinov: I don't think there is a way to overflow as we make sure that the total percent is at most 100.

agostbiro: I'm not sure I follow. :) The `apply_percent` function can overflow I think.

near-contract-standards/src/non_fungible_token/payout/payout_impl.rs
agostbiro: Ok thanks I think understand it better now. But shouldn't the sum of the percentages that each user gets be less than equal the total percent of incoming balance used for royalties? In other words, shouldn't the invariant be `require!(total <= self.percent` instead of `require!(total <= 100)`?

ruseinov: It's not how it works. First we take percent of the total balance. Then we distribute that percent to the accounts at their own specified percent. 

E.g. `incoming balance 100; total sum to distribute - 10% of incoming balance = 10; each account gets their share according to their individual percentage.`

ruseinov: Or at least that is how it's currently done.

near-contract-standards/src/non_fungible_token/payout/payout_impl.rs
ruseinov: Yeah, you are right. Because it first gets multiplied by 100. I have introduced an overflow test already to check this.

ruseinov: I would argue that we could change the behaviour though. Do something like
```
    match int.checked_mul(percent as u128) {
       Some(int) => int / 100,
       None => int / 100 * percent
     }
```
Then it will still be correct, but never overflowing.

near-contract-standards/src/non_fungible_token/payout/payout_impl.rs
```diff
@@ -0,0 +1,233 @@
...
+
+        let mut royalties = Royalties::new(KEY_PREFIX, 100);
+
+        mem::swap(&mut royalties.accounts, &mut map);
+        royalties.validate();
+    }
+
+    #[test]
+    #[should_panic(expected = "each royalty should be at most 100")]
+    fn validate_roalty_per_account_fails() {
```
agostbiro, line 192: Nit: typo in "royalty"

near-contract-standards/src/non_fungible_token/payout/payout_impl.rs
```diff
@@ -0,0 +1,233 @@
...
+
+        let mut royalties = Royalties::new(KEY_PREFIX, 100);
+
+        mem::swap(&mut royalties.accounts, &mut map);
+        royalties.validate();
+    }
+
+    #[test]
+    #[should_panic(expected = "total percent of each royalty split must be at most 100")]
+    fn validate_total_roalties_fails() {
```
agostbiro, line 206: Nit: typo in royalty

near-contract-standards/src/non_fungible_token/payout/payout_impl.rs
agostbiro: Thanks, understand it now 👍

agostbiro (dismissed):
Thanks for the changes @ruseinov ! LGTM with two small nits.

ruseinov:
> Thanks for the changes @ruseinov ! LGTM with two small nits.

Cheers! Fixed!

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -88,6 +92,10 @@ impl NonFungibleToken {
             tokens_per_owner: enumeration_prefix.map(LookupMap::new),
             approvals_by_id,
             next_approval_id_by_id,
+            royalties: royalties_prefix.map(|prefix| {
+                // TODO: Figure out where the percent has to be coming from.
+                Royalties::new(prefix, 0)
```
frol, line 97: So, currently the `royalties.payout.percent` is a public field, and it seems that the expectation is that developers using this extension will adjust it. I find this contract-level percentage quite inflexible - 0% shared between all payouts is useless, 1-99% could be limiting or confusing, and 100% seems to be the only reasonable option, but then payout to the token owner needs to be explicitly listed (storage wasted).

In either cases, please, provide usage-level documentation on how to use this stuff.

frol (changes requested):

near-contract-standards/src/non_fungible_token/core/core_impl.rs
ruseinov: I was thinking that this could be done in a different way. Why don't we sum all the payouts and send the rest to the owner:

```
    pub fn create_payout(&self, balance: Balance, owner_id: &AccountId) -> Payout {
        let mut payout = Payout {
            payout: self
                .accounts
                .iter()
                .map(|(account, percent)| (account.clone(), apply_percent(percent, balance).into()))
                .collect(),
        };
        let rest = balance - payout.payout.values().fold(0, |acc, &sum| acc + sum);
        let owner_payout: u128 = payout.payout.get(owner_id).map_or(0, |x| x.0) + rest;
        payout.payout.insert(owner_id.clone(), owner_payout.into());
        payout
    }
```

ruseinov:
There's one failing test that left me confused. When calling `create_payout` with `Balance::MAX` I end up with a smaller value, when it reaches `apply_percent`. The balance is not modified anywhere along the way.

ruseinov:
@frol @agostbiro This needs to be re-approved and merged if all good.

ruseinov:
> @frol @agostbiro This needs to be re-approved and merged if all good.



> @frol @agostbiro This needs to be re-approved and merged if all good.

Right, after updating to latest it needs some small fixes, I'll take care of those.

ruseinov:
Done.

near-contract-standards/src/non_fungible_token/payout/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/// This struct represents percentage of total royalty per receiver as well as the total percentage
+/// of distributed royalties based on incoming payment.
+#[derive(BorshDeserialize, BorshSerialize, Debug)]
+#[borsh(crate = "near_sdk::borsh")]
+pub struct Royalties {
+    /// A mapping of accounts to the percentage of total royalty to be distributed.
+    pub accounts: TreeMap<AccountId, BasisPoint>,
+}
```
frol, line 27: Weeeeell, I went through another round of review and noticed a MAJOR issue here - the royalties as of this implementation are applied to all the tokens on the contract. The original idea of payouts was that each individual NFT would have its own royalties setup, e.g. every previous owner of a specific NFT could receive % of the reselling price.

frol (changes requested):

near-contract-standards/src/non_fungible_token/payout/mod.rs
ruseinov: This has been fixed.

ruseinov:
@frol seems to be fixed, some flaky example tests due to `AddrInUse`, but not related to this PR.

near-contract-standards/src/non_fungible_token/payout/payout_impl.rs
```diff
@@ -0,0 +1,323 @@
...
+                            balance,
+                        )),
+                    )
+                })
+                .filter(|(_, payout)| payout.0 > 0)
+                .collect(),
+        };
+        let rest = balance - payout.payout.values().fold(0, |acc, &sum| acc + sum.0);
+        let owner_payout: u128 = payout.payout.get(owner_id).map_or(0, |x| x.0) + rest;
+        payout.payout.insert(owner_id.clone(), owner_payout.into());
```
frol, line 77: Something crazy is going on here. Royalties should be just an `UnorderedMap<TokenId, HashMap<AccountId, Percentage>>`, and then `create_payout` will be just: 

```rust
royalties.get(token_id)
    .items()
    .map(|(account_id, percentage)| (account_id, balance * percentage / 100))
    .collect()
```

near-contract-standards/src/non_fungible_token/payout/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+///     fn nft_payout(
+///         &self,
+///         token_id: String,
+///         balance: U128,
+///         max_len_payout: Option<u32>,
+///     ) -> Payout {
+///         self.tokens.nft_payout(token_id, balance, max_len_payout)
+///     }
+///     #[payable]
+///     fn nft_transfer_payout(
```
frol, line 77: I realized that most of the examples lack at least demo methods on how to mint tokens and, in this case, how to assign royalties to the tokens. This would highlight the problem that in order to assign royalties one would need to access internal fields deep down and manipulate with them manually:

```rust
self.tokens.royalties.unwrap().royalties.entry(token_id).insert(royalties_to_account_id, 5)
```

Without this example this implementation is most probably won't be used and we may even argue that we may not need to merge it in the first place.

frol (changes requested):

near-contract-standards/src/non_fungible_token/payout/mod.rs
ruseinov: Yes, I have actually been thinking about that too when I first glanced at the original PR, but then it got forgotten as I moved forward.
We do indeed have to expose an interface to assign royalties.

near-contract-standards/src/non_fungible_token/payout/payout_impl.rs
frol: @ruseinov Also, as a rule of thumb, if you write a code that iterates over near_sdk collections without bounds and potentially may exceed more than 100 lookups, you need to redesign the approach as that won't work with the growing number records. In this case, as I mentioned above, it needs a completely different layout

ruseinov: My mistake, I made a fast fix of a problem that completely redefined the solution.

ruseinov: Actually, the truth is that the original map is bound by `validate` method, it only allows for 10 accounts. So technically it's not an issue. 
The data structure and it's access are a little bit convoluted, but if we were to redesign this collection and the approach - we'd have to discuss how we are going to validate bounds going forward.

near-contract-standards/src/non_fungible_token/payout/mod.rs
ruseinov: Just going to leave this here: what we need to do is introduce a `create_royalties` method or similar that will only be accessible to nft owner.

PR #1078, opened 2023-08-21
feat/git-hook

shariffdev:
Addresses: https://github.com/near/near-sdk-rs/issues/539

The PR copies the `pre-commit` hook as defined in [grin](https://github.com/mimblewimble/grin/blob/94277bba9db2e31e713f2a4cfdc37b11e46a1630/.hooks/pre-commit#L1) with it's associated License.

### DIFF
--- a/.hooks/pre-commit
+++ b/.hooks/pre-commit
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+
+# Copyright 2021 The Grin Developers
+#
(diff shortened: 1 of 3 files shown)


uint (approved):
Okay, this seems handy and works even for compilation tests and things hidden behind non-default features. I couldn't find any way to horrendously break it.

I like it. Let's do it.

uint:
Thank you!

PR #1079, opened 2023-08-23
chore: replace manual value extraction  

shariffdev:
The `TODO` can be removed as the API discussed [here](https://github.com/near/workspaces-rs/issues/201) was implemented and in a separate PR the json API for `Value` was used.

Before: https://github.com/near/near-sdk-rs/blob/7d31a5b1ddd040407f646a5cf863fffb08729251/examples/fungible-token/tests/workspaces.rs#L220-L229
After: https://github.com/near/near-sdk-rs/blob/6c8556e9467127907b0b3fe5d9691c1657e8cc2b/examples/fungible-token/tests/workspaces.rs#L202-L210


uint (approved):
Good catch!

Issue #1071, opened 2023-08-17
Address audit reports properly.


ruseinov:
This is the original PR, that ignores one more audit advice: https://github.com/near/near-sdk-rs/pull/1068.

Needs to be addressed properly with reasoning going forward.

PR #1080, opened 2023-08-31
chore: adding an `audit.toml` file

shariffdev:
Closes https://github.com/near/near-sdk-rs/issues/1071
Using an config file for audit reports and comments.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -3,7 +3,3 @@ ignore = [
     "RUSTSEC-2020-0071", # `chrono` is a transient dependency used by `near-sandbox-utils` and other dependencies
     "RUSTSEC-2022-0093", # security audit error in ed25519-dalek which is not relevant for near-sdk-rs
 ]
-
-[output]
-quiet = false
-deny = ["warnings"]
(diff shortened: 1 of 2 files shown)


.cargo/audit.toml
```diff
@@ -0,0 +1,5 @@
+[advisories]
+ignore = [
+    "RUSTSEC-2020-0071", # `chrono` is a transient dependency used by `near-sandbox-utils` and other dependencies
```
frol, line 3: So it seems that we cannot upgrade it, but we also don't hit it, right?

frol (approved):
@ruseinov Could you take a look at this PR? Does it address your questions?

ruseinov:
> @ruseinov Could you take a look at this PR? Does it address your questions?

It would be nice to have a little comment on top of every RUSTSEC ignored stating why we are ignoring it and which dependency is the culprit.

.cargo/audit.toml
ruseinov: I mean, we could override the dependency if we wanted to, but that isn't the best solution. Or otherwise upgrade all the deps we control to a newer version.

ruseinov:
Looks good to me.

PR #1070, opened 2023-08-17
chore: release v5.0.0-alpha.1

frol:
## 🤖 New release
* `near-sdk`: 4.1.1 -> 5.0.0 (⚠️ API breaking changes)
* `near-sdk-macros`: 4.1.1 -> 5.0.0
* `near-sys`: 0.2.0 -> 0.2.1 (✓ API compatible changes)
* `near-contract-standards`: 4.1.1 -> 5.0.0 (✓ API compatible changes)

### ⚠️ `near-sdk` breaking changes

```
--- failure enum_variant_missing: pub enum variant removed or renamed ---

Description:
A publicly-visible enum has at least one variant that is no longer available under its prior name. It may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.24.2/src/lints/enum_variant_missing.ron

Failed in:
  variant PromiseError::NotReady, previously in file /tmp/.tmpGbdLOj/near-sdk/src/types/vm_types.rs:41
  variant PromiseResult::NotReady, previously in file /tmp/.tmpGbdLOj/near-sdk/src/types/vm_types.rs:18

--- failure pub_module_level_const_missing: pub module-level const is missing ---

Description:
A public const is missing, renamed, or changed from const to static.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.24.2/src/lints/pub_module_level_const_missing.ron

Failed in:
  ONE_YOCTO in file /tmp/.tmpGbdLOj/near-sdk/src/types/mod.rs:50
  STORAGE_PRICE_PER_BYTE in file /tmp/.tmpGbdLOj/near-sdk/src/environment/env.rs:804
  ONE_NEAR in file /tmp/.tmpGbdLOj/near-sdk/src/types/mod.rs:53

--- failure struct_missing: pub struct removed or renamed ---

Description:
A publicly-visible struct cannot be imported by its prior path. A `pub use` may have been removed, or the struct itself may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.24.2/src/lints/struct_missing.ron

Failed in:
[...]

PR #1092, opened 2023-10-04
ci: add examples `Cargo.lock`-s auto-update job

dj8yfo:
this attempts to address ci issue, mentioned in [comment](https://github.com/near/near-sdk-rs/pull/1070#issuecomment-1683070430). 

according to [a piece of github actions documentation](https://docs.github.com/en/github-ae@latest/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow), 
this workflow will not cause an infinite loop, because the push is made from [GITHUB_TOKEN](https://github.com/dj8yfo/near-sdk-rs/blob/add_auto_commit_with_locks_update/.github/workflows/release-plz-examples.yml#L31). 

At the same time, it will be triggered in the first place by [release-plz worklow](https://github.com/near/near-sdk-rs/blob/master/.github/workflows/release-plz.yml#L28), because it uses `CUSTOM_GITHUB_TOKEN`.

### DIFF
--- a/.github/workflows/release-plz-examples.yml
+++ b/.github/workflows/release-plz-examples.yml
@@ -1,0 +1,39 @@
+name: Release-plz auto-push Cargo.lock updates
+
+on:
+  push:
+    # Sequence of patterns matched against refs/heads
+    branches:    
+      - 'release-plz/*'
+jobs:
+  change-and-push:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: "1.69.0 rust toolchain"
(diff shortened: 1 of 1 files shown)


frol (changes requested):
> this workflow will not cause an infinite loop

But the release-plz CI will fail, and will need to be restarted. I think we should just drop Cargo.lock from examples (and git-ignore them) and call it a day.

dj8yfo:
> But the release-plz CI will fail, and will need to be restarted

no, the idea was, that the branch would be updated with next commit, for which the ci checks would pass

> think we should just drop Cargo.lock from examples (and git-ignore them) and call it a day

but this solution does look more straightforward

PR #1095, opened 2023-10-09
chore: update contract-builder container 1.69.0 -> 1.73.0

dj8yfo:


contract-builder/Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM rust:1.69.0
+FROM rust:1.72.1
```
frol, line 1: nearcore recently bumped their MSRV to 1.73, which will hit us with the next near-primitives release, so I guess we can already bump it to 1.73.

.github/workflows/test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest]
-        toolchain: [stable, 1.69.0]
+        toolchain: [stable]
```
frol, line 16: I think we should still check MSRV for SDK on CI to avoid surprises.

frol (approved):

Issue #1051, opened 2023-07-09
Example doc for random seed


agostbiro:
Add an example doc to

https://github.com/near/near-sdk-rs/blob/bb8a7f8c796e3dd2bd22b4066c97251288f7d84b/near-sdk/src/environment/env.rs#L261

The doc should showcase how to construct a struct that implements [SeedableRng](https://docs.rs/rand/latest/rand/trait.SeedableRng.html) from the `random_seed` and uses [Rng::gen_range](https://docs.rs/rand/latest/rand/trait.Rng.html#method.gen_range) to get a random value in a range with the seed.

The doc should also contain a warning that the random seed should not be used as input to cryptographic primitives.

frol:
> The doc should also contain a warning that the random seed should not be used as input to cryptographic primitives.

[Random seed in NEAR is quite well-thought](https://near.org/blog/randomness-in-blockchain-protocols), so it can be used for security-critical needs, but the public nature of the blockchain might be the blocker anyway. So I would not add that warning.

agostbiro:
Thanks for raising this @frol , the issue warrants more discussion, so I wrote it up in detail.

As explained in the [paper,](https://pages.near.org/downloads/NearRandomnessBeacon.pdf) the goal of the NEAR randomness beacon is twofold:

> 1. Unpredictable. No participant should be able to predict the resulting
> number or reason about any properties of it before the number is created.
> 2. Unbiasable. No participant should be able to influence the resulting
> number in any way.

These goals work well (with some [caveats](https://docs.near.org/develop/contracts/security/random)) for use cases such as on-chain betting where the quality of the randomness only matters during the execution of the transaction by the block producer.

Many cryptographic primitives such as encryption algorithms rely on random values as inputs. These random values are generated with cryptographically secure pseudorandom number generators (CSPRNG). The requirement on CSPRNGs is that their output should be undistinguishable from a truly random distribution by an adversary. 

CSPRNGs are called pseudorandom, because their outputs depend on a seed value. If the attacker learns about this seed value, they can distinguish the CSPRNG output from a random distribution, so the security of the CSPRNG and all the cryptographic primitives that depend on it is broken. 

[...]

PR #1096, opened 2023-10-16
random_seed_array documentation

iho:
Closes https://github.com/near/near-sdk-rs/issues/1051


frol (approved):
Looks good to me, but you need to make tests pass (probably, the best thing would be to include rand and near-rng as dev-dependencies)

frol (approved):

PR #1099, opened 2023-10-20
Have #[handle_result] support Result types reardless of how they're referred to

iho:
Port of https://github.com/near/near-sdk-rs/pull/1057 to syn2


uint (approved):
LGTM!

Issue #1103, opened 2023-10-27
Refactor NEAR tokens usages to use a strictly typed near-token crate


frol:
We recently introduced [near-token crate](https://crates.io/crates/near-token) to deal with NEAR Token values in an ergonomic way. I suggest we migrate to it in NEAR SDK.

PR #1104, opened 2023-10-28
refactor: Migrate to NearToken

iho:
Closes https://github.com/near/near-sdk-rs/issues/1103


examples/factory-contract/Cargo.toml
```diff
@@ -12,6 +12,7 @@ serde_json = "1.0"
 test-case = "2.0"
 tokio = { version = "1.14", features = ["full"] }
 near-workspaces = { version = "0.8.0" }
+near-token = { version = "0.1.0", features = ["serde", "borsh", "schemars"] }
```
frol, line 15: Use `near_workspaces::types::NearToken` instead of importing near-token separately

examples/factory-contract/tests/workspaces.rs
```diff
@@ -1,35 +1,30 @@
-
-use near_sdk::json_types::U128;
-use near_units::parse_near;
+use near_token::NearToken;
```
frol, line 1: ^

frol (changes requested):
@iho I would suggest to actually use NearToken and NearGas as function arguments types instead of u64/u128, and re-export them from `near_sdk` (see what is the best place for the type to be imported from the user perspective)

frol:
@iho Let’s also replace the Balance type alias with NearToken https://github.com/near/near-sdk-rs/blob/ee5cf867741d6d0d4db15857609b9e9268cc9b32/near-sdk/src/types/primitives.rs#L35, and remove ONE_YOCTO and ONE_NEAR consts: https://github.com/near/near-sdk-rs/blob/ee5cf867741d6d0d4db15857609b9e9268cc9b32/near-sdk/src/types/mod.rs#L48

near-contract-standards/src/fungible_token/core.rs
```diff
@@ -91,14 +91,14 @@ pub trait FungibleTokenCore {
     fn ft_transfer_call(
         &mut self,
         receiver_id: AccountId,
-        amount: U128,
+        amount: NearToken,
```
frol, line 94: Hold on, these are not NearTokens, so keep these U128.

Note: Most probably all the changes in FT implementation needs to be reverted.

near-contract-standards/src/fungible_token/core_impl.rs
```diff
@@ -28,10 +28,10 @@ const ERR_TOTAL_SUPPLY_OVERFLOW: &str = "Total supply overflow";
 #[borsh(crate = "near_sdk::borsh")]
 pub struct FungibleToken {
     /// AccountID -> Account balance.
-    pub accounts: LookupMap<AccountId, Balance>,
+    pub accounts: LookupMap<AccountId, NearToken>,
```
frol, line 31: Move Balance type alias to be part of FungibleToken module. It is not NearToken, and it was an abuse of Balance type in the first place.

near-contract-standards/src/storage_management/mod.rs
```diff
@@ -1,24 +1,24 @@
...
 
 #[derive(BorshDeserialize, BorshSerialize, Serialize, Deserialize)]
 #[serde(crate = "near_sdk::serde")]
 #[borsh(crate = "near_sdk::borsh")]
 #[cfg_attr(feature = "abi", derive(schemars::JsonSchema))]
 pub struct StorageBalance {
-    pub total: U128,
-    pub available: U128,
+    pub total: NearToken,
+    pub available: NearToken,
```
frol, line 12: Also update the doc strings below to use NearToken instead of U128.

near-sdk/src/environment/env.rs
```diff
@@ -12,10 +12,11 @@ use std::{convert::TryFrom, mem::MaybeUninit};
 use crate::mock::MockedBlockchain;
 use crate::promise::Allowance;
 use crate::types::{
-    AccountId, Balance, BlockHeight, Gas, PromiseIndex, PromiseResult, PublicKey, StorageUsage,
+    AccountId, BlockHeight, Gas, PromiseIndex, PromiseResult, PublicKey, StorageUsage,
 };
 use crate::{GasWeight, PromiseError};
 use near_sys as sys;
+use near_token::NearToken;
```
frol, line 19: Import NearToken from `crate::types`

frol (changes requested):

frol:
Please, apply the suggestions throughout the code, I only commented on a few places and didn’t go through every similar case in the code

frol (changes requested):

examples/fungible-token/ft/src/lib.rs
```diff
@@ -38,7 +38,7 @@ pub struct Contract {
     token: FungibleToken,
     metadata: LazyOption<FungibleTokenMetadata>,
 }
-
+pub type Balance = U128;
```
frol, line 41: You can import it from the core instead of redefining here again.

frol (changes requested):

frol (approved):

Issue #1102, opened 2023-10-24
Support Source Metadata out of the box


frol:
We need to implement https://github.com/near/NEPs/blob/master/neps/nep-0330.md into near-sdk-rs, so every contract built with near-sdk-rs would automatically get `contract_source_metadata` method exposed with the information populated from Cargo.toml and consider how to register near-contract-standards as available standards.

shariffdev:
Hey @frol assign me to this and I will take a look at this, thanks!

shariffdev:
I want to go about it by having `contract_source_metadata` created implicitly, as suggested using info from Cargo.toml, if not initialised by the contract author.

It can also be created explicitly if the metadata details are provided by a macro. The explicit implementation will be along the lines of:

```rust
use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
use near_sdk::near_bindgen;

#[derive(Default, BorshSerialize, BorshDeserialize)]
#[near_bindgen(contract_metadata(
    version = "39f2d2646f2f60e18ab53337501370dc02a5661c",
    link = "https://github.com/near-examples/nft-tutorial",
    standards = [
        {standard = "nep330", version = "1.1.0"},
        {standard = "nep171", version = "1.0.0"},
        {standard = "nep177", version = "2.0.0"},
    ]
))]
#[near_bindgen]
stuct Contract {}
```

frol:
@shariffdev The proposed solution sounds good to me. I would only consider if there is really a need for duplicative `#[near_bindgen]` - I don't see a good reason for it.

shariffdev:
Sure, I can work around not making it duplicative.

shariffdev:
The standards input changed to this to play well with the parser
```rust
#[near_bindgen(contract_metadata(
    version = "39f2d2646f2f60e18ab53337501370dc02a5661c",
    link = "https://github.com/near-examples/nft-tutorial",
    standard(standard = "nep330", version = "1.1.0"),
    standard(standard = "nep171", version = "1.0.0"),
    standard(standard = "nep177", version = "2.0.0"),
))]
stuct Contract {}
```

PR #1106, opened 2023-10-30
feat: adding `nep-0330` contract source metadata info

shariffdev:
Addresses: https://github.com/near/near-sdk-rs/issues/1102

### DIFF
--- a/near-contract-standards/Cargo.toml
+++ b/near-contract-standards/Cargo.toml
@@ -13,13 +13,10 @@ NEAR smart contracts standard library.
 """
 
 [dependencies]
-near-sdk = { path = "../near-sdk", version = "~4.1.1", default-features = false, features = [
-    "legacy",
-] }
+near-sdk = { path = "../near-sdk", version = "~4.1.1", default-features = false, features = ["legacy"] }
 serde = "1"
 serde_json = "1"
 schemars = "0.8"
-anyhow = "1"
 
 [features]
 default = ["abi"]
--- a/near-contract-standards/Cargo.toml
+++ b/near-contract-standards/Cargo.toml
@@ -13,10 +13,13 @@ NEAR smart contracts standard library.
 """
 
 [dependencies]
-near-sdk = { path = "../near-sdk", version = "~4.1.1", default-features = false, features = ["legacy"] }
+near-sdk = { path = "../near-sdk", version = "~4.1.1", default-features = false, features = [
+    "legacy",
+] }
 serde = "1"
 serde_json = "1"
 schemars = "0.8"
+anyhow = "1"
 
 [features]
 default = ["abi"]
--- a/near-contract-standards/src/contract_metadata.rs
+++ b/near-contract-standards/src/contract_metadata.rs
@@ -1,21 +1,13 @@
 use serde::{Deserialize, Serialize};
 
-#[derive(Serialize, Deserialize)]
-#[serde(tag = "standard")]
+#[derive(Serialize, Deserialize, Debug)]
 pub struct ContractSourceMetadata {
     pub version: Option<String>,
     pub link: Option<String>,
     pub standards: Vec<Standard>,
 }
 
-impl Default for ContractSourceMetadata {
-    // TODO: Use Cargo.toml from CARGO_MANIFEST_DIR to populate the version field.
-    fn default() -> Self {
-        Self { version: None, link: None, standards: vec![] }
-    }
-}
-
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Debug)]
 pub struct Standard {
     pub standard: String,
     pub version: String,
--- a/near-contract-standards/src/contract_metadata.rs
+++ b/near-contract-standards/src/contract_metadata.rs
@@ -1,33 +1,13 @@
 /// The contract source metadata is a standard interface that allows auditing and viewing source code for a deployed smart contract.
-#[derive(serde::Deserialize, Debug)]
+#[derive(Debug, serde:: Deserialize, PartialEq)]
 pub struct ContractSourceMetadata {
     pub version: Option<String>,
     pub link: Option<String>,
     pub standards: Vec<Standard>,
 }
 
-#[derive(serde:: Deserialize, Debug)]
+#[derive(Debug, serde::Deserialize, PartialEq)]
 pub struct Standard {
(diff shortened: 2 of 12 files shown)


near-sdk-macros/src/lib.rs
```diff
@@ -108,8 +138,37 @@ pub fn near_bindgen(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+                        )
+                        .to_compile_error(),
+                    );
+                }
+            }
+        }
+
+        if input.trait_.is_none() {
+            match syn::parse_str::<syn::ImplItem>(
+                "pub fn contract_source_metadata() -> String { CONTRACT_SOURCE_METADATA.to_string() }",
```
frol, line 161: It should return a JSON object, not a JSON object serialized to JSON string, so the return value has to be a unit type and then you can use `env::value_return()`:

```rust
                    pub fn contract_source_metadata() {
                        env::value_return(CONTRACT_SOURCE_METADATA);
```

near-sdk-macros/src/lib.rs
```diff
@@ -108,8 +138,37 @@ pub fn near_bindgen(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+                            "use of reserved contract method",
+                        )
+                        .to_compile_error(),
+                    );
+                }
+            }
+        }
+
+        if input.trait_.is_none() {
+            match syn::parse_str::<syn::ImplItem>(
```
frol, line 160: Why syn parse instead of `quote`? I find quote syntax more readable.

near-contract-standards/src/contract_metadata.rs
```diff
@@ -0,0 +1,33 @@
...
+    ///     .await?
+    ///     .json::<String>()?;
+    ///
+    /// let metadata = ContractSourceMetadata::from_payload(&payload)?;
+    /// ```
+    pub fn from_payload(payload: &str) -> anyhow::Result<Self> {
+        serde_json::from_str(payload)
+            .map_err(|e| anyhow::anyhow!("Failed to deserialize contract source metadata: {}", e))
+    }
+}
```
frol, line 33: This won't be needed would the method return a JSON object instead of JSON object encoded inside JSON string (see my other comment)

near-contract-standards/Cargo.toml
```diff
@@ -13,10 +13,13 @@ NEAR smart contracts standard library.
...
 
 [dependencies]
-near-sdk = { path = "../near-sdk", version = "~4.1.1", default-features = false, features = ["legacy"] }
+near-sdk = { path = "../near-sdk", version = "~4.1.1", default-features = false, features = [
+    "legacy",
+] }
 serde = "1"
 serde_json = "1"
 schemars = "0.8"
+anyhow = "1"
```
frol, line 22: This also won't be needed, see other comments

near-sdk-macros/src/lib.rs
```diff
@@ -81,13 +81,41 @@ use syn::{parse_quote, ItemEnum, ItemImpl, ItemStruct, ItemTrait, WhereClause};
 ///
 /// }
 /// ```
+///
+/// Contract Source Metadata Standard:
+///
+/// By using `contract_metadata` as an argument `near_bindgen` will populate the contract metadata
+/// according to NEP-330 standard. This still applies even when `#[near_bindgen]` is used without
```
frol, line 88: Make NEP-330 a link to the document, so it makes readers aware on where to find it.

frol (changes requested):

near-sdk-macros/src/core_impl/contract_metadata/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        if self.link.is_none() {
+            self.link = std::env::var("CARGO_PKG_REPOSITORY").ok();
+        }
+
+        if self.standards.is_empty() {
+            self.standards
+                .push(Standard { standard: "nep330".to_string(), version: "1.1.0".to_string() });
+        }
```
frol, line 37: Well, even if it is not empty, let's make sure that nep330 is listed

frol:
@shariffdev Thanks for working on it and providing great user documentation! It looks great to me, and I am looking forward to see more adoption to this new feature.

near-sdk-macros/src/lib.rs
```diff
@@ -108,8 +140,41 @@ pub fn near_bindgen(attr: TokenStream, item: TokenStream) -> TokenStream {
             #input
             #ext_gen
             #abi_embedded
+            #metadata
         })
     } else if let Ok(mut input) = syn::parse::<ItemImpl>(item) {
+        for method in input.items.iter() {
```
frol, line 146: nit:

```suggestion
        for method in &input.items {
```

near-sdk-macros/src/core_impl/contract_metadata/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+
+        // adding nep330 if it is not present
+        if self.standards.is_empty()
+            || self
+                .standards
+                .iter()
+                .find(|s| s.standard.to_ascii_lowercase().eq("nep330"))
+                .clone()
+                .is_none()
```
frol, line 47:
```suggestion
            || self.standards.iter().all(|s| s.standard.to_ascii_lowercase() != "nep330")
```

near-contract-standards/src/contract_metadata.rs
```diff
@@ -0,0 +1,13 @@
+/// The contract source metadata is a standard interface that allows auditing and viewing source code for a deployed smart contract.
+#[derive(Debug, serde:: Deserialize, PartialEq)]
```
frol, line 2: I don't see a reason to restrict users from Clone and serde::Serialize these structs

frol (changes requested):
Some final touches and it is good to go

near-sdk-macros/src/lib.rs
frol: @shariffdev Please, don't forget about this one

shariffdev: Done

frol (approved):

frol:
@shariffdev Cannot wait to build my first near-sdk-rs 5.0 contract and have this available out of the box!

PR #1109, opened 2023-11-10
fix: exceed prepaid gas in nft_approve

hanakannzashi:
When user use implicit account to sign tx, more gas will be used, it will cause that 10Tgas is not enough for `nft_approve`.

This is an example:
https://legacy.explorer.near.org/transactions/CwuFtBWYnr5xkdpvbbTtznvFUf1FneAwMVBpPi8Q8wyd
User can not list [NameSky NFT](https://namesky.app) in marketplace with implicit account as signer

I think we can remove static gas for x-call `nft_on_approve` because default gas weight 1 will distribute remaining gas to it.


frol (approved):
Overall, looks fine to me. Have you tested it? Does it work as expected on your end now?

PR #1110, opened 2023-11-10
refactor: gas distribution in transfer call

hanakannzashi:
Legacy gas distribution is not good because gas use of `ft_transfer_call` or `nft_tranasfer_call` itself is a  static value. This may cause panic when developers wrap this method and add their own logic which increase the gas use.

So I think we should remove static gas for `ft_transfer_call` or `nft_transfer_call` itself. Instead, we use gas weight to distribute the gas, that is more flexible and can solve previous problem.

Currently, I set `xx_on_transfer : xx_resolve_transfer` = `9 : 1` and attach a static gas `10 T` for resolver because resolver is more important and decrease its panic risk.

I am not sure `9 : 1` is good or not, we can disscuss.


hanakannzashi:
This change will cause log print when `nft_transfer_call` failed due to exceed gas.

PR #1114, opened 2023-11-21
refactor: Re-exported packages cleanup

iho:


near-contract-standards/src/fungible_token/metadata.rs
```diff
@@ -1,3 +1,4 @@
+use near_sdk::__private::schemars;
```
frol, line 1: 1. Instead of importing it for implicit use by the derived macro, use `#[schemars(crate=…]` https://github.com/GREsau/schemars/blob/e04e3a3a8191bac1f218539733110fcc26bdbf7c/CHANGELOG.md?plain=1#L121
2. Remove the re-export in __private since we promoted it to be part of the public re-export

near-sdk-macros/src/lib.rs
```diff
@@ -436,7 +436,7 @@ pub fn derive_near_schema(input: TokenStream) -> TokenStream {
             #[allow(non_camel_case_types)]
             type #input_ident_proxy = #input_ident;
             {
-                use ::near_sdk::__private::schemars;
+                use ::near_sdk::schemars;
```
frol, line 439: Can we completely remove this `use` and use fully qualified path to schemars types?

frol (changes requested):

frol (dismissed):

iho:
@frol LGTM!

Issue #1117, opened 2023-12-08
Compilation error due to duplicate definitions for 'contract_source_metadata*


frol:
![IMG_8574](https://github.com/near/near-sdk-rs/assets/304265/2228cfd7-1579-40be-a8d0-804381ecee52)

This happens when #[near_bindgen] is used several times (which is totally valid use case)

This is caused by https://github.com/near/near-sdk-rs/pull/1106 and only affects 5.0.0-alpha.1 release and onwards.

This blocks 5.0 stable release.

PR #1118, opened 2023-12-12
[Fix] Contract source metadata compilation issue.

ruseinov:
Fixes https://github.com/near/near-sdk-rs/issues/1117

### DIFF
--- a/near-sdk-macros/src/core_impl/code_generator/metadata.rs
+++ b/near-sdk-macros/src/core_impl/code_generator/metadata.rs
@@ -1,0 +1,17 @@
+use proc_macro2::Ident;
+use quote::quote;
+use syn::Generics;
+
+/// Generates a view method to retrieve the source metadata.
+pub(crate) fn generate_metadata_impl(
+    ident: &Ident,
+    generics: &Generics,
+) -> proc_macro2::TokenStream {
+    quote! {
+        impl #generics #ident #generics {
+            pub fn contract_source_metadata() {
+                near_sdk::env::value_return(CONTRACT_SOURCE_METADATA.as_bytes())
+            }
+        }
+    }
+}
--- a/near-sdk-macros/src/core_impl/code_generator/metadata.rs
+++ b/near-sdk-macros/src/core_impl/code_generator/metadata.rs
@@ -3,7 +3,7 @@ use quote::quote;
 use syn::Generics;
 
 /// Generates a view method to retrieve the source metadata.
-pub(crate) fn generate_metadata_impl(
+pub(crate) fn generate_contract_metadata_method(
     ident: &Ident,
     generics: &Generics,
 ) -> proc_macro2::TokenStream {
--- a/near-sdk-macros/src/lib.rs
+++ b/near-sdk-macros/src/lib.rs
@@ -283,9 +283,9 @@ pub fn init(_attr: TokenStream, item: TokenStream) -> TokenStream {
 }
 
 use crate::core_impl::metadata::generate_metadata_impl;
+
 #[cfg(feature = "abi")]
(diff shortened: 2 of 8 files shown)


frol (approved):

near-sdk-macros/src/lib.rs
```diff
@@ -116,6 +117,7 @@ pub fn near_bindgen(attr: TokenStream, item: TokenStream) -> TokenStream {
 
     if let Ok(input) = syn::parse::<ItemStruct>(item.clone()) {
         let metadata = core_impl::contract_source_metadata_const(attr);
+        let metadata_impl_gen = generate_metadata_impl(&input.ident, &input.generics);
```
frol, line 120: This way the generated function will only be implemented on the contract struct, but not exposed as contract methods and ABI is not generated for it. This has to be fixed and also test added as we obviously could have missed it unless I happened to spot it.

near-sdk-macros/src/core_impl/code_generator/metadata.rs
frol, line 1: * [ ] Merge contents of `near-sdk-macros/src/core_impl/contract_metadata/mod.rs` into this file
* [ ] Rename this file to `contract_metadata.rs`
* [ ] Rename `generate_metadata_impl` function to `generate_contract_metadata_method`

frol (changes requested):

near-sdk-macros/src/core_impl/code_generator/metadata.rs
ruseinov: This directory structure follows the same pattern as all the other traits that are being implemented via macro/codegen. I would argue that we should either convert all of them or none. I would personally find having those traits under `code_generator` a bit confusing. 
The rename is not a problem though.

near-sdk-macros/src/lib.rs
```diff
@@ -145,26 +164,58 @@ pub fn near_bindgen(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+//
+// # Arguments
+// * block - impl block to process.
+// * internal - whether or not this is an internal call, if the call is internal - reserved contract
+// method checks will be skipped.
+//
+// Returns None in case the given block is not an impl block. The underlying Result has a
+// TokenStream error type, because those need to be propagated to the compiler.
+fn process_impl_block(
+    block: TokenStream,
```
frol, line 194: Can we pass `ItemImpl` type here to avoid one level of nesting and also a need to return `Option<>`?

near-sdk-macros/src/lib.rs
```diff
@@ -145,26 +164,58 @@ pub fn near_bindgen(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+    internal: bool,
+) -> Option<Result<proc_macro2::TokenStream, proc_macro2::TokenStream>> {
+    if let Ok(mut input) = syn::parse::<ItemImpl>(block) {
         for method in &input.items {
             if let ImplItem::Fn(m) = method {
                 let ident = &m.sig.ident;
-                if ident.eq("__contract_abi") || ident.eq("contract_source_metadata") {
-                    return TokenStream::from(
+                if !internal && (ident.eq("__contract_abi") || ident.eq("contract_source_metadata"))
```
frol, line 201: I would extract this check into the `near_bindgen`, so we don't need to introduce an ugly boolean flag. When `internal` is true you don't need this loop at all.

frol (changes requested):

near-sdk-macros/src/lib.rs
ruseinov: Good call.

frol (approved):

Issue #1129, opened 2024-01-11
Compilation error on fresh Cargo.toml that depends on near-sdk-rs 5.0.0-alpha.1


fadeevab:
I tried to compile my project with `near-sdk-rs` `5.0.0-alpha.1` because of this issue https://github.com/near/near-sdk-rs/issues/1119, however, build fails with the errors:
```
error: proc-macro derive panicked
 --> src/main.rs
  |
6 | #[derive(BorshSerialize, BorshDeserialize)]
  |          ^^^^^^^^^^^^^^
  |
  = help: message: called `Result::unwrap()` on an `Err` value: Could not find `borsh` in `dependencies` or `dev-dependencies` in `/home/xxxx/yyyyy/Cargo.toml`!
```

frol:
Given that we re-export borsh from near-sdk, you need to hint it to borsh:

https://github.com/near/near-sdk-rs/blob/13430592a7be246dd5a29439791f4081e0107ff3/examples/adder/src/lib.rs#L6

PR #1132, opened 2024-01-16
doc: add `#[borsh(crate =...)]` mentions in readme (#1129)

dj8yfo:
this may help in cases like #1129


frol:
@dj8yfo I wonder if it would be nicer if `#[near_bindgen]` would add Borsh derives automatically in the first place. Since bindgen expects the state struct to be serialized to borsh, I feel it is not helpful to leave it up to the users to write derives manually. What do you think?

Issue #1124, opened 2023-12-19
Investigate automating ABI tests for internal implementations


ruseinov:
It would be nice to see failures when we introduce some more codegen for `near_bindgen` that does not take care of the ABI, similar to the PR mentioned below.

_Originally posted by @ruseinov in https://github.com/near/near-sdk-rs/pull/1118#discussion_r1431325142_

shariffdev:
Added https://github.com/shariffdev/near-sdk-rs/commit/20987c89e4047f3360c463a1d45f2f2926bc6dd3, haven't placed it as a PR as it's incomplete(run on CI vs not on CI).
We use the `cargo-near` tool to generate the abi spec file. Would we want to include that on the CI just for this test to run automatically?

frol:
@shariffdev I fear that use of `cargo-near` may lead to circular dependencies. `cargo-near` does not do much, it just uses abi feature flag on near-sdk-rs, so you should be able to replicate that without cargo-near

PR #1136, opened 2024-01-24
chore: adding abi tests

shariffdev:
Addresses https://github.com/near/near-sdk-rs/issues/1124


near-sdk-macros/src/lib.rs
```diff
@@ -643,3 +643,63 @@ pub fn derive_event_attributes(item: TokenStream) -> TokenStream {
         )
     }
 }
+
+#[cfg(test)]
+mod tests {
```
frol, line 648: @shariffdev 

1. This place does not make sense for these tests. Let's move them to `near-sdk/tests/`
2. I would like to encourage you to consider looking into cargo-near implementation of abi extraction to avoid sh script running, etc: https://github.com/near/cargo-near/blob/48c7b48fa0f7f3b4afda447ba0b04cce3609b0a9/cargo-near/src/commands/abi_command/abi.rs#L31

frol (changes requested):

near-sdk-macros/src/lib.rs
shariffdev: Updated

frol:
@shariffdev Would you be able to fix the issues highlighted by CI?

frol (dismissed):
Thank you for pushing it!

frol (approved):

Issue #1134, opened 2024-01-20
`Keys` iterator of the `store::UnorderedMap` has a huge performance drop after removing elements


fadeevab:
### Problem
1. Create `store::UnorderedMap`
2. Insert 1000 elements.
3. Remove 900 elements.
4. Now, iteration through 100 keys takes `O(1000)` time instead of `O(100)` expected.

Trying to take just a constant number of keys gives an inconsistent performance with increasing gas consumption 3TGas -> 10 TGas -> 20 TGas ->....., while removing bunches of elements from the map.

### Cause
`FreeList` is now used for the keys in the `store::UnorderedMap`, there is an iterator that goes through the "holes" in the vector of keys which never shrinks when elements are removed.

### Example

Here, iterating through 100 keys after removals gives `0.5 TGas` consumption instead of `0.04 TGas` if 100 elements were just inserted.

```rust
    pub fn test(&mut self) {
        let mut map = UnorderedMap::new(b'a');
        for i in 0..1000 {
            map.insert(i, i);
        }

        // Keep only 100 elements.
        for i in 0..900 {
            map.remove(&i);
        }

        let gas = env::used_gas();
        let _ = map.keys().collect::<Vec<_>>();
        let gas = (env::used_gas() - gas).0 as f32 / Gas::ONE_TERA.0 as f32;

        env::log_str(&format!("used gas: {:?}", gas));  // 0.5 TGas
    }
```

`0.5 TGas` - 100 keys, after removals.
`0.04 TGas` - 100 keys, no removals, just inserts.
x10 performance drop.
It's worse when there are thousands of elements.

fadeevab:
Another problem is that `skip` is not cheap (`FreeList::Iterator::skip`) because it's not lazy, it reads every value while moving to the nth position to figure out whether a slot is occupied or not. And it's another issue. E.g. even in a case with no removals, skipping over couple of thousand element may not be possible because gas will finish earlier.

fadeevab:
I found an enhancement ticket outlining the same issue: https://github.com/near/near-sdk-rs/issues/990.
`defrag` method was introduced to fight the fragmentation of the `FreeList` via removing the "holes" in it: https://github.com/near/near-sdk-rs/pull/1023/files#diff-793a8b595e5ccce11a932ce7a41e0ea3a313790edc40bc4ff0ad5943a112a673R230

[...]

PR #1139, opened 2024-02-08
fix: deprecated `store::UnorderedMap` and `store::UnorderedSet` due to not meeting the original requirements (iteration over a collection of more than 2k elements runs out of gas)

ruseinov:
- deprecated `store::UnorderedMap` due to its fragmentation property
- deprecated `store::UnorderedSet` due to its fragmentation property

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@
 ## [Unreleased]
 
 ### Fixed
-- deprecated `store::UnorderedMap` because of its fragmentation property([#1139](https://github.com/near/near-sdk-rs/pull/1139))
-- deprecated `store::UnorderedSet` because of its fragmentation property([#1139](https://github.com/near/near-sdk-rs/pull/1139))
+- deprecated `store::UnorderedMap` due to its fragmentation property([#1139](https://github.com/near/near-sdk-rs/pull/1139))
+- deprecated `store::UnorderedSet` due to its fragmentation property([#1139](https://github.com/near/near-sdk-rs/pull/1139))
 
 ## [5.0.0-alpha.2](https://github.com/near/near-sdk-rs/compare/near-sdk-v5.0.0-alpha.1...near-sdk-v5.0.0-alpha.2) - 2024-01-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Fixed
+- renamed `store::UnorderedMap` to `FrangibleUnorderedMap` to reflect its fragmentation property([#1139](https://github.com/near/near-sdk-rs/pull/1139))
+
 ## [5.0.0-alpha.2](https://github.com/near/near-sdk-rs/compare/near-sdk-v5.0.0-alpha.1...near-sdk-v5.0.0-alpha.2) - 2024-01-16
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 - renamed `store::UnorderedMap` to `FrangibleUnorderedMap` to reflect its fragmentation property([#1139](https://github.com/near/near-sdk-rs/pull/1139))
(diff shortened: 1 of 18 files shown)


near-sdk/src/store/frangible_map/mod.rs
```diff
@@ -18,23 +18,28 @@ use super::{FreeList, LookupMap, ERR_INCONSISTENT_STATE, ERR_NOT_EXIST};
...
-/// The default hash function for [`UnorderedMap`] is [`Sha256`] which uses a syscall
+/// The default hash function for [`FrangibleUnorderedMap`] is [`Sha256`] which uses a syscall
 /// (or host function) built into the NEAR runtime to hash the key. To use a custom function,
 /// use [`with_hasher`]. Alternative builtin hash functions can be found at
 /// [`near_sdk::store::key`](crate::store::key).
 ///
+/// # Performance considerations
+/// Note that this collection is optimized for fast removes at the expense of key management.
+/// If the amount of removes is significantly higher than the amount of inserts the iteration
+/// becomes more costly. See [`remove`](FrangibleUnorderedMap::remove) for details.
```
frol, line 34: As a developer I am left with an unanswered question in my head: "and now what? are there better options?"

near-sdk/src/store/unordered_set/mod.rs
frol, line 1: It looks to me that UnorderedSet has exactly the same problems, isn't it? Should it also be renamed?

frol (changes requested):

near-sdk/src/store/unordered_set/mod.rs
ruseinov: Yeah, at first I thought that it does not, but now I understand that it does. It also does not have a `defrag` method, but that's easy to fix.

ruseinov:
@frol this will need a test re-run due to a flaky test.

near-sdk/src/store/frangible_map/mod.rs
ruseinov: Fixed.

frol:
Renaming won't help us to introduce `UnorderedMap` name back (see my https://github.com/near/near-sdk-rs/issues/1134#issuecomment-1936963703). So I don't think renaming is the right move here.

ruseinov:
> Renaming won't help us to introduce `UnorderedMap` name back (see my [#1134 (comment)](https://github.com/near/near-sdk-rs/issues/1134#issuecomment-1936963703)). So I don't think renaming is the right move here.

Why not? Can't we introduce something like `UnorderedMapV2` and type alias it to `UnorederedMap`?

ruseinov:
> Renaming won't help us to introduce `UnorderedMap` name back (see my [#1134 (comment)](https://github.com/near/near-sdk-rs/issues/1134#issuecomment-1936963703)). So I don't think renaming is the right move here.

Right, let's deprecate them then.

examples/versioned/src/lib.rs
frol, line 1: I think the changes in this file has to be reverted.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+- deprecated `store::UnorderedMap` due to its fragmentation property ([#1139](https://github.com/near/near-sdk-rs/pull/1139))
+- deprecated `store::UnorderedSet` due to its fragmentation property ([#1139](https://github.com/near/near-sdk-rs/pull/1139))
```
frol, line 7: We don't need to manually edit CHANGELOG anymore, just write it in the PR description and release-plz will include it in the CHANGELOG, see the automated PR: https://github.com/near/near-sdk-rs/pull/1138

frol (changes requested):

ruseinov:
@frol all fixed.

examples/versioned/src/lib.rs
ruseinov: We should probably make sure this gets caught by the CI though.

frol (approved):

PR #1143, opened 2024-02-14
clean up features

njelich:
The project still compiles and can be used fine without these dependency features. 

On top of this, I removed abi and legacy from default near-sdk features.


frol:
1. Let's migrate examples that don't use `UnorderedMap` and `UnorderedSet` over to the `store` API (`store::UnorderedMap` and `store::UnorderedSet` got recently deprecated - see #1139 + discussion in #1134).
2. Let's keep the examples that use `collections::Unordered*` and use `legacy` feature for them until `store::IterableMap`/`store::IterableSet` are implemented.

Issue #1084, opened 2023-09-15
Cannot #[near_bindgen] expose a function with a duplicate name


encody:
Steps to reproduce:
1. Implement a function on your contract's default state struct, that is _not_ annotated with `#[near_bindgen]`. Wlog, this function is called `foo`. Whether this function is part of a trait or implemented directly in an `impl Contract {}` block is irrelevant.
2. Implement another function `foo` as part of a different trait or `impl` block, this time annotated with `#[near_bindgen]`.

Expected result: The WASM is generated, exporting & calling the correct function (that which _is_ annotated with `#[near_bindgen]`).

Actual result: Various compilation errors, including `multiple applicable items in scope`, `this method takes 1 argument but 0 arguments were supplied`, etc. depending on the nature of the functions involved.

Sample code:

```rust
use near_sdk::{
    borsh::{self, BorshDeserialize, BorshSerialize},
    log, near_bindgen, PanicOnDefault,
};

#[derive(BorshSerialize, BorshDeserialize, PanicOnDefault)]
#[near_bindgen]
pub struct Contract {}

trait T1 {
    fn foo(&self);
}

trait T2 {
    fn foo(&self);
}

impl T1 for Contract {
    fn foo(&self) {
        log!("foo from T1");
    }
}

#[near_bindgen]
impl T2 for Contract {
    fn foo(&self) {
        log!("foo from T2");
    }
}
```

Error output:

```text
error[E0034]: multiple applicable items in scope
  --> workspaces-tests/src/bin/testy.rs:30:8
   |
30 |     fn foo(&self) {
   |        ^^^ multiple `foo` found
   |
note: candidate #1 is defined in an impl of the trait `T1` for the type `Contract`
  --> workspaces-tests/src/bin/testy.rs:23:5
   |
23 |     fn foo(&self) {
   |     ^^^^^^^^^^^^^
note: candidate #2 is defined in an impl of the trait `T2` for the type `Contract`
  --> workspaces-tests/src/bin/testy.rs:30:5
   |
30 |     fn foo(&self) {
   |     ^^^^^^^^^^^^^
help: disambiguate the method for candidate #1
   |
28 | T1::foo(&#[near_bindgen])
   |
help: disambiguate the method for candidate #2
   |
28 | T2::foo(&#[near_bindgen])
   |

For more information about this error, try `rustc --explain E0034`.
```

Another, slightly more insidious example (`serialize` is the name of a function provided by the `BorshSerialize` derive macro):

[...]

PR #1144, opened 2024-02-14
feature: aliasing for functions with duplicate names

muse254:
The changes introduced addresses https://github.com/near/near-sdk-rs/issues/1084.
Concerned with aliasing colliding names in the generated abi using `#[abi_alias(...)]` as such:
```rust
trait T1{
   fn foo(&self);
}

trait T2 {
   fn foo(&self);
}

struct Contract;


impl T1 for Contract {
   fn foo(&self){
      log!("foo_one");
   }
}

impl T2 for Contract {
   #[abi_alias("foo_two")
   fn foo(&self){
      log!("foo_two");
   }
}
```

### DIFF
--- a/near-sdk-macros/src/core_impl/abi/abi_generator.rs
+++ b/near-sdk-macros/src/core_impl/abi/abi_generator.rs
@@ -339,7 +339,7 @@ mod tests {
             #[handle_result]
             pub fn f3(&mut self, arg0: FancyStruct, arg1: u64) -> Result<IsOk, Error> { }
         };
-        let method_info = ImplItemMethodInfo::new(&mut method, false, impl_type).unwrap().unwrap();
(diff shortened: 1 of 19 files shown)


frol:
@muse254 I believe it does not address the problem raised in #1084. They had a problem when there are two functions with exactly the same names implemented for the same struct (through `impl trait for ...`). The problem is that `#[near_bindgen]` generates an `extern "C" fn`, which does not use a fully-qualified (unambiguous) syntax, and as such `contract.foo()` raises a compilation error.

P.S. The CI error is fixed in the upstream now, so I merged the latest changes from there to your PR branch.

muse254:
I'll look into understanding the problem and the repo some more and rework the solution. Thank you for taking your time to review ❤️

frol:
This is taken over by #1189

Issue #1142, opened 2024-02-10
Investigate how to reduce boilerplate that needs to be added almost to every struct


frol:
Moving the discussion from https://github.com/near/borsh-rs/issues/277 since it is bigger than borsh. serde and schemars also contribute a lot to this ugly boilerplate:

```rust
use near_sdk::serde::{Deserialize, Serialize}
use near_sdk::borsh::{BorshDeserialize, BorshSerialize};
use near_sdk::schemars::JsonSchema;

#[derive(Deserialize, Serialize, JsonSchema, BorshDeserlialize, BorshSerialize)]
#[serde(crate = "near_sdk::serde")]
#[schemars(crate = "near_sdk::schemars")]
#[borsh(crate = "near_sdk::borsh")]
struct ...
```

frol:
I have chatted with @PolyProgrammist and we arrived to the following design:

```rust
#[near]
struct Contract {
    ...
}

#[near(serializers = [borsh, json])]
struct Post {
    author: AccountId,
    text: String,
}

#[near]
impl Contract {
    fn get_posts(&self) -> Vec<Post> { ... }
}
```

* `#[near]` for struct/enums would imply Borsh serialization by default.
* `#[near(serializers = [borsh, json])]` allows to control if borsh and/or json needs to be implemented for the struct/enum - some needs one of them, others may need both at once (if struct is persistent to the state and used in the function arguments / return value)
* `#[near]` for `impl` and `trait` blocks should behave almost like `#[near_bindgen]` (it seems that `#[near_bindgen]` would try to create the root struct if there is no struct marked with `#[near_bindgen]`, I would remove this behavior)
* `#[near]` should also enable JsonSchema and/or BorshSchema when abi feature is at play. Otherwise, schemars must be feature-gated

frol:
Just for the context, here is how I see [status-message contract](https://github.com/near/near-sdk-rs/blob/master/examples/status-message/src/lib.rs) to be updated after these improvements land:

Before:

```rs
use near_sdk::borsh::{BorshDeserialize, BorshSerialize};
use near_sdk::store::LookupMap;
use near_sdk::{env, near_bindgen, AccountId, BorshStorageKey};

#[derive(BorshSerialize, BorshStorageKey)]
#[borsh(crate = "near_sdk::borsh")]
enum StorageKey {
    StatusMessageRecords,
}

#[near_bindgen]
#[derive(BorshDeserialize, BorshSerialize)]
[...]

PR #1147, opened 2024-02-21
Introduce near macro

PolyProgrammist:
Resolves https://github.com/near/near-sdk-rs/issues/1142


near-sdk/src/collections/lazy_option.rs
```diff
@@ -9,12 +9,15 @@ use borsh::{to_vec, BorshDeserialize, BorshSerialize};
...
+use near_sdk_macros::NearSchema;
 
 const ERR_VALUE_SERIALIZATION: &str = "Cannot serialize value with Borsh";
 const ERR_VALUE_DESERIALIZATION: &str = "Cannot deserialize value with Borsh";
 
 /// An persistent lazy option, that stores a value in the storage.
-#[derive(BorshSerialize, BorshDeserialize)]
+#[derive(BorshSerialize, BorshDeserialize, NearSchema)]
+#[inside_nearsdk]
+#[abi(borsh)]
```
frol, line 20: 1. Do we need to manually derive Borsh/Schema? I would expect that `#[near]` would also work here
2. Instead of `#[inside_nearsdk]`, can we add a parameter, so you can just write `#[near(crate="crate")]`? This way we will also allow re-embedding near_sdk if necessary

frol (approved):

frol:
Overall, this looks great to me! I have only one question, but if it would take longer than a day to implement, let's ignore it since everything else is already implemented and looks great!

PolyProgrammist:
>Instead of #[inside_nearsdk], can we add a parameter, so you can just write #[near(crate="crate")]? This way we will also allow re-embedding near_sdk if necessary

Decided to skip it with Vlad for now. Since crate is a keyword and darling can't handle that. It requires a manual attribute parsing

frol (dismissed):

near-contract-standards/src/non_fungible_token/core/core_impl.rs
```diff
@@ -51,6 +50,7 @@ pub struct NonFungibleToken {
     pub next_approval_id_by_id: Option<LookupMap<TokenId, u64>>,
 }
 
+// TODO
```
frol, line 53: Should it be resolved?

frol (approved):

frol (approved):

PR #1151, opened 2024-02-25
fix: remove deprecated wee_alloc

njelich:
The wee_alloc library has been unmaintained for 3 years and has serious issues (e.g. a memory leak affecting some WASM compiles)

https://github.com/rustwasm/wee_alloc/issues/106

### DIFF
--- a/near-sdk/Cargo.toml
+++ b/near-sdk/Cargo.toml
@@ -34,8 +34,6 @@ near-account-id = { version="1.0.0", features = ["serde", "borsh"] }
 near-gas = { version = "0.2.3", features = ["serde", "borsh"] }
 near-token = { version = "0.2.0", features = ["serde", "borsh"] }
 
-[target.'cfg(target_arch = "wasm32")'.dependencies]
-wee_alloc = { version = "0.4.5", default-features = false, optional = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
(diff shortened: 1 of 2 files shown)


njelich:
Rust uses https://github.com/alexcrichton/dlmalloc-rs as a default allocator for WASM.

frol:
@njelich Are there any expected side-effects of this change?

njelich:
I ran the full test suite, and other than previously failing tests as mentioned in the other PR, everything passed. The build size for the cargo-template project increases from 133KB to 139KB, but that is a fixed increase of 5-6KB on any project.

njelich:
Based on discussion in the NEAR Tools Community channel, this change will not be implemented right now. For future maintainers, here are the benchmarks of different allocator options that were discussed:

Results of testing different allocators on cargo-near-new-project-template:

wee_alloc - 132.57KB
dlmalloc - 138.73KB
talc (unknown arena size) - 136.28KB
talc (known arena size) - 134.72KB
rlsf - 132.70KB
lol_alloc (single threaded) - 131.60KB
lol_alloc (generic) - 131.74KB
[custom_bump_alloc](https://github.com/Concordium/concordium-rust-smart-contracts/blob/aac71ec48e72611591f9b4afdc793e643a492a6b/concordium-std/src/bump_alloc.rs) - 130.97KB

PR #1152, opened 2024-02-25
ci: consolidate examples tests

njelich:
Taking a look at the CI, it was only checking some of the examples tests, and holding them under two separate workflows despite them having the same call conditions.

I found that the `callback-results` test has been broken for a while, because it was excluded from CI. I updated the workflows to cover all tests, but there is also a `test_all.sh` script that can be used for the same purpose instead of this matric approach.

Awaiting comment on which is preferred, as well as what to do with the failing test - should it be fixed or removed?


frol:
Closing as it got stale and it seems that we did not see the improvement with this PR

PR #1158, opened 2024-03-15
Add Borsh/Json -Schema to types

PolyProgrammist:
Adds ability to set borsh attributes as `#[near(serializers=[borsh(use_discriminant=false), json])`
Adds near macro which contain borsh/json schema for all the types


frol (approved):
@PolyProgrammist Awesome job!

PR #1159, opened 2024-03-17
chore: add typo checker in ci

tabVersion:
as described in title


frol (approved):
@tabVersion Beautiful! Thank you for the contribution

PR #1086, opened 2023-09-26
[experiment] Remove `#[handle_result]`

uint:
Pushing this early since I'm stupidly excited and figured it's interesting (:

Once I have a PoC, I expect this can be expanded into an ergonomic API for:
- structured errors with schemas (note there's already a way to choose the serialization format)
- opt-out rollback
- opt-in `Result` serialization (if asked for?)

PR #1165, opened 2024-04-08
Standartize errors

### LINKED ISSUE #936: Standardized errors

PolyProgrammist:
Closes #936

### DIFF
--- a/near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
+++ b/near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
@@ -2,7 +2,9 @@ use crate::core_impl::info_extractor::{ImplItemMethodInfo, SerializerType};
 use crate::core_impl::{MethodKind, ReturnKind};
 use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
+use serde::de::value;
 use syn::Receiver;
+use crate::core_impl::utils;
 
 impl ImplItemMethodInfo {
     /// Generate wrapper method for the given method of the contract.
@@ -54,6 +56,8 @@ impl ImplItemMethodInfo {
         let method_invocation = self.method_invocation_tokens();
         let contract_ser = self.contract_ser_tokens();
 
+        eprintln!("default returnkind");
+
         quote! {
             #contract_init
             #method_invocation;
@@ -67,6 +71,9 @@ impl ImplItemMethodInfo {
         let contract_ser = self.contract_ser_tokens();
         let value_ser = self.value_ser_tokens();
         let value_return = self.value_return_tokens();
+
+        eprintln!("general returnkind");
+
 
         quote! {
             #contract_init
@@ -84,6 +91,9 @@ impl ImplItemMethodInfo {
         let value_ser = self.value_ser_tokens();
         let value_return = self.value_return_tokens();
         let result_identifier = self.result_identifier();
+
+
+        eprintln!("handles result returnkind");
 
         quote! {
             #contract_init
@@ -291,6 +301,8 @@ impl ImplItemMethodInfo {
             }
         };
 
+        // self.attr_signature_info.returns.original
+
         match &self.attr_signature_info.method_kind {
             Call(call_method) => {
                 if call_method.receiver.is_some() {
@@ -336,8 +348,20 @@ impl ImplItemMethodInfo {
         use MethodKind::*;
 
         let value_ser = |result_serializer: &SerializerType| match result_serializer {
-            SerializerType::JSON => quote! {
-                let result = ::near_sdk::serde_json::to_vec(&result).expect("Failed to serialize the return value using JSON.");
+            SerializerType::JSON => {
+                eprintln!("{}", quote!{
+                    let result = ::near_sdk::serde_json::json!({
+                        "status": "Success",
+                        "result": result,
+                    });
+                });
+                quote! {
+                    let result = ::near_sdk::serde_json::json!({
+                        "status": "Success",
+                        "result": result,
+                    });
+                    let result = ::near_sdk::serde_json::to_vec(&result).expect("Failed to serialize the return value using JSON.");
+                }
             },
             SerializerType::Borsh => quote! {
                 let result = ::near_sdk::borsh::to_vec(&result).expect("Failed to serialize the return value using Borsh.");
--- a/near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
+++ b/near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
@@ -1,16 +1,45 @@
 use crate::core_impl::ext::generate_ext_function_wrappers;
 use crate::ItemImplInfo;
 use proc_macro2::TokenStream as TokenStream2;
-use quote::ToTokens;
+use quote::{ToTokens, quote, format_ident};
 use syn::{spanned::Spanned, Ident};
+use crate::core_impl::utils;
+use crate::core_impl::{ReturnKind};
 
 impl ItemImplInfo {
     /// Generate the code that wraps
     pub fn wrapper_code(&self) -> TokenStream2 {
         let mut res = TokenStream2::new();
+        let mut checks = quote!{};
         for method in &self.methods {
             res.extend(method.method_wrapper());
+            match method.attr_signature_info.returns.kind {
+                ReturnKind::HandlesResult {..} => {
+                    let error_type = match &method.attr_signature_info.returns.original {
+                        syn::ReturnType::Default => quote! { () },
+                        syn::ReturnType::Type(_, ty) => {
+                            let x = utils::extract_error_type(&ty);
+                            quote! { #x }
+                        },
+                    };
+                    let method_name = &method.attr_signature_info.ident;
+                    let concatenated = format_ident!("assert_implements_my_trait{}", method_name);
+        
+                    checks.extend(quote! {
+                        fn #concatenated() {
+                            let _ = near_sdk::check_trait as fn(&#error_type);
+                        }
+                    });
+                },
+                _ => {},
+            }
         }
+        let current_type = &self.ty;
+        res.extend(quote!{
+            impl #current_type {
+                #checks
+            }
+        });
         res
     }
 
--- a/near-sdk-macros/src/core_impl/info_extractor/visitor.rs
+++ b/near-sdk-macros/src/core_impl/info_extractor/visitor.rs
@@ -221,14 +221,7 @@ fn parse_return_kind(typ: &Type, handles_result: ResultHandling) -> syn::Result<
         }
         ResultHandling::None => {
             if utils::type_is_result(typ) {
-                Err(Error::new(
-                    typ.span(),
-                    "Serializing Result<T, E> has been deprecated. Consider marking your method \
-                    with #[handle_result] if the second generic represents a panicable error or \
-                replacing Result with another two type sum enum otherwise. If you really want \
-                to keep the legacy behavior, mark the method with #[handle_result] and make \
-                it return Result<Result<T, E>, near_sdk::Abort>.",
-                ))
+                Ok(ReturnKind::HandlesResult(typ.clone()))
             } else {
                 Ok(ReturnKind::General(typ.clone()))
             }
--- a/near-sdk-macros/src/core_impl/utils/mod.rs
+++ b/near-sdk-macros/src/core_impl/utils/mod.rs
@@ -47,6 +47,26 @@ pub(crate) fn extract_ok_type(ty: &Type) -> Option<&Type> {
     }
 }
 
+pub(crate) fn extract_error_type(ty: &Type) -> Option<&Type> {
+    match ty {
+        Type::Path(type_path) if type_path.qself.is_none() && path_is_result(&type_path.path) => {
+            // Get the first segment of the path (there should be only one, in fact: "Result"):
+            let type_params = &type_path.path.segments.first()?.arguments;
+            // We are interested in the first angle-bracketed param responsible for Ok type ("<String, _>"):
+            let generic_arg = match type_params {
+                PathArguments::AngleBracketed(params) => Some(params.args.last()?),
+                _ => None,
+            }?;
+            // This argument must be a type:
+            match generic_arg {
+                GenericArgument::Type(ty) => Some(ty),
+                _ => None,
+            }
+        }
+        _ => None,
+    }
+}
+
 /// Checks whether the given path is literally "Vec".
 /// Note that it won't match a fully qualified name `std::vec::Vec` or a type alias like
 /// `type MyVec = Vec<String>`.
--- a/near-sdk-macros/src/lib.rs
+++ b/near-sdk-macros/src/lib.rs
@@ -3,7 +3,7 @@ extern crate proc_macro;
 
 mod core_impl;
 
-use core_impl::{ext::generate_ext_structs, metadata::generate_contract_metadata_method};
+use core_impl::{ext::generate_ext_structs, metadata::generate_contract_metadata_method, extract_error_type};
 
 use proc_macro::TokenStream;
 
@@ -13,6 +13,60 @@ use darling::{Error, FromMeta};
 use proc_macro2::{Ident, Span};
 use quote::{quote, ToTokens};
 use syn::{parse_quote, Expr, ImplItem, ItemEnum, ItemImpl, ItemStruct, ItemTrait, WhereClause};
+
+
+#[proc_macro_derive(MyContractError)]
+pub fn derive_contract_error(input: TokenStream) -> TokenStream {
+    let item: ItemEnum = syn::parse(input).unwrap();
+    
+    let enum_name = &item.ident;
+    
+    let expanded = quote! {
+        impl near_sdk::MyContractErrorTrait for #enum_name {
+            // fn my_fn(&self) -> u64 {
+            //     1
+            // }
+        }
+    };
+    
+    expanded.into()
+}
+
+
+#[proc_macro_attribute]
+pub fn check_trait(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let mut input = syn::parse_macro_input!(item as syn::ItemImpl);
+
+    let self_ty = &input.self_ty;
+
+    let mut checks = quote!{};
+
+    for item in input.clone().items {
+        if let syn::ImplItem::Fn(method) = item {
+            let return_type = match &method.sig.output {
+                syn::ReturnType::Default => quote! { () },
+                syn::ReturnType::Type(_, ty) => {
+                    let x = extract_error_type(&ty);
+                    quote! { #x }
+                },
+            };
+
+            checks.extend(quote! {
+                        let _ = check_trait as fn(&#return_type);
+            });
+        }
+    }
+
+    return TokenStream::from(quote!{
+        #input
+        impl #self_ty {
+            fn assert_implements_my_trait() {
+                #checks
+            }
+        }
+    });
+}
+
 
 #[derive(Debug, Clone)]
 struct Serializers {
@@ -362,13 +416,15 @@ pub fn near_bindgen(attr: TokenStream, item: TokenStream) -> TokenStream {
         let abi_embedded = abi::embed();
         #[cfg(not(feature = "__abi-embed-checked"))]
         let abi_embedded = quote! {};
-        TokenStream::from(quote! {
+        let x = quote! {
             #input
             #ext_gen
             #abi_embedded
             #metadata
             #metadata_impl_gen
-        })
+        };
+        // eprintln!("x: {}", x);
+        TokenStream::from(x)
     } else if let Ok(input) = syn::parse::<ItemEnum>(item.clone()) {
         let metadata = core_impl::contract_source_metadata_const(attr);
         let metadata_impl_gen = generate_metadata(&input.ident, &input.generics);
@@ -383,6 +439,14 @@ pub fn near_bindgen(attr: TokenStream, item: TokenStream) -> TokenStream {
         let abi_embedded = abi::embed();
         #[cfg(not(feature = "__abi-embed-checked"))]
         let abi_embedded = quote! {};
+        let x = quote! {
+            #input
+            #ext_gen
+            #abi_embedded
+            #metadata
+            #metadata_impl_gen
+        };
+        // eprintln!("x: {}", x);
         TokenStream::from(quote! {
             #input
             #ext_gen
@@ -405,11 +469,13 @@ pub fn near_bindgen(attr: TokenStream, item: TokenStream) -> TokenStream {
                 }
             }
         }
-        match process_impl_block(input) {
+        let x = match process_impl_block(input) {
             Ok(output) => output,
             Err(output) => output,
         }
-        .into()
+        .into();
+    eprintln!("x: {}", x);
+        x
     } else {
         TokenStream::from(
             syn::Error::new(
@@ -445,6 +511,13 @@ fn process_impl_block(
     // Add wrapper methods for ext call API
     let ext_generated_code = item_impl_info.generate_ext_wrapper_code();
 
+    let x = quote! {
+        #ext_generated_code
+        #input
+        #generated_code
+        #abi_generated
+    };
+    // eprintln!("x: {}", x);
     Ok(TokenStream::from(quote! {
         #ext_generated_code
         #input
@@ -494,6 +567,11 @@ pub fn ext_contract(attr: TokenStream, item: TokenStream) -> TokenStream {
         };
         let ext_api = item_trait_info.wrap_trait_ext();
 
+        let x = quote! {
+            #input
+            #ext_api
+        };
+        // eprintln!("ext: {}", x);
         TokenStream::from(quote! {
             #input
             #ext_api
--- a/near-sdk/src/lib.rs
+++ b/near-sdk/src/lib.rs
@@ -8,11 +8,17 @@
 extern crate quickcheck;
 
 pub use near_sdk_macros::{
-    ext_contract, near, near_bindgen, BorshStorageKey, EventMetadata, FunctionError, NearSchema,
+    ext_contract, near, near_bindgen, MyContractError, BorshStorageKey, EventMetadata, FunctionError, NearSchema,
     PanicOnDefault,
 };
 
 pub mod store;
+
+pub trait MyContractErrorTrait {
+}
+
+pub fn check_trait<T: MyContractErrorTrait>(_: &T) {}
+
 
 #[cfg(feature = "legacy")]
 pub mod collections;
--- a/kek_callback.rs
+++ b/kek_callback.rs
@@ -1,0 +1,375 @@
+impl CallbackExt {
+    pub fn contract_source_metadata(self) -> ::near_sdk::Promise {
+        let __args = ::std::vec![];
+        ::near_sdk::Promise::new(self.account_id).function_call_weight(
+            ::std::string::String::from("contract_source_metadata"),
+            __args,
+            self.deposit,
+            self.static_gas,
+            self.gas_weight,
+        )
+    }
+}
+impl Callback {
+    pub fn contract_source_metadata() {
+        near_sdk::env::value_return(CONTRACT_SOURCE_METADATA.as_bytes())
+    }
+}
+#[cfg(target_arch = "wasm32")]
+#[no_mangle]
+pub extern "C" fn contract_source_metadata() {
+    ::near_sdk::env::setup_panic_hook();
+    Callback::contract_source_metadata();
+}
+#[derive(
+    :: near_sdk :: borsh :: BorshSerialize,
+    :: near_sdk :: borsh ::
+BorshDeserialize,
+)]
+#[borsh(crate = ":: near_sdk :: borsh")]
+pub struct Callback;
+#[must_use]
+pub struct CallbackExt {
+    pub(crate) account_id: ::near_sdk::AccountId,
+    pub(crate) deposit: ::near_sdk::NearToken,
+    pub(crate) static_gas: ::near_sdk::Gas,
+    pub(crate) gas_weight: ::near_sdk::GasWeight,
+}
+impl CallbackExt {
+    pub fn with_attached_deposit(mut self, amount: ::near_sdk::NearToken) -> Self {
+        self.deposit = amount;
+        self
+    }
+    pub fn with_static_gas(mut self, static_gas: ::near_sdk::Gas) -> Self {
+        self.static_gas = static_gas;
+        self
+    }
+    pub fn with_unused_gas_weight(mut self, gas_weight: u64) -> Self {
+        self.gas_weight = ::near_sdk::GasWeight(gas_weight);
+        self
+    }
+}
+impl Callback {
+    #[doc = r" API for calling this contract's functions in a subsequent execution."]
+    pub fn ext(account_id: ::near_sdk::AccountId) -> CallbackExt {
+        CallbackExt {
+            account_id,
+            deposit: ::near_sdk::NearToken::from_near(0),
+            static_gas: ::near_sdk::Gas::from_gas(0),
+            gas_weight: ::near_sdk::GasWeight::default(),
+        }
+    }
+}
+pub const CONTRACT_SOURCE_METADATA : & 'static str =
+"{\"version\":\"0.1.0\",\"link\":null,\"standards\":[{\"standard\":\"nep330\",\"version\":\"1.1.0\"}]}"
+;
+impl CallbackExt {
+    pub fn contract_source_metadata(self) -> ::near_sdk::Promise {
+        let __args = ::std::vec![];
+        ::near_sdk::Promise::new(self.account_id).function_call_weight(
+            ::std::string::String::from("contract_source_metadata"),
+            __args,
+            self.deposit,
+            self.static_gas,
+            self.gas_weight,
+        )
+    }
+}
+impl Callback {
+    pub fn contract_source_metadata() {
+        near_sdk::env::value_return(CONTRACT_SOURCE_METADATA.as_bytes())
+    }
+}
+#[cfg(target_arch = "wasm32")]
+#[no_mangle]
+pub extern "C" fn contract_source_metadata() {
+    ::near_sdk::env::setup_panic_hook();
+    Callback::contract_source_metadata();
+}
+impl CallbackExt {
+    pub fn call_all(self, fail_b: bool, c_value: u8, d_value: u8) -> ::near_sdk::Promise {
+        let __args = {
+            #[derive(:: near_sdk :: serde :: Serialize)]
+            #[serde(crate = "::near_sdk::serde")]
+            struct Input<'nearinput> {
+                fail_b: &'nearinput bool,
+                c_value: &'nearinput u8,
+                d_value: &'nearinput u8,
+            }
+            let __args = Input { fail_b: &fail_b, c_value: &c_value, d_value: &d_value };
+            ::near_sdk::serde_json::to_vec(&__args)
+                .expect("Failed to serialize the cross contract args using JSON.")
+        };
+        ::near_sdk::Promise::new(self.account_id).function_call_weight(
+            ::std::string::String::from("call_all"),
+            __args,
+            self.deposit,
+            self.static_gas,
+            self.gas_weight,
+        )
+    }
+    pub fn a(self) -> ::near_sdk::Promise {
+        let __args = ::std::vec![];
+        ::near_sdk::Promise::new(self.account_id).function_call_weight(
+            ::std::string::String::from("a"),
+            __args,
+            self.deposit,
+            self.static_gas,
+            self.gas_weight,
+        )
+    }
+    pub fn b(self, fail: bool) -> ::near_sdk::Promise {
+        let __args = {
+            #[derive(:: near_sdk :: serde :: Serialize)]
+            #[serde(crate = "::near_sdk::serde")]
+            struct Input<'nearinput> {
+                fail: &'nearinput bool,
+            }
+            let __args = Input { fail: &fail };
+            ::near_sdk::serde_json::to_vec(&__args)
+                .expect("Failed to serialize the cross contract args using JSON.")
+        };
+        ::near_sdk::Promise::new(self.account_id).function_call_weight(
+            ::std::string::String::from("b"),
+            __args,
+            self.deposit,
+            self.static_gas,
+            self.gas_weight,
+        )
+    }
+    pub fn c(self, value: u8) -> ::near_sdk::Promise {
+        let __args = {
+            #[derive(:: near_sdk :: serde :: Serialize)]
+            #[serde(crate = "::near_sdk::serde")]
+            struct Input<'nearinput> {
+                value: &'nearinput u8,
+            }
+            let __args = Input { value: &value };
+            ::near_sdk::serde_json::to_vec(&__args)
+                .expect("Failed to serialize the cross contract args using JSON.")
+        };
+        ::near_sdk::Promise::new(self.account_id).function_call_weight(
+            ::std::string::String::from("c"),
+            __args,
+            self.deposit,
+            self.static_gas,
+            self.gas_weight,
+        )
+    }
+    pub fn d(self, value: u8) -> ::near_sdk::Promise {
+        let __args = {
+            #[derive(:: near_sdk :: serde :: Serialize)]
+            #[serde(crate = "::near_sdk::serde")]
+            struct Input<'nearinput> {
+                value: &'nearinput u8,
+            }
+            let __args = Input { value: &value };
+            ::near_sdk::serde_json::to_vec(&__args)
+                .expect("Failed to serialize the cross contract args using JSON.")
+        };
+        ::near_sdk::Promise::new(self.account_id).function_call_weight(
+            ::std::string::String::from("d"),
+            __args,
+            self.deposit,
+            self.static_gas,
+            self.gas_weight,
+        )
+    }
+    pub fn handle_callbacks(self) -> ::near_sdk::Promise {
+        let __args = ::std::vec![];
+        ::near_sdk::Promise::new(self.account_id).function_call_weight(
+            ::std::string::String::from("handle_callbacks"),
+            __args,
+            self.deposit,
+            self.static_gas,
+            self.gas_weight,
+        )
+    }
+}
+impl Callback {
+    #[doc = " Call functions a, b, and c asynchronously and handle results with `handle_callbacks`."]
+    pub fn call_all(fail_b: bool, c_value: u8, d_value: u8) -> Promise {
+        Self::ext(env::current_account_id())
+            .a()
+            .and(Self::ext(env::current_account_id()).b(fail_b))
+            .and(Self::ext(env::current_account_id()).c(c_value))
+            .and(Self::ext(env::current_account_id()).d(d_value))
+            .then(Self::ext(env::current_account_id()).handle_callbacks())
+    }
+    #[doc = " Calls function c with a value that will always succeed"]
+    pub fn a() -> Promise {
+        Self::ext(env::current_account_id()).c(A_VALUE)
+    }
+    #[doc = " Returns a static string if fail is false, return"]
+    pub fn b(fail: bool) -> &'static str {
+        if fail {
+            env::panic_str("failed within function b");
+        }
+        "Some string"
+    }
+    #[doc = " Panics if value is 0, returns the value passed in otherwise."]
+    pub fn c(value: u8) -> u8 {
+        require!(value > 0, "Value must be positive");
+        value
+    }
+    #[doc = " Panics if value is 0."]
+    pub fn d(value: u8) {
+        require!(value > 0, "Value must be positive");
+    }
+    #[doc = " Receives the callbacks from the other promises called."]
+    pub fn handle_callbacks(
+        a: u8,
+        b: Result<String, PromiseError>,
+        c: Result<u8, PromiseError>,
+        d: Result<(), PromiseError>,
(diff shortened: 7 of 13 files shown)


near-sdk-macros/src/core_impl/info_extractor/visitor.rs
```diff
@@ -209,26 +220,26 @@ fn is_view(sig: &Signature) -> bool {
...
             if utils::type_is_result(typ) {
-                Err(Error::new(
-                    typ.span(),
-                    "Serializing Result<T, E> has been deprecated. Consider marking your method \
-                    with #[handle_result] if the second generic represents a panicable error or \
-                replacing Result with another two type sum enum otherwise. If you really want \
-                to keep the legacy behavior, mark the method with #[handle_result] and make \
-                it return Result<Result<T, E>, near_sdk::Abort>.",
-                ))
+                Ok(ReturnKind::HandlesResultImplicit(crate::StatusResult {
```
PolyProgrammist, line 239: I would like to mention this change: from some time using Result without #[handle_result] became deprecated. So introducing it again may break contracts that use that old sdk and move to the new version. Please let me know if this change should be reconsidered.

examples/error-handling/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+
+#[near]
+impl Contract {
+    #[handle_result]
+    pub fn inc_handle_result(&mut self, is_error: bool) -> Result<u32, &'static str> {
+        self.value += 1;
+        if is_error {
+            return Err("error in inc_handle_result");
+        } else {
+            return Ok(self.value);
```
frol, line 29: Let's assume that readers of the examples are familiar with the fact that the last statement in Rust block is the return value, and don't explicitly use `return` statements:

```suggestion
            Err("error in inc_handle_result")
        } else {
            Ok(self.value)
```

P.S. Apply it in all other occurences

examples/error-handling/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+#[near(contract_state)]
+#[derive(Default)]
+pub struct Contract {
+    value: u32,
+}
+
+#[near]
+impl Contract {
+    #[handle_result]
+    pub fn inc_handle_result(&mut self, is_error: bool) -> Result<u32, &'static str> {
```
frol, line 24: Let's provide doc strings that illustrate the potential return values from the RPC response point of view.

P.S. Apply the same reasoning below

examples/error-handling/tests/test_basics.rs
```diff
@@ -0,0 +1,57 @@
...
+    let contract =
+        worker.dev_deploy(&std::fs::read(format!("res/{}.wasm", "error_handling"))?).await?;
+
+    check_call(&contract, "inc_handle_result", false, 1, None).await;
+    check_call(&contract, "inc_persist_on_err", false, 2, None).await;
+    check_call(&contract, "inc_just_result", false, 3, None).await;
+    check_call(&contract, "inc_just_simple", false, 4, None).await;
+    check_call(&contract, "inc_handle_result", true, 4, None).await;
+    check_call(&contract, "inc_persist_on_err", true, 5, Some("Error { repr: Custom { kind: Execution, error: ActionError(ActionError { index: Some(0), kind: FunctionCallError(ExecutionError(\"Smart contract panicked: {\\\"error\\\":{\\\"error_type\\\":\\\"error_handling::MyErrorEnum\\\",\\\"value\\\":\\\"X\\\"}}\")) }) } }".to_string())).await;
+    check_call(&contract, "inc_just_result", true, 5, Some("Error { repr: Custom { kind: Execution, error: ActionError(ActionError { index: Some(0), kind: FunctionCallError(ExecutionError(\"Smart contract panicked: {\\\"error\\\":{\\\"error_type\\\":\\\"error_handling::MyErrorStruct\\\",\\\"value\\\":{\\\"x\\\":5}}}\")) }) } }".to_string())).await;
```
frol, line 53: I talked with some engineers on nearcore side and it seems that there is no capacity on their side to address this error handling on their side :-(

We will probably have to contribute it ourselves. I will guide you there once we will get to it.

near-sdk-macros/src/core_impl/info_extractor/visitor.rs
```diff
@@ -209,26 +220,26 @@ fn is_view(sig: &Signature) -> bool {
...
         ResultHandling::Check => {
             if !utils::type_is_result(typ) {
                 Err(Error::new(typ.span(), "Function marked with #[handle_result] should return Result<T, E> (where E implements FunctionError). If you're trying to use a type alias for `Result`, try `#[handle_result(aliased)]`."))
             } else {
-                Ok(ReturnKind::HandlesResult(typ.clone()))
+                Ok(ReturnKind::HandlesResultExplicit(typ.clone()))
             }
         }
         ResultHandling::None => {
             if utils::type_is_result(typ) {
```
frol, line 238: It seems that you rely on this function, but this means that we won't be able to support Result aliases that are [not named `Result`](https://github.com/near/near-sdk-rs/blob/e22f74ea9f6a0be8fe030a9225716eab811116ea/near-sdk-macros/src/core_impl/utils/mod.rs#L10-L25).

There was an effort to express it with the type system: https://github.com/near/near-sdk-rs/pull/1086

near-sdk-macros/src/core_impl/code_generator/snapshots/persist_on_error.snap
```diff
@@ -0,0 +1,29 @@
...
+        ::std::result::Result::Ok(result) => {
+            let result = ::near_sdk::serde_json::to_vec(&result)
+                .expect("Failed to serialize the return value using JSON.");
+            ::near_sdk::env::value_return(&result);
+            ::near_sdk::env::state_write(&contract);
+        }
+        ::std::result::Result::Err(err) => {
+            ::near_sdk::env::state_write(&contract);
+            let promise = Contract::ext(::near_sdk::env::current_account_id())
+                .method_error(err)
```
frol, line 25: Is there a test that demonstrates the code generated for this auxiliary method?

near-sdk-macros/src/core_impl/utils/mod.rs
```diff
@@ -152,6 +173,29 @@ pub struct SanitizeSelfResult {
...
+        struct ErrorWrapper {
+            error_type: &'static str,
+            value: #error_type,
+        };
+
+        let err = ErrorWrapper {
+            error_type: std::any::type_name::<#error_type>(),
+            value: err,
+        };
+        let err = ::near_sdk::serde_json::json!{{"error": err}};
```
frol, line 194: Let's copy the errors structure from NEAR RPC: https://docs.near.org/api/rpc/contracts#what-could-go-wrong

```json
{
  "error": {
    "name": <ERROR_TYPE>,
    "cause": {
      "info": {..},
      "name": <ERROR_CAUSE>
    }
  }
}
```

For example, for user-generated errors:

```json
{
  "error": {
    "name": "CUSTOM_CONTRACT_ERROR",
    "cause": {
      "info": {..},
      "name": "MY_ERROR_TYPE"
    }
  }
}
```

For SDK-generated errors:

```json
{
  "error": {
    "name": "SDK_CONTRACT_ERROR",
    "cause": {
      "info": {..},
      "name": "INVALID_INPUT" // "UNEXPECTED_ATTACHED_BALANCE" | "INVALID_INTERNAL_STATE" | ...
    }
  }
}
```

frol (changes requested):

near-sdk-macros/src/core_impl/info_extractor/visitor.rs
PolyProgrammist: It was initially called #[return_result] instead of #[handle_result] - change [applied](https://github.com/near/near-sdk-rs/pull/754/files)

Here is the first issue about this macro and [comment](https://github.com/near/near-sdk-rs/issues/574#issuecomment-923284947) by [austinabell](https://github.com/austinabell). According to the comment I believe it was decided to use explicit attribute so that it does not accidentally break existing contracts. So the macro was introduced in March 2022 along with forbidding to use Result without macro a bit [later](https://github.com/near/near-sdk-rs/pull/757/files).

near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
```diff
@@ -23,13 +52,41 @@ impl ItemImplInfo {
             Err(e) => syn::Error::new(self.ty.span(), e).to_compile_error(),
         }
     }
+
+    pub fn generate_error_methods(&self) -> TokenStream2 {
+        let mut error_methods = quote! {};
+
+        self.methods.iter().map(|m| &m.attr_signature_info).for_each(|method| {
+            let error_method_name = quote::format_ident!("{}_error", method.ident);
```
akorchyn, line 60: move into if

near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
```diff
@@ -23,13 +52,41 @@ impl ItemImplInfo {
             Err(e) => syn::Error::new(self.ty.span(), e).to_compile_error(),
         }
     }
+
+    pub fn generate_error_methods(&self) -> TokenStream2 {
+        let mut error_methods = quote! {};
+
+        self.methods.iter().map(|m| &m.attr_signature_info).for_each(|method| {
```
akorchyn, line 59: Let's add a warning that _error suffix in method identifier is reserved for our usage

akorchyn (changes requested):
Initial portion of the review: I have gone through near-sdk and near-sdk-macros (i.e., the internal implementation), but I haven't reviewed the examples or near-contract-standards yet.

To be fair, I would prefer to treat near-contract-standards separately, as I believe it's a high-risk venture to approve such a significant change all at once. I'd rather go through each standard individually with a careful review of each.

examples/error-handling/src/lib.rs
PolyProgrammist: docs provided

near-sdk-macros/src/core_impl/code_generator/snapshots/persist_on_error.snap
PolyProgrammist: [generated_method_error.snap](https://github.com/near/near-sdk-rs/pull/1165/files#diff-413175644291e9ab2285fdd16ef13c24203590b50188a988ce3fb2c44d77186e)

near-sdk-macros/src/lib.rs
```diff
@@ -927,3 +931,150 @@ pub fn derive_event_attributes(item: TokenStream) -> TokenStream {
...
+        #[#near_sdk_crate ::near(serializers=[json], inside_nearsdk=#bool_inside_nearsdk_for_macro)]
+        #input
+
+        impl #near_sdk_crate ::ContractErrorTrait for #ident {
+            fn error_type(&self) -> &'static str {
+                #error_type
+            }
+
+            fn wrap(&self) -> #near_sdk_crate ::serde_json::Value {
+                #[#near_sdk_crate ::near(inside_nearsdk=#bool_inside_nearsdk_for_macro, serializers = [json])]
```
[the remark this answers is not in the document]
PolyProgrammist: Where do you want that to be extracted? If there is a place, it should be under some macro. 
And do struct definitions really affect binary size?

near-sdk-macros/src/core_impl/info_extractor/mod.rs
```diff
@@ -85,7 +85,82 @@ pub struct Returns {
...
+    /// pub struct MyError;
+    ///
+    /// // if Ok() is returned, everything ok, otherwise panic with well-structured error
+    /// pub fn foo(&mut self) -> Result<u64, MyError>;
+    /// ```
+    ///
+    /// ```ignore
+    /// // Write state safely anyway.
+    /// // if Ok() is returned, just return. Otherwise call new Promise which will panic with well-structured error.
+    /// #[persist_on_error]
```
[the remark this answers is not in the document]
PolyProgrammist: Could you please provide an example of a security issue?
Do you mean renaming it to `unsafe_persist_on_error`?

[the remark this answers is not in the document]
akorchyn: yeah. I think we should flag that this is something that the user should understand very well before using.

Easy to create an issue:
* You have X
* You persist on error and revert X
* You add Y
* You forgot to revert Y

You might even forget to revert X. So it's better to flag for extra attention...

near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
PolyProgrammist: There is no such thing as compile_warning macro. https://github.com/rust-lang/rfcs/issues/3704

Do you think using the workaround provided by issue author is ok?

akorchyn: Probably yeah. We can put into the implementation of that method a warning variable. But come up with a more clever name. E.g., `near_warning_func_name` to avoid collisions

near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
```diff
@@ -94,8 +97,32 @@ impl ImplItemMethodInfo {
...
+            }
+        }
+    }
+
+    fn error_handling_tokens(&self) -> TokenStream2 {
+        match &self.attr_signature_info.returns.kind {
+            ReturnKind::HandlesResultImplicit(status_result) => {
+                if status_result.unsafe_persist_on_error {
+                    let error_method_name =
+                        quote::format_ident!("{}_error", self.attr_signature_info.ident);
```
akorchyn: maybe let's call it near_sdk_internal_x_error
or something like that?

near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
```diff
@@ -94,8 +97,32 @@ impl ImplItemMethodInfo {
...
                 }
-                ::std::result::Result::Err(err) => ::near_sdk::FunctionError::panic(&err)
+                ::std::result::Result::Err(err) => {
+                    #handle_error
+                }
+            }
+        }
+    }
+
+    fn error_handling_tokens(&self) -> TokenStream2 {
```
akorchyn, line 107: Can the user combine Promise API and unsafe_persist_on_error?

near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
```diff
@@ -23,13 +52,64 @@ impl ItemImplInfo {
...
+            }
+
+            if let ReturnKind::HandlesResultImplicit(status) = &method.returns.kind {
+                // if method.ident ends with _error, emit warning to avoid name clash
+                if method.ident.to_string().ends_with("_error") {
+                    let warning_message = format!(
+                        "Method '{}' ends with '_error'. This suffix in method identifier is reserved for our usage",
+                        method.ident
+                    );
+                    let warning_name = format_ident!("reserved_error_suffix_{}", method.ident);
```
akorchyn: Should we allow to supress it too? Probably not, but still up to discussion

near-sdk-macros/src/core_impl/info_extractor/mod.rs
```diff
@@ -91,7 +91,88 @@ pub struct Returns {
...
+    /// ```
+    ///
+    /// ```ignore
+    /// // Write state anyway.
+    /// // if Ok() is returned, just return. Otherwise call new Promise which will panic with well-structured error.
+    /// #[unsafe_persist_on_error]
+    /// pub fn foo(&mut self) -> Result<u64, MyError>;
+    /// ```
+    ///
+    HandlesResultImplicit(StatusResult),
```
akorchyn, line 166: awesome work on this documentation

near-sdk-macros/src/lib.rs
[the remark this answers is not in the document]
akorchyn: for some reason I thought that we can extract outside and just reference it here

near-sdk/tests/store_performance_tests.rs
```diff
@@ -479,7 +479,8 @@ async fn test_lazy() -> anyhow::Result<()> {
         .unwrap();
 
     // Override min gas to avoid constant tuning, it's pretty clear this is performant.
-    perform_asserts(res.total_gas_burnt.as_gas(), "lazy:get", Some(70));
+    println!("Total gas for lazy:get: {}", res.total_gas_burnt.as_gas());
```
akorchyn: nit: debug print

near-sdk/Cargo.toml
```diff
@@ -111,3 +111,5 @@ features = [
     "__macro-docs",
     "__abi-generate",
 ]
+
+
```
akorchyn: nit: empty lines

near-sdk/src/utils/mod.rs
```diff
@@ -75,6 +79,56 @@ macro_rules! require {
...
+        }
+    };
+}
+
+#[macro_export]
+macro_rules! compile_warning {
+    (
+    $name:ident, $message:expr $(,)*
+) => {
+        mod $name {
```
mitinarseny:
```suggestion
        const _: () = {
```

near-sdk-macros/src/core_impl/info_extractor/mod.rs
```diff
@@ -91,7 +91,88 @@ pub struct Returns {
...
+    ///
+    /// // if Ok() is returned, everything ok, otherwise panic with well-structured error
+    /// pub fn foo(&mut self) -> Result<u64, MyError>;
+    /// ```
+    ///
+    /// ```ignore
+    /// // Write state anyway.
+    /// // if Ok() is returned, just return. Otherwise call new Promise which will panic with well-structured error.
+    /// #[unsafe_persist_on_error]
+    /// pub fn foo(&mut self) -> Result<u64, MyError>;
```
mitinarseny, line 163: IMO, this is a very bad design choice: smart-contract developers should be responsible for panicking on some errors and omitting others , while, most importantly, never leave the contract in corrupted state.

Moreover, which `u64` would be returned here on error exactly?

Is there any real-world use-case where `#[unsafe_persist_on_error]` would actually be used?

near-sdk/src/errors/mod.rs
mitinarseny, line 1: Did you consider also adding `ResultExt` traits like [here](https://github.com/near/intents/blob/f573c251eeef545a0c3978c03bef8cfd30feac74/near-utils/src/panic.rs#L5-L102)? This is replicated in every second contract repo I encounter:
```rust
pub trait UnwrapOrPanic<T> {
    fn unwrap_or_panic(self) -> T;
}

impl<T> UnwrapOrPanic<T> for Option<T> {
    #[inline]
    #[track_caller]
    fn unwrap_or_panic(self) -> T {
        self.unwrap_or_else(|| env::abort())
    }
}

impl<T, E> UnwrapOrPanic<T> for Result<T, E>
where
    E: FunctionError,
{
    #[inline]
    #[track_caller]
    fn unwrap_or_panic(self) -> T
    where
        E: FunctionError,
    {
        self.unwrap_or_else(|err| err.panic())
    }
}
```

near-sdk/src/errors/custom.rs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+impl InsufficientBalance {
+    pub fn new(message: Option<&str>) -> Self {
+        Self { message: message.map(|s| s.to_string()) }
+    }
+}
+
+#[contract_error(inside_nearsdk)]
+pub struct InsufficientGas {}
```
mitinarseny, line 107:
```suggestion
pub struct InsufficientGas;
```
Same for other empty structs

near-sdk/src/errors/custom.rs
```diff
@@ -0,0 +1,127 @@
+use near_sdk_macros::contract_error;
+
+#[contract_error(inside_nearsdk)]
+pub struct InvalidArgument {
+    pub message: String,
+}
```
mitinarseny: General question: why did you chose to have per-error-type structs and not a single enum?
```rust
#[derive(Debug, thiserror::Error)]
pub enum ContractError {
    InvalidArgument(/*...*/),
    InvalidContractState(/*...*/),
    PermissionDenied(/*...*/),
    // ...

    #[error(transparent)]
    Custom(String), // <- here all custom errors from `fn method() -> Result<Value, MyErr> ` go
}
```

near-sdk/src/errors/mod.rs
frol: @mitinarseny How does it compare to the regular `.unwrap()`? On the surface it looks exactly the same to me

near-sdk-macros/src/core_impl/info_extractor/mod.rs
frol: > Is there any real-world use-case where [unsafe_persist_on_error] would actually be used?

@mitinarseny There were several usecases in Keypom and other linkdrop contracts that wanted to keep track of the action, but also signal with a transaction error to the cross-contract or transaction calls.

> IMO, this is a very bad design choice: smart-contract developers should be responsible for panicking on some errors and omitting others , while, most importantly, never leave the contract in corrupted state.

I don't see where this design contradicts your statement. This PR enables more granular error design instead of the current model where developers usually panic in random places of the code and as such potential errors are not encoded in the ABI at all.

near-sdk/src/errors/custom.rs
frol: I believe there was some limitation in serde serialization for the `;` structs, so we had to go with `{}`. Yet, we should document this detail clearly in the comment.

@PolyProgrammist Do you recall the details? I feel it was about `null` and `[]` (empty tuple) serialization or deserialization.

near-sdk/src/errors/custom.rs
frol: This would defeat the purpose of strictly typed error. We don't want users to have `any Error` type in every function signature. We encourage devs to use `thiserror` and construct their own error types where only the relevant errors can be thrown, so the enum will be defined on the contract level, not SDK level

near-sdk-macros/src/core_impl/info_extractor/mod.rs
mitinarseny: >  but also signal with a transaction error to the cross-contract or transaction calls

I see, thanks!

So then regarding auto-generated `<METHOD_NAME>_error` method: wouldn't it create too many per-method error-methods? While it could have been done via a single `panic_with` method (like [here](https://github.com/near/intents/blob/360abab593ad6d1a21b862c6d1881734353fc79b/escrow-swap/src/contract/return_value.rs#L5-L12)):
```rust
// a single optimized implementation without `#[near]` macro
#[unsafe(no_mangle)]
pub extern "C" fn panic_with() {
    if let Some(input) = ::near_sdk::env::input() {
        if let Ok(err) = str::from_utf8(&input) {
            ::near_sdk::env::panic_str(&input);
        }
    }
}

```

near-sdk/src/errors/mod.rs
mitinarseny: The version above is more efficient since it calls `near_sdk::env::abort()` directly instead of going through panic hook with constructed [PanicHookInfo](https://doc.rust-lang.org/stable/std/panic/struct.PanicHookInfo.html)

near-sdk/src/errors/custom.rs
mitinarseny: Ok, what about having it as type param?
```rust
#[derive(Debug, thiserror::Error)]
pub enum ContractError<T>
    where T: Into<crate::BaseError>
{
    InvalidArgument(/*...*/),
    InvalidContractState(/*...*/),
    PermissionDenied(/*...*/),
    // ...

    #[error("{}", serde_json::json!(.0.into()))]
    Other(T), //
}
```

near-sdk/src/errors/custom.rs
```diff
@@ -0,0 +1,127 @@
...
+impl ContractUpgradeError {
+    pub fn new(message: &str) -> Self {
+        Self { message: message.to_string() }
+    }
+}
+#[contract_error(inside_nearsdk)]
+#[derive(Default)]
+pub struct RequireFailed {
+    pub message: String,
+}
```
mitinarseny: Did you consider avoiding `String` allocation everywhere when the error is created from `&'static str`, let's say?
```suggestion
pub struct RequireFailed<'a> {
    pub message: Cow<'a, str>,
}
```

near-sdk-macros/src/core_impl/info_extractor/mod.rs
frol: @PolyProgrammist Can we have a single `__near_sdk_panic_callback` method shared for the whole contract instead of generating `{method}_error` functions? It seems to be a fair suggestion

near-sdk/src/errors/mod.rs
frol: @mitinarseny It is called once. Do you have an evidence that it has a noticeable gas overhead? I would not expect so. I don't want to require users to import the trait in their scopes everywhere with no good reason behind it.

near-sdk/src/errors/custom.rs
frol: Not every function will have all those possible errors (PermissionDenied, PromiseFailed, InvalidHashLength, ...). We may create a helper ContractError enum like you suggest, but it is beneficial to have all these errors separately, so one can actually construct a lean error type relevant for the function.

near-sdk/src/errors/mod.rs
mitinarseny: I've double-checked, with following Cargo config the size overhead is 30 bytes which is negligible, I suppose:
```toml
[profile.release]
codegen-units = 1
opt-level = 3
lto = true
debug = false
strip = "symbols"
panic = "abort"
overflow-checks = true
```
So I agree, regular `.unwrap()` can be used instead.
But still, why to even have such `require_or_err` macros? And why to implement them as macros and not the functions?

near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
PolyProgrammist: I think it's easy to change _error name to some other for the developer, so I would encourage that rather than allowing to suppress the warning

near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
PolyProgrammist: Currently yes. @frol what do you think about it?

PolyProgrammist: And what do you think about it @akorchyn

near-sdk/src/errors/mod.rs
PolyProgrammist: There is also `require!` macro. I added `require_or_err` in addition to `require` since it is used to return error, whilst `require` calls `panic_str`. That's not what you want when you want to return an error. So I believe require_or_err is necessary because it simplifies the code

PolyProgrammist: @mitinarseny

near-sdk/src/errors/custom.rs
PolyProgrammist: I am not sure, but maybe you are talking about progenitor. The issue was that it doesn't handle null types in the schema as it only supports openapi 3.0, not 3.1. Unit structs serialize to null, while empty structs serialize to {} which progenitor can process correctly.

PolyProgrammist: @frol

near-sdk/src/errors/custom.rs
PolyProgrammist: Let's add helper in a separate PR so not to overcomplicate the review https://github.com/near/near-sdk-rs/issues/1447

near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
PolyProgrammist: resolving as it's now single callback

near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
frol: Good catch! Tough question. Let me think outloud here. There are multiple scenarious I have just came up with

```rust
#[persist_on_error]
fn claim(&mut self) -> Result<Promise, MyErr> {
    let promise = Promise::new()...;
    if some_error {
        return Err(MyErr);
    }
    Ok(promise)
}
```

If Ok - the promise should be called next.
If Err - the promise MUST NOT be called, but it may actually get scheduled as promises are scheduled in the Drop implementation, and `promise` variable will get dropped on `return Err();`

That is problem no. 1.

There is a different scenario, what if I want to persist on error, report the error, and schedule the promise after it?

```rust
Promise::new(self).call_function("__near_sdk_on_error")
  .then(Promise::new(self).call_function("refund"))
  // the `refund` function will get the call with `is_promise_success` set to false
```

To be honest, I cannot think of the case where it would be needed as there seems to be no point in calling `refund` separately from the function that we already know going to fail.

PolyProgrammist: issue for that - https://github.com/near/near-sdk-rs/issues/1451

near-sdk/src/errors/mod.rs
mitinarseny: > How does it compare to the regular .unwrap()?

The regular `::core::result::Result::unwrap()` [wraps](https://doc.rust-lang.org/book/ch09-01-unrecoverable-errors-with-panic.html) the error message with unnecessary info unless `panic = "immediate-abort"` is set:

```rust
impl<T, E> Result<T, E> {
    // ...
    pub fn unwrap(self) -> T
    where
        E: fmt::Debug,
    {
        match self {
            Ok(t) => t,
            Err(e) => unwrap_failed("called `Result::unwrap()` on an `Err` value", &e),
        }
    }
}

// This is a separate function to reduce the code size of the methods
#[cfg(not(panic = "immediate-abort"))]
#[inline(never)]
#[cold]
#[track_caller]
fn unwrap_failed(msg: &str, error: &dyn fmt::Debug) -> ! {
    panic!("{msg}: {error:?}");
}

// ...

// This is a separate function to avoid constructing a `dyn Debug`
// that gets immediately thrown away, since vtables don't get cleaned up
// by dead code elimination if a trait object is constructed even if it goes
// unused
#[cfg(panic = "immediate-abort")]
#[inline]
#[cold]
#[track_caller]
const fn unwrap_failed<T>(_msg: &str, _error: &T) -> ! {
    panic!()
}
```

PolyProgrammist:
> __near_sdk_panic_callback

There is an if that says that [if](https://github.com/PolyProgrammist/near-sdk-rs/blob/a17a4ea64ca85450169adea05fdc5aaaea513059/near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs#L82) there is no `#[unsafe_persist_on_error]`, the callback is not in the code.

@mitinarseny

near-sdk/src/utils/mod.rs
```diff
@@ -75,6 +81,56 @@ macro_rules! require {
...
+        if !$cond {
+            return Err(::near_sdk::errors::RequireFailed::new().into());
+        }
+    };
+    ($cond:expr, $err:expr $(,)?) => {
+        if !$cond {
+            return Err($err.into());
+        }
+    };
+}
```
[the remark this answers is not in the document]
PolyProgrammist: PR description is outdated

near-sdk/src/private/mod.rs
```diff
@@ -5,10 +5,9 @@ pub use near_abi::{
     AbiBorshParameter, AbiFunction, AbiFunctionKind, AbiFunctionModifier, AbiJsonParameter,
     AbiParameters, AbiType,
 };
-#[cfg(feature = "abi")]
+
 mod result_type_ext;
 
-#[cfg(feature = "abi")]
 pub use result_type_ext::ResultTypeExt;
```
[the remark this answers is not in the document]
PolyProgrammist: Nice catch

near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
frol: I believe it is not a blocker.

Just a follow up thought:

> If Err - the promise MUST NOT be called, but it may actually get scheduled as promises are scheduled in the Drop implementation, and promise variable will get dropped on return Err();

It is only a problem for the `persist_on_error` case. Also, in near-sdk v5 we already have the warning for hanging promises and the developer must explicitly call `.detach()` to schedule the promise. In v6 we may drop the `Drop` implementation for the promises and this issue will be solved.

As to the second problem, I think it is not a blocker and in the worst case the dev will just implement the error handling and call the promises manually.

near-sdk-macros/src/core_impl/info_extractor/mod.rs
```diff
@@ -85,7 +85,82 @@ pub struct Returns {
...
+    /// Return type is Result<OkType, ErrType> and the function is marked with #[handle_result].
+    /// ErrType struct implements near_sdk::FunctionError. (i.e. used with #[derive(near_sdk::FunctionError)])
+    ///
+    /// When the contract call happens, in addition to the General:
+    ///  - In case Result value is Ok, the unwrapped object is returned
+    ///  - In case Result value is Err, panic is called and state is not written.
+    ///
+    /// # Example:
+    /// ```ignore
+    /// #[handle_result]
```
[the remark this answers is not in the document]
frol: The plan to completely deprecate `#[handle_result]` attribute in the future. @PolyProgrammist right? Or do we still need it?

near-sdk/src/errors/sdk.rs
```diff
@@ -0,0 +1,187 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[contract_error(inside_nearsdk, sdk)]
+pub struct IndexOutOfBounds {}
+
+#[contract_error(inside_nearsdk, sdk)]
+pub struct KeyNotFound {}
```
frol: Can it also include the key that was attempted to look up? It could be helpful for troubleshooting. Yet, if this makes things too complicated, I don't insist on adding it.

near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
```diff
@@ -23,13 +52,55 @@ impl ItemImplInfo {
...
+
+    pub fn generate_error_method(&self) -> TokenStream2 {
+        let mut result = quote! {};
+        let mut needs_panic_callback = false;
+
+        self.methods.iter().map(|m| &m.attr_signature_info).for_each(|method| {
+            if let ReturnKind::HandlesResultExplicit(explicit_result) = &method.returns.kind {
+                if !explicit_result.suppress_warnings {
+                    let warning_message = format!(
+                        "Method '{}' uses #[handle_result] attribute which is deprecated. Consider using implicit Result handling instead.",
```
frol, line 64: I believe it should not throw the warning if the result type is not "Result" as it implicit Result handling won't catch it. It seems that `#[handle_result]` is currently the only way to use a custom Result type, right?

near-sdk-macros/src/core_impl/code_generator/snapshots/unsafe_persist_on_error.snap
```diff
@@ -0,0 +1,38 @@
...
+        .unwrap_or_default();
+    let result = Hello::method(&mut contract);
+    match result {
+        ::std::result::Result::Ok(result) => {
+            let result = match near_sdk::serde_json::to_vec(&result) {
+                Ok(v) => v,
+                Err(_) => {
+                    ::near_sdk::env::panic_str(
+                        "Failed to serialize the return value using JSON.",
+                    )
```
frol: This should be a JSON instead of the string, right?

examples/lockable-fungible-token/src/lib.rs
```diff
@@ -58,7 +58,7 @@ pub struct FunToken {
 #[near]
 impl FunToken {
     #[init]
-    #[handle_result]
+    #[handle_result(suppress_warnings)]
```
frol, line 61: Just to align on the expectations: I believe that the expectation is that we drop `#[handle_result(suppress_warnings)]` lines in this example with v6 release, right? Maybe we can drop them in this PR already.

examples/error-handling/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+    // (changes value from 0 to 1)
+    //
+    // is_error = true
+    // Failed transaction
+    // Error:
+    // 0: Error: An error occurred during a `FunctionCall` Action, parameter is debug message.
+    // ExecutionError("Smart contract panicked: error in inc_handle_result")
+    // (does not change value)
+    #[handle_result(suppress_warnings)]
+    pub fn inc_handle_result(&mut self, is_error: bool) -> Result<u32, &'static str> {
```
frol, line 42: I believe we should not showcase `suppress_warnings`. I'd change the example to be useful for the contract devs:
1. a simple case (communicate error with panic, currently called `inc_just_simple`)
2. evolve the simple case into the implicit Result handling (`inc_just_result`)
3. showcase `unsafe_persist_on_error` (and document when it is useful - linkdrop contract that wants to keep track of the failing attempts in the contract state and yet signal the error for cross-contract callbacks)
4. showcase `handle_result` (without suppress_warnings) for a custom Result type

examples/error-handling/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+    // ------------------------------------
+    // (changes value from 3 to 4)
+    //
+    // is_error = true
+    // Failed transaction
+    // Error:
+    // 0: Error: An error occurred during a `FunctionCall` Action, parameter is debug message.
+    //  ExecutionError("Smart contract panicked: {\"error\":{\"error_type\":\"error_handling::MyErrorStruct\",\"value\":{\"x\":5}}}")
+    // (does not change value)
+    pub fn inc_just_result(&mut self, is_error: bool) -> Result<u32, MyErrorStruct> {
```
frol: @PolyProgrammist THIS IS CRITICAL. This creates misalignment as standard near-sdk call processing will return `{"error":{"name":"...","cause":{}}` (e.g. on input parameters parsing or STATE parsing). I believe this function must return the same error structure as `inc_base` (`{"error":{"name":"SDK_CONTRACT_ERROR","cause":{...}}`). The field names should also be aligned.

examples/error-handling/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+    // --- Result -------------------------
+    // 2
+    // ------------------------------------
+    // (changes value from 1 to 2)
+    //
+    // is_error = true
+    // Failed transaction
+    // Error:
+    // 0: Error: An error occurred during a `FunctionCall` Action, parameter is debug message.
+    // ExecutionError("Smart contract panicked: {\"error\":{\"error_type\":\"error_handling::MyErrorEnum\",\"value\":\"X\"}}")
```
frol: Looking at `"value": "X"` I feel that you apply the default serde serialization for the enum and this is really hard to maintain, so I suggest we default to ["Adjacently Tagged"](https://serde.rs/enum-representations.html#adjacently-tagged) and maybe even `rename_all` enum variants to SCREAMING_SNAKE_CASE

near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
PolyProgrammist: If it's not `Result` type, it will be a compilation error https://github.com/near/near-sdk-rs/blob/70d2576f84fbc1e21eff4faf0c524eaff9478c07/near-sdk-macros/src/core_impl/info_extractor/visitor.rs#L238

So I believe if there is an additional warning, it doesn't matter

examples/lockable-fungible-token/src/lib.rs
PolyProgrammist: `handle_result` still exist and I would show a warning first rather than turning it off at once without backward compatibility

examples/error-handling/src/lib.rs
PolyProgrammist: Why? `handle_result` is going to be deprecated. I wouldn't showcase just `handle_result` so that the developers don't use it for no reason

examples/error-handling/src/lib.rs
PolyProgrammist: This is just a problem in docs. See they are the same here https://github.com/PolyProgrammist/near-sdk-rs/blob/4bc36d7afb389898a789d11a77b99f6a8c71adf6/examples/error-handling/tests/test_basics.rs#L78-L80

examples/error-handling/src/lib.rs
PolyProgrammist: Do you mean `contract_error` macro should put `#[serde(tag = "type", content = "value", rename_all = "SCREAMING_SNAKE_CASE")]` for all the enums? What if the developer wants to use their own serialization? Currently, there are no error enums in the sdk. MyErrorEnum is from examples.

examples/error-handling/src/lib.rs
frol: @PolyProgrammist Put yourself in the shoes of a new developer using near-sdk-rs to build their app. What do you want to see first? I would love to see the best practices first (see 2), but there is a ton of existing code that look like (1), so it would be nice to showcase that (and ideally recommend (2)). Then, I want to see examples of advanced cases (persist_on_error and custom Result type, which is currently only possible to define with explicit `#[handle_result]`, right?). As a new developer there is no value in learning about `#[handle_result(suppress_warnings)]`; this option only exists for compatibility reasons for existing code, and maybe we don't even need this `suppress_warnings` flag at all, and instead recommend developers to remove the `#[handle_result]` altogether.

UPD: My point is that we should not confuse developers who are reading the example. We should not recommend them bad practices.

near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
frol: It becomes misleading as the compilation error suggests to use `#[handle_result(aliased)]`, but this messages says that the whole `#[handle_result]` attribute is deprecated.

examples/error-handling/src/lib.rs
frol: > Do you mean contract_error macro should put

Yep, exactly (except the "type" and "value" naming).

> What if the developer wants to use their own serialization? Currently, there are no error enums in the sdk.

Let's solve that problem if it will appear. I think good default is always better than flexibility with a bad default.

near-sdk-macros/src/core_impl/code_generator/snapshots/simple_init.snap
```diff
@@ -28,7 +30,8 @@ pub extern "C" fn method() {
         None => ::near_sdk::env::panic_str("Expected input since method has arguments."),
```
frol: Inputs parsing should also return the structured error (panic_err) instead of the panic_str

P.S. search for all `panic_str` in .snap files to see if we miss anything else.

near-sdk/src/lib.rs
```diff
@@ -1322,6 +1322,56 @@ pub use near_sdk_macros::near;
...
+///
+/// # Generated implementations
+///
+/// - [`ContractErrorTrait`] — enables structured error formatting
+/// - `From<ErrorStruct> for BaseError` — allows conversion to a polymorphic error type
+/// - `From<ErrorStruct> for String` — converts the error to its JSON string representation
+///
+/// # Attributes
+///
+/// - `#[contract_error(sdk)]` — sets the error name to `"SDK_CONTRACT_ERROR"` instead of `"CUSTOM_CONTRACT_ERROR"`
```
frol: It is an internal flag, right? Do we expect contract developers to use it?

examples/error-handling/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+
+#[contract_error]
+pub enum MyErrorEnum {
+    X,
+}
+
+#[contract_error(sdk)]
+pub struct MyErrorStruct {
+    x: u32,
+}
```
frol: It is related to my other question, but I believe we should not showcase/encourage to use internal flags.

examples/error-handling/src/lib.rs
```diff
@@ -0,0 +1,155 @@
+// Find all our documentation at https://docs.near.org
+use near_sdk::contract_error;
+use near_sdk::near;
+use near_sdk::BaseError;
+
+#[contract_error]
+pub enum MyErrorEnum {
+    X,
+}
```
frol: Can we bake `#[contract_error]` into `#[near(contract_error)]` macro, so it looks nicer as currently there is an API design misalignment with `#[near(contract_state)]`

near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
PolyProgrammist: I changed the compilation error message saying `#[handle_result]` is deprecated

examples/error-handling/src/lib.rs
PolyProgrammist: I see. The thing is, if I remove `suppress_warnings`, the CI fails cause the warning is produced. I wouldn't change the CI options

So there are couple options:

1. Leave it as is
2. Remove `suppress_warnings`. 
3. Remove examples with handle_result (use implicit Result handling, add more tests for `#[handle_result]` outside examples)

I prefer either 1 or 3. What do you think?
We also talked about updating examples in a separate PR not to overcomplicate the review

near-sdk/src/lib.rs
PolyProgrammist: I would say it's internal, yes. Contract developers should not use it

PR #1171, opened 2024-04-25
[Fix] Remove manual borsh ser/de implementations from 'store' collections.

### LINKED ISSUE #1169: Remove manual implementation of BorshSerialize from `store` collections.
Given that the referenced issue is resolved, can we remove the manual implementation now?

_Originally posted by @frol in https://github.com/near/near-sdk-rs/pull/1164#discussion_r1570778659_

ruseinov:
Fixes #1169 

This needs investigation in terms of bounds to make sure we're not too restrictive, while satisfying the derive requirements.


frol:
Fixed in #1176

PR #1178, opened 2024-05-03
feat: nep330 build info field of contract metadata

dj8yfo:


near-sdk-macros/src/core_impl/contract_metadata/build_info.rs
```diff
@@ -0,0 +1,84 @@
...
+            }
+            Self::UnsetOrEmptySourceSnapshot => {
+                write!(f, "{}{}",
+                    "`CARGO_NEAR_SOURCE_CODE_GIT_URL` is required, when `CARGO_NEAR_BUILD_ENVIRONMENT` is set, ",
+                    "but it's either not set or empty!"
+                )
+            }
+        }
+    }
+}
```
frol, line 43: I don't see a reason to overcomplicate things. Return `Result<Self, String>` from `from_env` and inline the error messages, and call it a day since anyway, you call `to_string` on it to unwrap during the compilation time.

near-sdk-macros/src/core_impl/contract_metadata/build_info.rs
```diff
@@ -0,0 +1,84 @@
...
+            FieldError::EmptyBuildEnvironment
+        );
+        let build_command = {
+            env_field!(
+                build_command,
+                "CARGO_NEAR_BUILD_COMMAND",
+                FieldError::UnsetOrEmptyBuildCommand
+            );
+            let build_command =
+                build_command.split_whitespace().map(|st| st.to_string()).collect::<Vec<_>>();
```
frol, line 68: Ouch, `split_whitespace` is not something we want to take chances with. Let's find a lossless way of passing BUILD_COMMAND. JSON array of strings? So, basically, `serde_json::from_str::<Vec<String>>(build_command)`

near-sdk-macros/src/core_impl/contract_metadata/build_info.rs
```diff
@@ -0,0 +1,84 @@
...
+                    return Err($error);
+                }
+            };
+        }
+
+        env_field!(
+            build_environment,
+            "CARGO_NEAR_BUILD_ENVIRONMENT",
+            FieldError::EmptyBuildEnvironment
+        );
```
frol, line 60: I feel that without the macro, the code would be more readable.

near-sdk-macros/src/core_impl/contract_metadata/mod.rs
```diff
@@ -47,26 +53,37 @@ struct Standard {
...
-            let repo = std::env::var("CARGO_PKG_REPOSITORY").unwrap_or(String::from(""));
-            if !repo.is_empty() {
-                self.link = Some(repo);
-            }
-        }
+        env_field!(
+            self.link,
+            std::env::var("CARGO_NEAR_REPO_LINK_HINT").or(std::env::var("CARGO_PKG_REPOSITORY"))
+        );
+        env_field!(self.version, std::env::var("CARGO_PKG_VERSION"));
```
frol, line 71: Let's remove the macro, it takes more space and only makes it harder to read the code

frol (changes requested):
This is the right direction! I have left a couple of review comments to address

near-sdk-macros/src/core_impl/contract_metadata/build_info.rs
dj8yfo: k, let me try, i vaguely remember the reason why this seemed like a better way to do this, not sure if it still holds true

dj8yfo: k, this was done to remove repeating parts of code, and arguably, make it more readable

dj8yfo: diff in https://github.com/dj8yfo/near-sdk-rs/commit/a34e7ef08f740709b50266768ad2e62993286e6d

frol (approved):
Looks good to me now. We just need to resolve the CI errors

frol (approved):

dj8yfo:
https://github.com/rust-lang/rust-clippy/issues/12967 silenced with https://github.com/near/near-sdk-rs/pull/1178/commits/39baf199d79a675ff6c78e1d1e2c2c56896e18ab

PR #1181, opened 2024-05-07
ci: fix new lints, introduced in 1.78

dj8yfo:
this new lint was silenced , it introduces 30+ locations to fix:  https://rust-lang.github.io/rust-clippy/master/index.html#/multiple_bound_locations ; it has no automatic fix yet with `cargo clippy --fix`

### DIFF
--- a/near-sdk/compilation_tests/schema_derive_invalids.stderr
+++ b/near-sdk/compilation_tests/schema_derive_invalids.stderr
@@ -87,7 +87,7 @@ error[E0277]: the trait bound `Inner: JsonSchema` is not satisfied
             i128
           and $N others
 note: required by a bound in `SchemaGenerator::subschema_for`
- --> $CARGO/schemars-0.8.17/src/gen.rs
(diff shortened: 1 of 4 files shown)


PolyProgrammist:
Looks perfect.
I would also consider fixing multiple_bound_locations instead of applying `#![allow(clippy::multiple_bound_locations)]`.

near-sdk/compilation_tests/schema_derive_invalids.stderr
```diff
@@ -87,7 +87,7 @@ error[E0277]: the trait bound `Inner: JsonSchema` is not satisfied
             i128
           and $N others
 note: required by a bound in `SchemaGenerator::subschema_for`
- --> $CARGO/schemars-0.8.17/src/gen.rs
+ --> $CARGO/schemars-0.8.19/src/gen.rs
```
PolyProgrammist, line 90: If you rerun, you get 20. Please consider either updating to 20 or making the test independent of the version / fixing schemars version

frol (approved):

PR #1133, opened 2024-01-17
doc: added example how to use `promise_yield_create` and `promise_yield_resume`

saketh-are:

PR #1183, opened 2024-05-16
add yield execution host functions

saketh-are:
Yield Execution will be stabilized in nearcore in the 1.40 release.

This PR adds imports for the `promise_yield_create` and `promise_yield_resume` host functions. They have been tested using the smart contract example in #1133.

In the future we can also merge a cleaned-up version of the example contract.

Not sure what's going on with the failing tests; it seems like master is red too if I run them locally?


frol (approved):
The CI issues are resolved in #1181. I will merge main into this branch once #1181 is merged. Once all CI tests pass, I will merge this PR

PR #1186, opened 2024-05-30
fix(macro): use FQDNs when calling contract methods

mitinarseny:
Hey Team!

I've noticed a small inconsistency while working on [`#[access_control]`](https://github.com/Near-One/near-plugins/blob/12d3feec8686a9200173d3c72dbf5159309ded13/near-plugins-derive/src/lib.rs#L54) from [near-plugins](https://github.com/Near-One/near-plugins), which [uses](https://github.com/Near-One/near-plugins/blob/12d3feec8686a9200173d3c72dbf5159309ded13/near-plugins-derive/src/access_controllable.rs#L587-L588) `#[near-sdk::near]` macro internally.

The inconsistency itself can be described as follows:
> The generated code by `#[near_bindgen]`  assumes that all necessary traits were imported into current scope.

To simplify the issue, consider the following example:
```rust
use near_sdk::near;

pub mod other {
    pub trait SomeTrait {
        fn method(&self);
    }
}

#[near(contract_state)]
struct Contract {
    // ..
}

#[near]
impl other::SomeTrait for Contract {
    fn method(&self) {
        // ...
    }
}
```

When you compile it with `cargo build --target wasm32-unknown-unknown`, you get the following error:
```
error[E0599]: no method named `method` found for struct `Contract` in the current scope
  --> controller/src/lib.rs:17:8
   |
5  |         fn method(&self);
   |            ------ the method is available for `Contract` here
...
11 | struct Contract {
   | --------------- method `method` not found for this struct
...
17 |     fn method(&self) {
   |        ^^^^^^ method not found in `Contract`
   |
   = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
   |
1  + use crate::other::SomeTrait;
   |

For more information about this error, try `rustc --explain E0599`.
error: could not compile `defuse-controller-contract` (lib) due to previous error
```

This happens due to `method` is not implemented on `Contract itself` but is a part of `impl other::SomeTrait for Contract`, so it's not visible in the current scope. 
Of course, It can be fixed by adding `use other::SomeTrait` at top of your module, but it's better to avoid it, so it can just work.

The current version of `near-sdk 5.1.0` generates following code:
```rust
#[no_mangle]
pub extern "C" fn method() {
    ::near_sdk::env::setup_panic_hook();
    let contract: Contract = ::near_sdk::env::state_read().unwrap_or_default();
    contract.method();
}
```

This PR introduces a fix for that by using [fully qualified syntax](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#fully-qualified-syntax-for-disambiguation-calling-methods-with-the-same-name) for calling methods that are part of trait implementation. 

Here is the code that this PR generates:
```rust
#[no_mangle]
pub extern "C" fn method() {
    ::near_sdk::env::setup_panic_hook();
    let contract: Contract = ::near_sdk::env::state_read().unwrap_or_default();
    <Contract as other::SomeTrait>::method(&contract);
}
```

### DIFF
--- a/near-sdk-macros/src/core_impl/abi/abi_generator.rs
+++ b/near-sdk-macros/src/core_impl/abi/abi_generator.rs
@@ -324,7 +324,7 @@ mod tests {
             #[handle_result]
             pub fn f3(&mut self, arg0: FancyStruct, arg1: u64) -> Result<IsOk, Error> { }
         };
-        let method_info = ImplItemMethodInfo::new(&mut method, false, impl_type).unwrap().unwrap();
+        let method_info = ImplItemMethodInfo::new(&mut method, None, impl_type).unwrap().unwrap();
         let actual = method_info.abi_struct();
 
         local_insta_assert_snapshot!(pretty_print_fn_body_syn_str(actual));
@@ -339,7 +339,7 @@ mod tests {
             #[handle_result]
             pub fn f3(&mut self, #[serializer(borsh)] arg0: FancyStruct) -> Result<IsOk, Error> { }
         };
-        let method_info = ImplItemMethodInfo::new(&mut method, false, impl_type).unwrap().unwrap();
+        let method_info = ImplItemMethodInfo::new(&mut method, None, impl_type).unwrap().unwrap();
         let actual = method_info.abi_struct();
 
         local_insta_assert_snapshot!(pretty_print_fn_body_syn_str(actual));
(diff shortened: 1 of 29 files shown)


frol (approved):

frol:
@mitinarseny Great job! Thank you for your contribution! It seems to resolve #1084

PR #1187, opened 2024-05-30
test: test ci

fospring:


frol:
@fospring I have fixed CI here: https://github.com/near/near-sdk-rs/pull/1186/commits/4f3785733427016b9c3e3c157620459832efd20e

Issue #1188, opened 2024-06-02
`log!("{x}")` treats input as a raw string


Sliman4:
When you write `log!("{x}")`, it outputs `{x}` literally instead of the value of variable `x` because `log!` is implemented differently if it has only 1 expression:

https://github.com/near/near-sdk-rs/blob/8a908c30ce43fbf1b1866486668a647fbc28b5bd/near-sdk/src/utils/mod.rs#L36-L43

`log!("{x}",)` (with a trailing comma) bypasses this, but I think it would be nice to maintain 100% compatibility with `std::fmt`.

If a string allocation makes a significant difference for some contract, is there a reason why they can't just use `env::log_str` directly? This call will also explicitly mark that this log outputs a lot of data and should not be changed to something that allocates, to avoid situations when at some point you add one `{}` to the log, and gas usage skyrockets for an unknown reason.

frol:
@Sliman4 The macro was implemented before string interpolation became a thing in Rust (stable). I agree with your reasoning. Would you like to prepare a PR?

PR #1189, opened 2024-06-03
fix: Make log macro fully compatible with std::format

Sliman4:
This PR makes `log!` macro 100% compatible with std `format!`, allowing string interpolation while passing just 1 string literal.

I also noticed that this macro uses `format` from local scope, now it's prefixed with `::std` to make sure it can't be accidentally overridden.

Fixes: #1188 

**BREAKING CHANGE**: `let x = "1"; log!(x)` doesn't work anymore, and `log!("str with no formatting")` now allocates, like `format!`. For logging `&str` with no additional formatting and no allocations, use `env::log_str` method instead.


frol (approved):
Ouch, the breaking change is unfortunate, but it is trivial to fix and I hope nobody actually used this syntax in proper contracts

PR #1191, opened 2024-06-03
docs: Fix dead links in markdown files

Sliman4:
- Replaced a relative link to `CONTRIBUTING.md` with an absolute link to this file on github, to make this link work on https://crates.io/crates/near-sdk. It didn't work because crates.io uses README from `/near-sdk/README.md`, which is a symlink to `/README.md`. I replaced it with an absolute link because I'm not sure if it's possible to get relative links to work in both `/README.md` and `near-sdk/README.md` without maintaining 2 copies of the same file.
- Replaced `FungibleTokenCore` (moved to `FungibleToken/Core`) nomicon link with `FungibleToken`. The example implements all Fungible Token standards, so no reason to link to `FungibleToken/Core`, and now it's consistent with the Non-Fungible Token example.
- Replaced `NonFungibleToken/README.html` link with `NonFungibleToken`, and changed the link text from `non-fungible token` to `Non-Fungible Token` to be consistent with the Fungible Token example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,4 +325,4 @@ The minimum supported Rust version is currently `1.72`. There are no guarantees 
 
 ## Contributing
 
-If you are interested in contributing, please look at the [contributing guidelines](CONTRIBUTING.md).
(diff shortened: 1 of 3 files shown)


frol (approved):
@Sliman4 Thank you for fixing these!

PR #1194, opened 2024-06-06
chore: bump MSRV to 1.73.0

dj8yfo:
otherwise was encountering ci error https://github.com/near/near-sdk-rs/actions/runs/9406335158/job/25909553039


frol:
After a review with @ruseinov, we decided to go straight to 1.76.0

PR #1195, opened 2024-06-07
[Testing] Performance tests for 'store' collections.

ruseinov:
This covers basic operations for `store` collections as well as comparison between `Unordered*` and `Iterable*` iteration performance.


ruseinov:
This works fine on other platforms, but fails the osx CI.
Locally on OS X it works, but can be reproduced by limiting the thread amount: `#[tokio::test(flavor = "multi_thread", worker_threads = 2)]`.
My guess is that the machine running mac os x tests just does not have enough CPU.

Potential fixes:
1. Remove parallelism - that'll result in very slow test runs.
2. Disable this on osx CI - it's enough to run those tests on linux.
3. Upgrade the macos GH runner: https://github.blog/2023-03-01-github-actions-introducing-faster-github-hosted-x64-macos-runners/

ruseinov:
> This works fine on other platforms, but fails the osx CI. Locally on OS X it works, but can be reproduced by limiting the thread amount: `#[tokio::test(flavor = "multi_thread", worker_threads = 2)]`. My guess is that the machine running mac os x tests just does not have enough CPU.
> 
> Potential fixes:
> 
> 1. Remove parallelism - that'll result in very slow test runs.
> 2. Disable this on osx CI - it's enough to run those tests on linux.
> 3. Upgrade the macos GH runner: https://github.blog/2023-03-01-github-actions-introducing-faster-github-hosted-x64-macos-runners/

I have fixed this by removing parallelism on account creation. However, it takes under a minute on linux builds and about 10 mins on mac os x. So it might still be interesting to investigate upgrading the runners.
It takes about the same time locally on my mac too, a bit long.

near-sdk/Cargo.toml
```diff
@@ -61,6 +61,11 @@ rand_chacha = "0.3.1"
 near-rng = "0.1.1"
 near-abi = { version = "0.4.0", features = ["__chunked-entries"] }
 symbolic-debuginfo = "12"
+near-workspaces = { git = "https://github.com/near/near-workspaces-rs", branch = "main"}
```
ruseinov, line 64: @frol not the git dependency here. It would actually be nice to release workspaces and fix this.

near-sdk/tests/test-contracts/store/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+            IterableMap => self.iterable_map.flush(),
+            UnorderedMap => self.unordered_map.flush(),
+            UnorderedSet => self.unordered_set.flush(),
+            LookupMap => self.lookup_map.flush(),
+            TreeMap => self.tree_map.flush(),
+            Vector => self.vec.flush(),
+            // No flush method.
+            LookupSet => unimplemented!(),
+        };
+    }
```
frol, line 156: This is pointless as a benchmark test. If there are no changes, there is nothing to flush.

near-sdk/tests/store_performance_tests.rs
```diff
@@ -0,0 +1,268 @@
...
+    let worker = Arc::new(near_workspaces::sandbox().await?);
+    let contract = worker.dev_deploy(include_bytes!("test-contracts/store/res/store.wasm")).await?;
+    let res = contract.call("new").max_gas().transact().await?;
+    assert!(res.is_success());
+    let contract_id = contract.id().clone();
+    let mut account_pool = Vec::new();
+
+    // Generate different accounts to avoid Nonce collisions when executing transactions in parallel.
+    // We could generate this per collection as well, but parallel execution for those proved to be
+    // too much for the sandbox, so we're keeping it simple.
```
frol, line 80: How exactly did it fail for you? I'd like to know when we should re-try lifting this restriction

near-sdk/tests/store_performance_tests.rs
```diff
@@ -0,0 +1,268 @@
...
+    for val in 0..=17 {
```
frol, line 81: This magic constant obviously depends on something below. It is really hard to follow this code.

near-sdk/tests/test-contracts/store/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+// This contract is designed for testing all of the `store` collections.
+pub struct StoreContract {
+    pub iterable_set: store::IterableSet<u32>,
+    pub iterable_map: store::IterableMap<u32, u32>,
+    pub unordered_set: store::UnorderedSet<u32>,
+    pub unordered_map: store::UnorderedMap<u32, u32>,
+    pub tree_map: store::TreeMap<u32, u32>,
+    pub lookup_map: store::LookupMap<u32, u32>,
+    pub lookup_set: store::LookupSet<u32>,
+    pub vec: store::Vector<u32>,
```
frol, line 20: Let's use more complex type than u32 for the load testing, since it is rare to see collections of primitive types in contracts.

near-sdk/tests/store_performance_tests.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    // remove
+    for col in Collection::iter() {
+        let account_pool = account_pool.clone();
+        let contract_id = contract_id.clone();
+        let mut total_gas: u64 = 0;
+        let mut futures = JoinSet::new();
+
+        // Can't use more than 15, because that's how much was inserted.
+        for val in 0..15 {
```
frol, line 241: This could be a const that is shared between insert code and this one rather than a comment

near-sdk/tests/store_performance_tests.rs
```diff
@@ -0,0 +1,268 @@
...
+        assert!(
+            total_gas < NearGas::from_tgas(100).as_gas(),
+            "performance regression: {}",
+            NearGas::from_gas(total_gas)
+        );
+        assert!(
+            total_gas > NearGas::from_tgas(90).as_gas(),
+            "not enough gas consumed: {}, adjust the number of iterations to spot regressions",
+            NearGas::from_gas(total_gas)
+        );
```
frol, line 118: Load so much data that each individual transaction execution consume 90..100TGas. No need to test total among several calls.

The hard limit for a single receipt is 300Tgas, so 90..100Tgas gives us room to catch the cases when it overflows

frol (changes requested):

near-sdk/tests/test-contracts/store/src/lib.rs
ruseinov: This is executed after insert tests.

near-sdk/tests/store_performance_tests.rs
ruseinov: It's failing with nonce errors.

near-sdk/tests/test-contracts/store/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "store"
+version = "0.1.0"
+edition = "2021"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+near-sdk = { git = "https://github.com/near/near-sdk-rs.git", branch = "master", features = ["default", "unstable"]}
```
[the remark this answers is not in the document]
ruseinov: Fixed

near-sdk/tests/test-contracts/store/src/lib.rs
ruseinov: Flush tests don't work even with pre-populated data, because `env::storage_write` does not charge any gas. I'd propose to ditch that for now.

near-sdk/tests/store_performance_tests.rs
ruseinov: In fact this is no longer relevant.

frol (changes requested):
Please, add a test for random access iteration which was the main motivation for these tests in the first place.

near-sdk/tests/test-contracts/store/res/store.wasm
frol, line 1: Let's not commit binary blobs to git. Please, use [near_workspace::compile_project](https://github.com/near/near-workspaces-rs/?tab=readme-ov-file#compiling-contracts-during-test-time)

near-sdk/tests/store_performance_tests.rs
```diff
@@ -0,0 +1,391 @@
...
+    {
+        let txn = account
+            .call(&contract_id, "insert")
+            .args_json((col, DEFAULT_INDEX_OFFSET, max_iterations))
+            .max_gas()
+            .transact()
+            .await;
+
+        let _ = txn?.unwrap();
+    }
```
frol, line 330: Is there something that blocks us from significantly simplifying the code?

```suggestion
    // We only care about Unordered* and Iterable* collections.
    let collection_types = &[
        Collection::UnorderedSet,
        Collection::UnorderedMap,
        Collection::IterableMap,
        Collection::IterableSet,
    ];

    // insert `element_number` elements.
    for col in &collection_types {
        account
            .call(&contract_id, "insert")
            .args_json((col, DEFAULT_INDEX_OFFSET, element_number))
            .max_gas()
            .transact()
            .await??;
    }
```

frol (changes requested):

near-sdk/tests/test-contracts/store/res/store.wasm
ruseinov: This means we should probably address this for the other contracts too in some other PRs.

frol (approved):
It is so much cleaner now! Great job

I will merge it now and will try to address near-workspaces-rs release sooner rather later. I hope it won't block publishing for near-sdk to crates.io

ruseinov:
> It is so much cleaner now! Great job
> 
> I will merge it now and will try to address near-workspaces-rs release sooner rather later. I hope it won't block publishing for near-sdk to crates.io

Thanks for great reviews!

frol (approved):

PR #1196, opened 2024-06-07
[Fix] Clap downgrade.

ruseinov:


frol (approved):

frol:
It is not needed anymore after #1197

PR #1198, opened 2024-06-10
doc: increase MSRV to 1.76 in readme.md

dj8yfo:


frol (approved):
Just for the record, there are no features that near-sdk-rs depends on in the newer Rust versions, but the transitive dependencies move on and it becomes harder to test it on 1.72 release. If you use Rust 1.72, there are good chances that near-sdk-rs will still work for your project that has older versions of the dependencies in Cargo.lock

PR #1202, opened 2024-06-18
ci: 1.79 fixes (replace `macos-latest stable` with `macos-latest 1.78`)

dj8yfo:
this passes if `macos-latest` is temporarily commented out https://github.com/near/near-sdk-rs/actions/runs/9566693799 ;
otherwise, `macos-latest` is having some serious trouble on 1.79

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest]
-        toolchain: [1.78.0, 1.76.0]
+        toolchain: [stable, 1.76.0]
     steps:
       - uses: actions/checkout@v3
       - name: "${{ matrix.toolchain }} with rustfmt, and wasm32"
@@ -23,6 +23,7 @@ jobs:
           toolchain: ${{ matrix.toolchain }}
           default: true
           target: wasm32-unknown-unknown
+          components: clippy, rust-docs
       - uses: Swatinem/rust-cache@v1
 #      - name: Downgrade dependencies
 #        run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest]
-        toolchain: [stable, 1.76.0]
+        toolchain: [1.78.0, 1.76.0]
     steps:
       - uses: actions/checkout@v3
       - name: "${{ matrix.toolchain }} with rustfmt, and wasm32"
@@ -23,7 +23,6 @@ jobs:
           toolchain: ${{ matrix.toolchain }}
           default: true
           target: wasm32-unknown-unknown
-          components: clippy, rust-docs
       - uses: Swatinem/rust-cache@v1
 #      - name: Downgrade dependencies
 #        run: |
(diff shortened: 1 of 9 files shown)


ruseinov:
I have this already fixed in my working branch.

ruseinov:
I'm looking into this in another branch and locally, if I make progress - I will let you know.

ruseinov:
This is not reproducible locally on my mac. Could it be the cache?

dj8yfo:
@ruseinov , what's your processor type? It might be some wicked bug of 1.79 affecting only `aarch64-*-*`, or it might be not

ruseinov:
> @ruseinov , what's the processor type of your machine? It might be some wicked bug of 1.79 affecting only `aarch64-*-*`, or it might be not

I have an m2-pro, so that should have popped up.

ruseinov:
I have tried investigating this with `macos-13` - that yields "test panicked while processing panic". There might indeed be something wrong with the setup somewhere.

ruseinov:
I guess you have already talked to @frol. Perhaps we should pin the version to 1.78 as a temp fix, either via `rust-toolchain.toml` or just via CI.

dj8yfo:
@ruseinov created an [issue](https://github.com/rust-lang/rust/issues/126643), as my somewhat limited research hasn't been particularly useful

ruseinov:
> @ruseinov created an [issue](https://github.com/rust-lang/rust/issues/126643), as my somewhat limited research hasn't been particularly useful

That’s fine, let’s so 1.78 to fix master for now. And solve the issue later, unless it solves itself with the next version. Can’t be debugging machine-specific failures with actual access to the machine.

near-sdk-macros/src/core_impl/info_extractor/arg_info.rs
```diff
@@ -18,12 +18,15 @@ pub enum BindgenArgType {
 /// A single argument of a function after it was processed by the bindgen.
 pub struct ArgInfo {
     /// Attributes not related to bindgen.
+    #[allow(unused)]
```
frol, line 21: Are these really unused or is it a clippy bug?

.github/workflows/test.yml
```diff
@@ -8,26 +8,27 @@ env:
...
   test:
-    runs-on: ${{ matrix.platform }}
-    name: "${{ matrix.platform }} ${{ matrix.toolchain }}"
+    runs-on: ${{ matrix.platform.os }}
+    name: "${{ matrix.platform.os }} ${{ matrix.platform.rs }}"
     strategy:
       matrix:
-        platform: [ubuntu-latest, macos-latest]
-        toolchain: [stable, 1.76.0]
+        platform: [{os: ubuntu-latest, rs: 1.76.0}, {os: ubuntu-latest, rs: stable}, {os: macos-latest, rs: 1.76.0}, {os: macos-latest, rs: 1.78.0}]
```
frol, line 15: I think this should work and it is easier to read and maintain:

```suggestion
        platform:
          - os: ubuntu-latest
            rs: 1.76.0
          - os: ubuntu-latest
            rs: stable
          - os: macos-latest
            rs: 1.76.0
          - os: macos-latest
            rs: 1.78.0
```

frol (dismissed):

near-sdk-macros/src/core_impl/info_extractor/arg_info.rs
dj8yfo: as far as go-to-references have told me, these are unused. looks like the code was copied from elsewhere, 
 but only some of the fields are used in the context of  `near-sdk-macros`

dj8yfo: triple-checked these; they are only created inside of `::new` constructors, and not touched by other code thereafter. So, as far as tooling can tell, these are unused

frol (approved):

PR #1206, opened 2024-06-19
feat(near-contract-standards): export ext_nft_*

mitinarseny:
This PR exports `ext_nft_*` modules, so NFT interfaces can be re-used in dependent contracts


frol (approved):
@mitinarseny Good point! Would you like to also add documentation to feature it?

P.S. I won't block this PR on documentation request, so I go head to merging it now

PR #1207, opened 2024-06-20
chore: Updated near-* dependencies to 0.23

akorchyn:
* Updated dependencies for near-* packages to 0.23 version.
* Fixed compilation


haski007:
Good job, these deps updates were needed so much

frol (approved):

frol:
I don't think these are breaking changes for the SDK itself. In the worst case, these could be breaking changes for the unit-tests, but that is also not necessarily true.

I have tested it locally on https://github.com/NEAR-DevHub/neardevhub-contract, and there is no regression with this PR, so I won't bump the major version to 6.x

PR #1208, opened 2024-06-24
feat(near-contract-standards): export ext_storage_management

aleksuss:
The PR exports the `ext_storage_management`, which allows the creation of promises based on the `StorageManagement` trait.


frol (approved):

frol:
@aleksuss Great! Thanks for contributing it.

Just for future reference, we slowly getting all near-contract-standards to receive the high-level APIs, here was the previous one: #1206

PR #1249, opened 2024-10-14
fix: populate default values for contract_metadata

mitinarseny:
Add `#[darling(default)]` attribute for `link` and `version` to make it possible to derive them from env:
```rust
#[near(contract_metadata(standard(standard = "nep141", version = "1.0.0")))
struct Contract {}
```

As current implementation returns an error:
```
error: expected an expression
 │   --> src/lib.rs:49:1
 │    |
 │ 49 | / #[near(
 │ 50 | |     contract_state,
 │ 51 | |     contract_metadata(standard(standard = "nep141", version = "1.0.0"))
 │ 52 | | )]
 │    | |__^
 │    |
 │    = note: this error originates in the attribute macro `near` (in Nightly builds, run with -Z macro-backtrace for more info)
 │ 
 │ error[E0425]: cannot find value `CONTRACT_SOURCE_METADATA` in this scope
 │   -->src/lib.rs:49:1
 │    |
 │ 49 | / #[near(
 │ 50 | |     contract_state,
 │ 51 | |     contract_metadata(standard(standard = "nep141", version = "1.0.0"))
 │ 52 | | )]
 │    | |__^ not found in this scope
 │    |
 │    = note: this error originates in the attribute macro `::near_sdk::near_bindgen` (in Nightly builds, run with -Z macro-backtrace for more info)
```

### DIFF
--- a/near-sdk-macros/src/core_impl/contract_metadata/mod.rs
+++ b/near-sdk-macros/src/core_impl/contract_metadata/mod.rs
@@ -13,7 +13,9 @@ struct MacroConfig {
 
 #[derive(serde::Serialize, Default, FromMeta)]
 pub(crate) struct ContractMetadata {
+    #[darling(default)]
     version: Option<String>,
+    #[darling(default)]
     link: Option<String>,
 
     #[darling(multiple, rename = "standard")]


akorchyn:
Hey, first of all, thanks for the contribution.

Please provide example to see how it helps. It would be nice to have a test repo with test contract to make review faster

dj8yfo (commented):
@frol, judging from https://github.com/Shitzu-Apes/validator-rewards/pull/2, logic has to be changed to such, that if `std::env::var("NEP330_LINK")` is set, then it should override value from `darling`. For such cases, when a legacy contract is updated with a `link` attribute lingering somewhere in `src/lib.rs`, forgotten or not. What do you think?

dj8yfo:
@mitinarseny can you add a unit test, demonstrating how parsing changes prior and after the change? It appears that both with `darling(default)` and without it a missing `link` or `version` are parsed as `None`, so there's not much changed with this pr. A unit test would help to clarify this

PR #1284, opened 2024-12-20
chore: updates near-workspaces to 0.16 version

akorchyn:
@race-of-sloths


dj8yfo (approved):
@race-of-sloths score 2

dj8yfo (changes requested):
some were missed

dj8yfo (approved):
nope, all good

PR #1332, opened 2025-03-21
docs: the discord link didn't work

Olexandr88:


akorchyn:
Thanks for submitting PR. Something is wrong with the SSL certificate (the HTTP version works).
So I would rather expedite that

PR #1338, opened 2025-03-26
docs(readme): update LICENSE-MIT

Olexandr88:


akorchyn:
Thanks for the PR, but it can be ignored and doesn't need to be updated. See [why](https://hynek.me/til/copyright-years/)

Olexandr88:
I understand, thanks for the info.

PR #1363, opened 2025-06-04
Add error information on deserialization errors

BonelessImpl:
There are many situations when error information is needed when deserialization fails. Currently, the error just states that deserialization failed without any hint to what happened. I added the error to the printed message. 

This is motivated by having received complaints from users, stating that the error, in its current state, is not helpful enough.

---

for 
```rust
        .args(b"{\"greeting\": \"Hello World!\",}".to_vec())
```
this pr changes error from 
```bash
Smart contract panicked: Failed to deserialize input from JSON.
```
to
```bash
Smart contract panicked: Failed to deserialize input from JSON. Error: `trailing comma at line 1 column 29`
```

for 
```rust
        .args(b"{\"wrong_key\": \"Hello World!\"}".to_vec())
```
this pr changes error from
```bash
Smart contract panicked: Failed to deserialize input from JSON.
```
to
```bash
Smart contract panicked: Failed to deserialize input from JSON. Error: `missing field `greeting` at line 1 column 29`
```

### DIFF
--- a/near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
+++ b/near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
@@ -122,9 +122,7 @@ impl ImplItemMethodInfo {
                         Some(input) => match ::near_sdk::serde_json::from_slice(&input) {
                             Ok(deserialized) => deserialized,
                             Err(e) => {
-                                let mut err = String::from("Failed to deserialize input from JSON. Error: ");
-                                err.push_str(&e.to_string());
-                                ::near_sdk::env::panic_str(&err);
+                                ::near_sdk::env::panic_str(&format!("Failed to deserialize input from JSON. Error: `{e}`"));
                             }
                         },
                         None => ::near_sdk::env::panic_str("Expected input since method has arguments.")
@@ -135,9 +133,7 @@ impl ImplItemMethodInfo {
                         Some(input) => match ::near_sdk::borsh::BorshDeserialize::try_from_slice(&input) {
                             Ok(deserialized) => deserialized,
                             Err(e) => {
-                                let mut err = String::from("Failed to deserialize input from Borsh. Error: ");
-                                err.push_str(&e.to_string());
-                                ::near_sdk::env::panic_str(&err);
+                                ::near_sdk::env::panic_str(&format!("Failed to deserialize input from Borsh. Error: `{e}`"));
                             }
                         },
                         None => ::near_sdk::env::panic_str("Expected input since method has arguments.")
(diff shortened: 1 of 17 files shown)


dj8yfo:
@BonelessImpl i like this change. 
@frol has previously been reluctant to add this due to reason of increased contract sizes.

dj8yfo:
trying to use https://doc.rust-lang.org/std/string/struct.String.html#method.push_str instead of `format!` might have a lesser impact on contract size, but might need to be tested, not sure about that

BonelessImpl:
> trying to use https://doc.rust-lang.org/std/string/struct.String.html#method.push_str instead of `format!` might have a lesser impact on contract size, but might need to be tested, not sure about that

This may not be a simple thing to do. Consider the following:

- A String is simply a `Vec<u8>` with runtime checks
- Appending a slice to a `Vec<u8>` may or may not introduce a new allocation, depending on whether the `Vec` already has enough allocated memory
- To minimize the number of allocations, we have to first allocate the `Vec`, and then add the u8 slice to it.
- To allocate the `Vec` only once, we have to pre-calculate all sizes, reserve the `Vec`, then push.
- But, `format` should be already optimized to do these calculations to minimize allocations.

I highly doubt `format` does more than one allocation.

BonelessImpl:
I'm thinking and I'm still not sure.

I can pre-calculate sizes, but this will involve writing more code to calculate things. So it will increase the contract size because there will be more code. On the other hand, it might save some gas on execution. I don't know which is better.

near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs
```diff
@@ -121,7 +121,9 @@ impl ImplItemMethodInfo {
                     match ::near_sdk::env::input() {
                         Some(input) => match ::near_sdk::serde_json::from_slice(&input) {
```
dj8yfo, line 122: if this is decided to be proceeded with, 
 
`near-sdk-macros` has to be grepped for `near_sdk::serde_json::from_slice` and checked how error is formatted there , as there's at least one more spot, where deserialize input is used - in `#[callback_...]` attributes (where https://docs.rs/near-sdk/5.14.0/near_sdk/env/fn.promise_result.html is used instead of https://docs.rs/near-sdk/5.14.0/near_sdk/env/fn.input.html). 

it appears to me the error should be printed uniformly in all such spots

BonelessImpl: Done

BonelessImpl:
Apparently me commenting `/compare` has no effect 😅...

dj8yfo:
@BonelessImpl please revert to `format!`. Looks like size reduction effect i had observed earlier was about formatting integers/floats into strings, not strings.

BonelessImpl:
> @BonelessImpl please revert to `format!`. Looks like size reduction effect i had observed earlier was about formatting integers/floats into strings, not strings.

Done.

BonelessImpl:
Hi. Is there any problem with this PR, or is it good for merging?

PR #1367, opened 2025-06-16
chore: expand cfgs' compilation error with complete reason of the error

dj8yfo:
this one is a small continuation on #1361 
image


dj8yfo:
@race-of-sloths include

frol (approved):

frol:
@race-of-sloths score 3

PR #1370, opened 2025-07-01
chore: non-contract-usage feature flag

frolvanya:


frol (approved):
@race-of-sloths score 2

PR #1376, opened 2025-07-13
feat(sft): reference implementation

mitinarseny:
Reference implementation of Sharded Fungible Token with deterministic addresses

### DIFF
--- a/near-contract-standards/Cargo.toml
+++ b/near-contract-standards/Cargo.toml
@@ -25,6 +25,8 @@ near-sdk = { path = "../near-sdk", default-features = false, features = [
 [features]
 default = []
 abi = ["near-sdk/__abi-generate"]
+# Compile Sharded Fungible Token wallet-contract implementation
+sharded_fungible_token_wallet_impl = []
 
 [package.metadata.docs.rs]
 features = ["abi"]
--- a/near-contract-standards/src/lib.rs
+++ b/near-contract-standards/src/lib.rs
@@ -3,6 +3,7 @@
 #![allow(clippy::needless_lifetimes)]
 
 pub mod fungible_token;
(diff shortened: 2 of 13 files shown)


examples/sharded-fungible-token/wallet/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+        notify: Option<TransferNotification>,
+        refund_to: Option<AccountId>,
+    ) -> PromiseOrValue<U128> {
+        let mut deposit_left = env::attached_deposit();
+        require!(deposit_left >= NearToken::from_yoctonear(1), Self::ERR_INSUFFICIENT_DEPOSIT);
+
+        let caller = env::predecessor_account_id();
+        // verify that the caller is a valid wallet-contract or the minter
+        require!(
+            caller == *self.minter_id || caller == self.sft_wallet_account_id_for(&caller),
```
jakmeier: Hm, shouldn't this be `self.sft_wallet_account_id_for(&sender_id)`?

```suggestion
            caller == *self.minter_id || caller == self.sft_wallet_account_id_for(&sender_id),
```

mitinarseny: Oops, thanks for noticing, fixed!

ilblackdragon:
Exciting progress here.

Given this is introducing new token standard - we should probably have the standard for that in NEP as well.
As part of that standard one request from my side is to think through how we can have private balances and transactions in the new standard.

Onchain privacy is becoming more important, especially with more adoption of crypto as payment rails. Given this new standard will require effectively full ecosystem overhaul - wallets, tooling, exchanges, it's a good time to think through all of the changes we would like to make.

PR #1387, opened 2025-08-22
refactor: Use CryptoHash wherever applicable and other small papercut fixes

frol:


near-sdk/src/promise.rs
```diff
@@ -368,14 +370,14 @@ impl Promise {
...
-    /// let code_hash = vec![0u8; 32]; // 32-byte hash
+    /// let code_hash = [0u8; 32]; // 32-byte hash (CryptoHash)
     /// Promise::new("alice.near".parse().unwrap())
     ///     .create_account()
     ///     .transfer(NearToken::from_yoctonear(1000))
     ///     .use_global_contract(code_hash);
     /// ```
-    pub fn use_global_contract(self, code_hash: Vec<u8>) -> Self {
-        self.add_action(PromiseAction::UseGlobalContract { code_hash })
+    pub fn use_global_contract(self, code_hash: impl Into<CryptoHash>) -> Self {
```
frol, line 385: Notes:
1. `CryptoHash` is [u8; 32], so it is a better type than a vector of arbitrary length, and it is also aligned with `sha256_array` return type.
2. `impl Into<>` this allows to pass Base58CryptoHash without a need to unwrap it into CryptoHash.

near-sdk/src/promise.rs
```diff
@@ -338,8 +340,8 @@ impl Promise {
     ///     .transfer(NearToken::from_yoctonear(1000))
     ///     .deploy_global_contract(code);
     /// ```
-    pub fn deploy_global_contract(self, code: Vec<u8>) -> Self {
-        self.add_action(PromiseAction::DeployGlobalContract { code })
+    pub fn deploy_global_contract(self, code: impl Into<Vec<u8>>) -> Self {
```
frol, line 349: Note: this allows to pass Base64VecU8 without a need to unwrap it into Vec<u8>.

examples/factory-contract-global/src/lib.rs
```diff
@@ -27,18 +27,14 @@ impl GlobalFactoryContract {
         code: Base64VecU8,
         account_id: AccountId,
     ) -> Promise {
-        // Store the code hash for later reference
-        let code_bytes: Vec<u8> = code.into();
-        let code_hash_vec = env::sha256(&code_bytes);
-        let code_hash: CryptoHash = code_hash_vec.try_into().unwrap();
+        let code_hash = env::sha256_array(&code);
         self.deployed_global_contracts.insert(name.clone(), code_hash);
-        self.global_contract_deployers.insert(name, account_id.clone());
```
PolyProgrammist, line 35: I don't get it, what's the reason for the change? 

With this, the result of `get_global_contract_hash` and `get_global_contract_deployer` methods changes. Why would it?

examples/factory-contract-global/src/lib.rs
```diff
@@ -100,17 +91,15 @@ impl GlobalFactoryContract {
     }
 
     /// List all deployed global contracts
-    pub fn list_global_contracts(&self) -> Vec<(String, Base58CryptoHash, AccountId)> {
-        self.deployed_global_contracts
+    pub fn list_global_contracts(&self) -> Vec<(String, String)> {
```
PolyProgrammist: So before you could get the hash and the account id. Now you can't. And types do mess.
Is the reason for that to distinguish between contracts deployed by hash or contracts deployed by account id? Maybe add boolean value for that instead?

near-sdk/src/environment/env.rs
```diff
@@ -2768,8 +2799,8 @@ mod tests {
 
         let promise_index = super::promise_batch_create(&"alice.near".parse().unwrap());
         let code = vec![0u8; 100]; // Mock contract bytecode
-        let code_hash = vec![0u8; 32]; // Mock 32-byte hash
```
PolyProgrammist, line 2771: why do you remove `vec!` here but not for `code` variable?

examples/factory-contract-global/src/lib.rs
frol: I find the original version misleading - it felt like the global contract can be referenced by either account id or hash regardless of how it was deployed (registered), but in reality you can only UseGlobalContract by hash if it was DeployGlobalContract by hash, and similarly for account id.

examples/factory-contract-global/src/lib.rs
frol: Good point. I tried to do minimal changes to the code, but in reality I should have just introduced two separate view methods and call it a day instead of trying to merge those together.

near-sdk/src/environment/env.rs
frol: Well, it was required to change the `code_hash` to be an array of exactly 32 bytes, but for `code` it is fine to be any `&[u8]`, so both would work.

examples/factory-contract-global/src/lib.rs
PolyProgrammist: I would rename the fields then. Maybe `deployed_global_contracts` -> `global_contracts_deployed_by_hash`, `global_contract_deployers` -> `global_contracts_deployed_by_account_id`

frol: Sounds good. Alternative naming suggestion: `global_contracts_registered_by_code_hash`, `global_contracts_registered_by_account_id`

PR #1393, opened 2025-10-16
feat(near-sdk): use #[serde_as] for #[near(serializers = [json])]

mitinarseny:
This PR lets contracts to use [serde_with](http://docs.rs/serde_with) crate by automatically adding `#[serde_as]` attribute when `json` serializer is enabled in `#[near]` macro.

[serde_with](http://docs.rs/serde_with) leverages Rust type system and lets to avoid introducing newtypes (such as [`U128`](https://docs.rs/near-sdk/latest/near_sdk/json_types/struct.U128.html)) to just redefine how to serialize/deserialize the underlying type. Newtype approach is very cumbersome and inconvenient to work with, especially when dealing with nested collections.

Right now smart-contract devs are forced to use some [dirty hacks](https://github.com/near/intents/blob/8e6b29e346899fe92bb4d90db9cda2768136b17a/core/src/intents/tokens.rs#L17-L32) in order to achieve it, as well as introduce fake features that are required for `cargo near build` to work due to [manual features activation](https://github.com/near/cargo-near/blob/723097bf82691bc2cd9dad4ff2f85303d4e6ec97/cargo-near-build/src/near/abi/generate/mod.rs#L61) via cli for `near-sdk`.

### DIFF
--- a/near-sdk/src/lib.rs
+++ b/near-sdk/src/lib.rs
@@ -352,7 +352,7 @@ compile_error!(
 /// as if `#[serde_as]` is added to the type (which is what actually happens under the hood of `#[near(serializers = [json])]` implementation).
 ///
 /// ```
-/// # use std::{collections::BTreeMap, u128};
+/// # use std::{collections::BTreeMap};
 /// use near_sdk::{
 ///     near,
 ///     serde_json::json,


PolyProgrammist (changes requested):
Hi! Could you please add a test that shows the benefit from the feature

PolyProgrammist (approved):

PR #1399, opened 2025-11-07
feat(near-sdk-macros): `#[near(contract_state(key = b"CUSTOM"))]`

mitinarseny:
This PR adds support for custom key in `#[near(contract_state)]` macro.  This becomes especially useful for [sharded contracts](https://github.com/near/NEPs/blob/master/neps/nep-0616.md#deterministic-accountids) where we want to optimize for storage size to, hopefully, fit into [ZBA limits](https://github.com/near/NEPs/blob/master/neps/nep-0448.md):
```rust
#[near(contract_state(key = b""))] // reduces storage by 5 bytes
pub struct ShardedContract {
    // ...
}
```

PS: The default key is kept to be `b"STATE"`.

### DIFF
--- a/near-sdk/tests/store_performance_tests.rs
+++ b/near-sdk/tests/store_performance_tests.rs
@@ -1,6 +1,6 @@
 // As wasm VM performance is tested, there is no need to test this on other types of OS.
 // This test runs only on Linux, as it's much slower on OS X due to an interpreted VM.
-// #![cfg(target_os = "linux")]
+#![cfg(target_os = "linux")]
 
 use near_account_id::AccountId;
 use near_gas::NearGas;


examples/adder/src/lib.rs
```diff
@@ -5,7 +5,7 @@ use near_sdk::{env, near, Promise};
 pub struct Pair(u32, u32);
 
 #[derive(Default)]
-#[near(serializers=[borsh, json], contract_state)]
+#[near(serializers=[borsh, json], contract_state(key = b""))]
```
frol: Added is one of the simplest examples, please, don't complicate it. Let's revert the change here. If you want to showcase it, please, make a new example.

mitinarseny:
@frol for some reason, `lazy:flush` store performance test started to consume less gas, so I've just adjusted number of operations for this test to pass

mitinarseny:
Hm, I've tried annotating `state_read()`, `state_write()` and `state_exists()` with `#[inline(always)]` and got the same result: gas consumption is still less than it was before... But that's the good news, right? 😅

PR #1403, opened 2025-11-12
feat: Add ergonomic API for global contracts

r-near:
Replaces the 4-function global contract API with a cleaner 2-function API.

## New API

```rust
// Publish contract code to global registry
publish_contract(code: Vec<u8>, account_id: Option<AccountId>)

// Deploy from published code (accepts hash OR account)
deploy_from_published(reference: impl IntoGlobalContractRef)
```

## Migration

Old functions are deprecated but still work for backwards compatibility:
- `deploy_global_contract()` → `publish_contract(code, None)`
- `deploy_global_contract_by_account_id()` → `publish_contract(code, Some(account_id))`
- `use_global_contract()` → `deploy_from_published(code_hash)`
- `use_global_contract_by_account_id()` → `deploy_from_published(account_id)`

### DIFF
--- a/near-sdk/src/promise.rs
+++ b/near-sdk/src/promise.rs
@@ -499,12 +499,12 @@ impl Promise {
     }
 
     #[cfg(feature = "global-contracts")]
-    /// Deploy an account using previously published global contract code.
+    /// Deploy previously published global contract code to an account.
     ///
-    /// This is a high-level API that creates a new account and deploys it using code that
-    /// was previously published to the global contract registry via [`Self::publish_contract_by_hash`]
-    /// or [`Self::publish_contract_by_account`]. This saves on deployment costs since the code
-    /// doesn't need to be uploaded again.
+    /// This function deploys code that was previously published to the global contract registry via
+    /// [`Self::publish_contract_by_hash`] or [`Self::publish_contract_by_account`] to the specified account.
+    /// Note: This function does **not** create a new account. If you need to create an account, use
+    /// `.create_account()` before calling this method, as shown in the examples.
     ///
     /// The `reference` parameter accepts either:
     /// - A code hash ([`CryptoHash`]) - deploys a pinned version of the contract; the deployed


near-sdk/src/promise.rs
```diff
@@ -30,6 +30,66 @@ impl Allowance {
     }
 }
 
+#[cfg(feature = "global-contracts")]
+/// Reference to a global contract, either by code hash or by the account that published it.
+///
+/// This enum is used internally by [`Promise::deploy_from_published`] to support
+/// ergonomic API where you can pass either a hash or an account ID.
+enum GlobalContractRef {
+    Hash(Vec<u8>),
```
mitinarseny: Please, avoid using `Vec` for fixed-length arrays, especially for public APIs:
```suggestion
    CodeHash(CryptoHash),
```
This type is likely to be exposed as public API as part of NEP-616 impl, which will account for its borsh representation, which is another point towards using an array.

near-sdk/src/promise.rs
```diff
@@ -367,9 +479,56 @@ impl Promise {
...
+    ///     .deploy_from_published(code_hash)
+    ///     .detach();
+    ///
+    /// // Deploy using publisher account (more ergonomic!)
+    /// Promise::new("bob.near".parse().unwrap())
+    ///     .create_account()
+    ///     .deploy_from_published(publisher)
+    ///     .detach();
+    /// ```
+    pub fn deploy_from_published(self, reference: impl IntoGlobalContractRef) -> Self {
```
mitinarseny: Why not just use `impl Into<GlobalContractRef>`?
This will automatically let you use `GlobalContractRef` itself as well

examples/factory-contract-global/src/lib.rs
```diff
@@ -38,7 +38,7 @@ impl GlobalFactoryContract {
             .create_account()
             .transfer(env::attached_deposit())
             .add_full_access_key(env::signer_account_pk())
-            .deploy_global_contract(code_bytes)
+            .publish_contract(code_bytes, None)
```
frol: 1. I don't like using Option to signal if the contract will be published by account id or by hash.
2. Some() must only receive the `current_account_id` or new account if the new account is created, but in reality it is not even passed to the host, so it only misleads developers getting them thinking that they may deploy it for a different account id here
3. `publish_contract` without the context of global contract is not easy to understand in my opinion. However, this naming highlighted the problem of this example - it is a meaningless example as I cannot think of a reason why anyone would want to create a new account to only register the global contract by hash.

r-near: For the account ID, we could hide that complexity from the user. But that might be too much "magic". 

I agree the example is a bit misleading.

Are you thinking of two `publish_contract` methods, then?

frol: > For the account ID, we could hide that complexity from the user. But that might be too much "magic".

Can you elaborate on the "magic" part? What is exactly the magical part? Note that the regular `deploy` action does not take account ID.

> Are you thinking of two `publish_contract` methods, then?

Yes, I think two separate methods are easier to grasp.

P.S. Please, resolve the merge conflicts

near-sdk/src/promise.rs
```diff
@@ -401,9 +496,57 @@ impl Promise {
...
+    /// Deploy an account using previously published global contract code.
+    ///
+    /// This is a high-level API that creates a new account and deploys it using code that
+    /// was previously published to the global contract registry via [`Self::publish_contract_by_hash`]
+    /// or [`Self::publish_contract_by_account`]. This saves on deployment costs since the code
+    /// doesn't need to be uploaded again.
+    ///
+    /// The `reference` parameter accepts either:
+    /// - A code hash ([`CryptoHash`]) - if you know the exact hash of the published code
+    /// - An account ID ([`AccountId`]) - if the code was published with an associated account
```
frol: This documentation is misleading. Those are two completely different usecases when the global contract is deployed by hash (pinned version) vs by account (it effectively enables immediate auto-update to the accounts that use this global contract). It is not about whether I know the contract code hash or not.

near-sdk/src/promise.rs
```diff
@@ -401,9 +496,57 @@ impl Promise {
...
+    /// let code_hash: CryptoHash = [0u8; 32]; // Hash from previously published code
+    /// let publisher: AccountId = "multisig-template.near".parse().unwrap();
+    ///
+    /// // Deploy using code hash
+    /// Promise::new("alice.near".parse().unwrap())
+    ///     .create_account()
+    ///     .deploy_from_published(code_hash)
+    ///     .detach();
+    ///
+    /// // Deploy using publisher account (more ergonomic!)
```
frol: "more ergonomic" is misleading, see above.

near-sdk/src/promise.rs
```diff
@@ -365,9 +398,65 @@ impl Promise {
...
+    /// # Examples
+    /// ```no_run
+    /// use near_sdk::{Promise, NearToken};
+    ///
+    /// let code = vec![0u8; 100]; // Contract bytecode
+    ///
+    /// // Publish contract code - can be referenced by its hash later
+    /// Promise::new("alice.near".parse().unwrap())
+    ///     .create_account()
+    ///     .publish_contract_by_hash(code)
```
frol: This is a bad example as there is zero reasons to create a separate account just to register a global contract by hash - you can register it under any account, including the current one.

near-sdk/src/promise.rs
```diff
@@ -365,9 +398,65 @@ impl Promise {
...
+    /// # Examples
+    /// ```no_run
+    /// use near_sdk::{Promise, NearToken};
+    ///
+    /// let code = vec![0u8; 100]; // Contract bytecode
+    ///
+    /// // Publish and associate with account - others can reference by account ID
+    /// Promise::new("multisig-template.near".parse().unwrap())
+    ///     .create_account()
+    ///     .publish_contract_by_account(code)
```
frol: I tried to look at this example without any prior context of global contracts and I find it impossible to understand. Let's brainstorm it more in a chat.

near-sdk/src/promise.rs
```diff
@@ -401,9 +496,57 @@ impl Promise {
...
+    /// // Deploy using code hash
+    /// Promise::new("alice.near".parse().unwrap())
+    ///     .create_account()
+    ///     .deploy_from_published(code_hash)
+    ///     .detach();
+    ///
+    /// // Deploy using publisher account (more ergonomic!)
+    /// Promise::new("bob.near".parse().unwrap())
+    ///     .create_account()
+    ///     .deploy_from_published(publisher)
```
frol, line 532: It is not a blocker for me, but it makes me uneasy:

1. It feels that it is too easy to pass CryptoHash instead of AccountId or vise versa by mistake (e.g. if I have it as part of some other struct or collection: `deploy_from_published(latest_contract)`)
2. No API symmetry - There are two separate methods for publishing/registering the global contracts, and they have quite different usecases, but there is only a single method to use/deploy it

examples/factory-contract-global/src/lib.rs
```diff
@@ -25,7 +25,7 @@ impl GlobalFactoryContract {
         let code_hash = env::sha256_array(&code);
         self.global_contracts_registered_by_code_hash.insert(name.clone(), code_hash);
 
-        Promise::new(env::current_account_id()).deploy_global_contract(code)
+        Promise::new(env::current_account_id()).publish_contract_by_hash(code.into())
```
frol: I don't think you need `.into()` or explicit `CryptoHash::from()` in this file anymore (I believe it is a merge conflict artifact).

near-sdk/src/promise.rs
r-near: Yeah I think you have a point here. The two operations are semantically very different
- By hash = immutable, pinned version (security-focused)
- By account = mutable, auto-updating (convenience-focused)

Hiding this distinction behind a single method with implicit conversion makes it too easy to accidentally pick the wrong one... Like which is this? Have to check the type of `reference`
```rust
.deploy_from_published(reference)
```


We could do something explicit, like:
```rust
deploy_from_hash(code_hash: CryptoHash)
deploy_from_account(account_id: AccountId)
 ```
 
This would match the publish APIs:
```rust
publish_contract_by_hash(code)
publish_contract_by_account(code)
```

Would be a 1:1 mapping with the canonical actions. Not sure how I feel about it yet, though. Thoughts?

frol: `deploy_from_hash` sounds like the old joke of "here is a hash of the movie, please, unarchive it". Also, I think it could be misleading unless you are in the context of global contracts - which hash do I need to provide? Transaction hash? Block hash?

Naming is hard, but neither `publish_contract_by_*` nor `deploy_from_*` help here. It is just a slightly better naming to the existing one, IMO, but it creates misalignment with the protocol naming, so the end result is worse or equal to what we already have.

Unfortunately, I don't have the capacity to come up with alternatives yet.

r-near: I agree. None of this seems **better** than what we currently have, it's all just...about the same.

PR #1410, opened 2025-11-26
fix(macros): implement #[private] attribute for init methods

### LINKED ISSUE #1409: Private init methods lack private code
Code:
```rs
use near_sdk::{near, PanicOnDefault};

#[near(contract_state)]
#[derive(PanicOnDefault)]
pub struct Contract {}

#[near]
impl Contract {
    #[init]
    #[private]
    pub fn new() -> Self {
        Self {}
    }
}
```

Expanded with `cargo-expand`:
```rs
#[no_mangle]
pub extern "C" fn new() {
    ::near_sdk::env::setup_panic_hook();
    if ::near_sdk::env::attached_deposit().as_yoctonear() != 0 {
        ::near_sdk::env::panic_str("Method new doesn't accept deposit");
    }
    if ::near_sdk::env::state_exists() {
        ::near_sdk::env::panic_str("The contract has already been initialized");
    }
    let contract = Contract::new();
    ::near_sdk::env::state_write(&contract);
}
```

There's also no snapshot of this attribute combination.

Not sure if this is disabled on purpose, but in that case the code should fail to compile instead of misleading the user. Generated ABI doesn't include private parameter either.

r-near:
## Problem

The `#[private]` attribute was silently ignored when used with `#[init]` methods. This is a security issue because developers believe their init methods are protected when they're not.

```rust
#[init]
#[private]
pub fn new() -> Self { ... }
```

**What should happen:**
- Runtime check that `current_account_id() == predecessor_account_id()`
- ABI includes `Private` modifier

**What actually happened:**
- No runtime check generated (anyone could call the method)
- ABI only showed `Init`, missing `Private`

## Root Cause

PR #1074 (August 2023) explicitly un-deprecated `#[private]` on init methods but never actually implemented the functionality. The `InitMethod` struct was missing an `is_private` field, and `is_private()` always returned `false` for init methods.

## Solution

Added `is_private` field to `InitMethod` struct and updated the code generator to properly handle it. This fixes both code generation and ABI generation.

**Generated code now includes the private check:**
```rust
if ::near_sdk::env::current_account_id() != ::near_sdk::env::predecessor_account_id() {
    ::near_sdk::env::panic_str("Method new is private");
}
```

**ABI now includes both modifiers:**
```rust
modifiers: vec![
    AbiFunctionModifier::Init,
    AbiFunctionModifier::Private
]
```

## References

Closes #1409
Related to #1040
Completes #1074

### DIFF
--- a/near-sdk-macros/src/core_impl/info_extractor/mod.rs
+++ b/near-sdk-macros/src/core_impl/info_extractor/mod.rs
@@ -47,11 +47,11 @@ pub enum MethodKind {
 
 #[derive(Clone, PartialEq, Eq)]
 pub struct CallMethod {
-    /// Whether method accepting $NEAR.
+    /// Whether method accepts attached $NEAR deposits (panic by default to avoid users attaching tokens to the function that do not handle/register them).
     pub is_payable: bool,
-    /// Whether method can accept calls from self (current account)
+    /// Whether method only accepts calls from self (current account)
     pub is_private: bool,
-    /// Whether method can accept unknown JSON fields
+    /// Whether method only accepts known JSON fields (useful for sensitive functions to prevent typos or malicious actors spoofing users with fields that are not going to be used)
     pub deny_unknown_arguments: bool,
     /// The serializer that we use for the return type.
     pub result_serializer: SerializerType,
@@ -61,9 +61,9 @@ pub struct CallMethod {
 
 #[derive(Clone, PartialEq, Eq)]
 pub struct ViewMethod {
-    /// Whether method can accept calls from self (current account)
+    /// Whether method only accepts calls from self (current account)
     pub is_private: bool,
-    /// Whether method can accept unknown JSON fields
+    /// Whether method only accepts known JSON fields (useful for sensitive functions to prevent typos or malicious actors spoofing users with fields that are not going to be used)
     pub deny_unknown_arguments: bool,
     /// The serializer that we use for the return type.
     pub result_serializer: SerializerType,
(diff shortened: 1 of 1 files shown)


near-sdk-macros/src/core_impl/info_extractor/mod.rs
```diff
@@ -75,6 +75,8 @@ pub struct ViewMethod {
 pub struct InitMethod {
     /// Whether method accepting $NEAR.
     pub is_payable: bool,
+    /// Whether method can accept calls from self (current account)
+    pub is_private: bool,
```
DSharifi:
```suggestion
    /// Whether method only accepts calls from self (current account)
    pub is_private: bool,
```

near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs
```diff
@@ -354,4 +354,17 @@ mod tests {
...
+        let impl_type: Type = syn::parse_str("Hello").unwrap();
+        let mut method: ImplItemFn = parse_quote! {
+            #[init]
+            #[private]
+            pub fn new() -> Self { }
+        };
+        let method_info = ImplItemMethodInfo::new(&mut method, None, impl_type).unwrap().unwrap();
+        let actual = method_info.method_wrapper();
+        local_insta_assert_snapshot!(pretty_print_syn_str(&actual).unwrap());
+    }
```
DSharifi, line 369: Would be good to test that no external accounts can call the method

frol: @r-near Could you, please, add that test? And also a test that `#[init] #[private]` generates the same snapshot as `#[private] #[init]` (e.g. the order does not make any difference)

PR #1425, opened 2025-12-02
feat: `StateInit`

mitinarseny:
This PR adds `StateInit` struct which will be heavily used by smart-contracts to pass it between function calls

### DIFF
--- a/near-sdk/src/promise.rs
+++ b/near-sdk/src/promise.rs
@@ -437,11 +437,7 @@ impl Promise {
 
     /// Creates a deterministic account with the given code, deposit, and data.
     #[cfg(feature = "deterministic-account-ids")]
-    pub fn deterministic_state_init(
-        self,
-        state_init: crate::state_init::StateInit,
-        deposit: NearToken,
-    ) -> Self {
+    pub fn state_init(self, state_init: crate::state_init::StateInit, deposit: NearToken) -> Self {
         self.add_action(PromiseAction::DeterministicStateInit { state_init, deposit })
     }
 


near-sdk/src/environment/env.rs
```diff
@@ -1170,7 +1155,7 @@ pub fn promise_set_refund_to(promise_index: PromiseIndex, account_id: &AccountId
 #[cfg(feature = "deterministic-account-ids")]
 pub fn promise_batch_action_state_init(
     promise_index: PromiseIndex,
-    code: CryptoHash,
+    code: &CryptoHash,
```
akorchyn: CryptoHash is 32 bytes. Don't see a reason not to copy.

near-sdk/src/types/contract_code.rs
```diff
@@ -0,0 +1,73 @@
+use near_account_id::AccountId;
+use near_sdk_macros::near;
+
+use crate::CryptoHash;
+
+#[near(inside_nearsdk)]
+#[derive(Debug, Clone, PartialEq, PartialOrd, Ord, Eq)]
+pub enum AccountContract {
+    None,
+    Local(CryptoHash),
```
akorchyn, line 9: probably we should use serde_as here too?

mitinarseny: Good catch! however, `AccountContract` doesn't implement `Serialize`
I suggest to add `serde_as` when it will make sense to derive `Serialize`, is it ok with you?

mitinarseny: Ah, I see. This enum shouldn't be used outside of a single contract context, at least for now I don't see any case for it. So, unless we understand the need for it, I suggest we remove `#[near]` here completely. I've pushed the change, lmk if you think that's still necessary.

akorchyn: Agree

PR #1428, opened 2025-12-04
docs: fix reference to UnorderedSet

VolodymyrBg:
The module docs incorrectly compared LookupSet to itself, which obscures the intended contrast. This aligns the wording with lookup_map.rs where the non-iterable type is contrasted with its iterable counterpart, and clarifies that UnorderedSet stores elements in a vector enabling iteration.


frol (approved):
@VolodymyrBg Good catch!

@race-of-sloths score 2

PR #1439, opened 2025-12-11
feat: promise yield api

pityjllk:
Adds high-level `Promise::yield_create` API for the promise yield/resume feature, making it easier to use coroutine-style async patterns in
  NEAR contracts.

  **Changes:**
  - Add `Promise::yield_create()` method that returns `(Promise, YieldId)` for creating yielded promises
  - Add `YieldId` type with `resume()` method for resuming yielded promises with data
  - Add helper `env::promise_yield_create_id()` that wraps the low-level API and returns the yield ID
  - Add proper error handling to prevent invalid operations on yielded promises (cannot add actions, cannot be used as continuations)
  - Add unit tests for the restriction behaviors
  
  @race-of-sloths

### DIFF
--- a/near-sdk/src/promise.rs
+++ b/near-sdk/src/promise.rs
@@ -1593,12 +1593,8 @@ mod tests {
         testing_env!(VMContextBuilder::new().signer_account_id(alice()).build());
 
         // Create a yielded promise
-        let (_promise, yield_id) = Promise::new_yield(
-            "on_callback",
-            b"test_args",
-            Gas::from_tgas(10),
-            GasWeight(1),
-        );
+        let (_promise, yield_id) =
+            Promise::new_yield("on_callback", b"test_args", Gas::from_tgas(10), GasWeight(1));
 
         // Verify yield_id is a valid 32-byte CryptoHash
         let yield_id_bytes: [u8; 32] = yield_id.0;


near-sdk/src/environment/env.rs
```diff
@@ -1968,6 +1968,25 @@ pub fn promise_yield_create(
...
+    arguments: impl AsRef<[u8]>,
+    gas: Gas,
+    weight: GasWeight,
+) -> (PromiseIndex, crate::YieldId) {
+    const REGISTER_ID: u64 = 66;
+    let promise_index = promise_yield_create(function_name, arguments, gas, weight, REGISTER_ID);
+    let yield_id = read_register(REGISTER_ID)
+        .and_then(|v| v.try_into().ok())
+        .map(crate::YieldId)
+        .expect("Failed to read yield ID from register");
```
mitinarseny:
```suggestion
    let yield_id = read_register_fixed(REGISTER_ID);
```

near-sdk/src/promise.rs
```diff
@@ -325,21 +325,71 @@ impl BorshSchema for Promise {
 enum PromiseSubtype {
     Single(Rc<PromiseSingle>),
     Joint(Rc<PromiseJoint>),
+    Yielded(PromiseIndex),
 }
```
mitinarseny, line 348: This way, you can't add other actions in the receipt that starts with yielded promise. As a solution, I'd suggest making it a part of `PromiseSingle`:
```rust
enum PromiseSingleSubtype {
    Regular {
        account_id: AccountId,
        after: RefCell<Option<Rc<Promise>>>,
        /// Promise index that is computed only once.
        promise_index: RefCell<Option<PromiseIndex>>,
    },
    Yielded(PromiseIndex),
}

struct PromiseSingle {
    pub subtype: PromiseSingleSubtype,
    pub actions: RefCell<Vec<PromiseAction>>, 
}
```
This way, you can also panic in runtime for `regular_promise.then(yielded_promise)`

near-sdk/src/environment/env.rs
```diff
@@ -1968,6 +1968,23 @@ pub fn promise_yield_create(
...
+///
+/// This is a convenience wrapper around [`promise_yield_create`] that automatically reads the
+/// yield ID from the register and returns it as a [`YieldId`].
+pub fn promise_yield_create_id(
+    function_name: &str,
+    arguments: impl AsRef<[u8]>,
+    gas: Gas,
+    weight: GasWeight,
+) -> (PromiseIndex, crate::YieldId) {
+    const REGISTER_ID: u64 = 66;
```
mitinarseny: Reuse ATOMIC_OP_REGISTER

near-sdk/src/promise.rs
```diff
@@ -792,6 +871,45 @@ impl schemars::JsonSchema for Promise {
...
+);
+
+impl YieldId {
+    /// Resume the yielded promise with the provided data.
+    ///
+    /// Returns `true` if the promise was successfully resumed, `false` if no promise
+    /// with this `YieldId` was found (e.g., if it was already resumed or timed out).
+    ///
+    /// Uses low-level [`crate::env::promise_yield_resume`]
+    pub fn resume(self, data: impl AsRef<[u8]>) -> bool {
```
frol: I'd like to suggest to use `Result<(), ResumeError>` as return value instead of boolean to leverage Rust type system expressiveness

Note: The error type can be as simple as `struct ResumeError;`

near-sdk/src/promise.rs
```diff
@@ -331,15 +351,69 @@ impl Promise {
...
+    /// Yielded promises have a restriction:
+    /// - **Cannot be used as continuations**: Using a yielded promise in `other.then(yielded)` will panic.
+    ///   Yielded promises must be first in the chain: `yielded.then(other)` is valid.
+    ///
+    /// # Example
+    /// ```ignore
+    /// use near_sdk::{Promise, Gas, GasWeight};
+    ///
+    /// // Create a yielded promise
+    /// let (promise, yield_id) = Promise::yield_create(
```
frol: I am thinking about the naming and while I think `Promise::yield_create` is quite neutral, it does not immediately bring any associations to me.

Have you considered something like `Promise::new_resumable`, `Promise::on_resume`, `Promise::with_resume`, `Promise::new_delayed`, `Promise::delay`, or even `Promise::new_yield`? Or even more agressive - introduce a method to turn an already created promise into a "resumable promise" / "delayed promise"

pityjllk: ideally i would go with just `Promise::yield` as its self explanatory, but its reserved keyword so i ended up with yield_create. from the suggestions you posted i think i like `new_yield` the most 👍

mitinarseny: > Or even more agressive - introduce a method to turn an already created promise into a "resumable promise" / "delayed promise"

This sounds great when you say it first but when you try to implement it ends up in poor API because it should panic if either promise receiver is not current_account_id or if first action is not a function_call.

I think current approach is the best we can do without introducing a statically-typed Promise API that would be a breaking change we discussed before.

near-sdk/src/promise.rs
mitinarseny: I would not recommend doing that because it would make it feel natural to panic on it, while in most cases we should encourage to handle it gracefully, such as [here](https://github.com/near/intents/blob/f4dbbdeaec9f3fa48d9f41ecb4a9bf8b21d54171/transfer-auth/src/contract/mod.rs#L58-L65).

frol: It does not make it feel natural to panic on Err for me any time outside of prototype and it is common for me to forbid expect, unwrap via clippy.

frol (approved):
@pityjllk Please, fix the CI errors and we are good to go. Thank you for the contribution!

@race-of-sloths score 8

PR #1443, opened 2025-12-15
fix(store): clarify LookupSet persistence behavior

ANtutov:
store::LookupSet was implemented as a non‑caching collection that writes directly to storage on each operation, but the store module level FAQ and a few test comments still implied Drop/flush-based persistence and cache-like states. This mismatch could confuse users who rely on the docs to reason about gas costs and storage semantics.

This change updates the store module documentation to state that most, but not all, collections persist changes on Drop/flush and explicitly calls out LookupSet as an exception that writes directly to storage. It also cleans up misleading comments in 
LookupSet tests so they no longer suggest deferred writes or cache-backed state transitions, while leaving the test logic and behavior unchanged.


frol:
@race-of-sloths score 3

PR #1444, opened 2025-12-16
fix: clarify deployer account comment in global factory example

GarmashAlex:
The comment above deploy_global_contract_by_account_id claimed the global contract is identified by the predecessor account, which is not how the API works. In practice the example uses the explicitly provided account_id as the deployer/owner, matching the Global Contracts “reference by account” model in NEAR docs.


frol (approved):
@race-of-sloths score 2

PR #1445, opened 2025-12-18
refactor: reuse shared TokenId alias in enumeration_impl

Forostovec:
Use the shared TokenId type alias exported from non_fungible_token::token in enumeration_impl.rs instead of redefining type TokenId = String. This removes the only local duplicate alias in the NFT standard implementation, keeps all modules consistent around a single definition of TokenId, and makes future changes to the underlying token ID type easier to manage.


frol (approved):
@race-of-sloths score 2

PR #1454, opened 2025-12-24
refactor: Use Base58CryptoHash instead of CryptoHash with base64 encoding (for JSON) in GlobalContractId

frol:


near-sdk/src/environment/env.rs
```diff
@@ -200,8 +200,12 @@ pub fn current_contract_code() -> AccountContract {
 ///
 /// # Examples
 /// ```
+/// # #[cfg(feature = "deterministic-account-ids")]
```
mitinarseny: would it work if we just move `#[cfg(feature = "deterministic-account-ids")]` above the comment?

.github/workflows/test.yml
```diff
@@ -22,7 +22,7 @@ jobs:
             rs: 1.85.0
           - os: macos-latest
             rs: 1.86.0
-        features: ["", "--features unstable,legacy,__abi-generate"]
+        features: ["", "--features unstable,legacy,__abi-generate,deterministic-account-ids"]
```
mitinarseny, line 25: How about running doctests in CI as well?
```suggestion
        features: ["", "--features unstable,legacy,__abi-generate,deterministic-account-ids"]
        targets: ["--all-targets", "--doc"]
```

And then add it to the step below:
```yaml
- name: test
   run: cargo test --workspace ${{ matrix.targets }} ${{ matrix.features }}
```

near-sdk/src/environment/env.rs
```diff
@@ -1136,21 +1140,25 @@ pub fn promise_batch_then(promise_index: PromiseIndex, account_id: &AccountId) -
...
 ///     NearToken::from_millinear(1),
-///     NearGas::from_tgas(1),
+///     Gas::from_tgas(1),
 /// );
-/// promise_set_refund_to(promise, "refund.near".parse().unwrap());
+/// promise_set_refund_to(promise, &"refund.near".parse().unwrap());
+/// # };
 /// ```
 #[cfg(feature = "deterministic-account-ids")]
 pub fn promise_set_refund_to(promise_index: PromiseIndex, account_id: &AccountId) {
```
mitinarseny, line 1158:
```suggestion
pub fn promise_set_refund_to(promise_index: PromiseIndex, account_id: impl AsRef<AccountIdRef>) {
```

mitinarseny (approved):

near-sdk/src/environment/env.rs
frol: I think you have already prepared the fix for this in #1458, right?

.github/workflows/test.yml
frol: Doctests are already executed (the whole reason for this massive PR is due to doctests were actually failing on CI) as `cargo test --all` covers it.

https://github.com/near/near-sdk-rs/actions/runs/20502766452/job/58912632026#step:7:2125

near-sdk/src/environment/env.rs
frol: It did not work without the feature guard, but I have just tested it again, and it works fine now. Let's see if it will pass on CI

PR #1455, opened 2025-12-24
feat: `env::current_global_contract_id()`

mitinarseny:
THis PR adds a helper method to optionally convert `AccountContract` to `GlobalContractId`:
```rust
let global_contract_id = env::current_contract_code().into_global()?;
```


frol (approved):
@race-of-sloths score 3

PR #1458, opened 2025-12-24
feat: `Promise::refund_to`

mitinarseny:
This PR adds `Promise::refund_to` method and removes unnecessary locks from inner `Promise` implementation:
```rust
Promise::new(account_id)
    .refund_to(refund_to)
    // ...
```

### DIFF
--- a/near-sdk/src/promise.rs
+++ b/near-sdk/src/promise.rs
@@ -1,18 +1,22 @@
 #[cfg(feature = "abi")]
-use borsh::BorshSchema;
-use std::cell::RefCell;
-#[cfg(feature = "abi")]
 use std::collections::BTreeMap;
-use std::collections::VecDeque;
-use std::io::{Error, Write};
-use std::mem;
-use std::num::NonZeroU128;
-use std::rc::Rc;
+use std::{
+    cell::RefCell,
+    collections::VecDeque,
+    io::{Error, Write},
+    mem,
+    num::NonZeroU128,
+    rc::Rc,
+};
 
-use crate::env::migrate_to_allowance;
-use crate::CryptoHash;
-use crate::{AccountId, Gas, GasWeight, NearToken, PromiseIndex, PublicKey};
+#[cfg(feature = "abi")]
+use borsh::BorshSchema;
 use near_sdk_macros::near;
+
+use crate::{
+    env::migrate_to_allowance, AccountId, CryptoHash, Gas, GasWeight, NearToken, PromiseIndex,
+    PublicKey,
+};
 
 /// Allow an access key to spend either an unlimited or limited amount of gas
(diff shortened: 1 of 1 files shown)


near-sdk/Cargo.toml
```diff
@@ -63,6 +63,7 @@ quickcheck = "1.0"
 arbitrary = { version = "1.2.3", features = ["derive"] }
 derive_arbitrary = "1.2.3"
 hex = { version = "0.4.3", features = ["serde"] }
+hex-literal = "1.1"
```
frol: It is only used in a single doctest, let's avoid adding dependencies for one-off things. Let's use Base58CryptoHash in that docstring to be consistent with the formatting that is usually used in NEAR

near-sdk/src/promise.rs
```diff
@@ -751,11 +735,11 @@ impl Promise {
     /// and finally `dave_near` and `eva_near` are created concurrently.
     ///
     /// ```no_run
-    /// # use near_sdk::Promise;
-    /// let p1 = Promise::new("bob_near".parse().unwrap()).create_account();
-    /// let p2 = Promise::new("carol_near".parse().unwrap()).create_account();
-    /// let p3 = Promise::new("dave_near".parse().unwrap()).create_account();
-    /// let p4 = Promise::new("eva_near".parse().unwrap()).create_account();
+    /// # use near_sdk::{AccountId, Promise};
+    /// let p1 = Promise::new("bob_near".parse::<AccountId>().unwrap()).create_account();
```
frol: On one hand, I love explicitness, on another, it becomes a bit vebose and technically a breaking change that many existing contracts will hit. Can we land this PR for 6.x release or do you depend on something here that you need to have in 5.x?

With this explicit `AccountId` import, maybe it will be easier to read if we use `AccountIdRef::new_or_panic()` instead of `"".parse::<AccountId>().unwrap()`

.github/workflows/test.yml
```diff
@@ -129,4 +129,4 @@ jobs:
           cargo doc -p near-sdk --features unstable,legacy,unit-testing,__macro-docs,__abi-generate
           cargo doc -p near-sdk-macros --features __abi-generate
           cargo doc -p near-contract-standards --no-deps --features abi
-          cargo doc -p near-sys
+          cargo doc -p near-sys
```
frol: Please, revert the change (probably it was introduced during the merge conflicts resolution)

near-sdk/src/promise.rs
mitinarseny: @frol I've removed breaking API changes from this PR and branched off another [draft PR](https://github.com/near/near-sdk-rs/pull/1459) with them. Would it work for you?

frol (approved):

PR #1479, opened 2026-01-28
fix(ci): use resolver v3 for MSRV-aware dependency resolution

r-near:
## Summary

Use Cargo resolver v3 which respects `rust-version` when resolving dependencies. This is the recommended approach from the Rust 1.84 release.

### Problem

`time v0.3.46` requires Rust 1.88, but our MSRV is 1.86. Since libraries don't commit lockfiles, and Cargo.toml version constraints don't propagate to downstream users, we need a resolver-level solution.

### Solution

Upgrade to resolver v3, which automatically selects dependency versions compatible with the declared `rust-version`. This works because:

1. Resolver v3 respects `rust-version` in dependency resolution
2. We declare `rust-version = "1.86"` in workspace and examples
3. Cargo picks `time v0.3.45` instead of `v0.3.46`

Changes:
- Set `resolver = "3"` in workspace Cargo.toml
- Add `rust-version = "1.86"` to all examples
- Add `resolver = "3"` to example workspaces
- Pin `cargo-llvm-cov` to 0.6.24 (separate issue - v0.7.0+ breaks wasm coverage tests)

### Note

Resolver v3 was stabilized in Cargo 1.84. Using it with edition 2021 is supported.

### DIFF
--- a/near-sdk/src/types/public_key.rs
+++ b/near-sdk/src/types/public_key.rs
@@ -92,7 +92,7 @@ impl TryFrom<PublicKey> for near_crypto::PublicKey {
 ///             .unwrap();
 ///
 /// // Uncompressed secp256k1 key
-/// let secp256k1: PublicKey = "secp256k1:qMoRgcoXai4mBPsdbHi1wfyxF9TdbPCF4qSDQTRP3TfescSRoUdSx6nmeQoN3aiwGzwMyGXAb1gUjBTv5AY8DXj"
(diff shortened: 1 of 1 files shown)


vsavchyn-dev (approved):
I checked manually if there is any increase in size for the examples compared to current master branch, and happy to say there is no difference. Plus, this will unblock slash command `compare` to check difference in sizes (bc of new resolver in examples).

There shouldn't be any problem merging this and following up with update of SDK's edition to 2024

PR #1483, opened 2026-02-05
ci: automate contract size comparison on PRs

r-near:
## Summary

Replaces the on-demand `/compare` workflow with automatic contract size tracking on every PR.

### Changes

- **New `sizes_baseline.yml`**: Runs on master push, builds all examples with reproducible Docker builds, caches sizes
- **Rewritten `compare_sizes.yml`**: Runs on PRs (with path filters), compares against cached baseline, posts comment only if sizes change ≥1%
- **Single Python script** (`ci/compare_sizes.py`): Handles both `build` and `compare` subcommands
- **Removed old Python scripts**: `compare_sizes/compare_sizes.py`, `project_instance.py`, `README.md`

### How it works

1. On master push → build examples, cache sizes
2. On PR → build examples, fetch baseline from cache, compare, comment if changes detected
3. Uses `peter-evans/create-or-update-comment` to update same comment (no spam)

### Key improvements

- **Automatic**: No need to manually type `/compare`
- **Faster**: Only builds PR branch (not both branches like before)  
- **Maintainable**: Example list in one place, pure Python (no GitPython dependency)
- **Quiet**: No comment if sizes unchanged
- **Accurate**: Uses reproducible Docker builds

### Testing

Tested locally - all examples build successfully with the v3 resolver.

```bash
python3 ci/compare_sizes.py build --output sizes.json
python3 ci/compare_sizes.py compare baseline.json sizes.json --threshold 1
```

### Notes

- The `sizes_baseline.yml` needs to run once on master to create the initial baseline cache
- Can manually trigger it via `workflow_dispatch` after merging
- There's a commented-out step to fail the job if sizes increase (can enable later if desired)

### DIFF
--- a/.github/workflows/release-plz.yml
+++ b/.github/workflows/release-plz.yml
@@ -18,7 +18,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
-          token: ${{ secrets.CUSTOM_GITHUB_TOKEN }}
+          token: ${{ secrets.NEARPROTOCOL_CI_PR_ACCESS }}
 
       - name: Install Rust toolchain 1.86 for builds
         uses: dtolnay/rust-toolchain@1.86
@@ -33,6 +33,6 @@ jobs:
         uses: release-plz/action@5ab144c9d67d4346240190d0f95ed08668677928
         env:
           # https://marcoieni.github.io/release-plz/github-action.html#triggering-further-workflow-runs
-          GITHUB_TOKEN: ${{ secrets.CUSTOM_GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.NEARPROTOCOL_CI_PR_ACCESS }}
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
           CARGO_BUILD_TARGET: wasm32-unknown-unknown
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ exclude = ["examples/"]
 
 [workspace.package]
 version = "5.24.1"
-edition = "2021"
+edition = "2024"
 license = "MIT OR Apache-2.0"
 categories = ["wasm"]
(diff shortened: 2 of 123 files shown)


.github/workflows/compare_sizes.yml
```diff
@@ -1,49 +1,165 @@
...
+      - '.github/workflows/sizes_baseline.yml'
+      - 'ci/compare_sizes.py'
+
+env:
+  SIZE_THRESHOLD: 1
+
 jobs:
-  generate_report:
-    runs-on: ubuntu-latest
-    if: ${{ (github.event.issue.pull_request) && (github.event.comment.body == '/compare') }}
```
vsavchyn-dev, line 9: Problem I have with this approach is that we run it with warp on each push to the branch, and it stacks up quickly:
- Price per minute warp-ubuntu-latest-x64-4x: 0.008$
- Num of jobs running together: 13
- Checking this job running, it's 1 min on AVG per job

This already gives us 0.104$ per push to branch. Let's say there are 5 pushes to the branch that trigger this job, which brings us to 0.52$ per branch. 

But I should admit, it's muuuch faster than 40 minutes before: https://github.com/near/near-sdk-rs/actions/runs/20560301708

ci/compare_sizes/compare_sizes.py
vsavchyn-dev, line 1: I think we should leave this just in case for local development

.github/workflows/compare_sizes.yml
r-near: Yeah that's a good point. 10 cents per push. Let's monitor this. If it turns out to be too expensive, we can go back to triggering it manually 🤷

ci/compare_sizes/compare_sizes.py
r-near: Fair enough, added it back. [`533d122` (this PR)](https://github.com/near/near-sdk-rs/pull/1483/commits/533d12251c977e79939f9a0ee559d6b14235ceb0)

vsavchyn-dev (approved):

Issue #239, opened 2020-10-08
Create macros that create new implementation that marshals method calls into PendingContractTx


willemneal:
For example, in the FungibleToken Contract:

```rust
struct FungibleTokenContract {
    account_id: AccountId,
}

impl FungibleTokenContract {
    pub fn new(&self, owner_id: AccountId, total_supply: Balance) -> PendingContractTx {
        let balance: U128 = total_supply.into();
        PendingContractTx::new(
            &self.account_id, // Contract is the receiver
            "new",
            json!({
              "owner_id": owner_id,
              "total_supply": balance
            }),
            false, // Is not a view function
        )
    }
...
}
```

willemneal:
see #235.  It now contains the marshaling.  Still need to improve the initialization experience.

PR #1486, opened 2026-02-05
fix(ci): pin release-plz/action to last working SHA

r-near:
## Summary

Pin `release-plz/action` to commit `5ab144c9d67d4346240190d0f95ed08668677928` (last version using `taiki-e/install-action` for prebuilt binaries).

## Problem

The `release-plz` CI has been failing since Jan 28 due to an upstream change:

| Date | Event |
|------|-------|
| Jan 23 | ✅ Last successful run - used `taiki-e/install-action` (prebuilt binary) |
| Jan 24 | `release-plz/action` [PR #239](https://github.com/release-plz/action/pull/239) switched to `cargo-binstall --strategies=crate-meta-data,compile` |
| Jan 28 | ❌ First failure - binstall compiles from source, hits MSRV conflicts |
| Feb 5 | ❌ PR #1485 added stable toolchain, but now hits `gix-sec` wasm32 compile error |

The `--strategies=crate-meta-data,compile` **excludes the `github-release` strategy**, so binstall can't download prebuilt binaries. When `crate-meta-data` fails (release-plz doesn't publish binstall metadata), it falls back to `cargo install --target wasm32-unknown-unknown`, which fails because `gix-sec` uses Unix APIs (`std::os::unix`) unavailable on wasm32.

### Why not pre-install release-plz?

The action uses `cargo binstall --force`, which reinstalls even if the binary exists.

## Solution

Pin to the last working SHA that uses `taiki-e/install-action` instead of `cargo-binstall`.

### DIFF
--- a/.github/workflows/release-plz.yml
+++ b/.github/workflows/release-plz.yml
@@ -24,10 +24,6 @@ jobs:
         uses: dtolnay/rust-toolchain@1.86
         with:
           targets: wasm32-unknown-unknown
-      - name: Install Rust toolchain stable for release-plz
-        uses: dtolnay/rust-toolchain@stable
-        with:
-          targets: wasm32-unknown-unknown
 
       # Pinned to last version using taiki-e/install-action for binary installation.
       # Newer versions use cargo-binstall which compiles from source and fails with


.github/workflows/release-plz.yml
```diff
@@ -29,8 +29,12 @@ jobs:
         with:
           targets: wasm32-unknown-unknown
```
vsavchyn-dev: Maybe we should remove installing of stable toolchain considering that using 1.86 worked with previous release-plz 0.3.151? Plus we install binary anyway which shouldn't require us to use stable toolchain for now.

vsavchyn-dev (approved):
LGTM, just one question.

PR #1499, opened 2026-02-19
fix: increase tolerance in flaky storage deposit test

### LINKED ISSUE #1353: todo: check flaky test `test_storage_deposit_minimal_deposit` in `examples/fungible-token/tests/workspaces.rs`
consider increasing threshold 1 -> 2 millinear in https://github.com/dj8yfo/near-sdk-rs/blob/396fbb8463e2cb98971d65783ae985e39a25ebaf/examples/fungible-token/tests/workspaces.rs#L191 , 
or otherwise investigating the reason of failure

Image

abhiraj-mengade:
Fixes #1353

## Changes
- Increase tolerance from 1 to 2 millinear for new account balance assertion
- Add detailed error messages to help debug future failures

## Problem
The  test was flaky due to gas price variations that could cause transaction fees to exceed the original 1 millinear tolerance.

## Solution  
Increasing to 2 millinear provides more headroom while maintaining test validity. The enhanced error messages will also help diagnose any future issues quickly.

## Testing
- All fungible token tests pass (13/13)
- Ran the specific test 5+ times consistently
- No other functionality affected

### DIFF
--- a/near-sdk-macros/src/core_impl/code_generator/ext.rs
+++ b/near-sdk-macros/src/core_impl/code_generator/ext.rs
@@ -75,7 +75,7 @@ pub(crate) fn generate_ext_structs(
 /// However, some attributes should be forwarded and they are defined here.
 ///
 /// [#959]: https://github.com/near/near-sdk-rs/pull/959
-const FN_ATTRIBUTES_TO_FORWARD: [&str; 1] = [
+const FN_ATTRIBUTES_TO_FORWARD: [&str; 2] = [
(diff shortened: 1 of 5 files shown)


r-near (changes requested):
Thanks for adding this, however, I'd prefer to move in the direction of a more deterministic test instead of allowing a wider gap. Would you mind refactoring this PR? I think some proactive steps would be figuring out what the exact number is, and how it relates to our other constants (such as gas, etc.)

This would also make a great smoke test if anything changes on the protocol level.

PR #1501, opened 2026-02-19
fix: `IterableSet` and `UnorderedSet` `Drain` trait

vsavchyn-dev:
This PR aims to fix bug in `IterableSet` and `UnorderedSet` drain iterators where backward iteration (`next_back()`) failed to remove index entries, causing silent state corruption.

### Problem

Both Set types use dual data structures that must stay synchronized:
- Elements: `Vector<T>` or `FreeList<T>` - primary storage
- Index: `LookupMap<T, Index>` - element position mapping

The `Drain` iterator's `next()` method correctly updated both:
```rust
fn next(&mut self) -> Option<Self::Item> {
    let key = self.elements.next()?;
    self.index.remove(&key);  // removes from index
    Some(key)
}
```

But next_back() only updated elements:
```rust
fn next_back(&mut self) -> Option<Self::Item> {
    self.elements.next_back()  // missing self.index.remove(&key)
}
```

### Impact

Where using `.drain().next_back()`:
- Orphaned entities remain in the index `LookupMap`, which might increase storage costs for accounts
- `contains()` returns true for drained elements
- `insert()` might fail silently for previously drained elements

### Verification

Added tests that expose the bug:
- `test_drain_next_back_removes_from_index()` - Verifies backward drain removes from index
- `test_drain_bidirectional()` - Verifies that forward and backward drain both work

### DIFF
--- a/near-sdk/src/store/iterable_set/iter.rs
+++ b/near-sdk/src/store/iterable_set/iter.rs
@@ -359,6 +359,8 @@ where
     H: ToKey,
 {
     fn next_back(&mut self) -> Option<Self::Item> {
-        self.elements.next_back()
+        let key = self.elements.next_back()?;
+        self.index.remove(&key);
+        Some(key)
     }
 }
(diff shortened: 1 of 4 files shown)


r-near (approved):
Nice work on tests 😁

PR #1508, opened 2026-02-22
feat: forward `allow` attribute

### LINKED ISSUE #1505: #[allow(clippy::too_many_arguments)] attached to #[near] impl methods is not retained
```rs
#[near]
impl Contract {
    #[allow(clippy::too_many_arguments)]
    pub fn do_something(&self, arg1: i8, arg2: i8, arg3: i8, arg4: i8, arg5: i8, arg6: i8, arg7 : i8) {}
}
```

Clippy still warns even though there is #[allow]. But the warning is removed if it's a normal impl (without #[near])

frolvanya:
I see that we stopped forwarding all non-bindgen attributes in #959 with a good reasoning behind, but I think it's safe to forward `allow` attribute, since it's used only for linting and shouldn't cause any problems or breaking changes. However, I don't have a strong opinion here, so feel free to close this PR if I've missed something, since crate-level `#![allow(...)]` still works

closes #1505

### DIFF
--- a/near-sdk-macros/src/core_impl/code_generator/ext.rs
+++ b/near-sdk-macros/src/core_impl/code_generator/ext.rs
@@ -75,7 +75,7 @@ pub(crate) fn generate_ext_structs(
 /// However, some attributes should be forwarded and they are defined here.
 ///
 /// [#959]: https://github.com/near/near-sdk-rs/pull/959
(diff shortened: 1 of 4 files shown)


r-near:
Thank you for fixing this!

PR #1511, opened 2026-03-09
feat: allow crypto-related `env::*` functions for non-contract-usage

mitinarseny:
This PR allows third-party projects to use crypto-related `env::*` functions from near-sdk even if `non-contract-usage` feature is enabled.

For example, this lets other projects to use `StateInit::derive_account_id` without enabling a heavy `unit-testing` feature.

Implementations for these `env::*` functions follow the behavior from `near-vm-runner` crate but without unnecessary conversions.

### DIFF
--- a/near-sdk/Cargo.toml
+++ b/near-sdk/Cargo.toml
@@ -106,7 +106,14 @@ unit-testing = [
     "near-crypto",
     "near-parameters",
 ]
-non-contract-usage = ["ed25519-dalek", "near-crypto", "ripemd", "sha2", "sha3"]
+non-contract-usage = [
+    "ed25519-dalek",
+    "near-crypto",
+    "near-primitives-core",
+    "ripemd",
+    "sha2",
+    "sha3",
+]
 
 __abi-embed = ["near-sdk-macros/__abi-embed"]
 __abi-generate = ["abi", "near-sdk-macros/__abi-generate"]
--- a/near-sdk/src/environment/env.rs
+++ b/near-sdk/src/environment/env.rs
@@ -607,21 +607,10 @@ pub fn keccak512(value: impl AsRef<[u8]>) -> Vec<u8> {
 /// );
 /// ```
 pub fn sha256_array(value: impl AsRef<[u8]>) -> CryptoHash {
-    #[cfg(all(
-        feature = "non-contract-usage",
-        not(target_arch = "wasm32"),
-        not(feature = "unit-testing"),
-    ))]
-    {
-        use sha2::Digest;
-
-        sha2::Sha256::digest(value).into()
-    }
-
     #[cfg(any(
-        not(feature = "non-contract-usage"),
         target_arch = "wasm32",
-        feature = "unit-testing",
+        not(feature = "non-contract-usage"),
+        all(feature = "unit-testing", not(test)),
     ))]
     {
         let value = value.as_ref();
@@ -632,6 +621,17 @@ pub fn sha256_array(value: impl AsRef<[u8]>) -> CryptoHash {
             sys::sha256(value.len() as _, value.as_ptr() as _, ATOMIC_OP_REGISTER);
             read_register_fixed(ATOMIC_OP_REGISTER)
         }
+    }
+
+    #[cfg(all(
+        not(target_arch = "wasm32"),
+        feature = "non-contract-usage",
+        any(not(feature = "unit-testing"), test),
+    ))]
+    {
+        use sha2::Digest;
+
+        sha2::Sha256::digest(value).into()
     }
 }
 
@@ -650,21 +650,10 @@ pub fn sha256_array(value: impl AsRef<[u8]>) -> CryptoHash {
 /// );
 /// ```
 pub fn keccak256_array(value: impl AsRef<[u8]>) -> CryptoHash {
-    #[cfg(all(
-        feature = "non-contract-usage",
-        not(target_arch = "wasm32"),
-        not(feature = "unit-testing"),
-    ))]
-    {
-        use sha3::Digest;
-
-        sha3::Keccak256::digest(value).into()
-    }
-
     #[cfg(any(
-        not(feature = "non-contract-usage"),
         target_arch = "wasm32",
-        feature = "unit-testing",
+        not(feature = "non-contract-usage"),
+        all(feature = "unit-testing", not(test)),
     ))]
     {
         let value = value.as_ref();
@@ -675,6 +664,17 @@ pub fn keccak256_array(value: impl AsRef<[u8]>) -> CryptoHash {
             sys::keccak256(value.len() as _, value.as_ptr() as _, ATOMIC_OP_REGISTER);
             read_register_fixed(ATOMIC_OP_REGISTER)
         }
+    }
+
+    #[cfg(all(
+        not(target_arch = "wasm32"),
+        feature = "non-contract-usage",
+        any(not(feature = "unit-testing"), test),
+    ))]
+    {
+        use sha3::Digest;
+
+        sha3::Keccak256::digest(value).into()
     }
 }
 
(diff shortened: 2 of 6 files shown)


near-sdk/src/environment/env.rs
```diff
@@ -607,13 +607,23 @@ pub fn keccak512(value: impl AsRef<[u8]>) -> Vec<u8> {
...
 /// ```
 pub fn sha256_array(value: impl AsRef<[u8]>) -> CryptoHash {
-    let value = value.as_ref();
-    //* SAFETY: sha256 syscall will always generate 32 bytes inside of the atomic op register
-    //*         so the read will have a sufficient buffer of 32, and can transmute from uninit
-    //*         because all bytes are filled. This assumes a valid sha256 implementation.
-    unsafe {
-        sys::sha256(value.len() as _, value.as_ptr() as _, ATOMIC_OP_REGISTER);
-        read_register_fixed(ATOMIC_OP_REGISTER)
+    #[cfg(feature = "non-contract-usage")]
```
frolvanya: I think it will break things in omni-bridge. We have omni-types project which is basically a collection of types that are used in **both** on-chain and off-chain (e.g for relayer's and indexers to catch event, construct transaction and for contract method to receive and parse these arguments). In order to do this, we have a workaround to import `near-sdk`  with `non-contract-usage` (even though it's used on-chain as well): https://github.com/Near-One/omni-bridge/blob/cdbb1a03ce8bc446fe065302e6bead18562a8d2c/near/omni-types/Cargo.toml#L8

And we're using `#[cfg(target_arch = "wasm32")]` to determine what's the current scope: https://github.com/Near-One/omni-bridge/blob/main/near/omni-types/src/utils.rs. So on-chain behaviour will go into `target_arch = "wasm32"` branch that calls `near_sdk::env::sha256_array(data)`, but since we have `non-contract-usage` it'll trigger `sha2::Sha256::digest(value).into()` branch in this PR instead of proper sys call

I think it's much safer to use same approach here

mitinarseny: So, you're using
`non-contract-usage` feature in contract's code? 😅

What prevents you from enabling `near-sdk/non-contract-usage` feature only in your relayer code?

frolvanya: When I was adding `non-contract-usage` feature in sdk, I mostly thought about enabling option to use sdk in non contract context. Not just saying that if project is compiled with this flag, then it **must** be only used for off-chain. This was a purely a compilation guard bypass

That's being said, by enabling `non-contract-usage` in `omni-types` we ensure that every crate that depends on `omni-types` transitively gets near-sdk with this feature active already and we don't need to specify it in: relayers, indexers, monitoring services, etc.

Moreover, we also compile omni-types natively in CI, so in order to make it work, we'd need to add `non-contract-usage` to the `omni-types` crate itself (not a big deal, but doesn't sound right to me). This would just create indirection without any real gain, that's why I decided to put `non-contract-usage` directly in `omni-types`

P.S: I still think that relying on `#[cfg(target_arch = "wasm32")]` is much safer approach than relying on a flag that was created as an option to bypass compilation guard

P.S2: Maybe it was a bad naming after all, since I treat it more like `not-only-contract-usage`

mitinarseny: > we'd need to add non-contract-usage to the omni-types crate itself (not a big deal, but doesn't sound right to me)

This is the idiomatic way how to use features in Rust and the reason why they're so powerful when used right.

Having transitive feature is not always the best option, but it works too. Otherwise, you can add a dependency on `near-sdk` in your off-chain components and enable this feature there. This is the same approach recommended by [getrandom](https://docs.rs/getrandom/0.2.3/getrandom/#indirect-dependencies) crate (they have a full backstory with reasoning why transitive features can lead to problems)

As of now, you can enable this feature only in needed crates, while we're unable to call `StateInit::derive_account_id` from our Relayer API (it causes compilation error that functions from `near-sys` were not linked)

frolvanya: I still don't understand why you want to rely on `non-contract-usage` flag instead of proper target check Not user should choose if code is running in near-vm or no, but the platform should decide this. And the native crypto deps (sha2, sha3, etc.) should become gated with `[target.'cfg(not(target_arch = "wasm32"))'.dependencies]` in `Cargo.toml` rather than behind `non-contract-usage`. This will both work for your relayer and for our use case

mitinarseny: But what if you want to compile it as a WASM module and use it in JS environment?

frolvanya: This scenario is essentially fictional for near-sdk. The entire near-sdk crate is built for NEAR smart contracts. It's full of near-sys FFI calls (storage, promises, validators, etc.) that only work inside the near-vm. If you compile near-sdk to wasm32 and run it in a browser, you'd fail on hundreds of other unresolved imports, not just the crypto functions. Providing native fallbacks for 6 crypto functions doesn't make near-sdk usable outside near-vm. It just masks the problem for those 6 functions

Please correct me if I'm wrong on this one, since I've never considered such use case for near-sdk before

mitinarseny: I see, thanks for explaining!
I'll rewrite it to `#[cfg(any(target_family = "wasm"), feature = "unit-testing"))]` then.

However, it might be a good idea to fix all of it in the future, so that we support not only NEAR VM but other WASM runtimes, too.

frolvanya: Thanks! And I'll raise the issue with transitive dependency in my team internally and if everyone will be okay with this, we'll try to replace current approach with a proper feature forwarding

mitinarseny: @frolvanya I ended up with following since we still need to check if `non-contract-usage` feature was enabled as it is the one enabling these dependencies (already pushed):
```rust
#[cfg(all(
    feature = "non-contract-usage",
    not(target_arch = "wasm32"),
    not(feature = "unit-testing"),
))]
```
Yes, technically it would not compile without `non-contract-usage` feature due to `compile_error!(...)` in `lib.rs`, but this is just a safeguard if we ever want to add another acceptable combination of features and targets that would not imply enabling this functionality.
What do you think?

frolvanya: Yep, looks good. Thanks!

near-sdk/Cargo.toml
```diff
@@ -105,7 +106,7 @@ unit-testing = [
     "near-crypto",
     "near-parameters",
 ]
-non-contract-usage = []
+non-contract-usage = ["ed25519-dalek", "near-crypto", "ripemd", "sha2", "sha3"]
```
frolvanya: Should we enable them only for `[target.'cfg(not(target_arch = "wasm32"))'.dependencies]`? Or is it better to leave it as it is?

It'll look something like:
```toml
[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
sha2 = { version = "0.10", optional = true }
sha3 = { version = "0.10", optional = true }
# ...
[features]
non-contract-usage = ["dep:sha2", "dep:sha3", ...]
```

It would ensure that even if contract is using `non-contract-usage` flag, then no extra deps will be included

mitinarseny: Yes, it is already like that ^

frolvanya: My bad, missed it in diff mode

near-sdk/Cargo.toml
frolvanya, line 62: Because of this `unit-testing` flag running `cargo test hash_smoke_tests --features non-contract-usage -- --nocapture` (or any other test that checks `env::*` method) still goes to mocked sys call methods path. It would be nice to be able to test new `non-contract-usage` path in env methods in CI as well

mitinarseny: @frolvanya this turned out to be not that trivial one 😅
Tried several approaches and ended up with the easiest one: made `non-contract-usage` to take precedence over `unit-testing` **only in tests**, while for third-party dependent crates, if they enable `unit-testing` then it will call `near-sys::*` functions regardless of whether `non-contract-usage` feature was enabled or not.

What do you think? I'd appreciate your thoughts on this 🙏🏻

PS: Also extended feature gates for conversions between `near-sdk` and `nearcore`-related crates for `non-contract-usage` as well. This is needed for our relayer too, so we can avoid copy-pasting these from near-sdk in our code

frolvanya: I can't think of any better approach here for now, so should be good

Thanks for fixing it!

PR #1524, opened 2026-03-21
refactor: extract `near-sdk` types into `near-sdk-core` crate

vsavchyn-dev:
This pr aims to extract types in `near-sdk` that developers might need in non-wasm environment (e.g. indexers, off-cahin apps) into `near-sdk-core` crate.

Related links:

- Initial discussion about `non-contract-usage`: https://nearone.slack.com/archives/C0A64SAMDNG/p1773415116540219
- Final decision: https://nearone.slack.com/archives/C0A64SAMDNG/p1773616628763249


near-sdk-core/src/di.rs
```diff
@@ -0,0 +1,168 @@
+#[cfg(any(target_arch = "wasm32", all(feature = "__near-sdk-unit-testing", not(test))))]
```
r-near: Curious about the name `di.rs`, is this dependency injection? It's doing the wasm-vs-native dispatch for crypto functions and abort/panic. Something like `crypto_env.rs` or `env_impl.rs` might be more self-documenting

near-sdk-core/Cargo.toml
```diff
@@ -0,0 +1,70 @@
+[package]
+name = "near-sdk-core"
+version = "0.1.0"
+authors = ["Near Inc <vladyslav.savchyn@nearone.org>"]
```
r-near: The other workspace crates don't set authors explicitly, might wanna either drop this or pull it from the workspace config for consistency

near-sdk-core/src/json_types/vector.rs
```diff
@@ -1,10 +1,12 @@
 use near_sdk_macros::near;
 use serde::{Deserialize, Deserializer, Serializer};
 
+// TODO: rewrite those examples for non-contract usage. Contract usage will be added to the
+// near-sdk directly inside the doctests.
 /// Helper class to serialize/deserialize `Vec<u8>` to base64 string.
 ///
 /// # Example
-/// ```rust
+/// ```ignore
```
r-near: Would be nice to actually fix these for the new import path (`near_sdk_core::json_types::Base64VecU8`) rather than shipping ignore'd examples. Same for the ones in `error.rs`

near-sdk-core/Cargo.toml
```diff
@@ -0,0 +1,70 @@
...
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+near-crypto = { version = "0.34", default-features = false, optional = true }
+# Pin: unicode-segmentation 1.13.1 requires Rust 1.87 but does not declare rust-version
+unicode-segmentation = { version = ">=1, <1.13.1", optional = true }
+near-vm-runner = { version = "0.34", default-features = false, optional = true }
+near-primitives-core = { version = "0.34", optional = true }
+near-parameters = { version = "0.34", optional = true }
+ripemd = "0.1.1"
+sha2 = "0.10"
+sha3 = "0.10"
```
r-near: These are only used by `di.rs` but they're unconditional for any non-wasm consumer. Someone pulling in `near-sdk-core` just for the types would still get all three hash crate families. Could these go behind an optional feature flag?

r-near (commented):
Also we should figure out what to do with `non-contract-usage`, probably remove it?

near-sdk-core/src/di.rs
vsavchyn-dev: I thought about deeper dependency injection at first via traits, but then saw that we just don't need it here. Changed the name to `env_impl.rs` and removed struct as we don't need it there: 
https://github.com/near/near-sdk-rs/pull/1524/commits/b3d064ff05d4ec06ea7a0470d54df8230c08bd05

near-sdk-core/Cargo.toml
vsavchyn-dev: They do, actually, we don't have workspace-level authors...:
https://github.com/near/near-sdk-rs/blob/de25f8c0421fa4c409532b060326ff8747cd060d/near-sdk/Cargo.toml#L4

near-sdk-core/src/json_types/vector.rs
vsavchyn-dev: Enabled them by putting near-sdk into dev-dependencies: https://github.com/near/near-sdk-rs/pull/1524/commits/b3d064ff05d4ec06ea7a0470d54df8230c08bd05

near-sdk-core/Cargo.toml
vsavchyn-dev: One of the questions I have if we should expose them to user actually - from what I see, people might need it in `non-contract-usage`? Wdyt?

It would be also nice to feature-gate this with the following block, as it enables conversion from `near_crypto::PublicKey` (not sure how useful it is though):
https://github.com/near/near-sdk-rs/pull/1524/changes#diff-045dd73cbeb9b93294a7712826664e1ec29bedbe5c5ceafe0e6c363d7b46c7f5R56-R99

Otherwise - I would remove them fully.

near-sdk-core/src/types/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+/// Weight of unused gas to use with [`promise_batch_action_function_call_weight`].
+///
+/// This weight will be used relative to other weights supplied in the function to distribute
+/// unused gas to those function calls. The default weight is 1.
+///
+/// For example, if 40 gas is leftover from the current method call and three functions specify
+/// the weights 1, 5, 2 then 5, 25, 10 gas will be added to each function call respectively,
+/// using up all remaining available gas.
+///
+/// [`promise_batch_action_function_call_weight`]: `near-sdk::env::promise_batch_action_function_call_weight`
```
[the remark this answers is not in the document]
vsavchyn-dev: edit: you actually don't need this - it was fine before :)

near-sdk/src/types/mod.rs
```diff
@@ -1,95 +1,9 @@
...
-    fn test_gas_weight_hash() {
-        use std::collections::HashSet;
-        let mut set = HashSet::new();
-        set.insert(GasWeight(1));
-        set.insert(GasWeight(2));
-        set.insert(GasWeight(1)); // duplicate
-        assert_eq!(set.len(), 2);
-    }
-}
+// VM types mostly are unused in non-sdk environment, hence, we didn't export them
```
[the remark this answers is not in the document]
vsavchyn-dev: done: https://github.com/near/near-sdk-rs/pull/1524/commits/91598f75647367a113158f7455a089bd35865d1b

near-sdk-core/README.md
```diff
@@ -0,0 +1,22 @@
+# near-sdk-core
+
+Core NEAR types for off-chain and on-chain usage.
+
+This crate provides the foundational types (`PublicKey`, `AccountId`, `U128`, `Base64VecU8`, etc.) used by the [NEAR SDK](https://crates.io/near-sdk), extracted so that off-chain applications (indexers, CLIs, frontends) can use them without pulling in the full smart contract SDK.
```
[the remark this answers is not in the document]
vsavchyn-dev: done: https://github.com/near/near-sdk-rs/pull/1524/commits/91598f75647367a113158f7455a089bd35865d1b

near-sdk-core/Cargo.toml
r-near: Let's gate them behind `__near-sdk-unit-testing` alongside `near-crypto`

They're only needed for the non-wasm test/simulation codepaths anyway, and nobody outside the crate can call them 🤷

near-sdk-core/src/state_init.rs
```diff
@@ -0,0 +1,123 @@
...
+        Self::V1(state_init)
+    }
+}
+
+#[cfg(all(not(target_arch = "wasm32"), feature = "__near-sdk-unit-testing"))]
+const _: () = {
+    use near_primitives_core::deterministic_account_id::{
+        DeterministicAccountStateInit, DeterministicAccountStateInitV1,
+    };
+
```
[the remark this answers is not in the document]
vsavchyn-dev: Makes sense, added interop to mimic `near-sdk/non-contract-usage` in https://github.com/near/near-sdk-rs/pull/1524/commits/6df64198bc6e612b6039e2a343fa3352d5fd1136

near-sdk-core/Cargo.toml
vsavchyn-dev: Looking into it, I just removed them as we are not using nor exposing these functions to a user. https://github.com/near/near-sdk-rs/pull/1524/commits/6df64198bc6e612b6039e2a343fa3352d5fd1136

r-near (approved):

near-sdk-core/src/allowance.rs
```diff
@@ -0,0 +1,153 @@
+use std::num::NonZeroU128;
+
+use borsh::{BorshDeserialize, BorshSerialize};
+use near_token::NearToken;
+use serde::{Deserialize, Serialize};
+
+/// Allow an access key to spend either an unlimited or limited amount of gas
```
vsavchyn-dev, line 7: @r-near is it "amount of gas" or "amount of tokens"? From what I remember it is token allowance for the function-call key, right?

r-near: Yeah it should be amount of tokens

near-state-init/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+pub enum StateInit {
+    V1(StateInitV1) = 0,
+}
+
+impl StateInit {
+    /// Derives [`AccountId`] deterministically, according to NEP-616.
+    #[inline]
+    pub fn derive_account_id(&self) -> AccountId {
+        let serialized = borsh::to_vec(self).unwrap_or_else(|_| unreachable!());
+        format!("0s{}", hex::encode(&near_env::keccak256_array(&serialized)[12..32]))
```
mitinarseny: Can we avoid depending on near-env at all unless `near` feature is enabled?

near-global-contracts/src/state_init/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+        #[cfg(not(feature = "near-contracts"))]
+        {
+            use sha3::Digest;
+
+            let mut hasher = sha3::Keccak256::new();
+            borsh::to_writer(&mut hasher, self).unwrap_or_else(|_| unreachable!());
+            let hash = hasher.finalize();
+            format!("0s{}", hex::encode(&hash[12..32])).parse().unwrap_or_else(|_| unreachable!())
+        }
+    }
```
mitinarseny: 1. don't duplicate `format!("0s{}", ...).parse()...`, reuse it
2. I think it's now safe to create `AccountId` without validation to improve performance
3. use [`digest-io::IoWrapper`](https://docs.rs/digest-io/latest/digest_io/struct.IoWrapper.html) and serialize directly to it to avoid Vec allocations. See an example [here](https://github.com/near/intents/blob/2583d1c266f63ca9471eaa25233aeb45ba08e70f/crates/outlayer/primitives/src/crypto.rs#L35-L49)

near-global-contracts/src/state_init.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+#[cfg_attr(feature = "borsh", derive(borsh::BorshSerialize, borsh::BorshDeserialize))]
+#[cfg_attr(feature = "abi", derive(borsh::BorshSchema))]
+pub struct StateInitV1 {
+    pub code: GlobalContractId,
+    pub data: BTreeMap<Vec<u8>, Vec<u8>>,
+}
```
mitinarseny, line 98: derive `serde` traits, don't forget `#[serde_as(as = "BTreeMap<Base64, Base64>")]` for `data`

near-global-contracts/src/state_init/serde_impl.rs
```diff
@@ -0,0 +1,58 @@
...
+use serde_with::base64::Base64;
+use std::collections::BTreeMap;
+
+#[serde_with::serde_as]
+#[derive(serde::Serialize, serde::Deserialize)]
+struct StateInitV1Helper {
+    code: GlobalContractId,
+    #[serde_as(as = "BTreeMap<Base64, Base64>")]
+    data: BTreeMap<Vec<u8>, Vec<u8>>,
+}
```
mitinarseny: Avoid using helpers for deserialization, derive it directly on `StateInitV1`

near-global-contracts/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+mod global_contract_identifier;
+pub use global_contract_identifier::*;
+
+mod state_init;
+pub use state_init::*;
+
+#[cfg(test)]
+// XXX: `near-sdk` was added in order to enable tests and doctests compiling with mockchain
+use near_sdk as _;
```
mitinarseny: So, you're not testing without `near-contract` feature enabled?

```suggestion
#[cfg(all(test, feature = "near-contracts"))]
// XXX: `near-sdk` was added in order to enable tests and doctests compiling with mockchain
use near_sdk as _;
```

near-global-contracts/src/global_contract_identifier.rs
```diff
@@ -1,10 +1,9 @@
 use near_account_id::AccountId;
-use near_sdk_macros::near;
 
-use crate::CryptoHash;
-use crate::json_types::Base58CryptoHash;
+use near_sdk_core::json_types::Base58CryptoHash;
+use near_sdk_core::types::CryptoHash;
```
mitinarseny: Please, avoid any dependencies on `near-sdk*` crates here

near-global-contracts/src/state_init/schemars_impl.rs
```diff
@@ -0,0 +1,28 @@
...
+            instance_type: Some(schemars::schema::InstanceType::Object.into()),
+            object: Some(Box::new(schemars::schema::ObjectValidation {
+                properties,
+                required: ["code".to_string(), "data".to_string()].into(),
+                ..Default::default()
+            })),
+            ..Default::default()
+        })
+    }
+}
```
mitinarseny: why not just `#[cfg_attr(all(feature = "abi", feature = "serde"), derive(schemars::JsonSchema))]` on `StateInitV1`?

near-global-contracts/Cargo.toml
```diff
@@ -0,0 +1,77 @@
...
+categories.workspace = true
+repository.workspace = true
+homepage.workspace = true
+rust-version.workspace = true
+
+[dependencies]
+near-account-id = "2.6.0"
+near-sdk-core = { path = "../near-sdk-core/", version = "0.1.0" }
+
+sha3 = "0.10"
```
mitinarseny: This should be optional, only if `near-contracts` feature is not enabled. Maybe add `digest` feature?

mitinarseny: I would also suggest to guard all near-sdk-related crates behind `[target.'cfg(target_arch = "wasm32")'.dependencies]`, as an addition to enabled `near-contract` feature

near-env/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+        not(target_arch = "wasm32"),
+        any(not(feature = "__near-sdk-unit-testing"), test, doctest)
+    ))]
+    {
+        panic!()
+    }
+}
+
+pub fn panic_str(message: &str) -> ! {
+    #[cfg(any(target_arch = "wasm32", all(feature = "__near-sdk-unit-testing", not(test))))]
```
r-near: Do we need `not(doctest)` here (like abort above)?

near-sdk-core/Cargo.toml
```diff
@@ -0,0 +1,106 @@
...
+    "json",
+], optional = true }
+borsh = { version = "1.0.0", features = ["derive"], optional = true }
+base64 = "0.22"
+bs58 = "0.5"
+hex = "0.4"
+
+arbitrary = { version = "1.4", optional = true }
+
+near-account-id = { version = "2.3.0" }
```
r-near, line 32: Should probably update this to match the version in global contracts

```suggestion
near-account-id = { version = "2.6.0" }
```

near-sdk-core/src/types/public_key.rs
```diff
@@ -162,7 +159,16 @@ impl PublicKey {
...
     pub fn curve_type(&self) -> CurveType {
-        CurveType::from_u8(self.data[0]).unwrap_or_else(|_| crate::env::abort())
+        CurveType::from_u8(self.data[0]).unwrap_or_else(|_| {
+            #[cfg(feature = "near-contracts")]
+            {
+                near_env::abort()
+            }
+            #[cfg(not(feature = "near-contracts"))]
+            {
+                panic!()
```
r-near, line 173: `abort()` shows the bad byte, but I don't think `panic!()` does...

Maybe this so it's debuggable off-chain?
```suggestion
                panic!("invalid curve type byte: {}", self.data[0])
```

Just double-check, I might be missing something.

near-env/src/hash.rs
```diff
@@ -0,0 +1,231 @@
...
+            read_register_fixed(ATOMIC_OP_REGISTER)
+        }
+    }
+
+    #[cfg(all(
+        not(target_arch = "wasm32"),
+        any(not(feature = "__near-sdk-unit-testing"), test, doctest)
+    ))]
+    {
+        use sha2::Digest;
```
r-near: Shouldn't this be ripemd?


```suggestion
        use ripemd::Digest;
```

EDIT: I guess it's the same trait re-exported 🤷

near-sdk-core/src/types/error.rs
```diff
@@ -80,7 +88,14 @@ impl std::fmt::Display for Abort {
...
 impl FunctionError for Abort {
     fn panic(&self) -> ! {
-        crate::env::abort()
+        #[cfg(feature = "near-contracts")]
+        {
+            near_env::abort()
+        }
+        #[cfg(not(feature = "near-contracts"))]
+        {
+            panic!()
```
r-near, line 97: Nice, does this run cleanly on non-wasm? I can't recall if we're running doctests in CI

near-env/README.md
```diff
@@ -0,0 +1,16 @@
...
+- wasm32 (contract builds): routed through NEAR VM host calls
+- non-wasm32 (off-chain tools, tests): fallback to Rust implementations (`sha2`, `sha3`, `ripemd`)
+
+Please be mindful that some of the functions like `random_seed_array()` or `block_height()` that are available in `near-sdk` might not be available in `near-env`, as they require direct interaction with blockchain and it would be impossible to reliably mimic this behaviour for off-chain services.
+
+## License
+
+Licensed under either of:
+
+- Apache License, Version 2.0 ([LICENSE_APACHE](../LICENSE-APACHE))
```
r-near: nit: hyphen instead of underscore
```suggestion
- Apache License, Version 2.0 ([LICENSE-APACHE](../LICENSE-APACHE))
```

near-sdk/src/state_init.rs
```diff
@@ -1,126 +1,5 @@
-use std::collections::BTreeMap;
+//! Deterministic account state initialization types (NEP-616)
+//!
+//! Types are defined in [near_global-contracts] and re-exported here for the ease of use.
```
r-near: nit: does this render on the docs site?

```suggestion
//! Types are defined in [near_global_contracts] and re-exported here for the ease of use.
```

r-near (commented):
Minor comments

near-sdk/src/lib.rs
```diff
@@ -1624,6 +1624,9 @@ pub mod state;
 #[cfg(feature = "deterministic-account-ids")]
 pub mod state_init;
 
+#[cfg(feature = "deterministic-account-ids")]
+pub use near_global_contracts::{AccountContract, GlobalContractId};
```
r-near: Worth flagging in the Changelog that the import paths will break unless users explicitly enable this feature

near-sdk-core/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+pub use serde_with;
+
+#[cfg(feature = "borsh")]
+pub use borsh;
+
+#[cfg(feature = "abi")]
+pub use schemars;
+
+#[cfg(test)]
+// XXX: `near-sdk` was added in order to enable tests and doctests compiling with mockchain
```
r-near: This is kinda awkward...is there a way to do this without the circular dep (e.g., feature-gate the doctests, or move them into integration tests in near-sdk or smth)?

If it adds too much complexity, don't bother.

near-sdk-core/src/lib.rs
```diff
@@ -0,0 +1,27 @@
+pub mod json_types;
+
+pub mod types;
```
r-near, line 7: Thoughts on adding some top-level re-exports? for convenience?

near-global-contracts/src/state_init.rs
```diff
@@ -0,0 +1,114 @@
...
+            encoded_acc_id = hex::encode(&near_env::keccak256_array(&serialized)[12..32]);
+        }
+        #[cfg(all(feature = "digest", not(feature = "near-contracts")))]
+        {
+            use sha3::Digest;
+
+            let mut hasher = sha3::Keccak256::new();
+            borsh::to_writer(&mut hasher, self).unwrap_or_else(|_| unreachable!());
+            encoded_acc_id = hex::encode(&hasher.finalize()[12..32]);
+        }
```
r-near: Doesn't near-env handle this now? Or do we still need two feature-gated implementations?

near-crypto-hash/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
```
r-near, line 1: Sorry if you already addressed this, but why do we have a separate crate for this?

near-env/src/lib.rs
```diff
@@ -0,0 +1,78 @@
+//! Low-level abstraction over `near-sys` host functions. Provides the same
```
r-near, line 1: The dispatch design is half-cfg, half-feature...
- `cfg(target_arch = "wasm32")` _and_
- `feature = "__near-sdk-unit-testing"` _and_ 
- `cfg(test)` _and_
- `cfg(doctest)` _and_

Maybe we could consolidate this?

near-global-contracts/src/state_init/mod.rs
vsavchyn-dev: 1. Fixed in in https://github.com/near/near-sdk-rs/pull/1524/commits/853f6d233e253c9317b079236002888eac769d52

2. Not sure how to do it... You mean using something like [`new_unvalidated`](https://github.com/near/near-account-id-rs/blob/main/src/account_id_ref.rs#L95-L110)?

3. I guess it would make sense to use it for `near-env` as well then?

near-global-contracts/src/state_init.rs
vsavchyn-dev: Fixed in https://github.com/near/near-sdk-rs/pull/1524/commits/853f6d233e253c9317b079236002888eac769d52

Thx for showing the `cfg_eval`!

near-global-contracts/src/state_init/serde_impl.rs
vsavchyn-dev: Removed in https://github.com/near/near-sdk-rs/pull/1524/commits/853f6d233e253c9317b079236002888eac769d52

near-global-contracts/src/lib.rs
vsavchyn-dev: `near-sdk` already enables `near-contracts` for `near-global-contracts`.

Still, as I removed `near-sdk` as dev-dep in https://github.com/near/near-sdk-rs/pull/1524/changes/853f6d233e253c9317b079236002888eac769d52, it shouldn't matter now.

near-global-contracts/src/global_contract_identifier.rs
vsavchyn-dev: Done in https://github.com/near/near-sdk-rs/pull/1524/changes/853f6d233e253c9317b079236002888eac769d52

near-global-contracts/src/state_init/schemars_impl.rs
vsavchyn-dev: There was an error when adding the `#[cfg_attr(feature = "abi"), derive(schemars::JsonSchema))]`, which you showed how to fix in Slack.

Hence, removed in https://github.com/near/near-sdk-rs/pull/1524/changes/853f6d233e253c9317b079236002888eac769d52

near-global-contracts/Cargo.toml
vsavchyn-dev: > This should be optional, only if near-contracts feature is not enabled. Maybe add digest feature?

Introduced `digest` feature (might call it something else later, as it is not just about digest), in https://github.com/near/near-sdk-rs/pull/1524/changes/853f6d233e253c9317b079236002888eac769d52


> I would also suggest to guard all near-sdk-related crates behind [target.'cfg(target_arch = "wasm32")'.dependencies], as an addition to enabled near-contract feature

I don't want to do that, for a simple reason of having `non-contract-usage` feature in sdk (for now), that relies on `near-env`'s non-wasm implementation (and I just want to keep things consistent with `near-sdk-core`).

near-sdk-core/src/lib.rs
vsavchyn-dev: Current docs that we have (aka copy-pasted ones) would either require circular dep or rewrite.

near-sdk-core/src/lib.rs
vsavchyn-dev: Well, I guess it's easier from sdk standpoint to see what goes where, and it's a bit more straightforward?

But I have no objections for top-level re-export. Will do it

near-global-contracts/src/state_init.rs
vsavchyn-dev: It does, but we don't want to pull additional `ripemd` and `sha2`.

Also, it was discussed [here](https://nearone.slack.com/archives/C0A64SAMDNG/p1777758407705199?thread_ts=1775998883.678839&cid=C0A64SAMDNG)

near-crypto-hash/Cargo.toml
vsavchyn-dev: Check [this discussion](https://nearone.slack.com/archives/C0A64SAMDNG/p1777758407705199?thread_ts=1775998883.678839&cid=C0A64SAMDNG)

TLDR; people don't want to rely on `near-sdk-core` + `near-env`

near-env/src/lib.rs
vsavchyn-dev: Hmm, I guess it would've flagged it during `cargo-doc` CI job, no?

But good catch! Fixed in https://github.com/near/near-sdk-rs/pull/1524/commits/f58881649a2b7af9861fc4d1c0927185d5fe32a3

near-sdk-core/Cargo.toml
vsavchyn-dev: It will resolve to latest version, I made near-global-contract's version match sdk in general in https://github.com/near/near-sdk-rs/pull/1524/commits/7fc350c0d66583d9391be28c43d8f1536a4b02d0

near-sdk-core/src/types/public_key.rs
vsavchyn-dev: Not sure, will check it

near-env/src/hash.rs
vsavchyn-dev: It can be, but they [reuse](https://docs.rs/ripemd/latest/ripemd/#:~:text=See%20the%20digest%20crate%20docs%20for%20additional%20examples.) the same `digest` crate.

I don't think it hurts to change it (https://github.com/near/near-sdk-rs/pull/1524/commits/e3d1213924e9b09c57fa6b1fc72530420175ab84)

near-sdk-core/src/types/error.rs
vsavchyn-dev: Yes, it runs on CI - we do doctests as far as I remember (`cargo-doc` step inside `TestCore` job).

near-env/README.md
vsavchyn-dev: Fixed in https://github.com/near/near-sdk-rs/pull/1524/commits/0862f06c2e4a3d9885ccb2ec234c7a89fdd6c0b0

near-sdk/src/state_init.rs
vsavchyn-dev: Nope, it doesn't :)

I guess adding better docs for re-exports would be the next step in this PR...

Fixed in https://github.com/near/near-sdk-rs/pull/1524/commits/e628a6f0f6fd271ad4161177a3aaa961589baa2f

near-sdk/src/lib.rs
vsavchyn-dev: ~~It was like this before, no...?~~
Edit: it wasn't like this before :)

`AccountContract` is re-exported from `contract_code` into [`src/types/mod.rs`](https://github.com/near/near-sdk-rs/blob/addaa0c28cd208b0df41a2341e1aa5a0df6b6ad7/near-sdk/src/types/mod.rs#L10-L11), and then from `src/types/mod.rs` into [`src/lib.rs`](https://github.com/near/near-sdk-rs/blob/addaa0c28cd208b0df41a2341e1aa5a0df6b6ad7/near-sdk/src/lib.rs#L1616-L1617).

But maybe gating it behind the feature doesn't make sense. Fixed in https://github.com/near/near-sdk-rs/pull/1524/commits/2512dcc8ad85c504d5fa63aed9c3c4482d5b62eb

near-env/src/lib.rs
vsavchyn-dev: Consolidated in https://github.com/near/near-sdk-rs/pull/1524/commits/af92731f1591bdadd553607614002e184cfe2947

near-sdk-core/src/types/public_key.rs
vsavchyn-dev: Well, while we parse the `CurveType`, `abort()` doesn't show bad byte here, as for contracts we call `core::arch::wasm32::unreachable()`, which shouldn't display anything (`panic_str` should've been used here then).

Matching this behaviour, `panic!()` doesn't display anything.

near-crypto-hash/src/base58_hash.rs
mitinarseny, line 1: This will be not needed anymore thanks to [base58 support](https://github.com/jonasbb/serde_with/pull/943) in `serde_with` to be released this weekend.

near-env/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+sha2 = { version = "0.10" }
+sha3 = { version = "0.10" }
+
+[dev-dependencies]
+# XXX: `near-sdk` was added in order to enable testing and doctests
+near-sdk = { path = "../near-sdk", features = ["unit-testing"] }
+hex = { version = "0.4" }
+
+[features]
+__near-sdk-unit-testing = ["dep:near-sys"]
```
mitinarseny, line 38: Can we move all mocking stuff into `near-env` and expose it as a real feature, so that crates depending on it can enable it for testing purposes without having to add additional dev-dependency on `near-sdk`?

```suggestion
unit-testing = ["dep:near-sys"]
```

near-global-contracts/src/global_contract_identifier.rs
```diff
@@ -13,17 +12,18 @@ pub enum AccountContract {
...
+#[cfg_attr(feature = "borsh", derive(borsh::BorshSerialize, borsh::BorshDeserialize))]
+#[cfg_attr(feature = "borsh", borsh(use_discriminant = true))]
+#[cfg_attr(all(feature = "abi", not(target_arch = "wasm32")), derive(borsh::BorshSchema))]
+#[cfg_attr(all(feature = "abi", not(target_arch = "wasm32")), derive(schemars::JsonSchema))]
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[repr(u8)]
 pub enum GlobalContractId {
-    #[serde(rename = "hash")]
+    #[cfg_attr(feature = "serde", serde(rename = "hash"))]
     CodeHash(Base58CryptoHash) = 0,
```
mitinarseny, line 32: Can we please avoid using newtype pattern for only deserialization overwrite? It's non-idiomatic and doesn't make sense without `serde` feature enabled. Moreover, this [won't be needed soon](https://github.com/near/near-sdk-rs/pull/1524/changes#r3211280034).

near-global-contracts/src/state_init.rs
```diff
@@ -0,0 +1,114 @@
...
+use std::collections::BTreeMap;
+
+#[cfg(feature = "serde")]
+use serde_with::base64::Base64;
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "borsh", derive(borsh::BorshSerialize, borsh::BorshDeserialize))]
+#[cfg_attr(feature = "borsh", borsh(use_discriminant = true))]
```
mitinarseny: nit
```suggestion
#[cfg_attr(
    feature = "borsh",
    derive(borsh::BorshSerialize, borsh::BorshDeserialize),
    borsh(use_discriminant = true)
)]
```

near-global-contracts/src/state_init.rs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            use sha3::Digest;
+
+            let mut hasher = sha3::Keccak256::new();
+            borsh::to_writer(&mut hasher, self).unwrap_or_else(|_| unreachable!());
+            encoded_acc_id = hex::encode(&hasher.finalize()[12..32]);
+        }
+
+        // SAFETY: Keccak will always generate 32 bytes
+        format!("0s{encoded_acc_id}").parse().unwrap_or_else(|_| unreachable!())
```
mitinarseny: You can avoid encoding in two places and skip AccountId validation:
```suggestion
        let hash: [u8; 32];

        #[cfg(feature = "near-contracts")]
        {
            let serialized = borsh::to_vec(self).unwrap_or_else(|_| unreachable!());
            hash = near_env::keccak256_array(&serialized)[12..32];
        }
        #[cfg(all(feature = "digest", not(feature = "near-contracts")))]
        {
            use sha3::Digest;

            let mut hasher = sha3::Keccak256::new();
            borsh::to_writer(&mut hasher, self).unwrap_or_else(|_| unreachable!());
            hash = hasher.finalize()[12..32];
        }

        // SAFETY: Keccak will always generate 32 bytes
        AccountId::new_unvalidated(format!("0s{}", hex::encode(&hash)))
```

near-global-contracts/src/state_init.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+#[cfg_attr(feature = "serde", cfg_eval::cfg_eval, serde_with::serde_as)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
```
mitinarseny: nit
```suggestion
#[cfg_attr(
    feature = "serde",
    cfg_eval::cfg_eval,
    serde_with::serde_as,
    derive(serde::Serialize, serde::Deserialize)
)]
```

near-global-contracts/Cargo.toml
```diff
@@ -0,0 +1,81 @@
...
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+schemars = { version = "0.8.22", optional = true }
+
+[features]
+default = []
+
+serde = [
+    "dep:serde",
+    "dep:serde_json",
```
mitinarseny: What do you need `serde_json` here for?

near-global-contracts/Cargo.toml
```diff
@@ -0,0 +1,81 @@
...
+
+borsh = ["dep:borsh", "near-account-id/borsh", "near-crypto-hash/borsh"]
+
+abi = [
+    "dep:schemars",
+    "borsh",
+    "borsh/unstable__schema",
+    "near-account-id/abi",
+    "near-crypto-hash/abi",
+]
```
mitinarseny, line 83: Can we have separate features for `schemars_0_8` and `borsh_schema`?

near-global-contracts/Cargo.toml
```diff
@@ -0,0 +1,81 @@
...
+
+near-primitives-interop = ["dep:near-primitives-core"]
+
+digest = ["borsh", "dep:sha3", "dep:hex"]
+
+near-contracts = [
+    "dep:near-env",
+    "dep:hex",
+    "serde",
+    "borsh",
```
mitinarseny, line 97: Why do you want to enable `serde` and `borsh` for `near-contracts` unconditionally?

near-global-contracts/Cargo.toml
```diff
@@ -0,0 +1,81 @@
...
+homepage.workspace = true
+rust-version.workspace = true
+
+[dependencies]
+near-account-id = "2.3.0"
+near-crypto-hash = { path = "../near-crypto-hash/", version = "~0.1.0" }
+
+hex = { version = "0.4", optional = true }
+
+cfg_eval = { version = "0.1" }
```
mitinarseny: This is only needed when `serde` feature enabled

near-global-contracts/Cargo.toml
```diff
@@ -0,0 +1,81 @@
...
+serde = { version = "1", features = ["derive"], optional = true }
+serde_json = { version = "1", features = ["preserve_order"], optional = true }
+serde_with = { version = "3", features = [
+    "base64",
+    "hex",
+    "json",
```
mitinarseny: JSON is not used in this crate
```suggestion
```

near-global-contracts/Cargo.toml
```diff
@@ -0,0 +1,81 @@
...
+], optional = true }
+borsh = { version = "1.0.0", features = ["derive"], optional = true }
+
+arbitrary = { version = "1.4", optional = true }
+
+sha3 = { version = "0.10", optional = true }
```
mitinarseny: There is a newer version available, should we use it?
```suggestion
sha3 = { version = "0.11", optional = true }
```

near-global-contracts/Cargo.toml
```diff
@@ -0,0 +1,81 @@
...
+    "borsh",
+    "near-crypto-hash/near-contracts",
+]
+
+__near-sdk-unit-testing = [
+    "near-contracts",
+    "near-primitives-interop",
+    "near-crypto-hash/__near-sdk-unit-testing",
+    "near-env/__near-sdk-unit-testing",
+]
```
mitinarseny: [Same](https://github.com/near/near-sdk-rs/pull/1524/changes#r3211295906)

near-global-contracts/src/state_init.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+impl StateInitV1 {
+    #[inline]
+    pub const fn code(code: GlobalContractId) -> Self {
+        Self { code, data: BTreeMap::new() }
+    }
+
+    #[inline]
+    pub fn with_data_entry(mut self, key: Vec<u8>, value: Vec<u8>) -> Self {
```
mitinarseny: Let's have a nicer API, so it can be used as `state_init.with_data_entry(b"key", b"value")`:
```suggestion
    pub fn with_data_entry(mut self, key: impl Into<Vec<u8>>, value: impl Into<Vec<u8>>) -> Self {
```

near-global-contracts/src/global_contract_identifier.rs
```diff
@@ -1,10 +1,9 @@
...
 
-use crate::CryptoHash;
-use crate::json_types::Base58CryptoHash;
+use near_crypto_hash::Base58CryptoHash;
+use near_crypto_hash::CryptoHash;
 
-#[cfg_attr(feature = "arbitrary", derive(::arbitrary::Arbitrary))]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum AccountContract {
```
mitinarseny, line 8: It seems that this type should be in `near-env`, because it's a result of `current_contract_code()` host-function and does not play any role in StateInit derivation

near-sdk/Cargo.toml
```diff
@@ -108,15 +120,18 @@ unit-testing = [
...
     "ed25519-dalek",
     "near-crypto",
     "unicode-segmentation",
     "near-primitives-core",
-    "ripemd",
-    "sha2",
-    "sha3",
+    "near-sdk-core/near-crypto-interop",
+    "near-sdk-core/near-primitives-interop",
+    "near-global-contracts/near-primitives-interop",
```
mitinarseny, line 138: Maybe separate into another feature? So that projects that depend on this feature to be lightweight can continue to do so without heavy-lifting of migrating to separate dependencies on `near-account-id`, `near-token`, etc...

mitinarseny (commented):
Thanks for working on it!
Looking forward on making these crates truly reusable in all use cases soon

near-global-contracts/src/state_init.rs
mitinarseny: Also, just an idea: maybe to introduce `near-digest` crate with wrappers around host-functions like [this one](https://github.com/near/intents/blob/1823243edc5f8093ec1225393d35854251d2ef07/crates/near/utils/src/digest.rs#L4-L32)? Then it can be nicely re-used here and we will [use it](https://github.com/near/intents/pull/268#discussion_r3211255219) in our repos as well!

near-sdk-core/src/types/public_key.rs
r-near: Ah...welp, we can fix this in a follow-up

near-global-contracts/src/global_contract_identifier.rs
mitinarseny: `serde_with v3.20.0` has been [released](https://github.com/jonasbb/serde_with/releases/tag/v3.20.0)! You can now use `#[serde_as(as = "Base58")]` and avoid `#[schemars(with = "String")]` for both base58 and base64

near-global-contracts/Cargo.toml
```diff
@@ -0,0 +1,94 @@
...
+near-crypto-hash = { path = "../near-crypto-hash/", version = "~0.1.0" }
+
+hex = { version = "0.4", optional = true }
+
+cfg_eval = { version = "0.1", optional = true }
+serde = { version = "1", features = ["derive"], optional = true }
+serde_with = { version = "3", features = ["base64", "hex"], optional = true }
+borsh = { version = "1.0.0", features = ["derive"], optional = true }
+
+arbitrary = { version = "1.4", optional = true }
```
r-near: Do we need to add the `derive` feature here?

near-global-contracts/src/state_init.rs
```diff
@@ -0,0 +1,129 @@
...
+#[repr(u8)]
+pub enum StateInit {
+    V1(StateInitV1) = 0,
+}
+
+impl StateInit {
+    /// Derives [`AccountId`](near_account_id::AccountId) deterministically, according to NEP-616.
+    #[inline]
+    #[cfg(all(feature = "borsh", any(near, feature = "digest")))]
+    pub fn derive_account_id(&self) -> near_account_id::AccountId {
```
r-near, line 43: Can we get a parity test here? Just a round-trip test with a hardcoded accountid/stateinit pair.

near-sdk-core/Cargo.toml
```diff
@@ -0,0 +1,115 @@
...
+borsh = [
+    "dep:borsh",
+    "near-account-id/borsh",
+    "near-gas/borsh",
+    "near-token/borsh",
+    "near-crypto-hash/borsh",
+]
+
+abi = [
+    "borsh/unstable__schema",
```
r-near, line 89: Don't we need to add the local `borsh` feature as well?

.github/workflows/test.yml
```diff
@@ -8,6 +8,60 @@ env:
...
+        with:
+          cache-provider: warpbuild
+          shared-key: "${{ matrix.crate }}-${{ matrix.platform.name }}-${{ matrix.platform.rs }}"
+          cache-all-crates: true
+
+      - name: Print rustc && rustdoc version
+        run: rustc --version && rustdoc --version
+
+      - name: Cargo hack check non-wasm32 powerset
+        run: cargo hack check -p ${{ matrix.crate }} --feature-powerset --no-dev-deps --mutually-exclusive-features digest,near-contracts
```
r-near: `near-contracts` no longer exists, right? not sure how this passed 🤷

near-crypto-hash/src/base58_hash.rs
vsavchyn-dev: `serde_with` v3.20.0 requires rustc 1.88, and we are stuck at 1.86 (for now). With wasmtime release, we should be able to use it. So I created a reminder #1535

near-env/Cargo.toml
vsavchyn-dev: I think we can do that later with a quick follow-up. I am still trying to understand if we can do this somehow better and if we should move most of non-sdk-related types from `near-sdk-core` into `near-env`.

near-global-contracts/src/global_contract_identifier.rs
vsavchyn-dev: serde_with v3.20.0 requires rustc 1.88, and we are stuck at 1.86 (for now). With wasmtime release, we should be able to use it. So I created a reminder https://github.com/near/near-sdk-rs/issues/1535

But I think we can change it later with update of the nearcore to wasmtime (aka bundle it with update of serde_with)

near-global-contracts/src/state_init.rs
vsavchyn-dev: Fixed in https://github.com/near/near-sdk-rs/pull/1524/commits/19215f41b8d9f1938611cac5ac4b002cd42e61e1

near-global-contracts/src/state_init.rs
vsavchyn-dev: Fixed in https://github.com/near/near-sdk-rs/pull/1524/commits/fb34f519fe5aec10ef3c50445d3886cf285b66a4

Regarding `near-digest` - I will either create a new repo or put it here in a follow-up pr, but it does sound like a nice idea for contracts (considering that `digest` released it's own `std::io` wrappers).

near-global-contracts/src/state_init.rs
vsavchyn-dev: Fixed in https://github.com/near/near-sdk-rs/pull/1524/commits/19215f41b8d9f1938611cac5ac4b002cd42e61e1

near-global-contracts/Cargo.toml
vsavchyn-dev: No need, you are right, fixed it https://github.com/near/near-sdk-rs/pull/1524/changes/9027f30fe19cb4b94e180aa1207a91cde4a8b4fa

near-global-contracts/Cargo.toml
vsavchyn-dev: I made it similar to `near-account-id` with only `schemars-v0_8` feature in: https://github.com/near/near-sdk-rs/pull/1524/commits/0c736e2d83aba7898f15418566c2bf1a04c7a195

vsavchyn-dev: I am curious about `borsh_schema` feature though - what is other use case of it instead of ABI?
I just don't have much knowledge why would it be used without abi...

near-global-contracts/Cargo.toml
vsavchyn-dev: Idea was to nicely bundle this `near-contract` feature for nicer usage in `near-sdk`. But now it's not used anymore with introduction of `--cfg near`: https://github.com/near/near-sdk-rs/pull/1524/commits/1f82e7f951d23ef4d4f517bd2bbe21e735270ed0

near-global-contracts/Cargo.toml
vsavchyn-dev: fixed in https://github.com/near/near-sdk-rs/pull/1524/changes/9027f30fe19cb4b94e180aa1207a91cde4a8b4fa

near-global-contracts/Cargo.toml
vsavchyn-dev: fixed in https://github.com/near/near-sdk-rs/pull/1524/changes/9027f30fe19cb4b94e180aa1207a91cde4a8b4fa

near-global-contracts/Cargo.toml
vsavchyn-dev: I tried, but apparently they introduce some new wrappers that do not have `std::io::Writer` implemented (they moved them into `std::io` into `digest-io` here https://github.com/RustCrypto/traits/pull/1809). I will double-check this one and hopefully can resolve it, but any help is appreciated as I don't have much knowledge about RustCrypto's repos...

near-global-contracts/src/state_init.rs
vsavchyn-dev: Fixed in https://github.com/near/near-sdk-rs/pull/1524/changes/9027f30fe19cb4b94e180aa1207a91cde4a8b4fa

near-global-contracts/Cargo.toml
vsavchyn-dev: Yup, we need it for `derive(borsh::BorshDeserialize, borsh::BorshSerialize)`

near-sdk-core/Cargo.toml
vsavchyn-dev: Fixed in https://github.com/near/near-sdk-rs/pull/1524/commits/88b1fcd9b4f83c022522182e85a86bdd70a5c595

.github/workflows/test.yml
vsavchyn-dev: Fixed in https://github.com/near/near-sdk-rs/pull/1524/changes/948750a47500b63f772e3345990aa492db4ebd7f

near-sdk/Cargo.toml
vsavchyn-dev: Not sure what you mean by that, dependencies for `near-primitives-interop` and `near-crypto-interop`  features are pulled only when you enable previously mentioned features or when you use `non-contract-usage`/`unit-testing`/`__near-sdk-unit-testing`.
just in case it was like that before, so nothing "new" was added: https://github.com/near/near-sdk-rs/blob/master/near-sdk/src/types/contract_code.rs#L51-L55

vsavchyn-dev: I just wanted to be sure that people that use `non-contract-usage` have some time to adapt and see if there is some problems with new leaf crates

near-global-contracts/src/global_contract_identifier.rs
vsavchyn-dev: We don't have the `current_contract_code()` host-function in `near-env` yet and we would to export some types unrelated to `near-sdk` from `near-sdk-core` to `near-env`. Also, it was easier to make only one crate optional for the `deterministic-account-ids` feature.

Hence, I think we can bundle this with `unit-testing`/mockchain work for the `near-env`

near-global-contracts/Cargo.toml
r-near: Yes, I mean, should we add it?
```suggestion
arbitrary = { version = "1.4", features = ["derive"], optional = true }
```

vsavchyn-dev: For some reason we don't need it for ˋarbitrary`

mitinarseny: It's a coincidence, because enabling `near-account-id/arbitrary` already enables `arbitrary/derive`. But I would add it here anyway, so that if those dependencies are incompatible by semver, you'd still enable this feature on your `arbitrary` dependency anyway

near-global-contracts/Cargo.toml
```diff
@@ -0,0 +1,100 @@
...
+near-account-id = { version = "2.3.0", features = [
+    "internal_unstable",
+    "serde",
+    "borsh",
+] }
+near-crypto-hash = { path = "../near-crypto-hash/", version = "~0.1.0" }
+near-env = { path = "../near-env/", version = "~0.1.0" }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+schemars-v0_8 = { version = "0.8.22", optional = true, package = "schemars" }
```
mitinarseny: 1. Why to unconditionally enable serde, hex, borsh, cfg_eval, etc...? Can it be enabled by `near-sdk` itself?
2. Why to make optional schemars accessible only for non-wasm builds? Can we enable it by enabling a corresponding feature, regardless of the architecture being used? We might want to access it in WASM environment.  Thus, the question is: can we avoid target-specific deps everywhere except for `near-sdk` itself?

near-global-contracts/Cargo.toml
```diff
@@ -0,0 +1,100 @@
...
+
+arbitrary = [
+    "dep:arbitrary",
+    "near-account-id/arbitrary",
+    "near-crypto-hash/arbitrary",
+]
+
+near-primitives-interop = ["dep:near-primitives-core"]
+
+digest = ["borsh", "dep:sha3", "dep:hex"]
```
mitinarseny: Having a separate `digest` feature would lead to transitive features hell. Actually, it's not needed anymore thanks to `cfg(near)`. Instead, `sha3` can be a required dependency for `not(cfg(near))`.

near-env/Cargo.toml
mitinarseny: > I think we can do that later with a quick follow-up

Then, we wouldn't be able to test crates that depend on `near-env` alone... Adding additional dependency on `near-sdk` doesn't help in general case: you can have two versions of `near-sdk` compiled if you use incompatible semvers of it

How about gating "mock functionality" by `cfg(all(near, not(target_family = "wasm")))`?

near-global-contracts/Cargo.toml
```diff
@@ -0,0 +1,100 @@
...
+
+__near-sdk-unit-testing = [
```
mitinarseny, line 93: [Same](https://github.com/near/near-sdk-rs/pull/1524/changes#r3298893078): gate it by `cfg(all(near, not(target_family = "wasm")))`, not by feature?

near-global-contracts/src/state_init.rs
```diff
@@ -0,0 +1,134 @@
+use crate::GlobalContractId;
+use std::collections::BTreeMap;
+
+#[cfg(any(near, feature = "serde"))]
+use serde_with::base64::Base64;
+
+#[cfg(feature = "schemars-v0_8")]
+use schemars_v0_8 as schemars;
```
mitinarseny: This will cause compilation errors when you'd want to add `schemars_v0_9` and `schemars_v1`. Better use FQDNs instead: `#[cfg_attr(feature = "schemars_v0_8", derive(::schemars_v0_8::JsonSchema), schemars(crate = "::schemars_v0_8"))]`

near-global-contracts/src/state_init.rs
```diff
@@ -0,0 +1,134 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+#[cfg_attr(any(near, feature = "serde"), derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(
+    any(near, feature = "borsh"),
+    derive(borsh::BorshSerialize, borsh::BorshDeserialize),
+    borsh(use_discriminant = true)
+)]
```
mitinarseny: Enable only when `serde`/`borsh` feature is enabled, `cfg(near)` should only gate near host functions

near-global-contracts/src/state_init.rs
```diff
@@ -0,0 +1,134 @@
...
+            // 20 bytes, matching [u8; 20]
+            hash = hasher.finalize()[12..32].try_into().unwrap_or_else(|_| unreachable!());
+        }
+
+        // SAFETY: 20 bytes-long hash will produce 40 hex chars.
+        // "0s" + 40 hex chars = 42 chars, which is within `AccountId` length bounds (2-64).
+        // `hex::encode` always produces valid [0-9a-f] characters, hence, we can construct
+        // AccountId without validation
+        #[allow(deprecated)]
+        near_account_id::AccountId::new_unvalidated(format!("0s{}", hex::encode(hash)))
```
mitinarseny: Do not repeat yourself:

```suggestion
        let hash: [u8; 32];

        #[cfg(near)]
        {
            let serialized = borsh::to_vec(self).unwrap_or_else(|_| unreachable!());
            // SAFETY: keccak256 hash will always generate 32 bytes; [12..32] is exactly
            // 20 bytes, matching [u8; 20]
            hash = near_env::keccak256_array(&serialized);
        }
        #[cfg(all(feature = "digest", not(near)))]
        {
            use sha3::Digest;

            let mut hasher = sha3::Keccak256::new();
            borsh::to_writer(&mut hasher, self).unwrap_or_else(|_| unreachable!());
            // SAFETY: keccak256 hash will always generate 32 bytes; [12..32] is exactly
            // 20 bytes, matching [u8; 20]
            hash = hasher.finalize();
        }

        // SAFETY: 20 bytes-long hash will produce 40 hex chars.
        // "0s" + 40 hex chars = 42 chars, which is within `AccountId` length bounds (2-64).
        // `hex::encode` always produces valid [0-9a-f] characters, hence, we can construct
        // AccountId without validation
        #[allow(deprecated)]
        near_account_id::AccountId::new_unvalidated(
            format!("0s{}", hex::encode(
                hash[12..32].try_into().unwrap_or_else(|_| unreachable!())
            ))
        )
```

near-global-contracts/Cargo.toml
mitinarseny:
```rust
use digest_io::IoWrapper;
use sha3::Keccak256;

let mut hasher = IoWrapper(Keccak256::new());
borsh::to_writer(&mut hasher, self).unwrap();
hasher.0.finalize()
```

near-global-contracts/src/state_init.rs
```diff
@@ -0,0 +1,134 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+#[cfg_attr(
+    any(near, feature = "serde"),
+    cfg_eval::cfg_eval,
+    serde_with::serde_as,
+    derive(serde::Serialize, serde::Deserialize)
+)]
+#[cfg_attr(any(near, feature = "borsh"), derive(borsh::BorshSerialize, borsh::BorshDeserialize))]
```
mitinarseny: Enable only when `serde`/`borsh` feature is enabled, `cfg(near)` should only gate near host functions

near-global-contracts/src/state_init.rs
```diff
@@ -0,0 +1,134 @@
...
+pub struct StateInitV1 {
+    pub code: GlobalContractId,
+    #[cfg_attr(any(near, feature = "serde"), serde_as(as = "BTreeMap<Base64, Base64>"))]
+    #[cfg_attr(feature = "schemars-v0_8", schemars(with = "BTreeMap<String, String>"))]
+    pub data: BTreeMap<Vec<u8>, Vec<u8>>,
+}
+
+impl StateInitV1 {
+    #[inline]
+    pub const fn code(code: GlobalContractId) -> Self {
```
mitinarseny: How about a nicer API?
```suggestion
    pub fn code(code: impl Into<GlobalContractId>) -> Self {
```

near-global-contracts/src/state_init.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+impl From<StateInitV1> for StateInit {
+    #[inline]
+    fn from(state_init: StateInitV1) -> Self {
+        Self::V1(state_init)
+    }
+}
+
+#[cfg(all(not(target_arch = "wasm32"), feature = "near-primitives-interop"))]
```
mitinarseny: Please, do not gate anything behind `target_arch`. It's ok to use `cfg(not(near))` though

near-crypto-hash/src/base58_hash.rs
```diff
@@ -64,27 +63,29 @@ impl AsRef<[u8]> for Base58CryptoHash {
...
-        S: ser::Serializer,
+        S: serde::Serializer,
     {
         serializer.serialize_str(&String::from(self))
     }
 }
 
-impl<'de> de::Deserialize<'de> for Base58CryptoHash {
+#[cfg(any(near, feature = "serde"))]
+impl<'de> serde::Deserialize<'de> for Base58CryptoHash {
```
mitinarseny: You can avoid implementing any of `serde` traits by hands by using `#[derive(serde_with::SerializeDisplay, serde_with::DeserializeFromStr)]`. Don't forget about `schemars(with = "String")`, too

near-sdk-core/Cargo.toml
```diff
@@ -0,0 +1,123 @@
...
+near-env = { path = "../near-env/", version = "~0.1.0" }
+near-crypto-hash = { path = "../near-crypto-hash/", version = "~0.1.0" }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+schemars-v0_8 = { version = "0.8.22", optional = true, package = "schemars" }
+near-primitives-core = { version = "0.35", optional = true }
+near-crypto = { version = "0.35", default-features = false, optional = true }
+# XXX: unicode-segmentation 1.13.1 requires Rust 1.87 but does not declare rust-version
+unicode-segmentation = { version = ">=1, <1.13.1", optional = true }
+near-parameters = { version = "0.35", optional = true }
```
mitinarseny: Do not gate anything (except in `near-sdk`) behind `cfg(not(target_arch = "wasm32")`, use `cfg(not(near))` instead

near-sdk-core/Cargo.toml
```diff
@@ -0,0 +1,123 @@
...
+    "borsh/unstable__schema",
+    "near-account-id/abi",
+    "near-gas/abi",
+    "near-crypto-hash/abi",
+    "serde_with/schemars_0_8",
+]
+
+arbitrary = [
+    "dep:arbitrary",
+    "arbitrary/derive",
```
mitinarseny: nit: `arbitrary/derive` is already specified in `[dependencies]`
```suggestion
```

mitinarseny (changes requested):

near-global-contracts/Cargo.toml
vsavchyn-dev: Fixed in [1e06a33](https://github.com/near/near-sdk-rs/pull/1524/commits/1e06a3374ec9974a0757d654eeedd1a10a9d83cd)

near-global-contracts/src/state_init.rs
vsavchyn-dev: Fixed in [4c95dea](https://github.com/near/near-sdk-rs/pull/1524/commits/4c95dea295848cf221e2e1437d3e23aef374b90f)

near-global-contracts/Cargo.toml
vsavchyn-dev: 1. I was thinking to have people use `--cfg near` similar to `near-contracts` feature. That's how I imagined it originally. However, that wouldn't be a good DevEx, you are right. Easier to enable those things in `near-sdk` itself. Fixed in [aa6aa97](https://github.com/near/near-sdk-rs/pull/1524/commits/aa6aa977b8628e665c2ad2121c420adef7643c92)
2. Maybe yes, but I had an issue with `near-crypto-interop` in some of the CI runs. Apparently old LLVM that is shipped with MacOS does not have wasm compiler(?). But I fixed it for our CI runners in [c229916](https://github.com/near/near-sdk-rs/pull/1524/commits/c229916074fca7ad3803835bd48d587b8e71b91f). The optional, not-platform-gated schemars was created in [aa6aa97](https://github.com/near/near-sdk-rs/pull/1524/commits/aa6aa977b8628e665c2ad2121c420adef7643c92).

near-global-contracts/Cargo.toml
vsavchyn-dev: Fixed in [aa6aa97](https://github.com/near/near-sdk-rs/pull/1524/commits/aa6aa977b8628e665c2ad2121c420adef7643c92)

near-global-contracts/src/state_init.rs
vsavchyn-dev: Fixed in [aa6aa97](https://github.com/near/near-sdk-rs/pull/1524/commits/aa6aa977b8628e665c2ad2121c420adef7643c92)

near-global-contracts/src/state_init.rs
vsavchyn-dev: [c06af38](https://github.com/near/near-sdk-rs/pull/1524/commits/c06af3819d528bae648b5f678b7a638192583c22)

near-global-contracts/src/state_init.rs
vsavchyn-dev: Fixed in [aa6aa97](https://github.com/near/near-sdk-rs/pull/1524/commits/aa6aa977b8628e665c2ad2121c420adef7643c92)

near-global-contracts/src/state_init.rs
vsavchyn-dev: Fixed in [741959a](https://github.com/near/near-sdk-rs/pull/1524/commits/741959afb389aa3ad4d553b0d6705f468bacabbb)

near-sdk-core/Cargo.toml
vsavchyn-dev: Fixed in [aa6aa97](https://github.com/near/near-sdk-rs/pull/1524/commits/aa6aa977b8628e665c2ad2121c420adef7643c92)

near-sdk-core/Cargo.toml
vsavchyn-dev: Fixed in [1b2b6d0](https://github.com/near/near-sdk-rs/pull/1524/commits/1b2b6d0c6ddc29c60f43a2da87056410f4807603)

near-global-contracts/src/state_init.rs
vsavchyn-dev: Fixed in [6731f07](https://github.com/near/near-sdk-rs/pull/1524/commits/6731f07fcd1844ab777c958da1c2a5103a9de6ce)

I am wondering, are there any other places in near-sdk-core or near-crypto-hash I can utilize this pattern? Wdyt?

near-global-contracts/src/state_init.rs
vsavchyn-dev: Fixed in [02bd5c9](https://github.com/near/near-sdk-rs/pull/1524/commits/02bd5c9fc98def3c983f4c9e62bfd988d754d158)

Thanks! Didn't know you can do that and I frankly couldn't find examples from near-* crates (like near-account-id, near-api-rs, etc)

near-crypto-hash/src/base58_hash.rs
vsavchyn-dev: Fixed in: [5312f73](https://github.com/near/near-sdk-rs/pull/1524/commits/5312f7398e84d82a6ef56bd8b90ca0eb145bbcaf)

Issue #1535, opened 2026-05-22
Reminder: Replace Base58CryptoHash newtype with serde_with v3.20.0 Base58 support


vsavchyn-dev:
In #1524, @mitinarseny pointed out several improvements that are gated behind `serde_with` v 3.20.0:

1. Remove Base58CryptoHash newtype pattern in near-crypto-hash - the newtype only exists for deserialization overwrite, which is non-idiomatic. With `serde_with` v3.20.0, `#[serde_as(as = "Base58")]` can be used directly, also eliminating the need for `#[schemars(with = "String")]` workarounds for both base58 and base64.
2. Remove manual base58 logic in near-crypto-hash/src/base58_hash.rs — no longer needed with native base58 support in `serde_with`.

Reference: https://github.com/near/near-sdk-rs/pull/1524

Blocker:
`serde_with` v3.20.0 requires a newer rustc than the current near-vm supports. This should be unblocked with the nearcore 2.12 (wasmtime) release, which will allow bumping to a newer Rust version.

PR #1544, opened 2026-06-02
refactor: relocate Base58CryptoHash to near-sdk-core and modernize near-global-contracts serde

r-near:
Picks up @mitinarseny's review feedback from #1524 / #1535. Three related changes:

- `near-global-contracts`' internal enums (`GlobalContractId`, `AccountContract`) now hold a raw `CryptoHash` with `#[serde_as(as = "Base58")]` instead of the `Base58CryptoHash` newtype.
- Dropped the hand-written `schemars(with = ...)` shims — `serde_with`'s native schemars support covers them now, and the data map even picks up a proper `contentEncoding: base64`.
- Moved `Base58CryptoHash` (and `ParseCryptoHashError`) out of `near-crypto-hash` into `near-sdk-core`, the only thing that uses them. `near-crypto-hash` is now just `pub type CryptoHash = [u8; 32]` with zero dependencies, so it stops pulling `bs58` into crates that only want the hash alias.

`near_sdk::json_types::Base58CryptoHash` is unchanged for users — same path, same behavior. It stays as the public type because `serde_as` is field-level and can't tag a `#[near]` method's return value back to base58. Wire format is untouched (base58 JSON, 32-byte borsh).

Versioning: `near-crypto-hash` and `near-global-contracts` carry real breaking changes, so they go to `0.2.0`. The near-sdk family stays a minor (`5.29.0`) — its only change is the re-exported enums, which is low-impact given how new these APIs are, so it's not worth a major. (`near-sdk-core` is just a `4.1.2` patch — its public API is unchanged.)

Verified across feature combos — builds, clippy, fmt, and tests all green.

Refs #1535


near-global-contracts/src/global_contract_identifier.rs
```diff
@@ -29,21 +36,18 @@ pub enum AccountContract {
 #[repr(u8)]
 pub enum GlobalContractId {
     #[cfg_attr(feature = "serde", serde(rename = "hash"))]
-    CodeHash(Base58CryptoHash) = 0,
+    CodeHash(
+        #[cfg_attr(feature = "serde", serde_as(as = "Base58"))]
+        #[cfg_attr(feature = "schemars-v0_8", schemars(with = "String"))]
```
mitinarseny: No need for manual `schemars` [anymore](https://github.com/jonasbb/serde_with/pull/943), just activating `serde_with/schemars_0_8` feature should work
```suggestion
        #[cfg_attr(feature = "serde", serde_as(as = "::serde_with::base58::Base58"))]
```

near-global-contracts/src/global_contract_identifier.rs
```diff
@@ -1,19 +1,26 @@
 use near_account_id::AccountId;
 
-use near_crypto_hash::Base58CryptoHash;
 use near_crypto_hash::CryptoHash;
 
+#[cfg(feature = "serde")]
+use serde_with::base58::Base58;
```
mitinarseny: better use FQDNs IMO
```suggestion
```

near-global-contracts/Cargo.toml
mitinarseny, line 73:
```suggestion
schemars-v0_8 = [
    "dep:schemars-v0_8",
    "near-account-id/schemars-v0_8",
    "near-crypto-hash/schemars-v0_8",
    "serde_with/schemars_0_8",
]
```

mitinarseny (commented):
@r-near thanks for the PR!

PS: Also, `StateInitV1` [doesn't need](https://github.com/jonasbb/serde_with/pull/949) `schemars(with = "BTreeMap<String, String>")` anymore

mitinarseny (commented):
Can we now move `Base58CryptoHash` from `near-crypto-hash` to the crate that actually needs it? Or make `base58` a feature of `near-crypto-hash`, so that it doesn't pull in `bs58` and other unnecessary stuff when not needed (e.g. when `near-global-contract/serde` is not enabled)?

r-near:
> Can we now move `Base58CryptoHash` from `near-crypto-hash` to the crate that actually needs it? Or make `base58` a feature of `near-crypto-hash`, so that it doesn't pull in `bs58` and other unnecessary stuff when not needed (e.g. when `near-global-contract/serde` is not enabled)?

Yep just updated

near-global-contracts/Cargo.toml
r-near: Fixed

near-global-contracts/src/global_contract_identifier.rs
r-near: Done: [`96579ed` (this PR)](https://github.com/near/near-sdk-rs/pull/1544/commits/96579edd271f5e2978ab85b201d7eb5a338a184f)

near-global-contracts/src/global_contract_identifier.rs
r-near: Fixed [`27d594f` (this PR)](https://github.com/near/near-sdk-rs/pull/1544/commits/27d594f62f69fdac0e9b0bc1f87f8fea3e7710cd)

near-sdk-core/Cargo.toml
```diff
@@ -27,12 +27,12 @@ base64 = { version = "0.22", optional = true }
...
 
-arbitrary = { version = "1.4", feature = ["derive"], optional = true }
+arbitrary = { version = "1.4", features = ["derive"], optional = true }
 
 near-account-id = { version = "2.3.0" }
 near-gas = { version = "0.3" }
 near-token = { version = "0.3" }
-near-crypto-hash = { path = "../near-crypto-hash/", version = "~0.1.0" }
+near-crypto-hash = { path = "../near-crypto-hash/", version = "~0.2.0" }
 near-sdk-env = { path = "../near-sdk-env/", version = "~0.1.0", optional = true }
```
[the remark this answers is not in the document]
r-near: release-plz will handle this

PR #1546, opened 2026-06-03
ci: drop NEAR_SANDBOX_BIN_PATH workaround now that near-sandbox defaults to 2.12

r-near:
While 2.12 was still an RC, CI manually downloaded the `2.12.0` sandbox tarball and pointed `near-workspaces` at it via `NEAR_SANDBOX_BIN_PATH`, because the `near-sandbox` version we pulled in still defaulted to the 2.11 (PV-83) VM, which can't run wasm produced by rustc 1.93.

That's no longer necessary. `near-sandbox` 0.3.11 is published and its `DEFAULT_NEAR_SANDBOX_VERSION` is now `2.12.0` (protocol 84). `near-workspaces` 0.22.2 (used by our example/integration tests) depends on `near-sandbox ^0.3.9`, which resolves to 0.3.11, so the tests get the 2.12 VM by default — no manual download or env override needed.

This removes the download step and the `NEAR_SANDBOX_BIN_PATH` env from `test.yml`, `coverage.yml`, `test_examples.yml`, and `test_examples_small.yml`. CI-only change; no crate source touched.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -41,16 +41,6 @@ jobs:
       - name: Install cargo-near CLI
         run: curl --proto '=https' --tlsv1.2 -LsSf https://github.com/near/cargo-near/releases/download/cargo-near-v0.20.2/cargo-near-installer.sh | sh
 
-      # 2.12 RC: point `near-workspaces` at a PV-84 sandbox so wasm produced by
-      # rustc 1.93 is accepted. See the same step in `test.yml` for the rationale.
-      - name: Download near-sandbox 2.12.0
-        run: |
-          mkdir -p "$HOME/.near-sandbox-2.12.0"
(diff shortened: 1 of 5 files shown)


r-near:
Good catch @chatgpt-codex-connector — that was exactly it. The `factory-contract-global` test was pinning `sandbox_with_version("2.7.0")`, and the `NEAR_SANDBOX_BIN_PATH` override was quietly masking it by forcing the 2.12 binary regardless. Once the override was gone the test really did spin up 2.7.0, so the 1.93 wasm got rejected (`PrepareError(Deserialization)`) and it hit the old 300 TGas cap.

Switched all four call sites to the default `near_workspaces::sandbox()`, which resolves to near-sandbox 0.3.11's 2.12.0 default. factory-contract-global linux + macos both go green now without the override, so the default path genuinely reaches the 2.12 VM. Merged.

PR #1553, opened 2026-06-08
refactor: use idiomatic serde_with for Base58CryptoHash

### LINKED ISSUE #1535: Reminder: Replace Base58CryptoHash newtype with serde_with v3.20.0 Base58 support

codeboost-tr:
## Summary
This PR refactors `Base58CryptoHash` to use idiomatic `serde_with` attributes instead of manual trait implementations for Base58 serialization, as requested in #1535.

## Changes
- Updated `near-sdk-core/Cargo.toml` to include the `base58` feature for `serde_with`.
- Refactored `Base58CryptoHash` in `near-sdk-core/src/json_types/hash.rs` to use `#[serde_as(as = "DisplayFromStr")]`.
- Verified changes with existing unit tests in `near-sdk-core`.

Fixes #1535

### DIFF
--- a/near-sdk-core/Cargo.toml
+++ b/near-sdk-core/Cargo.toml
@@ -21,6 +21,7 @@ serde_with = { version = "3", features = [
     "base64",
     "hex",
     "json",
+    "base58",
 ], optional = true }
 borsh = { version = "1.0.0", features = ["derive"], optional = true }
 base64 = { version = "0.22", optional = true }
--- a/near-sdk-core/Cargo.toml
+++ b/near-sdk-core/Cargo.toml
@@ -21,7 +21,6 @@ serde_with = { version = "3", features = [
     "base64",
     "hex",
     "json",
-    "base58",
 ], optional = true }
 borsh = { version = "1.0.0", features = ["derive"], optional = true }
 base64 = { version = "0.22", optional = true }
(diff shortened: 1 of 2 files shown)


codeboost-tr:
Re-ping on this — the `serde_with::serde_as` swap for `Base58CryptoHash` is a clean drop-in (no public API change beyond a derive attribute). CI green, codecov report attached. Issue #1535 has been tracked for a while; happy to expand the migration to the other newtype wrappers (`MerklizedItemHash`, `EpochId`, etc.) in a follow-up PR if this lands. Open to any naming or scope feedback.

PR #1571, opened 2026-06-26
feat: `near-digest` crate

vsavchyn-dev:
This PR introduces a `near-digest` crate implementing the RustCrypto `digest` traits with both on-chain and off-chain usage in mind:

- `--cfg near`: backed by NEAR host functions via `near-sdk-env`
- otherwise: falls back to RustCrypto implementations (`sha2`, `sha3`, `ripemd`)

### DIFF
--- a/near-digest/Cargo.toml
+++ b/near-digest/Cargo.toml
@@ -7,6 +7,9 @@ categories.workspace = true
 repository.workspace = true
 homepage.workspace = true
 rust-version.workspace = true
+description = """
+`digest` trait implementations for hash functions in NEAR smart contracts: backed by NEAR host functions on-chain, pure Rust off-chain.
+"""
 
 [dependencies]
 digest = "0.11.3"
--- a/near-digest/Cargo.toml
+++ b/near-digest/Cargo.toml
@@ -21,6 +21,8 @@ sha3 = { version = "0.11.0", optional = true }
 [target.'cfg(near)'.dependencies]
 near-sdk-env = { path = "../near-sdk-env/" }
 impl-tools = "0.12.0"
+# for `unstable` Sha3-256/512
+sha3 = { version = "0.11.0", optional = true }
 
 [features]
 ripemd = ["dep:ripemd"]
@@ -53,3 +55,6 @@ near-sdk = { path = "../near-sdk/", features = ["unit-testing"] }
 
 [lints]
 workspace = true
+
+[package.metadata.docs.rs]
+all-features = true
--- a/near-digest/Cargo.toml
+++ b/near-digest/Cargo.toml
@@ -1,0 +1,47 @@
+[package]
+name = "near-digest"
+version = "0.1.0"
+edition.workspace = true
+license.workspace = true
+categories.workspace = true
+repository.workspace = true
+homepage.workspace = true
+rust-version.workspace = true
+
+[dependencies]
+digest = "0.11.3"
+
+zeroize = { version = "1.9.0", features = ["derive"], optional = true }
+
+[target.'cfg(not(near))'.dependencies]
+ripemd = { version = "0.2.0", optional = true }
+sha2 = { version = "0.11.0", optional = true }
+sha3 = { version = "0.11.0", optional = true }
+
+[target.'cfg(near)'.dependencies]
+near-sdk-env = { path = "../near-sdk-env/" }
+impl-tools = "0.12.0"
+
+[features]
+ripemd = ["dep:ripemd"]
+sha2 = ["dep:sha2"]
+sha3 = ["dep:sha3"]
+
+zeroize = [
+    "dep:zeroize",
+    "digest/zeroize",
+    "ripemd?/zeroize",
+    "sha2?/zeroize",
+    "sha3?/zeroize",
+]
+
+[dev-dependencies]
+near-digest = { path = ".", features = ["ripemd", "sha2", "sha3", "zeroize"] }
+rstest = "0.26.1"
+hex = "0.4.3"
+
+[target.'cfg(near)'.dev-dependencies]
+near-sdk = { path = "../near-sdk/", features = ["unit-testing"] }
+
+[lints]
+workspace = true
--- a/near-digest/Cargo.toml
+++ b/near-digest/Cargo.toml
@@ -46,7 +46,7 @@ near-digest = { path = ".", features = [
     "unstable",
 ] }
 rstest = "0.26.1"
-hex = "0.4.3"
+hex-literal = "1.1.0"
 
 [target.'cfg(near)'.dev-dependencies]
 near-sdk = { path = "../near-sdk/", features = ["unit-testing"] }
--- a/near-digest/Cargo.toml
+++ b/near-digest/Cargo.toml
@@ -35,8 +35,16 @@ zeroize = [
     "sha3?/zeroize",
 ]
 
+unstable = []
+
 [dev-dependencies]
-near-digest = { path = ".", features = ["ripemd", "sha2", "sha3", "zeroize"] }
+near-digest = { path = ".", features = [
+    "ripemd",
+    "sha2",
+    "sha3",
+    "zeroize",
+    "unstable",
+] }
 rstest = "0.26.1"
 hex = "0.4.3"
 
--- a/near-digest/Cargo.toml
+++ b/near-digest/Cargo.toml
@@ -19,7 +19,7 @@ sha2 = { version = "0.11.0", optional = true }
 sha3 = { version = "0.11.0", optional = true }
 
 [target.'cfg(near)'.dependencies]
-near-sdk-env = { path = "../near-sdk-env/" }
+near-sdk-env = { path = "../near-sdk-env/", version = "~0.1.1" }
 impl-tools = "0.12.0"
 # for `unstable` Sha3-256/512
 sha3 = { version = "0.11.0", optional = true }
@@ -40,6 +40,8 @@ zeroize = [
 unstable = []
 
 [dev-dependencies]
+# XXX: `near-sdk` was added in order to enable tests and doctests compiling with mockchain
+near-sdk = { path = "../near-sdk", features = ["unit-testing"] }
 near-digest = { path = ".", features = [
     "ripemd",
     "sha2",
--- a/near-digest/src/lib.rs
+++ b/near-digest/src/lib.rs
@@ -1,0 +1,65 @@
+//! Helper crate to automatically chose a backend for digest implementations.
+//! Currently supported backends are:
+//!
+//! * `cfg(near)`: via Near host-function
+//! * default: fallback to pure Rust implementation
+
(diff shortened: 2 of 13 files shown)


near-digest/src/sha2/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+    /// Backed by the `sha256_array` host function when compiled as a NEAR contract (`cfg(near)`),
+    /// and by the pure-Rust implementation from the [`sha2`](https://docs.rs/sha2) crate otherwise.
+    pub struct Sha256 {
+        near => self::near::Sha256,
+        _ => ::sha2::Sha256,
+    }
+}
+
+#[cfg(test)]
+mod test {
```
r-near, line 20: Tried it locally and these tests don't actually run under `--cfg near` (fails to link). Might need the same `use near_sdk as _` fix we use.

Maybe it's also worth adding a CI job that runs tests with `--cfg near`

near-digest/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+
+zeroize = { version = "1.9.0", features = ["derive"], optional = true }
+
+[target.'cfg(not(near))'.dependencies]
+ripemd = { version = "0.2.0", optional = true }
+sha2 = { version = "0.11.0", optional = true }
+sha3 = { version = "0.11.0", optional = true }
+
+[target.'cfg(near)'.dependencies]
+near-sdk-env = { path = "../near-sdk-env/" }
```
r-near: Will this work when publishing? The path dep?

near-digest/Cargo.toml
```diff
@@ -0,0 +1,60 @@
+[package]
+name = "near-digest"
+version = "0.1.0"
```
r-near, line 3: Do we need a description for cargo publish?

near-digest/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+//! | `sha3::Keccak256` | `sha3` | `keccak256_array` |
+//! | `sha3::Keccak512` | `sha3` | `keccak512_array` |
+//! | `sha3::Sha3_256` | `sha3` + `unstable` | - (pure Rust on all targets for now) |
+//! | `sha3::Sha3_512` | `sha3` + `unstable` | - (pure Rust on all targets for now) |
+//! | `ripemd::Ripemd160` | `ripemd` | `ripemd160_array` |
+//!
+//! Additional features:
+//!
+//! * `zeroize` — implements `zeroize::ZeroizeOnDrop` for all hash types, clearing
+//!   buffered input from memory when a hasher is dropped.
```
r-near, line 56: nit: doesn't this only clear the final buffer? probably fine regardless 🤷

near-digest/src/utils.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use digest::{FixedOutput, HashMarker, Output, OutputSizeUser, Update};
+use impl_tools::autoimpl;
+
+pub trait DigestFinalizer: OutputSizeUser {
+    fn digest(bytes: &[u8]) -> Output<Self>;
+}
+
+#[cfg_attr(feature = "zeroize", derive(::zeroize::ZeroizeOnDrop))]
+#[autoimpl(Debug, Clone, Default, PartialEq, Eq)]
```
r-near: autoimpl is interesting...is this from Intents? idk if we should pull in the dep just for this. PartialEq/Eq are unused anyway, right?

and comparing hash buffers is kinda weird anyway. maybe we just hand-write Debug/Clone/Default?

near-digest/src/ripemd/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+    #[rstest]
+    #[case(
+        b"",
+        hex!("9c1185a5c5e9fc54612808977ee8f548b2258d31"),
+    )]
+    #[case(
+        b"near is cool!",
+        hex!("320214cbbb6821fb23d3cd96dc0731ff5644323b"),
+    )]
+    fn ripemd10_has_not_changed(#[case] data: &[u8], #[case] output: [u8; 20]) {
```
r-near:
```suggestion
    fn ripemd160_has_not_changed(#[case] data: &[u8], #[case] output: [u8; 20]) {
```

typo?

near-digest/src/ripemd/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+        assert!(Ripemd160::digest(data) == output, "has changed")
```
r-near: Should we do `assert_eq!` so it prints the value?

near-sdk/src/lib.rs
```diff
@@ -1630,6 +1628,9 @@ pub mod state;
 #[cfg(feature = "deterministic-account-ids")]
 pub mod state_init;
 
+#[cfg(feature = "digest")]
+pub use near_digest as digest;
```
r-near, line 1632: Would this be referenced as `near_sdk::digest::digest::Digest`? Maybe re-export the trait at the top level too?

r-near (changes requested):

r-near:
@codex review

near-digest/src/sha2/mod.rs
vsavchyn-dev: So, I didn't find a reliable way to actually have this crate going through the `--cfg near` reliably except if we use this crate in example.

However, now everything should work with `use near_sdk as _;`. https://github.com/near/near-sdk-rs/pull/1571/commits/490a541d06e21710d9d88f375a1a1b3c1a1a459e

near-digest/Cargo.toml
vsavchyn-dev: Yup, we do, thx for catching. Fixed in https://github.com/near/near-sdk-rs/pull/1571/commits/675fa9bcff8c41c3af582afc29bbdfd654566ef9

lmk if I need to change desc in some way

near-digest/Cargo.toml
vsavchyn-dev: Fixed in https://github.com/near/near-sdk-rs/commit/490a541d06e21710d9d88f375a1a1b3c1a1a459e

near-digest/src/utils.rs
vsavchyn-dev: Yes, it's from Intents. 

Anyway we are using this `autoimpl` in `--cfg near path`, and `impl-tools` relies on `quote`, `syn` and `proc-macro2` which are already in `near-sdk-macros`. Hence, I think it shouldn't be a problem to include this crate.

near-digest/src/ripemd/mod.rs
vsavchyn-dev: Fixed in https://github.com/near/near-sdk-rs/pull/1571/commits/490a541d06e21710d9d88f375a1a1b3c1a1a459e

near-digest/Cargo.toml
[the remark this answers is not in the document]
vsavchyn-dev: Fixed in https://github.com/near/near-sdk-rs/pull/1571/commits/490a541d06e21710d9d88f375a1a1b3c1a1a459e

near-digest/src/ripemd/mod.rs
vsavchyn-dev: Fixed in https://github.com/near/near-sdk-rs/pull/1571/commits/c7f2b5a12d2c8dbe8b95f16cb9c93f2ce372bab1

near-sdk/src/lib.rs
vsavchyn-dev: Yup, you're right. I re-exported some traits so it would look something like `near_sdk::digest::Digest` in following commit https://github.com/near/near-sdk-rs/pull/1571/commits/01969e656b1a386dfd3a81ea29b4bafda1a3b280

near-digest/src/lib.rs
mitinarseny: The idea was to let to use this code in off-chain applications, which might want to zeroize memory since there can be some secrets there...

near-digest/src/utils.rs
mitinarseny: The idea was to mimic traits implemented on real `sha2::*`, `sha3::*`, etc.., so that this crate can be used as drop-in replacement for them. `#[autoimpl(...)]` lets you ignore trait bounds on generic type params. However, you can implement it by hands, too.

Also,  just checked, `PartialEq` and `Eq` are not implemented there, so feel free to remove them.

near-sdk/src/lib.rs
mitinarseny: IMO, there is no point in re-exporting it from `near_sdk`, since `near-sdk` brings a lot of unnecessary stuff that even doesn't compile for non-wasm targets properly... While the main motivation behind this crate is to write re-usable code. So, re-exporting it from near-sdk would not discourage people from using near-sdk in non-contract code and we will end up in the same situation as we have now: people writing more and more non-reusable code.

near-digest/src/utils.rs
vsavchyn-dev: Removed those traits in https://github.com/near/near-sdk-rs/pull/1571/commits/eacac9cc4c14353f05773c1c71102e0c4f84f304

near-digest/src/utils.rs
```diff
@@ -0,0 +1,35 @@
...
+}
+
+impl<F: DigestFinalizer> FixedOutput for DigestFn<F> {
+    #[inline]
+    fn finalize_into(self, out: &mut digest::Output<Self>) {
+        *out = F::digest(&self.data);
+    }
+}
+
+impl<F: DigestFinalizer> HashMarker for DigestFn<F> {}
```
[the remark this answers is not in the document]
vsavchyn-dev: Fixed in https://github.com/near/near-sdk-rs/pull/1571/commits/a371c18d6193aa9f9cb42d1583d6c1281c5dc7fa

near-digest/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        impl ::digest::FixedOutput for $name {
+            #[inline]
+            fn finalize_into(self, out: &mut ::digest::Output<Self>) {
+                ::digest::FixedOutput::finalize_into(self.0, out);
+            }
+        }
+
+        impl ::digest::HashMarker for $name {}
```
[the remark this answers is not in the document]
vsavchyn-dev: Fixed in https://github.com/near/near-sdk-rs/pull/1571/commits/a371c18d6193aa9f9cb42d1583d6c1281c5dc7fa

near-digest/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
```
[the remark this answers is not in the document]
vsavchyn-dev: Fixed in https://github.com/near/near-sdk-rs/pull/1571/commits/a371c18d6193aa9f9cb42d1583d6c1281c5dc7fa

r-near (approved):

PR #1573, opened 2026-06-30
refactor(global-contracts): derive From for GlobalContractId via derive_more

r-near:
## What

Replace the two hand-written `From` impls on `GlobalContractId` with `#[derive(derive_more::From)]`:

```rust
// before
impl From<CryptoHash> for GlobalContractId { fn from(h) -> Self { Self::CodeHash(h) } }
impl From<AccountId>  for GlobalContractId { fn from(a) -> Self { Self::AccountId(a) } }

// after
#[derive(..., derive_more::From)]
pub enum GlobalContractId { CodeHash(CryptoHash) = 0, AccountId(AccountId) = 1 }
```

`derive_more::From` generates one `From` per single-field variant, so this produces byte-identical `From<CryptoHash>` / `From<AccountId>` conversions. **No public API change** — purely removes boilerplate.

## Dependency

`derive_more = { version = "2", default-features = false, features = ["from"] }`. `derive_more` 2.1.1 is already in the workspace lockfile (transitive), so nothing new is compiled; `default-features = false` keeps it `no_std`-safe and avoids pulling in `std`.

## Context

Suggested by a downstream user wiring NEP-616 global contracts into integration tests via [`near-kit`](https://github.com/r-near/near-kit-rs) (the defuse / intents test suite).

## Testing

- `cargo check` across all feature combos: `{}`, `serde`, `borsh`, `serde,borsh`, `abi`, `near-primitives-interop`, and all-together — the `serde` + `serde_as` field-attribute path compiles fine.
- `cargo test -p near-global-contracts --features serde,borsh,abi,near-primitives-interop` — green.
- `cargo fmt --check` + `cargo clippy` clean. `cargo check --locked` passes.

### DIFF
--- a/near-global-contracts/Cargo.toml
+++ b/near-global-contracts/Cargo.toml
@@ -21,6 +21,7 @@ features = ["serde", "borsh", "abi", "near-primitives-interop"]
 [dependencies]
 near-account-id = { version = "2.3.0", features = ["internal_unstable"] }
 near-crypto-hash = { path = "../near-crypto-hash/", version = "~0.2.0" }
+derive_more = { version = "2", default-features = false, features = ["from"] }
 
 hex = { version = "0.4", optional = true }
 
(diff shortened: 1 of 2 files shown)


r-near:
Closing — on reflection this is a pure style refactor: `derive_more::From` regenerates the same `From<CryptoHash>`/`From<AccountId>` impls that have existed since 0.2.0, so it adds a dependency to delete ~12 lines with no behavior or API change. Not worth the trade. Thanks for the suggestion regardless.

PR #1577, opened 2026-07-03
feat: add `p256_verify`

vsavchyn-dev:
Adds bindings for the `p256_verify` host function ([NEP-635](https://github.com/near/NEPs/pull/635), https://github.com/near/nearcore/pull/15588, protocol v85 / nearcore 2.13):

- `near-sys` extern declaration
- `env::p256_verify(signature, message, public_key)` wrapper
- mock shim for unit tests
- native fallback via the `p256` crate for `non-contract-usage`

Note: unlike `ed25519_verify`, the host does not hash the message - the caller hashes it first (SHA-256 for WebAuthn). Doc examples use `sha256_array` to show that.

### DIFF
--- a/near-sdk/src/environment/env.rs
+++ b/near-sdk/src/environment/env.rs
@@ -689,6 +689,54 @@ pub fn ed25519_verify(
     }
 }
 
+/// Verifies a P-256 (secp256r1) ECDSA signature of a prehashed message using provided public key
+///
+/// Unlike [`ed25519_verify`], the message is **not** hashed by the host function: the caller is
+/// responsible for applying the appropriate hash (for example SHA-256) before calling, per [NEP-635](https://github.com/near/NEPs/pull/635).
+///
+/// - `signature` - 64 bytes, encoded as `r || s` (32 bytes each, big-endian).
+/// - `message` - the prehashed digest to verify, typically a 32-byte SHA-256 digest.
+/// - `public_key` - a 33-byte compressed SEC1 encoding.
+///
+/// # Examples
+/// ```
+/// use near_sdk::env::{p256_verify, sha256_array};
+/// use hex;
+///
+/// assert_eq!(
+///     p256_verify(
+///         hex::decode("EFD48B2AACB6A8FD1140DD9CD45E81D69D2C877B56AAF991C34D0EA84EAF3716F7CB1C942D657C41D436C7A1B6E29F65F3E900DBB9AFF4064DC4AB2F843ACDA8")
+///             .expect("Decoding failed")
+///             .as_slice()
+///             .try_into()
+///             .unwrap(),
+///         sha256_array(b"sample"),
+///         hex::decode("0360FED4BA255A9D31C961EB74C6356D68C049B8923B61FA6CE669622E60F29FB6")
+///             .expect("Decoding failed")
+///             .as_slice()
+///             .try_into()
+///             .unwrap(),
+///     ),
+///     true
+/// );
+///
+/// assert_eq!(
+///     p256_verify(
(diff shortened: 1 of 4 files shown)


r-near (approved):

near-sdk/src/environment/env.rs
```diff
@@ -689,6 +689,94 @@ pub fn ed25519_verify(
...
+///         hex::decode("0360FED4BA255A9D31C961EB74C6356D68C049B8923B61FA6CE669622E60F29FB6")
+///             .expect("Decoding failed")
+///             .as_slice()
+///             .try_into()
+///             .unwrap(),
+///     ),
+///     false
+/// );
+/// ```
+pub fn p256_verify(signature: &[u8; 64], message: impl AsRef<[u8]>, public_key: &[u8; 33]) -> bool {
```
r-near, line 782: Can you update the rustdoc to mention the malleability thing?

near-sdk/src/environment/env.rs
```diff
@@ -689,6 +689,94 @@ pub fn ed25519_verify(
...
+        let Ok(verifying_key) = VerifyingKey::from_sec1_bytes(public_key) else {
+            return false;
+        };
+        let Ok(signature) = Signature::from_bytes(signature.into()) else {
+            return false;
+        };
+        verifying_key.verify_prehash(message, &signature).is_ok()
+    }
+}
+
```
r-near, line 821: Could we also get the low-s helper we discussed in Slack, with its own docs? Also a WebAuthn-flavored example would be nice too 🤷

near-sdk/src/environment/env.rs
vsavchyn-dev: Done in https://github.com/near/near-sdk-rs/pull/1577/commits/531739d59556c6d28e8397b6afef6c55519d7b29

near-sdk/src/environment/env.rs
vsavchyn-dev: Done in https://github.com/near/near-sdk-rs/pull/1577/commits/531739d59556c6d28e8397b6afef6c55519d7b29

I am not sure what to put for the "WebAuthn-flavored" example - I guess I will mention that in rustdoc?

r-near: Sorry, that was vague... I mean like a worked example of building the digest for a passkey assertion, since the message construction is the part people get wrong with a prehashed API (and it's the whole reason the host fn doesn't hash)

Per the WebAuthn spec the authenticator signs `authenticator_data || SHA-256(client_data_json)` with ES256

r-near: So like:

```rust
use near_sdk::env::{p256_verify, sha256_array};
/// Verifies a WebAuthn (passkey) assertion signature.
///
/// The authenticator signs `authenticator_data || SHA-256(client_data_json)`
/// with ECDSA P-256 / SHA-256 (COSE alg ES256), so that concatenation gets
/// hashed once more to produce the digest for `p256_verify`.
///
/// Note: browsers return the assertion signature as ASN.1 DER — clients must
/// convert it to raw `r || s` (each padded to 32 bytes) before submitting.
pub fn verify_passkey_assertion(
    authenticator_data: &[u8],
    client_data_json: &[u8],
    signature: &[u8; 64],
    public_key: &[u8; 33],
) -> bool {
    let mut signed_payload = Vec::with_capacity(authenticator_data.len() + 32);
    signed_payload.extend_from_slice(authenticator_data);
    signed_payload.extend_from_slice(&sha256_array(client_data_json));
    p256_verify(signature, sha256_array(&signed_payload), public_key)
}
```

An ignored doctest is fine, or you can make it self-checking by generating the signature with the p256 crate inside the doctest (it's already a dep for non-contract-usage)

vsavchyn-dev: Got it, added it in https://github.com/near/near-sdk-rs/pull/1577/commits/633ab558f518b62d9b2e69e213c0aa29a9797f50

r-near (approved):

r-near:
Looks good 👍

PR #1584, opened 2026-07-08
refactor: hash StateInit::derive_account_id via near-digest

r-near:
## What

Follow-up to the `near-digest` crate (#1571). Per [Arseny's request in Slack](https://nearone.slack.com/archives/C0A64SAMDNG/p1783531653206219?thread_ts=1782478837.131179&cid=C0A64SAMDNG), now that `near-digest` v0.1.0 is [published](https://crates.io/crates/near-digest), this collapses the two keccak256 code paths in `near_global_contracts::StateInit::derive_account_id` into a single one backed by `near-digest`:

```rust
let mut hasher = IoWrapper(Keccak256::new());
borsh::to_writer(&mut hasher, self).unwrap_or_else(|_| unreachable!());
let hash: [u8; 32] = hasher.0.finalize().into();
```

`near-digest` selects the backend automatically — the `keccak256` host function on-chain (`--cfg near`) and pure-Rust off-chain — so the previous `#[cfg(any(near, feature = "__near-sdk-unit-testing"))]` / `#[cfg(not(...))]` split is no longer needed.

## Changes

- **near-global-contracts**
  - Single `derive_account_id` path via `near-digest`; drop the direct `near-sdk-env`, `sha3`, and `cfg(near)`/`cfg(not(near))` dependency blocks.
  - `near-digest` (`sha3` feature) + `digest-io` are now regular deps, gated behind `borsh` (the only feature that exposes `derive_account_id`) instead of being pulled in unconditionally per target.
  - Remove the now-dead `__near-sdk-unit-testing` feature (its only job was routing hashing through the `near-sdk-env` mock; the pure-Rust fallback produces identical output in unit tests).
  - Docs/README updated.
- **near-sdk**
  - `unit-testing` feature now enables `near-global-contracts/near-primitives-interop` directly (it previously got it transitively via `__near-sdk-unit-testing`). The unit-testing mock's `DeterministicAccountStateInit -> StateInit` conversion in `environment/mock/receipt.rs` needs those `From` impls.
  - Fixed a stale doc that referenced a non-existent `digest` feature on `near-global-contracts`.
- **near-digest**
  - Pin MSRV to **1.88** (previously inherited the workspace's 1.93). Its off-chain surface is pure-Rust RustCrypto and only needs edition 2024's 1.85 floor; the `cfg(near)` deps that require 1.93 are only ever built on-chain, where 1.93+ is enforced by cargo-near / the protocol version. This keeps `near-global-contracts` at **1.88** (unchanged) so off-chain consumers like near-kit keep working — same low-MSRV-leaf arrangement `near-global-contracts` and `near-crypto-hash` already use.
- **CI**
  - Add an `msrv` job that runs the off-chain feature powerset for the low-MSRV leaf crates (`near-crypto-hash`, `near-digest`, `near-global-contracts`) against each crate's *declared* `rust-version`, so the 1.88 promise is actually verified and can't silently rot (previously nothing tested these crates below 1.93).

## MSRV note

Earlier revision briefly bumped `near-global-contracts` to 1.93; that's reverted. The floor stays at 1.88, achieved by lowering `near-digest` to 1.88 instead — cargo's resolver keys off the *declared* MSRV, so both crates had to move together. Confirmed locally that `near-crypto-hash`, `near-digest`, and `near-global-contracts` all build the full off-chain feature powerset on the 1.88 toolchain with `-D warnings`.

## Verification

- `cargo test -p near-global-contracts --features serde,borsh` — derivation test passes; **derived account IDs are byte-for-byte unchanged** (the test asserts exact values).
- `cargo hack check` feature powersets for `near-global-contracts` on host, off-chain wasm, and on-chain (`--cfg near`) wasm targets.
- `cargo +1.88 hack check -p {near-crypto-hash,near-digest,near-global-contracts} --feature-powerset --no-dev-deps` with `-D warnings` — all green (this is what the new `msrv` CI job runs).
- `cargo test -p near-sdk --features unit-testing,deterministic-account-ids` deterministic-account-id mock tests pass; base `unit-testing` build clean.
- `cargo fmt --check` and `cargo clippy -p near-global-contracts --tests --all-features` clean.

Note: this touches audit-sensitive code (cryptographic account derivation), but the hash output is provably identical — same keccak256 over the same borsh bytes, just via a different backend selector.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -75,6 +75,50 @@ jobs:
           RUSTFLAGS: "-D warnings --cfg near"
         run: cargo hack check -p ${{ matrix.crate }} --target "wasm32-unknown-unknown" --feature-powerset --exclude-features schemars-v0_8,abi,__near-sdk-unit-testing --no-dev-deps
 
+  # These leaf crates deliberately support an older MSRV than the workspace 1.93 so off-chain
+  # consumers on an older toolchain (indexers, CLIs, near-kit) can use them. Their `cfg(near)`
+  # on-chain path does need 1.93, but it is only ever built with a newer toolchain (cargo-near /
+  # the protocol version enforce it), so only the off-chain surface is checked here.
+  msrv:
+    runs-on: warp-ubuntu-latest-x64-4x
+    name: "MSRV off-chain ${{ matrix.crate }}"
+    strategy:
+      fail-fast: false
+      matrix:
+        crate:
+          - "near-crypto-hash"
+          - "near-digest"
+          - "near-global-contracts"
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install MSRV toolchain
+        run: |
+          # Read the crate's declared MSRV so this job always tests exactly what is published.
+          MSRV=$(grep -m1 '^rust-version = ' ${{ matrix.crate }}/Cargo.toml | sed -E 's/.*"([0-9.]+)".*/\1/')
+          if [ -z "$MSRV" ]; then echo "::error::could not read rust-version from ${{ matrix.crate }}/Cargo.toml"; exit 1; fi
+          echo "Package MSRV: $MSRV"
+          rustup toolchain install "$MSRV" --profile minimal
+          rustup override set "$MSRV"
+
+      - name: Install cargo-hack
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-hack
+
+      - uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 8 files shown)


vsavchyn-dev:
@r-near I believe we should leave the near-env and non-near-env path for now - as we don't expose `near-digest` for now fully in `near-sdk` (without feature flag), we should see if people use `near-digest` and only then add it for the `near-global-contracts`...

r-near:
Good point, moving this to draft

PR #1608, opened 2026-07-16
chore: use `near-digest` in `near-global-contracts`

mitinarseny:


near-global-contracts/Cargo.toml
```diff
@@ -28,21 +28,15 @@ cfg_eval = { version = "0.1", optional = true }
 serde = { version = "1", features = ["derive"], optional = true }
 serde_with = { version = "3.20", features = ["base58", "base64", "hex"], optional = true }
 borsh = { version = "1.0.0", features = ["derive"], optional = true }
+near-digest = { path = "../near-digest/", features = ["sha2", "sha3"], optional = true }
```
r-near, line 31: Can't have path-deps for cargo publish

near-global-contracts/Cargo.toml
```diff
@@ -52,7 +46,7 @@ serde_json = { version = "1", features = ["preserve_order"] }
 workspace = true
 
 [features]
-default = []
+default = ["digest"]
```
r-near, line 49: I guess this is technically breaking? 

Anyone on `default-features = false, features = ["borsh"]` loses `derive_account_id`

Buuuuut is should be fine, just gotta bump it to 0.30

r-near (changes requested):
Looks like cargo-doc is failing as well

PR #1609, opened 2026-07-16
docs: document GlobalContractId and StateInit types

r-near:
Adds rustdoc for the two public types that were missing it in `near-global-contracts`:

- `GlobalContractId` + both variants (`CodeHash` vs `AccountId`, i.e. reference by immutable code hash vs by deployer account).
- `StateInit` / `StateInitV1` (variant, `code`/`data` fields, and the `code`/`with_data_entry` constructors).

Docs-only, no behavior change. `cargo doc` builds clean with `-D warnings` (docs.rs feature set).

### DIFF
--- a/near-global-contracts/src/global_contract_identifier.rs
+++ b/near-global-contracts/src/global_contract_identifier.rs
@@ -2,13 +2,6 @@ use near_account_id::AccountId;
 
 use near_crypto_hash::CryptoHash;
 
-#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub enum AccountContract {
-    None,
-    Local(CryptoHash),
-    Global(CryptoHash),
-    GlobalByAccount(AccountId),
 /// What contract code an account currently runs.
 ///
 /// An account either has its own copy of the code (`Local`) or points to shared
--- a/near-global-contracts/src/global_contract_identifier.rs
+++ b/near-global-contracts/src/global_contract_identifier.rs
@@ -15,7 +15,7 @@ pub enum AccountContract {
 ///
 /// Global contracts let contract code be deployed once and shared across the whole
 /// network, so many accounts can run the same code without each storing its own copy.
-/// A global contract is referenced in one of two ways, one per variant below.
+/// A global contract is referenced in one of two ways: by hash of its code or by `AccountId` that deployed it.
 ///
 /// [global contract]: https://github.com/near/NEPs/pull/591
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
--- a/near-global-contracts/src/global_contract_identifier.rs
+++ b/near-global-contracts/src/global_contract_identifier.rs
@@ -11,6 +11,13 @@ pub enum AccountContract {
     GlobalByAccount(AccountId),
 }
 
+/// Identifies which [global contract] an account should run.
+///
+/// Global contracts let contract code be deployed once and shared across the whole
+/// network, so many accounts can run the same code without each storing its own copy.
+/// A global contract is referenced in one of two ways, one per variant below.
+///
+/// [global contract]: https://github.com/near/NEPs/pull/591
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 #[cfg_attr(
     feature = "serde",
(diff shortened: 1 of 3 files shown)


near-global-contracts/src/state_init.rs
```diff
@@ -4,6 +4,19 @@ use std::collections::BTreeMap;
...
+///
+/// A deterministic account ([NEP-616]) lives at an address computed from its own initial
+/// state instead of one chosen up front, so anyone who knows the state can work out the
+/// address ahead of time. This type is that state; pass it to
+/// [`StateInit::derive_account_id`] to get the resulting [`AccountId`].
+///
+/// It is versioned so the layout can change later without breaking existing data.
+/// [`V1`](StateInit::V1) is the only version today.
+///
+/// [deterministic account]: https://github.com/near/NEPs/pull/616
```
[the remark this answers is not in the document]
r-near: Fixed in 5598b16b. Both unconditional references to the `borsh`-gated method now use feature-safe code formatting and explicitly document the `borsh` requirement. Verified `cargo doc -p near-global-contracts --no-deps` with `RUSTDOCFLAGS='-D warnings'` both without and with the `borsh` feature.

near-global-contracts/src/global_contract_identifier.rs
```diff
@@ -11,6 +11,13 @@ pub enum AccountContract {
     GlobalByAccount(AccountId),
 }
 
+/// Identifies which [global contract] an account should run.
+///
+/// Global contracts let contract code be deployed once and shared across the whole
+/// network, so many accounts can run the same code without each storing its own copy.
+/// A global contract is referenced in one of two ways, one per variant below.
```
vsavchyn-dev: rustdoc already renders two ways we can reference global contracts:

Image

We should delete this line or change it to something like this:
```suggestion
/// A global contract is referenced in one of two ways: by hash of its code or by `AccountId` that deployed it.
```

near-global-contracts/src/state_init.rs
vsavchyn-dev: There is no `digest` feature

near-global-contracts/src/state_init.rs
```diff
@@ -4,6 +4,19 @@ use std::collections::BTreeMap;
...
 
+/// Initial on-chain state used to derive a [deterministic account].
+///
+/// A deterministic account ([NEP-616]) lives at an address computed from its own initial
+/// state instead of one chosen up front, so anyone who knows the state can work out the
+/// address ahead of time. This type is that state; enable the `borsh` feature and call
+/// `StateInit::derive_account_id` to get the resulting [`AccountId`].
+///
+/// It is versioned so the layout can change later without breaking existing data.
+/// [`V1`](StateInit::V1) is the only version today.
```
vsavchyn-dev:
```suggestion
/// Initial on-chain state used to derive a [deterministic account] ([NEP-616]).
///
/// A deterministic account lives at an address computed from its own initial state, so
/// anyone who knows the state can work out the address ahead of time. Enable the `borsh`
/// feature and call `StateInit::derive_account_id` to get the resulting [`AccountId`].
///
/// Version so the layout can evolve; [`V1`](StateInit::V1) is the only version today.
```

near-global-contracts/src/global_contract_identifier.rs
```diff
@@ -11,6 +11,13 @@ pub enum AccountContract {
     GlobalByAccount(AccountId),
 }
```
vsavchyn-dev: Maybe we should document this as well then :)
```suggestion
/// What contract code an account currently runs.
///
/// An account either has its own copy of the code (`Local`) or points to shared
/// [`GlobalContractId`] (`Global` / `GlobalByAccount`), letting many accounts run the
/// same code without each storing it.
#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
pub enum AccountContract {
    /// The account has no contract deployed and cannot be called.
    None,
    /// The code deployed directly to this account and stored in its own state, identified
    /// by the code hash.
    Local(CryptoHash),
    /// A global contract referenced by code hash - pins exact bytecode.
    /// Corresponds to [`GlobalContractId::CodeHash`]
    Global(CryptoHash),
    /// A global contract referenced by deployer account - follows that account's latest
    /// globally deployed code.
    /// Corresponds to [`GlobalContractId::AccountId`]
    GlobalByAccount(AccountId),
}
```

near-global-contracts/src/global_contract_identifier.rs
```diff
@@ -9,8 +9,35 @@ pub enum AccountContract {
...
     Global(CryptoHash),
     GlobalByAccount(AccountId),
+/// What contract code an account currently runs.
+///
+/// An account either has its own copy of the code (`Local`) or points to shared
+/// [`GlobalContractId`] (`Global` / `GlobalByAccount`), letting many accounts run the
+/// same code without each storing it.
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum AccountContract {
```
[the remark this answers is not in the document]
r-near: Fixed in f5ae5ada. Applying the suggestion replaced only the enum's closing brace, leaving the original declaration wrapped around the documented one — removed the leftover lines.

near-global-contracts/src/state_init.rs
r-near: Fixed in 5f0a430c

near-global-contracts/src/state_init.rs
```diff
@@ -4,6 +4,17 @@ use std::collections::BTreeMap;
 #[cfg(feature = "serde")]
 use serde_with::base64::Base64;
 
+/// Initial on-chain state used to derive a [deterministic account] ([NEP-616]).
+///
+/// A deterministic account lives at an address computed from its own initial state, so
+/// anyone who knows the state can work out the address ahead of time. Enable the `borsh`
+/// feature and call `StateInit::derive_account_id` to get the resulting [`AccountId`].
+///
+/// Version so the layout can evolve; [`V1`](StateInit::V1) is the only version today.
```
vsavchyn-dev, line 13: "Version" -> "Versioned"?

vsavchyn-dev (approved):

PR #1611, opened 2026-07-17
fix: disable near-vm-runner default features for unit-testing

r-near:
### Why

`unit-testing` depends on `near-vm-runner` without `default-features = false`, so it inherits the default `wasmtime_vm` feature — dragging in the wasmtime/cranelift backend and, via `prepare → metrics → near-o11y`, the full OpenTelemetry OTLP stack. None of it is used at test time: the mock env only touches near-vm-runner's always-compiled `logic` layer, never real wasm.

### Fix

Disable default features on `near-vm-runner`. Drops **339 → 202 crates (−137)**: wasmtime/cranelift/winch, opentelemetry + tonic/prost/hyper/tokio, finite-wasm, prometheus. All lib unit tests still pass.

### DIFF
--- a/near-sdk/Cargo.toml
+++ b/near-sdk/Cargo.toml
@@ -67,7 +67,7 @@ p256 = { version = "0.13.2", default-features = false, features = [
 near-abi = { version = "0.4.0", features = [
     "__chunked-entries",
 ], optional = true }
-near-vm-runner = { version = "0.37.0", optional = true }
+near-vm-runner = { version = "0.37.0", default-features = false, optional = true }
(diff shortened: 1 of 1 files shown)


vsavchyn-dev (approved):
I don't see a reason why we shouldn't remove default features. From what I see, we use `External` trait and just in general mock blockchain and don't use wasmtime VM or metrics.

Just as a remark - `prepare` is also enabled by default for wasmtime which seems to be responsible for calculating gas. Doesn't look like we were using it in any case, but we can improve unit-testing by somehow mocking `prepare` module for better gas usage estimations in future...?

PR #1613, opened 2026-07-17
feat: typed `env::bls12381` module

### LINKED ISSUE #1612: Improve ergonomics of BLS12-381 wrappers in near_sdk::env
The `bls12381_*` host-function wrappers in `near-sdk/src/environment/env.rs` are hard to use correctly.

**1. No documentation.** The wrappers have one-line doc comments and don't describe the byte layout callers must produce (big-endian x‖y points, `c1‖c0` for Fp², little-endian scalars, a leading sign byte for the sum ops). The only way to learn the encoding today is to read the [nearcore macros](https://github.com/near/nearcore/blob/50464a6e382627ba77f02e6c18d0459c0c2d1692/runtime/near-vm-runner/src/wasmtime_runner/logic.rs#L1379-L1421).

**2. Untyped slices in, `Vec<u8>` out.** Every fn takes `impl AsRef<[u8]>` and returns `Vec<u8>`, so callers hand-pack a flat buffer with exact offsets and re-slice the output. Nothing about the encoding is discoverable from the signature.

**3. Error codes dropped → panics instead of `Result`/`Option`.** Host fns return `0 = ok`, `1 = invalid input`. Most wrappers ignore the return value, so on invalid input the register is never written and `read_register` panics with a cryptic `"register expected"` (e.g. `bls12381_p1_decompress`, env.rs:1012-1021) instead of returning an error. Separately, `pairing_check` (env.rs:1006-1009) collapses three host states — `0 = pass`, `2 = fail`, `1 = invalid input` — into a bool via `== 0`, so malformed input can't be told apart from a legitimately-failing pairing.

**Open question — how far to take the API?** A spectrum, roughly:
- **Minimal:** just fix (3) — check the return code and return `Result`/`Option` — plus real docs. Smallest change, unblocks the worst footgun.
- **Typed wrappers:** newtypes that name the encoding (`G1`, `G2`, `Fp`, `Fp2`, `Scalar`, `Sign`, compressed variants) with the SDK owning buffer packing; batch fns returning `Result`.
- **High-level API:** actual point/scalar types with arithmetic, validated (de)serialization, pairing as a method. Nicest to use, but note these host fns are batch primitives (`sum`/`multiexp` fold N inputs in one call) — an operator-overloaded API risks turning one batched call into N, so it partly fights the gas model.

Worth deciding which level we want. Also: `alt_bn128_*` and `ecrecover` share the same untyped shape, so there's a consistency question of whether this stays BLS-scoped or covers all curve wrappers.

r-near:
Closes #1612.

## Motivation

The `bls12381_*` wrappers in `near_sdk::env` are hard to use correctly:

1. **Undocumented byte layout.** #1623 documents input sizes and failure modes, but not the encoding inside each element.
2. **Untyped** — every fn takes `impl AsRef<[u8]>` and returns raw bytes, so callers hand-pack a flat buffer at exact offsets (big-endian `x‖y` points, `c1‖c0` for `Fp2`, little-endian scalars, a leading sign byte for the sum ops).

Error handling was fixed separately in #1623: the byte-slice functions now check the host status code and return `None` on invalid input, and `pairing_check` returns `Option<bool>`, so a malformed input is distinguishable from a failing check. This PR builds on that.

## What this does

Adds a `near_sdk::env::bls12381` module (the "typed wrappers" direction, kept intentionally close to the host primitives — no in-wasm curve arithmetic, no eager validation, so there's no added gas cost):

- **Named types**: `G1`, `G2`, `G1Compressed`, `G2Compressed`, `Fp`, `Fp2`, `Scalar`, `Sign` — each (except `Sign`) a newtype over a fixed-size byte array, so the length is enforced by the type and the encoding is documented, with `From<[u8; N]>` / `TryFrom<&[u8]>` / `as_bytes()`. Contents are validated by the host when passed to a function, not on construction.
- **Batch functions** that own the buffer packing and return `Result<_, Error>`: `p1_sum`, `p2_sum`, `g1_multiexp`, `g2_multiexp`, `map_fp_to_g1`, `map_fp2_to_g2`, `p1_decompress`, `p2_decompress`, `pairing_check`. Plus per-element convenience methods (`G1Compressed::decompress`, `Fp::map_to_g1`, …).
- `pairing_check -> Result<bool, Error>`: `Ok(true)` = passes, `Ok(false)` = fails, `Err` = malformed input.
- A module-level `no_run` doc example showing signature verification via a batched pairing check.

The old `env::bls12381_*` functions are **kept and `#[deprecated]`** (pointing at the new module), so this PR is non-breaking. (The signature change to those functions landed in #1623.)

## Testing

- 14 new unit tests under `mod bls12381_typed` covering success paths, error paths (`Err(InvalidInput)`), the three-state `pairing_check`, decompress round-trips + convenience methods, empty-batch behavior, and `TryFrom` length checks. Valid points are bootstrapped via `map_fp_to_g1` / known-good vectors.
- Existing byte-slice tests, including the rejection and stale-register tests from #1623, retained under `#[allow(deprecated)]` as back-compat coverage.
- `cargo fmt`, `cargo clippy`, and `cargo doc` (with `-D warnings`) are clean; the `no_run` doc example compiles under the non-`unit-testing` feature sets CI uses for doctests.

One gotcha surfaced by the tests and now documented: the BLS12-381 **point at infinity** is encoded with the infinity flag (`0x40` in the first byte), *not* as all-zeros.

## Notes / follow-ups

- Rebased onto #1623. The last commit adapts to it: `#[allow(deprecated)]` on the tests #1623 added (they fail to build under `-D warnings` otherwise), and the deprecation notes and module docs no longer describe the old functions as panicking on invalid input.
- Design discussion (minimal fix vs. typed wrappers vs. a full high-level point API) is in #1612. This PR is the typed-wrappers option.
- `alt_bn128_*` and `ecrecover` share the same untyped-slice shape; giving them the same treatment is a natural follow-up but out of scope here.
- Possible future additions: `borsh`/`serde` derives on the point types, and feature-gated conversions to/from `bls12_381`/`blst` (the encoding is the ZCash BLS12-381 layout those crates use — note it is *not* EIP-2537, which zero-pads to 64-byte field elements).

### DIFF
--- a/near-sdk/src/environment/env.rs
+++ b/near-sdk/src/environment/env.rs
@@ -1086,8 +1086,9 @@ pub mod bls12381 {
     //!
     //! # Encoding
     //!
-    //! All encodings match the ones used by the NEAR protocol host functions (big-endian field
-    //! elements; the same serialization as [EIP-2537] / the ZCash BLS12-381 spec):
+    //! All encodings match the ones used by the NEAR protocol host functions: the "ZCash"
+    //! BLS12-381 byte layout with big-endian field elements (the same serialization the zkcrypto
+    //! `bls12_381` crate produces):
     //!
     //! | Type             | Size (bytes) | Layout                                          |
     //! |------------------|-------------:|-------------------------------------------------|
@@ -1104,6 +1105,10 @@ pub mod bls12381 {
     //!
     //! The point at infinity is encoded with the infinity flag set (`0x40` in the first byte),
     //! not as all zeros.
+    //!
+    //! Note this is **not** the EIP-2537 encoding used by the Ethereum precompiles: that spec
+    //! zero-pads field elements to 64 bytes (128-byte `G1`, 256-byte `G2`) and defines no
+    //! compressed form, so EIP-2537 byte vectors are not interchangeable with these types.
     //!
     //! These functions are batch primitives — folding or mapping every element in a single host
     //! call. Prefer passing all your inputs to one call over calling them in a loop.
@@ -1133,7 +1138,6 @@ pub mod bls12381 {
     //! ```
     //!
     //! [BLS12-381]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-pairing-friendly-curves
-    //! [EIP-2537]: https://eips.ethereum.org/EIPS/eip-2537
     use super::{ATOMIC_OP_REGISTER, read_register, read_register_fixed, sys};
 
     /// Size in bytes of an uncompressed G1 point.
--- a/near-sdk/src/environment/env.rs
+++ b/near-sdk/src/environment/env.rs
@@ -931,6 +931,10 @@ pub fn alt_bn128_pairing_check(value: impl AsRef<[u8]>) -> bool {
 /// Compute BLS12-381 G1 sum.
 ///
 /// See also: [IETF draft-irtf-cfrg-pairing-friendly-curves](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-pairing-friendly-curves)
+#[deprecated(
+    since = "5.30.0",
+    note = "use the typed `near_sdk::env::bls12381::p1_sum`, which names the byte layout and returns a `Result` instead of panicking on invalid input"
+)]
 pub fn bls12381_p1_sum(value: impl AsRef<[u8]>) -> Vec<u8> {
     let value = value.as_ref();
     unsafe {
@@ -943,6 +947,10 @@ pub fn bls12381_p1_sum(value: impl AsRef<[u8]>) -> Vec<u8> {
 }
 
 /// Compute BLS12-381 G2 sum.
+#[deprecated(
+    since = "5.30.0",
+    note = "use the typed `near_sdk::env::bls12381::p2_sum`, which names the byte layout and returns a `Result` instead of panicking on invalid input"
+)]
 pub fn bls12381_p2_sum(value: impl AsRef<[u8]>) -> Vec<u8> {
     let value = value.as_ref();
     unsafe {
@@ -955,6 +963,10 @@ pub fn bls12381_p2_sum(value: impl AsRef<[u8]>) -> Vec<u8> {
 }
 
 /// Compute BLS12-381 G1 multiexponentiation.
+#[deprecated(
+    since = "5.30.0",
+    note = "use the typed `near_sdk::env::bls12381::g1_multiexp`, which names the byte layout and returns a `Result` instead of panicking on invalid input"
+)]
 pub fn bls12381_g1_multiexp(value: impl AsRef<[u8]>) -> Vec<u8> {
     let value = value.as_ref();
     unsafe {
@@ -967,6 +979,10 @@ pub fn bls12381_g1_multiexp(value: impl AsRef<[u8]>) -> Vec<u8> {
 }
 
 /// Compute BLS12-381 G2 multiexponentiation.
+#[deprecated(
+    since = "5.30.0",
+    note = "use the typed `near_sdk::env::bls12381::g2_multiexp`, which names the byte layout and returns a `Result` instead of panicking on invalid input"
+)]
 pub fn bls12381_g2_multiexp(value: impl AsRef<[u8]>) -> Vec<u8> {
     let value = value.as_ref();
     unsafe {
@@ -979,6 +995,10 @@ pub fn bls12381_g2_multiexp(value: impl AsRef<[u8]>) -> Vec<u8> {
 }
 
 /// Map an Fp element to a BLS12-381 G1 point.
+#[deprecated(
+    since = "5.30.0",
+    note = "use the typed `near_sdk::env::bls12381::map_fp_to_g1`, which names the byte layout and returns a `Result` instead of panicking on invalid input"
+)]
 pub fn bls12381_map_fp_to_g1(value: impl AsRef<[u8]>) -> Vec<u8> {
     let value = value.as_ref();
     unsafe {
@@ -991,6 +1011,10 @@ pub fn bls12381_map_fp_to_g1(value: impl AsRef<[u8]>) -> Vec<u8> {
 }
 
 /// Map an Fp2 element to a BLS12-381 G2 point.
+#[deprecated(
+    since = "5.30.0",
+    note = "use the typed `near_sdk::env::bls12381::map_fp2_to_g2`, which names the byte layout and returns a `Result` instead of panicking on invalid input"
+)]
 pub fn bls12381_map_fp2_to_g2(value: impl AsRef<[u8]>) -> Vec<u8> {
     let value = value.as_ref();
     unsafe {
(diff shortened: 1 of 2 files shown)


r-near:
@codex review

near-sdk/src/environment/env.rs
```diff
@@ -1032,6 +1071,450 @@ pub fn bls12381_p2_decompress(value: impl AsRef<[u8]>) -> Vec<u8> {
...
+    //!   caller has to hand-pack a flat buffer at exact offsets.
+    //! * Every fallible operation returns a [`Result`] with an [`Error`] on malformed input,
+    //!   instead of panicking with an opaque "register expected" message.
+    //! * [`pairing_check`] returns `Result<bool, Error>`, so a malformed input (`Err`) is
+    //!   distinguishable from a pairing that simply does not hold (`Ok(false)`).
+    //!
+    //! # Encoding
+    //!
+    //! All encodings match the ones used by the NEAR protocol host functions (big-endian field
+    //! elements; the same serialization as [EIP-2537] / the ZCash BLS12-381 spec):
```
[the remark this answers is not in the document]
r-near: Good catch — fixed in d841aa55. EIP-2537 pads field elements to 64 bytes (128-byte G1 / 256-byte G2) and has no compressed form, so it isn't interchangeable with these types. Removed the EIP-2537 equivalence, kept the (correct) ZCash/`bls12_381` layout description, and added an explicit note that EIP-2537 vectors are not compatible.

near-sdk/src/environment/env.rs
```diff
@@ -950,6 +950,10 @@ fn bls12381_register_result(status: u64) -> Option<Vec<u8>> {
 /// See also: [IETF draft-irtf-cfrg-pairing-friendly-curves](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-pairing-friendly-curves)
 ///
 /// [NEP-488]: https://github.com/near/NEPs/blob/master/neps/nep-0488.md
+#[deprecated(
+    since = "5.30.0",
+    note = "use the typed `near_sdk::env::bls12381::p1_sum`, which takes and returns typed values instead of hand-packed byte buffers"
+)]
```
vsavchyn-dev:
```suggestion

```

Having these "deprecated" macros on bls12381 wrappers creates ambiguity, as we don't have any typed structs for other functions like `ed25519_verify`.

They should be removed.

near-sdk/src/environment/env.rs
```diff
@@ -1081,6 +1117,453 @@ pub fn bls12381_p2_decompress(value: impl AsRef<[u8]>) -> Option<Vec<u8>> {
     bls12381_register_result(status)
 }
 
+pub mod bls12381 {
```
vsavchyn-dev: Having this module inside `env.rs` that provides wrappers around low-level `near-sys` looks weird and out of place. I would rather have it in it's own file and let the people choose if they want to use these typed structs or simple slices that host function returns.

near-sdk/src/environment/env.rs
```diff
@@ -928,107 +922,645 @@ pub fn alt_bn128_pairing_check(value: impl AsRef<[u8]>) -> bool {
...
+            }
+        }
+    }
+
+    /// Defines a fixed-size byte-array newtype with the common conversions.
+    macro_rules! byte_newtype {
+        ($(#[$meta:meta])* $name:ident, $len:expr) => {
+            $(#[$meta])*
+            #[derive(Clone, Copy, PartialEq, Eq, Hash)]
+            pub struct $name(pub [u8; $len]);
```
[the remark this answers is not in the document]
vsavchyn-dev: Quite valid comment imo. Let's say we want to include this struct into another one that deserializes with serde. If it's transparent, it will deserialize into array. Without `transaparent` representation it will be a nested struct, no?

Correct me if I`m wrong here @r-near

near-sdk/src/environment/env.rs
```diff
@@ -1081,6 +1117,453 @@ pub fn bls12381_p2_decompress(value: impl AsRef<[u8]>) -> Option<Vec<u8>> {
...
+        debug_assert_eq!(raw.len() % N, 0, "register length must be a multiple of the point size");
+        raw.chunks_exact(N)
+            .map(|chunk| {
+                let arr: [u8; N] = chunk.try_into().expect("chunks_exact yields N-byte chunks");
+                T::from(arr)
+            })
+            .collect()
+    }
+
+    /// Compute the BLS12-381 G1 sum `Σ ±pᵢ` over the given signed points.
```
vsavchyn-dev: I would rather not have this "Σ ±pᵢ" formula, considering that text already provides us with what this function does (aka a BLS112-381 G1 sum over the given signed points `i`). Same goes for other usage of special unicode math chars that appear in other functions.

near-sdk/src/environment/env.rs
```diff
@@ -3185,6 +3668,7 @@ mod tests {
     }
 
     #[test]
+    #[allow(deprecated)]
```
vsavchyn-dev:
```suggestion
```

Removing `#[deprecated]` macro on original bls wrapper functions will also remove the need to have `#[allow(deprecated)]` in tests... Less code :)

vsavchyn-dev (changes requested):
I believe we shouldn't enforce people to use this typed module by deprecating original host functions wrappers.

I think we should put this module outside the `env.rs` file into it's own file, considering that env module is just a set of wrappers around host functions. This way, people will be able to choose if they want to use low-level wrappers or actual typed wrappers with their own optimizations...

near-sdk/src/environment/env.rs
r-near: Removed here: [`a3382d4` (this PR)](https://github.com/near/near-sdk-rs/pull/1613/commits/a3382d49ba602142b813ff3ca09214f80a324f3e)

near-sdk/src/environment/env.rs
r-near: Moved it out: [`f279e89` (this PR)](https://github.com/near/near-sdk-rs/pull/1613/commits/f279e8902612bf81b52ba0468694dcf0aa33aada)

near-sdk/src/environment/env.rs
r-near: Cleaned it up: [`71d2d55` (this PR)](https://github.com/near/near-sdk-rs/pull/1613/commits/71d2d558464f795f9f7db913ca6682e2ad74b4aa)

near-sdk/src/environment/env.rs
r-near: Yep 😁 added here: [`a3382d4` (this PR)](https://github.com/near/near-sdk-rs/pull/1613/commits/a3382d49ba602142b813ff3ca09214f80a324f3e)

vsavchyn-dev (approved):
LGTM! I would still love to hear feedback from @gilcu3 for instance, as I only know nothing about the usage of these functions in actual code :)

near-sdk/src/bls12381.rs
```diff
@@ -0,0 +1,631 @@
...
+//!
+//! ```no_run
+//! use near_sdk::bls12381::{self, Error, G1Compressed, G2Compressed, G1, G2};
+//!
+//! fn verify_signature(
+//!     pubkey_compressed: [u8; 48],    // public key, a compressed G1 point
+//!     signature_compressed: [u8; 96], // signature, a compressed G2 point
+//!     hashed_message: G2,             // H(m), the message hashed to a G2 point
+//!     neg_g1_generator: G1,           // the negated G1 generator
+//! ) -> Result<bool, Error> {
```
gilcu3, line 55: I don't understand this example. Why would it take the `neg_g1_generator` parameter if that's a constant anyway?

near-sdk/src/bls12381.rs
```diff
@@ -0,0 +1,631 @@
...
+            let arr: [u8; N] = chunk.try_into().expect("chunks_exact yields N-byte chunks");
+            T::from(arr)
+        })
+        .collect()
+}
+
+/// Compute the BLS12-381 G1 sum over the given signed points.
+///
+/// Returns [`Error::InvalidInput`] if any point is not correctly encoded or not on the curve.
+pub fn p1_sum(summands: &[(Sign, G1)]) -> Result<G1, Error> {
```
gilcu3, line 258: these functions should probably also carry the explicit caveat that they do not enforce the point being in the prime subgroup

near-sdk/src/bls12381.rs
```diff
@@ -0,0 +1,631 @@
...
+    // exactly `N` bytes.
+    unsafe { read_register_fixed::<N>(ATOMIC_OP_REGISTER) }
+}
+
+/// Reads a batch of fixed-size points out of the atomic register after a successful host call.
+#[inline]
+fn read_points<const N: usize, T: From<[u8; N]>>() -> Vec<T> {
+    // The host wrote `k * N` bytes for `k` output points (possibly zero). `read_register`
+    // returns `Some` even for an empty register once it has been written to.
+    let raw = read_register(ATOMIC_OP_REGISTER).unwrap_or_default();
```
gilcu3, line 243: question: why is this better than just crashing or returning error?

gilcu3 (commented):
Post merge review, LGTM, left some minor comments

PR #1623, opened 2026-09-15
fix(env)!: bls host functions should return None on error

gilcu3:
PS: certainly related to #1613, but more pressing IMO


gilcu3:
@r-near could you take a look when you have a chance?

r-near (approved):

r-near:
Looks good, I'm hoping we can deprecate these soon, but thanks for the correctness fix!

PR #1625, opened 2026-09-16
test: back the unit-testing mock with nearcore's host functions (validation, git-pinned)

r-near:
Validates near/nearcore#16428 from the SDK side. With `HostCtx` and the host functions
callable without wasmtime, `unit-testing` no longer needs its own host: the mock becomes
argument marshalling over `near_vm_runner::logic::host::*`.

**Not mergeable as-is.** Every nearcore crate is git-pinned to the #16428 commit, plus a
revert of nearcore#16420 whose exact `aws-lc-rs = "=1.18.1"` collides with the
`near-workspaces 0.23` dev-dependency's `near-crypto 0.37` (`=1.16.2`). That collision
disappears once near-workspaces moves to nearcore 0.38. Swap back to `=0.38.x` from
crates.io once #16428 ships.

`MockedBlockchain` owns a `HostCtx<'static>` instead of a `VMLogic`. Each of the 89
`#[no_mangle]` shims copies its `(ptr, len)` arguments into a per-call scratch buffer,
passes offsets, and copies back what the host wrote; `mocked_memory.rs` is gone. No
hand-written host behaviour is left behind the shims.

Source-compatible: `testing_env!`, `VMContextBuilder`, `VMContext`, `get_logs`,
`get_created_receipts`, `Receipt`/`MockAction`, `MockedBlockchain::new`,
`set_blockchain_interface`, `with_mocked_blockchain`, `take_storage`, `created_receipts`,
`logs`, `gas`, `test_vm_config()`. `MockedBlockchain` loses its unused `<Memory>` type
parameter; the `memory` argument stays so `::new(.., None)` call sites are unchanged.

Verified locally: workspace tests on the three CI feature sets, wasm32 and `--cfg near`
checks, `cargo fmt` and both clippy invocations on the pinned 1.93 toolchain. The gas
upper bounds in `event_tests.rs` still hold.


r-near:
Closing in favor of larger gh stack

PR #1628, opened 2026-09-17
feat(near-sys): add ml_dsa_verify, sha3, and universal state init bindings (nearcore 2.14)

r-near:
## What

Adds the protocol 87 host function declarations to `near-sys`, with the signatures from near-vm-runner `0.38.0-rc.2`:

- `ml_dsa_verify`
- `sha3_256`, `sha3_384`, `sha3_512`
- `universal_state_init_to_account_id`
- `promise_batch_action_universal_state_init`

No versions move here: the `near-sys` bump and the workspace requirement on it are left to release-plz.

## Why

`near-sys` is the raw binding layer, so it has to carry the new imports before anything in `near-sdk` can call them. Nothing else changes: these are `extern` declarations with no call sites yet, and the crate builds on master today.


r-near:
@greptileai

vsavchyn-dev (approved):

PR #1630, opened 2026-09-17
test(examples): pin sandbox to 2.14.0-rc.2

r-near:
## What

Replaces every `near_workspaces::sandbox()` call in `examples/` and in the two `near-sdk` integration tests with `sandbox_with_version("2.14.0-rc.2")`.

## Why

The next PR makes `Promise::add_action_to_batch` grow an unconditional `UniversalStateInit` arm, so every example contract's wasm imports `promise_batch_action_universal_state_init`. near-sandbox still defaults to 2.13.4, protocol 86, which rejects that import with "Link Error: unknown or invalid import".

Every call site is pinned, not just the ones that need it today, so the eventual revert is one commit. This mirrors the 2.13 campaign. A 2.14 sandbox runs 2.13-era wasm unchanged, so on its own this PR is a no-op, and `store_performance_tests` gas budgets are unaffected.

Revert to bare `sandbox()` once near-sandbox defaults to 2.14, near/near-sandbox-rs#97.


r-near:
@greptileai

vsavchyn-dev (approved):

PR #1631, opened 2026-09-17
feat: add ml_dsa_verify, sha3 hashes, and universal state init (nearcore 2.14)

r-near:
## What

The user-facing protocol 87 surface.

- `env::ml_dsa_verify(&[u8; 3309], msg, &[u8; 1952]) -> bool`, mirroring `ed25519_verify` and `p256_verify`. Off-chain it verifies through `near-crypto`.
- `env::sha3_256`, `env::sha3_384` and `env::sha3_512`, returning `[u8; 32/48/64]`, provided by `near-sdk-env` with a pure-Rust fallback like `keccak256_array`.
- `universal_state_init::{UniversalStateInit, UniversalStateInitV1, PublicKeyHandle}`, a typed state init whose borsh matches `near-primitives::universal_state_init` byte for byte, pinned by tests against nearcore's own vectors, with `derive_account_id()` producing the `0u` id. `UniversalStateInit` and `UniversalStateInitV1` are defined in `near-global-contracts` next to `StateInit`, `PublicKeyHandle` in `near-sdk-core` next to `PublicKey`, and all three are re-exported from `near_sdk::universal_state_init`, so contract code is unchanged. Off-chain consumers can derive a `0u` account id from `near-global-contracts` alone, without near-sdk.
- `env::universal_state_init_to_account_id` and `env::promise_batch_action_universal_state_init`, plus `*_raw` byte variants so a contract can pass through a state-init version this SDK predates, and `Promise::universal_state_init`.
- `near-digest`'s `Sha3_256`, `Sha3_384` and `Sha3_512` route to the host functions under `cfg(near)` and no longer need `unstable`. The feature is kept, now empty, so existing dependents keep building.
- `package.metadata.near.min_protocol_version` moves to 87.

## Why

nearcore 2.14 ships protocol 87. The bindings are unconditional, so contracts built with this release require nearcore 2.14 or newer, the same way 5.29 raised the floor to 85 with the gas-key imports. cargo-near reads `min_protocol_version` to decide what a dependent contract targets, so it has to move with the floor.

New hash functions follow a new convention: the bare hash name returns the fixed-size array, and the legacy `Vec` and `_array` pair is not extended to new hashes.

This is the only PR whose example wasm imports the new host function, which is why it sits on top of the sandbox pin.


CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- support the nearcore 2.14 / protocol 87 host functions: `env::ml_dsa_verify`, `env::sha3_256`/`sha3_384`/`sha3_512`, `env::universal_state_init_to_account_id`, `env::promise_batch_action_universal_state_init` (plus `*_raw` byte variants), the typed `universal_state_init::{UniversalStateInit, UniversalStateInitV1, PublicKeyHandle}` and `Promise::universal_state_init`; `near-digest`'s `Sha3_256`/`Sha3_384`/`Sha3_512` use the host functions on-chain and are no longer behind `unstable`. New hash functions follow a new convention: the bare hash name returns the fixed-size array, and the legacy `Vec`/`_array` pair is not extended to new hashes ([#1624](https://github.com/near/near-sdk-rs/pull/1624))
+
+### Changed
+
+- **protocol floor is now 87**: the new host-function imports are unconditional, so contracts built with this release require nearcore 2.14+ (protocol 87), the same way 5.29 raised the floor to 85 with the gas-key imports. nearcore crates are pinned to `0.38.0-rc.2` and `near-account-id` moves to 3 ([#1624](https://github.com/near/near-sdk-rs/pull/1624))
```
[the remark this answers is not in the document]
r-near: Agreed, fixed in adc6f173: `package.metadata.near.min_protocol_version` is now 87, with the comment updated alongside it. Verified that cargo-near still builds `examples/adder` with the new value.

r-near:
@greptileai

CHANGELOG.md
```diff
@@ -7,8 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- support the nearcore 2.14 / protocol 87 host functions: `env::ml_dsa_verify`, `env::sha3_256`/`sha3_384`/`sha3_512`, `env::universal_state_init_to_account_id`, `env::promise_batch_action_universal_state_init` (plus `*_raw` byte variants), the typed `universal_state_init::{UniversalStateInit, UniversalStateInitV1, PublicKeyHandle}` and `Promise::universal_state_init`; `near-digest`'s `Sha3_256`/`Sha3_384`/`Sha3_512` use the host functions on-chain and are no longer behind `unstable`. New hash functions follow a new convention: the bare hash name returns the fixed-size array, and the legacy `Vec`/`_array` pair is not extended to new hashes ([#1631](https://github.com/near/near-sdk-rs/pull/1631))
+
```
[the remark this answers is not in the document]
r-near: Not an issue for this stack: the BLS `Option` change is commit ade3de49 (`fix(env)!: bls host functions should return None on error`, #1623), which is already on master and outside these PRs, and release-plz generates CHANGELOG.md from that breaking-change-marked subject at release time.

r-near:
@greptileai

r-near:
@greptileai

r-near:
@greptileai

near-digest/src/lib.rs
```diff
@@ -46,8 +46,9 @@
 //! | `sha2::Sha256` | `sha2` | `sha256_array` |
 //! | `sha3::Keccak256` | `sha3` | `keccak256_array` |
 //! | `sha3::Keccak512` | `sha3` | `keccak512_array` |
-//! | `sha3::Sha3_256` | `sha3` + `unstable` | - (pure Rust on all targets for now) |
-//! | `sha3::Sha3_512` | `sha3` + `unstable` | - (pure Rust on all targets for now) |
+//! | `sha3::Sha3_256` | `sha3` | `sha3_256` (nearcore 2.14+) |
+//! | `sha3::Sha3_384` | `sha3` | `sha3_384` (nearcore 2.14+) |
+//! | `sha3::Sha3_512` | `sha3` | `sha3_512` (nearcore 2.14+) |
```
vsavchyn-dev: We'll need to remove references to "2.14+" in actual release - I don't think people really care :)

near-global-contracts/src/universal_state_init.rs
```diff
@@ -0,0 +1,337 @@
...
+    /// Canonical borsh encoding of this state init: the bytes the account id commits to, and what
+    /// the `promise_batch_action_universal_state_init` host function sends to the runtime.
+    ///
+    /// # Availability
+    ///
+    /// Requires the `borsh` feature.
+    #[cfg(feature = "borsh")]
+    pub fn to_bytes(&self) -> Vec<u8> {
+        borsh::to_vec(self).unwrap_or_else(|_| unreachable!())
+    }
```
vsavchyn-dev, line 161: I am a bit confused, is this an accessor method similar to `to_raw` in [near-primitives](https://github.com/near/nearcore/blob/21ec809d1530c73f68b61509263e0c7c0f6dbb06/core/primitives/src/universal_state_init.rs#L108-L111)? If that's the case, shouldn't we have `RawStateInit` type and `from_bytes`?

Not sure if we need to use `RawStateInit` at all in sdk code, considering that it is used in `near-vm-runner` only.

near-global-contracts/src/universal_state_init.rs
```diff
@@ -0,0 +1,337 @@
...
+    #[cfg(any(near, feature = "__near-sdk-unit-testing"))]
+    {
+        near_sdk_env::sha3_256(input)
+    }
+    #[cfg(not(any(near, feature = "__near-sdk-unit-testing")))]
+    {
+        use sha3::Digest;
+        sha3::Sha3_256::digest(input).into()
+    }
+}
```
vsavchyn-dev, line 213:
```suggestion
pub fn derive_universal_account_id(state_init: &[u8]) -> near_account_id::AccountId {
    let hash: [u8; 32];

    #[cfg(any(near, feature = "__near-sdk-unit-testing"))]
    {
        hash = near_sdk_env::sha3_256(state_init);
    }
    #[cfg(not(any(near, feature = "__near-sdk-unit-testing")))]
    {
        use sha3::Digest;
        hash = sha3::Sha3_256::digest(state_init).into();
    }

    UniversalAccountId::from_hash(hash).into_account_id()
}
```

I don't think it's worth to have this function if we can simply have it written down like that.

near-sdk-core/src/types/public_key.rs
```diff
@@ -313,6 +313,183 @@ impl From<B58Error> for ParsePublicKeyError {
...
+        }
+
+        if let Some(data) = value.strip_prefix(ML_DSA_65_HASH_PREFIX) {
+            return decode(data).map(Self::MLDSA65Hash);
+        }
+        if value.starts_with("ml-dsa-65:") {
+            return Err(ParsePublicKeyHandleError {
+                message: "a full ml-dsa-65 public key is not a handle; convert it with \
+                          `PublicKeyHandle::from(PublicKey)` or use the `ml-dsa-65-hash:` form",
+            });
```
vsavchyn-dev: Why this cannot go to `match curve` block? Also, I believe it deserves it's own error type (aka `Self::Err::InvalidData`).

near-sdk-core/src/types/public_key.rs
```diff
@@ -313,6 +313,183 @@ impl From<B58Error> for ParsePublicKeyError {
...
+    }
+}
+
+impl std::error::Error for ParsePublicKeyHandleError {}
+
+impl std::str::FromStr for PublicKeyHandle {
+    type Err = ParsePublicKeyHandleError;
+
+    fn from_str(value: &str) -> Result<Self, Self::Err> {
+        fn decode<const N: usize>(data: &str) -> Result<[u8; N], ParsePublicKeyHandleError> {
```
vsavchyn-dev: 1. Let's rename this function to `decode_base58`
2. How about putting it outside PublicKeyHandle parsing? I think we could reuse it in `PublicKey` parsing if needed, no?

near-sdk-core/src/types/public_key.rs
```diff
@@ -313,6 +313,183 @@ impl From<B58Error> for ParsePublicKeyError {
...
+            return Err(ParsePublicKeyHandleError {
+                message: "a full ml-dsa-65 public key is not a handle; convert it with \
+                          `PublicKeyHandle::from(PublicKey)` or use the `ml-dsa-65-hash:` form",
+            });
+        }
+        let (curve, data) = match value.split_once(':') {
+            Some((curve, data)) => (curve.parse::<CurveType>(), data),
+            // No prefix: default to ed25519, like `PublicKey`.
+            None => (Ok(CurveType::ED25519), value),
+        };
```
vsavchyn-dev: Can't `split_key_type_data` be used here?

near-sdk-core/src/types/public_key.rs
```diff
@@ -313,6 +313,183 @@ impl From<B58Error> for ParsePublicKeyError {
...
+/// into its on-trie handle.
+const ML_DSA_65_PUBKEY_HASH_DOMAIN: &[u8] = b"near:ml-dsa-65-pubkey-hash:v1";
+
+/// Length of a full ML-DSA-65 public key.
+const ML_DSA_65_PUBLIC_KEY_LENGTH: usize = 1952;
+
+const ML_DSA_65_HASH_PREFIX: &str = "ml-dsa-65-hash:";
+
+/// SHA3-256, through the host function on-chain and pure Rust everywhere else.
+fn sha3_256(input: &[u8]) -> [u8; 32] {
```
vsavchyn-dev: Either we should put `#[inline]` directive here, or inline this `sha3_256` function ourselves.

Similar to one of the previous comments, I think we don't need it and can inline `sha3_256` ourselves (it's used 2 times here only).

near-sdk-core/src/types/public_key.rs
```diff
@@ -313,6 +313,183 @@ impl From<B58Error> for ParsePublicKeyError {
...
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde_with::SerializeDisplay, serde_with::DeserializeFromStr))]
+#[cfg_attr(
+    feature = "borsh",
+    derive(borsh::BorshSerialize, borsh::BorshDeserialize),
+    borsh(use_discriminant = true)
+)]
+#[cfg_attr(feature = "abi", derive(borsh::BorshSchema))]
+#[repr(u8)]
+pub enum PublicKeyHandle {
```
vsavchyn-dev, line 378: Shouldn't there be `near-crypto-interop` for `PublicKeyHandle`?

Situation with `PublicKeyHandle` is similar to `PublicKey` after all and we need to ensure our in-house types work with `near-crypto`, so `From<near_crypto::PublicKey> for PublicKeyHandle` and `From<near_crypto::PublicKeyHandle> for PublicKeyHandle` must be implemented.

near-sdk-core/src/types/public_key.rs
```diff
@@ -313,6 +313,183 @@ impl From<B58Error> for ParsePublicKeyError {
...
+    #[cfg(any(near, feature = "__near-sdk-unit-testing"))]
+    {
+        near_sdk_env::sha3_256(input)
+    }
+    #[cfg(not(any(near, feature = "__near-sdk-unit-testing")))]
+    {
+        use sha3::Digest;
+        sha3::Sha3_256::digest(input).into()
+    }
+}
```
vsavchyn-dev, line 346: Damn, I really hate that we cannot use `near-sdk-env` bc `near-global-contracts` relies on the fact that `near-sdk-env` will be pulled only in `cfg(near)` case...

We will need to add ability to enable only specific host fn's and improve `near-sdk-env` integration with `near-digest`, so that `near-sdk-env` can be used as a single import (instead of duplicating `sha3` all the time :|).

near-sdk/src/environment/env.rs
```diff
@@ -2463,6 +2546,142 @@ pub fn promise_yield_resume_with_yield_id(yield_id: &[u8], data: impl AsRef<[u8]
...
+        assert_valid_account_id(expect_register(read_register(ATOMIC_OP_REGISTER)))
+    }
+
+    #[cfg(all(
+        not(target_arch = "wasm32"),
+        feature = "non-contract-usage",
+        any(not(feature = "unit-testing"), test),
+    ))]
+    {
+        near_global_contracts::universal_state_init::derive_universal_account_id(state_init)
```
vsavchyn-dev:
```suggestion
       UniversalAccountId::from_hash(sha3_256(state_init)).into_account_id()
```

How about we use `UniversalAccountId` here directly? I just don't want to have issues if `derive_universal_account_id` changes in any form.

near-global-contracts/src/lib.rs
```diff
@@ -55,6 +55,8 @@ pub use global_contract_identifier::*;
 mod state_init;
 pub use state_init::*;
 
+pub mod universal_state_init;
+
```
vsavchyn-dev:
```suggestion
mod universal_state_init;
pub use universal_state_init;
```

Shouldn't we export `universal_state_init` module similar to `state_init` one?

near-sdk/src/lib.rs
```diff
@@ -1630,6 +1630,8 @@ pub mod state;
 #[cfg(feature = "deterministic-account-ids")]
 pub mod state_init;
 
+pub mod universal_state_init;
+
```
vsavchyn-dev, line 1634:
```suggestion
#[cfg(feature = "univesal-state-init")]
pub mod universal_state_init;

```
Maybe we should gate `universal_state_init` similar to `state-init` above?

vsavchyn-dev (changes requested):
I like direction of universal state init, but we need to do something about `near-sdk-env` and duplication of crypto crates - it bugs me that we can't simply write `near_sdk_env::sha3_256` and should have this ugly duplicate check for `#[cfg(....)]`...

near-digest/src/lib.rs
r-near: Yeah good point, done here: [`9f1e55c` (this PR)](https://github.com/near/near-sdk-rs/pull/1631/changes/9f1e55c0c8770ab4903ab84f455172eef8f2c0ef)

near-global-contracts/src/universal_state_init.rs
r-near: Yes it's meant to be the `to_raw` equivalent. [Added `from_bytes` here](https://github.com/near/near-sdk-rs/pull/1631/commits/79ebeb94349c18619e118a7b59ff1f39fa2d0bf0) so it would match. 

IMO we don't need to mess with `RawStateInit` b/c in nearcore it's an unvalidated `Vec<u8>` wrapper, and we already take in `&[u8]` 🤷

near-global-contracts/src/universal_state_init.rs
r-near: Yea might as well inline it. Done here: [`81dd80d` (this PR)](https://github.com/near/near-sdk-rs/pull/1631/commits/81dd80df9b5e535e88beb0e857bd79b5db46b56c)

near-sdk-core/src/types/public_key.rs
r-near: No good reason. It can. Moved it here: [`a9736fb` (this PR)](https://github.com/near/near-sdk-rs/pull/1631/commits/a9736fb45c98e3431e909848337d83c10b1f535b)
For the errors, what do you think about passing through `ParsePublicKeyError`? Might save us some code instead of `ParsePublicKeyHandleError`?

But maybe we keep it separate for DevEx purposes? Thoughts?

near-sdk-core/src/types/public_key.rs
r-near: 1. Done in a9736fb4
2. That's a bit more tricky... `PublicKey::from_str` picks the length at runtime and holds a Vec

near-sdk-core/src/types/public_key.rs
r-near: Yeah good point. Done here: [`a9736fb` (this PR)](https://github.com/near/near-sdk-rs/pull/1631/commits/a9736fb45c98e3431e909848337d83c10b1f535b)

near-sdk-core/src/types/public_key.rs
r-near: Yeah, good call. Added here: [`c9b7cd8` (this PR)](https://github.com/near/near-sdk-rs/pull/1631/commits/c9b7cd87f650902acbc13abc9dd96840acf813bb)

near-sdk-core/src/types/public_key.rs
r-near: Yeah, I don't love it either... 

After inlining the helper on the near-global-contracts side, each crate at least only has the `cfg` pair once.

Let's make a follow-up issue to pull in single host functions from near-sdk-env

near-sdk/src/environment/env.rs
r-near: Yeah fair point. Done here: [`4e0b459` (this PR)](https://github.com/near/near-sdk-rs/pull/1631/commits/4e0b459a020951675f5385f2a5c24f5546554ff9)

near-global-contracts/src/lib.rs
r-near: Yeah it should match. Done here: [`3daec97` (this PR)](https://github.com/near/near-sdk-rs/pull/1631/commits/3daec97ca069a25addf5db82009ea4f1752b5569)

near-sdk/src/lib.rs
r-near: Why'd we gate state-init in the first place? IMO we should probably leave it alone. But curious to hear your thoughts.

near-global-contracts/src/universal_state_init.rs
vsavchyn-dev: While we do not need `RawStateInit`, just note that is serialized as Base64 with serde. That's why I asked

near-sdk-core/src/types/public_key.rs
vsavchyn-dev: I believe if we want to implement these `PublicKey` and `PublicKeyHandle` types, we at least should ensure that there is some kind of parsing errors that are sensible for devs. So yeah, I think `ParsePublicKeyError` would work for both `PublicKey` and `PublicKeyHandle`.

One thing to note is that we need interop with `near-crypto` error maybe.

near-sdk-core/src/types/public_key.rs
vsavchyn-dev: Welp, can't do anything with adding this to `PublicKey::from_str` then...
not a big deal though

near-sdk/Cargo.toml
```diff
@@ -42,8 +42,9 @@ hex = { version = "0.4", optional = true }
 # Used for caching, might be worth porting only functionality needed.
 once_cell = { version = "1.17", default-features = false }
 
-# nearcore 0.38 requires near-account-id 3 (adds `AccountType::UniversalAccount`).
-near-account-id = { version = "3", features = ["serde", "borsh"] }
+# nearcore 0.38 requires near-account-id 3 (adds `AccountType::UniversalAccount`); 3.1 adds
+# `UniversalAccountId::from_hash`, which the off-chain `universal_state_init_to_account_id` uses.
+near-account-id = { version = "3.1", features = ["serde", "borsh"] }
```
vsavchyn-dev, line 47: Another thing for future - unify dependencies under "workspace"

near-sdk/src/lib.rs
vsavchyn-dev: Well, it doesn't make sense to have `universal_state_init` pulled in if:
1. `deterministic_account_id`s are not available
2. normal `state_init` is not available

Plus, I am not sure a lot of people would use this feature in near future.

With reasons described above I thought it would be logical to gate `universal_state_init`, no?

near-sdk-core/src/types/public_key.rs
```diff
@@ -313,6 +315,202 @@ impl From<B58Error> for ParsePublicKeyError {
...
+                }
+            }
+        }
+    }
+
+    impl From<near_crypto::PublicKey> for PublicKeyHandle {
+        fn from(public_key: near_crypto::PublicKey) -> Self {
+            near_crypto::PublicKeyHandle::from(public_key).into()
+        }
+    }
```
vsavchyn-dev, line 468: What about `From<&near_crypto::PublicKey>`? I think it's worth implementing...

vsavchyn-dev (approved):
Other than additional `From<...> for PublicKeyHandle` that we may or may not need, this PR looks ok to me

near-sdk/src/lib.rs
r-near: What if we ungated those features then? Since they're officially part of the protocol now?

I think if we add a feature flag, we'd have to cover all the env functions, mock actions, the promise, etc. It's a lot of `cfg` sites...

near-sdk-core/src/types/public_key.rs
r-near: Added here: [`658bc69` (this PR)](https://github.com/near/near-sdk-rs/pull/1631/commits/658bc69d78278219991d116ffcc41d08650347cb)

near-sdk/Cargo.toml
r-near: Yeah good idea, let's open an issue for this as well

near-sdk-core/src/types/public_key.rs
r-near: Opened #1635

near-sdk/Cargo.toml
r-near: Opened #1636

near-global-contracts/src/universal_state_init.rs
```diff
@@ -0,0 +1,364 @@
...
+#[cfg_attr(
+    feature = "borsh",
+    derive(borsh::BorshSerialize, borsh::BorshDeserialize),
+    borsh(use_discriminant = true)
+)]
+#[cfg_attr(
+    feature = "schemars-v0_8",
+    derive(::schemars_v0_8::JsonSchema),
+    schemars(crate = "::schemars_v0_8")
+)]
```
mitinarseny: Are you sure that this ordering is correct, especially that `derive(JsonSchema)` comes after `serde(rename_all = "snake_case")` attribute?

near-global-contracts/src/universal_state_init.rs
```diff
@@ -0,0 +1,364 @@
...
+    /// Initial storage; empty unless seeded. Sorted keys give a canonical encoding.
+    #[cfg_attr(feature = "serde", serde_as(as = "BTreeMap<Base64, Base64>"))]
+    #[cfg_attr(
+        feature = "schemars-v0_8",
+        schemars(with = "::std::collections::BTreeMap<String, String>")
+    )]
+    #[cfg_attr(
+        feature = "serde",
+        serde(default, skip_serializing_if = "::std::collections::BTreeMap::is_empty")
+    )]
```
mitinarseny: These attributes can be merged and `schemars(with = "...")` is not needed:
```suggestion
    #[cfg_attr(
        feature = "serde",
        serde(default, skip_serializing_if = "BTreeMap::is_empty"),
        serde_as(as = "BTreeMap<Base64, Base64>")
    )]
```

near-global-contracts/src/universal_state_init.rs
```diff
@@ -0,0 +1,364 @@
...
+        derive_universal_account_id(&self.to_bytes())
+    }
+}
+
+/// Derives the `0u` account id from raw state-init bytes, like nearcore's
+/// `derive_universal_account_id`: SHA3-256 over exactly those bytes, Crockford-base32 encoded.
+///
+/// Takes the bytes rather than a typed value, so a caller can pass through a state-init version
+/// this crate predates.
+pub fn derive_universal_account_id(state_init: &[u8]) -> near_account_id::AccountId {
```
mitinarseny:
```suggestion
pub fn derive_universal_account_id(state_init: impl AsRef<[u8]>) -> near_account_id::AccountId {
```

near-global-contracts/src/universal_state_init.rs
```diff
@@ -0,0 +1,364 @@
...
+    {
+        hash = near_sdk_env::sha3_256(state_init);
+    }
+    #[cfg(not(any(near, feature = "__near-sdk-unit-testing")))]
+    {
+        use sha3::Digest;
+        hash = sha3::Sha3_256::digest(state_init).into();
+    }
+
+    UniversalAccountId::from_hash(hash).into_account_id()
```
mitinarseny: Can we simplify and use `near_digest::sha3::Sha3_256` instead?

near-global-contracts/src/lib.rs
```diff
@@ -55,6 +58,9 @@ pub use global_contract_identifier::*;
 mod state_init;
 pub use state_init::*;
 
+mod universal_state_init;
+pub use universal_state_init::*;
```
mitinarseny: Can we also re-export `near_sdk_core::PublicKeyHandle`?

near-sdk/src/environment/env.rs
```diff
@@ -2463,6 +2546,142 @@ pub fn promise_yield_resume_with_yield_id(yield_id: &[u8], data: impl AsRef<[u8]
...
+/// directly, so a contract can derive the id for a state-init version this SDK predates.
+///
+/// The account id commits to exactly these bytes: two encodings of the same logical state init
+/// are two different accounts.
+///
+/// # Requirements
+///
+/// Requires the host to support universal accounts (nearcore protocol version 87+, shipped in
+/// nearcore 2.14).
+pub fn universal_state_init_to_account_id_raw(state_init: &[u8]) -> AccountId {
```
mitinarseny:
```suggestion
pub fn universal_state_init_to_account_id_raw(state_init: impl AsRef<[u8]>) -> AccountId {
```

near-sdk/src/environment/env.rs
```diff
@@ -2463,6 +2546,142 @@ pub fn promise_yield_resume_with_yield_id(yield_id: &[u8], data: impl AsRef<[u8]
...
+///
+/// # Requirements
+///
+/// Requires the host to support universal accounts (nearcore protocol version 87+, shipped in
+/// nearcore 2.14).
+pub fn promise_batch_action_universal_state_init_raw(
+    promise_index: PromiseIndex,
+    state_init: &[u8],
+    amount: NearToken,
+) {
```
mitinarseny:
```suggestion
pub fn promise_batch_action_universal_state_init_raw(
    promise_index: PromiseIndex,
    state_init: impl AsRef<[u8]>,
    amount: NearToken,
) {
```

near-sdk/src/promise.rs
```diff
@@ -588,6 +599,41 @@ impl Promise {
...
+    ///         .with_access_key(env::signer_account_pk()),
+    /// );
+    /// Promise::new(state_init.derive_account_id())
+    ///     .universal_state_init(state_init, NearToken::from_millinear(10));
+    /// ```
+    pub fn universal_state_init(
+        self,
+        state_init: crate::universal_state_init::UniversalStateInit,
+        deposit: NearToken,
+    ) -> Self {
```
mitinarseny: This method MUST accept already borsh-serialized `UniversalStateInit`. See NEP-655.

near-global-contracts/src/universal_state_init.rs
r-near: Moot now: #1642 drops serde and JsonSchema from the typed `UniversalStateInit` (JSON callers carry `RawStateInit` as base64), [ba43194](https://github.com/near/near-sdk-rs/pull/1642/commits/ba431941). For the record we compiled it and the dumped schema had the snake_case `"v1"` tag, so the ordering was fine.

near-global-contracts/src/universal_state_init.rs
r-near: Moot now: #1642 removes serde from the typed struct, [ba43194](https://github.com/near/near-sdk-rs/pull/1642/commits/ba431941). The new `RawStateInit` relies on `serde_as` adding the schemars attribute as you suggested, so its schema is a base64 string, [106ae57](https://github.com/near/near-sdk-rs/pull/1642/commits/106ae579).

near-global-contracts/src/universal_state_init.rs
r-near: Done in https://github.com/near/near-sdk-rs/pull/1642: `derive_universal_account_id(state_init: impl AsRef<[u8]>)` with a non-generic inner fn, [6162167](https://github.com/near/near-sdk-rs/pull/1642/commits/61621671).

near-global-contracts/src/universal_state_init.rs
r-near: Done in https://github.com/near/near-sdk-rs/pull/1642: it hashes with `near_digest::sha3::Sha3_256` now, [6162167](https://github.com/near/near-sdk-rs/pull/1642/commits/61621671). near-digest also routes to the mocked host functions under near-sdk `unit-testing`, so nothing is lost there, [68e0c77](https://github.com/near/near-sdk-rs/pull/1642/commits/68e0c771).

near-global-contracts/src/lib.rs
r-near: Done in https://github.com/near/near-sdk-rs/pull/1642, [dec77a8](https://github.com/near/near-sdk-rs/pull/1642/commits/dec77a85).

near-sdk/src/environment/env.rs
r-near: Done in https://github.com/near/near-sdk-rs/pull/1642: raw is the only API now, so it's `universal_state_init_to_account_id(state_init: impl AsRef<[u8]>)` without the `_raw` suffix and the typed variant is gone, [cbea62d](https://github.com/near/near-sdk-rs/pull/1642/commits/cbea62d5).

near-sdk/src/environment/env.rs
r-near: Done in https://github.com/near/near-sdk-rs/pull/1642: `promise_batch_action_universal_state_init(promise_index, state_init: impl AsRef<[u8]>, amount)` is the only variant now, [cbea62d](https://github.com/near/near-sdk-rs/pull/1642/commits/cbea62d5).

near-sdk/src/promise.rs
r-near: Done in https://github.com/near/near-sdk-rs/pull/1642: `Promise::universal_state_init` takes `impl Into<RawStateInit>` and stores the bytes verbatim, and `Promise::new_universal_account` derives the receiver from the same bytes, [a3a7587](https://github.com/near/near-sdk-rs/pull/1642/commits/a3a75876). `RawStateInit` mirrors nearcore's type, [106ae57](https://github.com/near/near-sdk-rs/pull/1642/commits/106ae579).

